PR #252, opened 2013-12-01
Gist Comments API

khellang:
Here's an implementation of the gist comments API. The only thing missing should be some integration tests, I can add that later if you want :smile:

### DIFF
--- a/Octokit.Reactive/Clients/IObservableGistCommentsClient.cs
+++ b/Octokit.Reactive/Clients/IObservableGistCommentsClient.cs
@@ -2,7 +2,7 @@
 using System.Diagnostics.CodeAnalysis;
 using System.Reactive;
 
-namespace Octokit.Reactive.Clients
+namespace Octokit.Reactive
 {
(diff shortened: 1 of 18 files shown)


Haacked:
@khellang Can you remind me why we needed the `BodyWrapper` class?

khellang:
Uuuh. I don't really remember :confused: I think it might be because the API [expects an object and not just a string](https://developer.github.com/v3/gists/comments/#create-a-comment);

```json
{
  "body": "Just commenting for the sake of commenting"
}
```

Haacked:
Ah, right. Maybe `BodyResponse` is a better name, but no need to change it for now.

khellang:
If anything, it's a `BodyRequest` :wink: https://github.com/octokit/octokit.net/blob/1cea52eefa67d094e165b5f06a2d9dbad0330490/Octokit/Clients/GistCommentsClient.cs#L56

Haacked:
Or `RequestBody`!

PR #561, opened 2014-08-18
All models used in PATCH verbs should have nullable fields

thedillonb:
There were a few update models, used in PATCH requests, which had fields that were not nullable. This pull requests makes sure all fields within PATCH models can null every field (unless otherwise specified by the API docs).
1. Cloned the ReleaseUpdate (used in PATCH requests) model into a NewRelease model (used in PUT requests) since the two models should not have been shared between the request types as they had different field requirements.
2. Cleaned up the PullRequestUpdate object as it looks like there was a copy and paste mistake since it contained a few too many fields that don't exist on the update.

### DIFF
--- a/Octokit.Reactive/Clients/IObservableReleasesClient.cs
+++ b/Octokit.Reactive/Clients/IObservableReleasesClient.cs
@@ -43,7 +43,7 @@ namespace Octokit.Reactive
         /// <param name="data">A description of the release to create</param>
         /// <exception cref="ApiException">Thrown when a general API error occurs.</exception>
         /// <returns>The created <see cref="Release"/>.</returns>
-        IObservable<Release> Create(string owner, string name, ReleaseUpdate data);
+        IObservable<Release> Create(string owner, string name, NewRelease data);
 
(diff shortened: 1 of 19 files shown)


thedillonb:
@Haacked @shiftkey, Hey guys, with the new release 0.7 it looks like the changes I made to the IssueUpdate were reverted (f3d4b751cc7c255748c972a19b76467f0a50f39f - I think). With the reversion, it looks like you are unable to set the Labels to null, which as we discussed above, prevents you from *not* modifying that field during the PATCH.

shiftkey:
@thedillonb apologies, I'm looking to put together another release later this week - I'll make sure we get that back in https://github.com/octokit/octokit.net/issues/715

thedillonb:
@shiftkey, It's all good. Just wanted to make sure someone was aware of it. Let me know if there's any way I can assist.

shiftkey:
@thedillonb at least this time I'll make sure to keep a regression test around for Future Me to ensure I remember this :trollface:

PR #571, opened 2014-09-19
WIP: Dogfood

forki:
This is WIP and for discussion.

The release target would create a git tag push it to github and create a "github release" using the latest octokit version.

Open issues:

* Can we use the currently build Octokit.dll?
* Which files do you want to release


shiftkey:
Not sure why I missed this initial discussion, but here's my thoughts:

 - why are we doing this? to let those who want to manage their dependencies by hand do bypass NuGet
 - don't bother with PDBs or other files, just the binaries
 - the root folder should contain the platform flavours, with `Octokit.dll` and `Octokit.Reactive.dll` (if applicable) in place
 - the PR just uploads Octokit.dll, which might not be suitable for non-desktop apps
 
I'm not sure of the demand for this for non-desktop apps, so I'm happy to just start with a `net45/Octokit.dll` zip file if that makes your situation easier to work with...

shiftkey:
Taking this off my watch-list. @forki feel free to get in touch if you want to revive this idea.

forki:
> why are we doing this? to let those who want to manage their dependencies by hand do bypass NuGet

Actually the idea was to dogfood Octokit Relases API since basically whole F# ecosystem (and all the new C# projects that use https://github.com/fsprojects/ProjectScaffold) now uses Octokit with this script ;-)

It would probably have shown that https://github.com/fsharp/FAKE/commit/7fcbd023443c20bacaf658548334b5e19b48c5ed is needed with the latest octokit release. A breaking change which is not listed in the release notes.

shiftkey:
@forki apologies for that - the class itself is in the wrong namespace and so it was updated incorrectly when we changed the response models over to be readonly properties.

If you like I can revert that change for 0.7.1 which I was aiming to ship tomorrow my time.

forki:
No all is well. Paket takes care of that. Everybody I know is using the
github link feature and automatically gets my "patch".

Still I'd love to Octokit dogfooding itself.

shiftkey:
I've opened up https://github.com/octokit/octokit.net/issues/721 to address this anyway - it's wasn't in the right location, which lead to the problem at hand...

Once I've dealt with some of the urgent issues (like Mono* support) and the other outstanding PRs, I'll come back to this. Thanks again for the initial spike.

forki:
Sorry if this sounded angry. Actually I'm really really happy about the combination of FAKE+Paket+Octokit. It works very well in practice. It works so well that according to https://github.com/fsprojects/Paket/releases I have already pushed 548 releases of Paket alone.

shiftkey:
@forki it's all good, I'm mostly annoyed at myself for not picking that issue (and also for letting the release languish for so long).

PR #603, opened 2014-11-05
Implement merge client

tabro:
There is no implementation of the github merging api in octokit, so I decided to implement it.
The base functionality is done, but I need to figure out how to make a suitable integration tests (which seem to be hard since there is no handles to create branches - or am I missing something?).


tabro:
I consider the implementation ready for review.

The pullrequest contains the following:

- Implementation of a MergingClient that implements the github v3 merging api (https://developer.github.com/v3/repos/merging/)
- Extension of the RepositoryClient and ObservableRepositoryClient with the MergingClient/ObservableMergingClient
- UnitTests of the MergingClient
- IntegrationsTest of the MergingClient

tabro:
Some context on why I added the Merging api:

At job I wanted to optimize our development process, by automating mergin of our main development branch into a branch that reflects the currently live baseline.

I tried using the PullRequest api, however since pull requests creates no-ff merges, we get bubble-commits instead of a simple ff merge.

So since there was no implementation of the merging API i decided to write it.

Octokit/Clients/MergingClient.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// <param name="name">The name of the repository</param>
+        /// <param name="merge">The merge to create</param>
+        /// <returns></returns>
+        public Task<Merge> Create(string owner, string name, NewMerge merge)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
+            Ensure.ArgumentNotNullOrEmptyString(name, "name");
+            Ensure.ArgumentNotNull(merge, "merge");
+            Ensure.ArgumentNotNullOrEmptyString(merge.Base, "baseBranch");
+            Ensure.ArgumentNotNullOrEmptyString(merge.Head, "head");
```
Haacked: These last two `Ensure.ArgumentNotNullOrEmptyString` calls should be in the constructor of `NewMerge` not here. We reserve argument null exceptions for when the argument is null, not when a property of the argument is null.

Octokit/Models/Request/NewMerge.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// <summary>
+        /// Create a new commit which has multiple parents (i.e. a merge commit)
+        /// </summary>
+        /// <param name="base">The name of the base branch that the head will be merged into</param>
+        /// <param name="head">The head to merge. This can be a branch name or a commit SHA1.</param>
+        /// <param name="commitMessage">Commit message to use for the merge commit. If omitted, a default message will be used.</param>
+        public NewMerge(string @base, string head, string commitMessage)
+        {
+            Base = @base;
+            Head = head;
```
Haacked: The pattern we use for request objects like this is that the `ctor` should only accept values that are absolutely required. Thus we should ensure that `@base` and `head` are not null nor empty. `commitMessage` is optional so it shouldn't be a parameter to the ctor.

Octokit/Models/Request/NewMerge.cs
```diff
@@ -0,0 +1,35 @@
...
+            CommitMessage = commitMessage;
+        }
+
+        public string CommitMessage { get; set; }
+        public string Base { get; set; }
```
Haacked: Since `Base` is required and set by the ctor, make this a `readonly` property.

Octokit/Models/Request/NewMerge.cs
```diff
@@ -0,0 +1,35 @@
...
+        public string Head { get; set; }
```
Haacked: Make this `readonly`

Haacked:
This is great! I just have a few nitpicky comments, but otherwise this looks very well done! :+1:

tabro:
@Haacked Cool - I corrected the findings! :+1:

Octokit.Tests.Integration/Clients/MergingClientTests.cs
```diff
@@ -0,0 +1,103 @@
...
+        _repository = _client.Repository.Create(new NewRepository { Name = repoName, AutoInit = true }).Result;
+        _owner = _repository.Owner.Login;
+    }
+
+    [IntegrationTest]
+    public async Task CanCreateMerge()
+    {
+        await CreateTheWorld();
+        
+        var newMerge = new NewMerge("master", branchName, "merge commit to master from integrationtests");
```
Haacked: Ah, this won't compile with your last changes.

tabro: Ah - doh - I will fix it quickly..

tabro:
@Haacked Fixed - build is green again.

Haacked:
selfie-0

PR #630, opened 2014-12-14
writing some docs

shiftkey:
It's a bit copy-paste, but thankfully the build will propagate the xml-docs from interface to implementation, so less :poop: work all round

 - [x] document `IGitHubClient`
 - [x] document `IObservableGitHubClient`
 - [x] document `Exceptions` namespace
 - [x] document `Authentication` namespace
 - [x] document `Extensions` namespace


Haacked:
Nice work. One minor nit that can be addressed later. I have a bad memory for these things, but I thought we all agreed that summaries are full sentences and thus require punctuation. But parameters, for example, are fragments and don't end in periods. We should memorialize whatever we decide into a style guide for comments. :smile: I seem to remember @half-ogre having opinions on this.

shiftkey:
> We should memorialize whatever we decide into a style guide for comments.

:thumbsup::thumbsup::thumbsup::thumbsup::thumbsup:

PR #649, opened 2014-12-31
Repository Contents API

Haacked:
This incorporates and supersedes #434 

I took a slightly different approach. In the interest of keeping close to the API, there's only one method called `GetContents` that maps to the ["Get contents" API](https://developer.github.com/v3/repos/contents/#get-contents). I was able to accomplish this by changing the serializer a bit. When it's trying to deserialize a JSON object into an `IEnumerable<T>` but the JSON text is clearly a `T` and not an array, I simply wrap the json text with square brackets turning it into an array of one item and then deserialize it.

The other bit I had to do is when creating content, the text of the content needs to be base64 encoded. Also, when requesting content, the data is base64 encoded. So I changed the serializer to handle this transparently by simply applying the `SerializeAsBase64` attribute. So the consumer of the client doesn't have to worry about this.


Octokit/Models/Response/RepositoryContent.cs
```diff
@@ -0,0 +1,43 @@
...
+        /// </summary>
+        [Parameter(Key = "content")]
+        protected string EncodedContent { get; set; }
+
+        /// <summary>
+        /// The unencoded content. Only access this if the content is expected to be text and not binary content.
+        /// </summary>
+        public string Content
+        {
+            get { return EncodedContent.FromBase64String(); }
```
shiftkey: Should this guard against `null` for `EncodedContent`?

Octokit/Models/Request/CreateFileRequest.cs
```diff
@@ -0,0 +1,126 @@
...
+    /// </summary>
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
+    public class DeleteFileRequest : ContentRequest
+    {
+        public DeleteFileRequest(string message, string sha) : base(message)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(sha, "sha");
+
+            Sha = sha;
+        }
```
shiftkey: dat newline

Octokit/Http/SimpleJsonSerializer.cs
```diff
@@ -23,55 +23,28 @@ public T Deserialize<T>(string json)
...
-
-                // sometimes Octokit needs to send a null with the payload so the user
-                // can unset the value of a property.
-                // This method uses the same checks as PocoJsonSerializerStrategy
-                // to identify the right fields and properties to serialize
-                // but it then filters on the presence of SerializeNullAttribute.
+                var propertiesAndFields = type.GetPropertiesAndFields().Where(p => p.CanSerialize).ToList();
 
-                foreach (var propertyInfo in ReflectionUtils.GetProperties(type))
+                foreach (var property in propertiesAndFields.Where(p => p.SerializeNull))
```
shiftkey: You could probably :fire: this `foreach` loop and populate `_membersWhichShouldPublishNull` directly:

```
_membersWhichShouldPublishNull = propertiesAndFields.Where(p => p.SerializeNull)
    .Select(member => type.FullName + "-" + member.JsonFieldName)
    .ToList();
```

But this code is already much more simpler than what was there before :sparkles:

shiftkey: Or even make my hacky workaround go away completely `¯\(°_o)/¯`

Octokit/Octokit-Monotouch.csproj
```diff
@@ -351,6 +351,19 @@
...
+    <Compile Include="Helpers\SerializeAsBase64Attribute.cs" />
+    <Compile Include="Models\Response\ContentType.cs" />
+    <Compile Include="Models\Response\RepositoryContent.cs" />
+    <Compile Include="Models\Response\RepositoryContentChangeSet.cs" />
+    <Compile Include="Models\Response\CommitContent.cs" />
+    <Compile Include="Models\Response\SignatureResponse.cs" />
+    <Compile Include="Helpers\PropertyOrField.cs" />
   </ItemGroup>
   <Import Project="$(MSBuildExtensionsPath)\Xamarin\iOS\Xamarin.MonoTouch.CSharp.targets" />
+  <Import Project="$(MSBuildBinPath)\Microsoft.CSharp.targets" />
```
shiftkey: Urgh, I need to renew my Xamarin license and ensure everything still builds there...

Octokit/Http/JsonHttpPipeline.cs
```diff
@@ -44,11 +46,18 @@ public void DeserializeResponse<T>(IResponse<T> response)
...
                 // simple json does not support the root node being empty. Will submit a pr but in the mean time....
-                if (response.Body != "{}") 
+                if (body != "{}") 
                 {
-                    var json = _serializer.Deserialize<T>(response.Body);
-                    response.BodyAsObject = json;
+                    // If we're expecting an array, but we get a single object, just wrap it.
+                    // This supports an api that dynamically changes the return type based on the content.
+                    if ((typeof(IEnumerable).IsAssignableFrom(typeof(T)))
+                        && body.StartsWith("{", StringComparison.Ordinal))
```
shiftkey: Paranoia check: could this break if we get whitespace around the `{}`? (it's probably unlikely)

Octokit.Tests.Integration/Clients/RepositoryContentsClientTests.cs
```diff
@@ -0,0 +1,133 @@
...
+                    "somefile.txt",
+                    new DeleteFileRequest("Deleted file", fileSha));
+
+                await Assert.ThrowsAsync<NotFoundException>(
+                    async () => await fixture.GetContents(repository.Owner.Login, repository.Name, "somefile.txt"));
+            }
+            finally
+            {
+                if (repository != null)
+                    client.Repository.Delete(repository.Owner.Login, repository.Name);
```
shiftkey: We've got a couple of different ways to dispose of repositories when we create them in the scope of an integration test (mostly I believe it's around using `IDisposable` in tests to use xUnit hooks.

Not a blocker on shipping this feature, but I'd like to unify these: https://github.com/octokit/octokit.net/issues/655

shiftkey:
Code and tests look pretty great. Just a few nagging questions :metal:

Octokit/Http/JsonHttpPipeline.cs
Haacked: Whitespace before the "{" would be an invalid JSON response so we'd be fubar anyways.

Octokit/Http/SimpleJsonSerializer.cs
Haacked: In another PR. I have some upcoming improvements to how we do serialization.

Haacked:
Something keeps fucking with the whitespace. I suspect it's the documentation rewriter. Anyways, this should be good I hope. I'm up too late again. :smile: so I'm going to sleep and hope the build is green.

:package:

Octokit/Http/JsonHttpPipeline.cs
shiftkey: :ok_hand:

Octokit/Http/SimpleJsonSerializer.cs
shiftkey: :ok_hand:

PR #656, opened 2015-01-02
support for paid account test scenarios

### LINKED ISSUE #631: Integration tests relying on test account having at least a micro plan
Some integration tests are trying to create and update a private repository. This feature is only available when the user has a micro plan (or better).

Right now these tests are being run alongside the `IntegrationTestAttribute`. To fix this we could follow the same approach than for the `OrganizationTestAttribute`:

1. Create a `OCTOKIT_GITHUB_HAS_PAID_PLAN` environment variable set to true if the test account has at least a micro plan.
2. Read this environment variable in `Helper` and assign it to the newly created boolean `HasPaidPlan`.
3. Create a `PaidPlanTestAttribute` that return `SkipCommand` if `HasPaidPlan` is false, or the tests otherwise.
4. Apply this attribute to the required test methods (listed below).

Tests relying on a paid account:
* TheCreateMethodForUser.CreatesANewPrivateRepository
* TheEditMethod.UpdatesPrivate

What do you think of this approach and the naming convention (`OCTOKIT_GITHUB_HAS_PAID_PLAN`, `PaidPlanTestAttribute` and `HasPaidPlan`)?

shiftkey:
Fixes #631

 - [x] added `[PaidAccountTest]` attribute
 - [x] ported `TheEditMethod.UpdatesPrivate` to new attribute
 - [x] ported `TheCreateMethodForUser.CreatesANewPrivateRepository` to new attribute
 - [x] fixes bug with error message not being extracted from message https://github.com/octokit/octokit.net/commit/b73775b9aaaf7f62d1bc5eb8dd7437bf1eff4eb9 and https://github.com/octokit/octokit.net/commit/c76cc9402042ea881816b3944217d554dce568ca
 - [x] update docs for new environment variable

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -105,6 +105,7 @@ owned by that account. Then set the following environment variables:
 `OCTOKIT_GITHUBUSERNAME` (set this to the test account's username)
 `OCTOKIT_GITHUBPASSWORD` (set this to the test account's password)
 `OCTOKIT_GITHUBORGANIZATION` (set this to the test account's organization)
+`OCTOKIT_PRIVATEREPOSITORIES` (set this to `TRUE` to indicate account has access to private repositories)
 
 Once these are set, the integration tests will be executed both when 
 running the FullBuild MSBuild target, and when running the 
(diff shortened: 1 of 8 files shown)


shiftkey:
Ready for review :cow2:

Haacked:
Code looks good. Just a question, don't we have tests that require a free account? If we do, we might need to have 2-test accounts rather than this boolean conditional.

shiftkey:
> Just a question, don't we have tests that require a free account?

I don't believe so - the test account I use has a micro plan, so I would be seeing different failures here.

PR #658, opened 2015-01-02
Readonly response models

### LINKED ISSUE #650: Response models should be readonly
It's bothered me that our response models are mutable. They really shouldn't be. Now that I've dug into the serialization code a bit, I think we can make our models have readonly properties. Right now, I'm planning to make the properties have a `protected` setter instead of `private` to provide flexibility for others who might be deriving from these classes or testing them.

Let me know what you think. I'll have a PR ready soon-ish. This depends on the work I did in #649

Haacked:
This makes our response models readonly. I made the setters `protected` instead of `private` to provide flexibility for others such as inheritors, testers, etc.. Also, it gets rid of that warning that the setter is never used. Let me know what you think.

Fixes #650


Octokit/Models/Response/Deployment.cs
```diff
@@ -14,48 +14,48 @@ public class Deployment
...
         /// <summary>
         /// The <seealso cref="User"/> that created the deployment.
         /// </summary>
-        public User Creator { get; set; }
+        public User Creator { get; protected set; }
 
         /// <summary>
         /// JSON payload with extra information about the deployment.
         /// </summary>
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Usage", "CA2227:CollectionPropertiesShouldBeReadOnly")]
```
khellang: Can these be removed now?

khellang: Thinking about it, shouldn't collection properties be initialized in the constructors and have `private` setters? I assume SimpleJson supports it since it supports `protected` setters?

khellang:
Also, does https://github.com/octokit/octokit.net/commit/1e1039ff244fda2d8ca4b78f4da6f136160b6b4f mean that all `List<T>` properties etc. can be converted to `IReadOnlyList<T>`?

Octokit/Models/Response/Deployment.cs
Haacked: We shouldn't initialize them in the ctors for these types because they'll just get replaced when SimpleJson deserializes the json into an object. It uses the private setters to set the collection.

Haacked:
Quite possibly! I'll test it out.

Haacked:
Also, it looks like there were some new model types I missed when I merged master in.

Haacked:
SimpleJson doesn't support readonly dictionaries. And yet another yak appears!

Haacked:
Ok, I created a [PR to simple-json](https://github.com/facebook-csharp-sdk/simple-json/pull/65). In the meanwhile, I'll just edit our copy directly.

Haacked:
Ok, I think this is ready. Would love some :eyes: on it. Thanks for the excellent review so far @khellang.

Octokit/Models/Response/Deployment.cs
Haacked: Just a follow-up. For "request models" what you described is exactly correct. The ctor should instantiate the collection and the setter should be private. I made that change. Let me know if you find any cases I missed.

Octokit/Models/Response/Commit.cs
```diff
@@ -6,11 +6,11 @@ namespace Octokit
...
-        public SignatureResponse Author { get; set; }
-        public SignatureResponse Committer { get; set; }
-        public GitReference Tree { get; set; }
-        public IEnumerable<GitReference> Parents { get; set; }
-        public int CommentCount { get; set; }
+        public string Message { get; protected set; }
+        public SignatureResponse Author { get; protected set; }
+        public SignatureResponse Committer { get; protected set; }
+        public GitReference Tree { get; protected set; }
+        public IEnumerable<GitReference> Parents { get; protected set; }
```
khellang: `IReadOnlyList<GitReference>`?

Octokit/Models/Response/CompareResult.cs
```diff
@@ -8,18 +8,18 @@ namespace Octokit
...
+        public string PermalinkUrl { get; protected set; }
+        public string DiffUrl { get; protected set; }
+        public string PatchUrl { get; protected set; }
+        public GitHubCommit BaseCommit { get; protected set; }
+        public GitHubCommit MergedBaseCommit { get; protected set; }
+        public string Status { get; protected set; }
+        public int AheadBy { get; protected set; }
+        public int BehindBy { get; protected set; }
+        public int TotalCommits { get; protected set; }
+        public IReadOnlyCollection<GitHubCommit> Commits { get; protected set; }
```
khellang: Maybe it's a good idea to standardize on either `IReadOnlyList<T>` or `IReadOnlyCollection<T>`?

Octokit/Models/Response/Contributor.cs
```diff
@@ -11,11 +11,11 @@ namespace Octokit
...
     public class Contributor
     {
-        public Author Author { get; set; }
+        public Author Author { get; protected set; }
 
-        public int Total { get; set; }
+        public int Total { get; protected set; }
 
-        public IEnumerable<WeeklyHash> Weeks { get; set; }
+        public IEnumerable<WeeklyHash> Weeks { get; protected set; }
```
khellang: `IReadOnllyList<WeeklyHash>`?

Octokit/Models/Response/Issue.cs
```diff
@@ -13,71 +13,71 @@ public class Issue
...
         /// </summary>
-        public User User { get; set; }
+        public User User { get; protected set; }
         
         /// <summary>
         /// The set of labels applied to the issue
         /// </summary>
-        [SuppressMessage("Microsoft.Usage", "CA2227:CollectionPropertiesShouldBeReadOnly")]
-        public ICollection<Label> Labels { get; set; }
+        public IReadOnlyCollection<Label> Labels { get; protected set; }
```
khellang: Same here...

Octokit/Models/Response/TreeResponse.cs
```diff
@@ -12,18 +12,17 @@ public class TreeResponse
...
         /// </summary>
-        public Uri Url { get; set; }
+        public Uri Url { get; protected set; }
 
         /// <summary>
         /// The list of Tree Items for this Tree response.
         /// </summary>
-        [SuppressMessage("Microsoft.Usage", "CA2227:CollectionPropertiesShouldBeReadOnly")]
-        public ICollection<TreeItem> Tree { get; set; }
+        public IReadOnlyCollection<TreeItem> Tree { get; protected set; }
```
khellang: And here :smile:

khellang:
Changed test in https://github.com/khellang/octokit.net/commit/3d16def7ec515a4deb1a6c816ab8e04b13684ab5 to verify handling of readonly collections, and it works beatifully :grin: Maybe there should be a convention test that verifies that collections are of the right type? Is there a reason for sometimes using `IReadOnlyCollection<T>` over `IReadOnlyList<T>`?

Haacked:
> Is there a reason for sometimes using IReadOnlyCollection<T> over IReadOnlyList<T>

Good question. I don't see any benefit to using collection over list here. List provides an indexer which can be handy.

Haacked:
:shaved_ice:

shiftkey:
> Maybe there should be a convention test that verifies that collections are of the right type? 

:heart::heart::heart::heart::heart:

Haacked:
:pineapple:

PR #659, opened 2015-01-04
Release v0.6.2

shiftkey:
- [x] update release notes
- [x] version bump
- [x] :thumbsup: on release notes
- [x] integration tests all :ok_hand: 
- [x] :ship: to MyGet
- [x] :shipit:

### DIFF
--- a/Octokit/Clients/RepositoriesClient.cs
+++ b/Octokit/Clients/RepositoriesClient.cs
@@ -386,18 +386,17 @@ namespace Octokit
         /// <param name="owner">The owner of the repository</param>
         /// <param name="name">The name of the repository</param>
         /// <returns>All languages used in the repository and the number of bytes of each language.</returns>
-        public Task<IReadOnlyList<RepositoryLanguage>> GetAllLanguages(string owner, string name)
+        public async Task<IReadOnlyList<RepositoryLanguage>> GetAllLanguages(string owner, string name)
         {
             Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
             Ensure.ArgumentNotNullOrEmptyString(name, "name");
 
-            return ApiConnection
-                .Get<IDictionary<string, long>>(ApiUrls.RepositoryLanguages(owner, name))
-                .ContinueWith<IReadOnlyList<RepositoryLanguage>>(t => 
-                    new ReadOnlyCollection<RepositoryLanguage>(
-                        t.Result.Select(kvp => new RepositoryLanguage(kvp.Key, kvp.Value)).ToList()
-                    ),
-                    TaskContinuationOptions.OnlyOnRanToCompletion);
+            var data = await ApiConnection
+                .Get<Dictionary<string, long>>(ApiUrls.RepositoryLanguages(owner, name))
+                .ConfigureAwait(false);
+
+            return new ReadOnlyCollection<RepositoryLanguage>(
+                data.Select(kvp => new RepositoryLanguage(kvp.Key, kvp.Value)).ToList());
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


Octokit/Http/InMemoryCredentialStore.cs
```diff
@@ -15,7 +15,7 @@ public InMemoryCredentialStore(Credentials credentials)
 
         public Task<Credentials> GetCredentials()
         {
-            return Task.Factory.StartNew(() => _credentials);
+            return Task.FromResult(_credentials);
```
shiftkey: This actually introduces some resource contention when parallelizing tests (but isn't the only one).

Do you have any objections to this @haacked?

Octokit/Http/JsonHttpPipeline.cs
```diff
@@ -48,12 +48,15 @@ public void DeserializeResponse<T>(IResponse<T> response)
...
-                if (body != "{}") 
+                if (body != "{}")
                 {
+                    var typeIsDictionary = typeof(IDictionary).IsAssignableFrom(typeof(T));
+                    var typeIsEnumerable = typeof(IEnumerable).IsAssignableFrom(typeof(T));
+                    var responseIsArray = body.StartsWith("{", StringComparison.Ordinal);
+
                     // If we're expecting an array, but we get a single object, just wrap it.
                     // This supports an api that dynamically changes the return type based on the content.
-                    if ((typeof(IEnumerable).IsAssignableFrom(typeof(T)))
```
shiftkey: `IDictionary` subclasses `IEnumerable`, so some of our tests which return `Dictionary<T1,T2>` were being caught here and deserialized incorrectly.

Feel free to submit a PR to make this a bit more cleaner.

Octokit/Clients/RepositoriesClient.cs
```diff
@@ -386,18 +386,17 @@ public Task<IReadOnlyList<RepositoryContributor>> GetAllContributors(string owne
...
             Ensure.ArgumentNotNullOrEmptyString(name, "name");
 
-            return ApiConnection
-                .Get<IDictionary<string, long>>(ApiUrls.RepositoryLanguages(owner, name))
-                .ContinueWith<IReadOnlyList<RepositoryLanguage>>(t => 
-                    new ReadOnlyCollection<RepositoryLanguage>(
-                        t.Result.Select(kvp => new RepositoryLanguage(kvp.Key, kvp.Value)).ToList()
-                    ),
-                    TaskContinuationOptions.OnlyOnRanToCompletion);
+            var data = await ApiConnection
```
shiftkey: Just tidied up this function to behave like `GetEmoji` while I was in changing `IDictionary` to `Dictionary`

Octokit/Http/InMemoryCredentialStore.cs
Haacked: No objections. In fact, I meant to make this change everywhere. We should never be using `StartNew` just to wrap an object in a task.

shiftkey: I'll clean that up against `master` once this is merged in.

shiftkey: Also, I can see a `Task.Factory.StartNew` inside `HttpClientHandler.SendAsync` that also causes contention (I see four requests enter, but none complete).

shiftkey:
Can I get a :thumbsup: on the release notes? Everything else is ready to roll out after that...

Haacked:
:thumbsup: to release notes

shiftkey:
And we're live: https://www.nuget.org/packages/Octokit/0.6.2

PR #662, opened 2015-01-05
Make the response objects readonly.

Haacked:
:warning: BREAKING CHANGE ALERT! :warning: 

One of the haacks we've had in the API design is that `IResponse` was not a readonly interface. We made the `BodyAsObject` property settable. The reason was it made it convenient to separate the creation of the response from the deserialization.

But in thinking about this, I think the proper approach to that is not to mutate the response object, but to wrap it with another object that accepts the deserialized type (look at how the `Connection.Run<T>` method changes as well as `JsonHttpPipeline` object).

So in this PR, I went all out and we have a `Response` object that's pretty simple. And we have a separate `ApiResponse<T>` object that has two properties, the deserialized `Body` and the original `IResponse`. So `IHttpClient.Send` just returns a `IResponse` and we then deserialize the body and wrap both things in `ApiResponse<T>`.

An alternative approach I considered is to simplify this by getting rid of `IResponse` altogether and only having an `ApiResponse<T>`. However that means that any implementations of `IHttpClient` would be responsible not only for creating a response, but deserializing it. With the current implementation, an implementation of `IHttpClient` doesn't have to worry about deserialization. It's handled separately.

Thoughts?


Octokit/Http/JsonHttpPipeline.cs
```diff
@@ -40,15 +39,15 @@ public void SerializeRequest(IRequest request)
             request.Body = _serializer.Serialize(request.Body);
         }
 
-        public void DeserializeResponse<T>(IResponse<T> response)
+        public IApiResponse<T> DeserializeResponse<T>(IResponse response)
```
shiftkey: :heart::heart::heart::heart::heart:

Octokit/Http/HttpClientAdapter.cs
```diff
@@ -140,7 +137,7 @@ protected virtual HttpRequestMessage BuildRequestMessage(IRequest request)
             return requestMessage;
         }
 
-        static string GetContentType(HttpContent httpContent)
+        static string GetContentMediaTypeType(HttpContent httpContent)
```
shiftkey: Is `MediaTypeType` necessary here, or just a refactoring mistype?

Octokit/Http/IResponse.cs
```diff
@@ -4,19 +4,50 @@
...
 
+    /// <summary>
+    /// Represents a generic HTTP response
+    /// </summary>
     public interface IResponse
     {
-        object BodyAsObject { get; set; }
-        string Body { get; set; }
-        Dictionary<string, string> Headers { get; }
-        Uri ResponseUri { get; set; }
```
shiftkey: This seems to be the only property that didn't survive the refactor intact. Any particular reasons why?

Octokit/Helpers/HttpClientExtensions.cs
```diff
@@ -7,16 +7,12 @@ namespace Octokit
 {
     public static class HttpClientExtensions
     {
-        /// <summary>
-        /// OBSOLETE
-        /// </summary>
-        [Obsolete("This will be removed in a future release")]
-        public static Task<IResponse<T>> Send<T>(this IHttpClient httpClient, IRequest request)
+        public static Task<IResponse> Send(this IHttpClient httpClient, IRequest request)
```
shiftkey: This was marked as `[Obsolete]` (probably by me). Do we need it?

shiftkey:
:apple:

Octokit/Http/IResponse.cs
Haacked: I think because it wasn't being used anywhere. Do we need it? Easier to add it later than remove it.

Octokit/Helpers/HttpClientExtensions.cs
Haacked: I don't think _we_ need it, but it may be useful to people who call `IHttpClient` directly. For example, we use this extension method in GHfW.

Octokit/Http/IResponse.cs
shiftkey: Yeah, that's fine. I'll run the tests for this and confirm it's all kosher.

Haacked:
:wolf:

Octokit/Helpers/HttpClientExtensions.cs
shiftkey: Yeah, this bit me upgrading to vLatest this week. 

I didn't mind calling out `CancellationToken.None` for these few usages, so I guess I'm debating cosmetics and shortcuts here. Happy to bring it back :tada:

Haacked: We can always add the extension method to the GHfW side and remove it from here. I don't feel strongly either way other than leaning towards being lazy. :stuck_out_tongue:

shiftkey: :thumbsup: to lazy, that work hasn't made it into a GHfW PR yet.

If I get caremad about it again, I'll PR it separately.

Octokit.sln
```diff
@@ -101,6 +101,7 @@ Global
 		HideSolutionNode = FALSE
 	EndGlobalSection
 	GlobalSection(NestedProjects) = preSolution
+		{01687D54-1D87-4562-A721-C57F1C94052C} = {CEC9D451-6291-4EDF-971A-D398144FBF96}
```
shiftkey: This is just the integration test project being nested. I'll undo this, not a blocker.

Octokit/Http/ApiInfoParser.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace Octokit.Internal
 {
-    internal static class ApiInfoParser
+    public static class ApiInfoParser
```
shiftkey: Do we need this to be `public`? The namespace intent is confusing me...

Octokit/Http/IResponse.cs
shiftkey: Tests are :sparkles:, just giving this another :eyes:

Octokit/Http/Response.cs
```diff
@@ -0,0 +1,56 @@
+﻿using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Net;
+
+namespace Octokit.Internal
+{
+    /// <summary>
+    /// Represents a generic HTTP response
+    /// </summary>
+    public class Response : IResponse
```
shiftkey: As above, do we need to make this `public`?

shiftkey:
A couple of thoughts on `public`/`internal` changes, but this is looking pretty amazing.

Needs a merge with `master`, but aside from that I think we're close to :ship:ing this.

If you're interested in drafting up the **Breaking Changes** for this we can start setting up a 0.7 release asap.

Octokit/Http/Response.cs
Haacked: Hmm, I guess not. I could make it internal till someone complains.

shiftkey: Ideally I'd like everything under `Octokit.Internal` to be marked as internal - my thinking on this is that if there's a compelling case to make a type public, it shouldn't be in this namespace.

Happy to do a once-over on this after this PR (before the next major release), just calling this out as it's a new type.

shiftkey:
:moon:

Octokit/Http/Response.cs
Haacked: > Ideally I'd like everything under Octokit.Internal to be marked as internal - my thinking on this is that if there's a compelling case to make a type public, it shouldn't be in this namespace.

Hmm, that's not the original intent I had for this namespace. It wasn't to mark types as `internal`. The keyword suffices for that.

It's specifically for types that most people don't need to see directly, but many, if not most, need to use. For example, `ApiResponse` is in this namespace.

Most users can get by importing just the `Octokit` namespace and these other types won't "pollute" their namespace, but are still usable as return values. If they need these types directly, they can import `Octokit.Internal`.

I'm not opposed to making these specific case internal. I just wanted to clarify why we have `Octokit.Internal` in the first place. Maybe a better name for that namespace is in order. I wonder if `Octokit.Internals` is slightly better to distinguish between the keyword and the fact that these types represent the internal workings of the lib. :smile:

Haacked:
Ok, merged master in. Hopefully this is good. :smile:

:tanabata_tree:

Haacked:
*bump* to Mr. @shiftkey

Haacked:
:poop: I just merged another PR into master that breaks this PR. Doh!

Haacked:
selfie-0

PR #667, opened 2015-01-08
Refactored RepositoryExistsException 

### LINKED ISSUE #633: TheCreateMethodForUser.ThrowsRepositoryExistsExceptionForExistingRepository is failing
It is failing because `RepositoriesClient.Create` is calling its overload with `null` as the value for `organizationLogin` argument. This value is then used to instantiate the `RepositoryExistsException`.

gabrielweyer:
Fixed #633

### DIFF
--- a/Octokit/Clients/RepositoriesClient.cs
+++ b/Octokit/Clients/RepositoriesClient.cs
@@ -86,6 +86,11 @@ namespace Octokit
                     errorMessage,
                     StringComparison.OrdinalIgnoreCase))
                 {
+                    if (string.IsNullOrEmpty(organizationLogin))
+                    {
+                        throw new RepositoryExistsException(newRepository.Name, e);
+                    }
+
                     var baseAddress = Connection.BaseAddress.Host != GitHubClient.GitHubApiUrl.Host
                         ? Connection.BaseAddress
                         : new Uri("https://github.com/");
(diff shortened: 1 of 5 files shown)


Octokit/Clients/RepositoriesClient.cs
```diff
@@ -86,6 +86,11 @@ public Task<Repository> Create(string organizationLogin, NewRepository newReposi
                     errorMessage,
                     StringComparison.OrdinalIgnoreCase))
                 {
+                    if (string.IsNullOrEmpty(organizationLogin))
+                    {
+                        throw new RepositoryExistsException(newRepository.Name, e);
+                    }
```
Haacked: In looking at this, shouldn't we always be passing in `baseAddress` to `RepositoryExistsException`. Why would we only pass it in if we know your org?

gabrielweyer: In the previous version of the code `ExistingRepositoryWebUrl` was being set only if the owner is an organization. This is due to fact that we don't have access to the owner's name if it's an account and not an organization.

Haacked: > This is due to fact that we don't have access to the owner's name if it's an account and not an organization.

Ah. And we don't have access to the name because when we do operations for the _current_ account, the name isn't needed as part of the URL. Right?

That was confusing to me and I should know that. :stuck_out_tongue:

Haacked:
selfie-0

Octokit/Clients/RepositoriesClient.cs
gabrielweyer: > Ah. And we don't have access to the name because when we do operations for the current account, the name isn't needed as part of the URL. Right?

Exactly :smile: at least the integration test is passing now!

PR #672, opened 2015-01-11
Adds roslyn temporary files to the gitignore file.

Zoltu:
The roslyn compiler creates a folder name [solutionname].sln.ide next to the solution file.  This folder is used for storing temporary files and should not be committed to source control.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,9 @@ bin/**/*.pdb
 *.vspscc
 .builds
 
+# Roslyn cache directories
+*.ide/
+
 # Visual C++ cache files
 ipch/
 *.aps


Haacked:
Looks good to me. Perhaps a PR to https://github.com/github/gitignore/blob/master/VisualStudio.gitignore is also in order.

.gitignore
```diff
@@ -10,6 +10,9 @@ Desktop.ini
 ## Ignore Visual Studio temporary files, build results, and
 ## files generated by popular Visual Studio add-ons.
 
+# Roslyn temporary files
+*.sln.ide/
```
shiftkey: Let's use the same rule that's in the official repo: `*.ide/`

https://github.com/github/gitignore/blob/master/VisualStudio.gitignore#L22-L23

shiftkey:
No upstream changes necessary, as this was added in https://github.com/github/gitignore/pull/1035 and we found a duplicate rule in https://github.com/github/gitignore/pull/1292 which was reverted.

.gitignore
Haacked: Ah! Good catch.

Zoltu:
Fixed.  Copied and pasted straight from https://github.com/github/gitignore/blob/master/VisualStudio.gitignore.

shiftkey:
Thanks @Zoltu!

PR #673, opened 2015-01-11
Adds TreeClient.GetRecursive method.

Zoltu:
Initially I implemented this as an optional boolean parameter on `TreeClient.Get` that would append `?recursive=1` to the end of the URI.  Unfortunately, code analysis complained about that (optional parameters) so I went with a set of new overloads instead.

Let me know if a different approach is preferred for adding optional query string parameters to API methods or if this is the preferred solution.  The problem with this solution is that it doesn't scale out well with multiple optional parameters.

### DIFF
--- a/Octokit/Helpers/ApiUrls.cs
+++ b/Octokit/Helpers/ApiUrls.cs
@@ -1034,6 +1034,18 @@ namespace Octokit
         }
 
         /// <summary>
+        /// Returns the <see cref="Uri"/> for the specified tree.
+        /// </summary>
+        /// <param name="owner">The owner of the repository</param>
+        /// <param name="name">The name of the repository</param>
+        /// <param name="reference">The tree reference (SHA)</param>
+        /// <returns></returns>
+        public static Uri TreeRecursive(string owner, string name, string reference)
+        {
+            return "repos/{0}/{1}/git/trees/{2}?recursive=1".FormatUri(owner, name, reference);
+        }
+
+        /// <summary>
         /// returns the <see cref="Uri"/> for org teams 
         /// use for both Get and Create methods
         /// </summary>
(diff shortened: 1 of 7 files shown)


Zoltu:
Investigating test failures.

Zoltu:
Test failures fixed.

Zoltu:
Fixing missing observable methods (and some bad tests)...

Octokit/Helpers/ApiUrls.cs
```diff
@@ -1027,6 +1027,7 @@ public static Uri Tree(string owner, string name)
         /// <param name="owner">The owner of the repository</param>
         /// <param name="name">The name of the repository</param>
         /// <param name="reference">The tree reference (SHA)</param>
+        /// <param name="recursive">Whether to include recursive subtrees in the result</param>
```
shiftkey: Remove this line

Octokit/Helpers/ApiUrls.cs
```diff
@@ -1034,6 +1035,19 @@ public static Uri Tree(string owner, string name, string reference)
         }
 
         /// <summary>
+        /// Returns the <see cref="Uri"/> for the specified tree.
+        /// </summary>
+        /// <param name="owner">The owner of the repository</param>
+        /// <param name="name">The name of the repository</param>
+        /// <param name="reference">The tree reference (SHA)</param>
+        /// <param name="recursive">Whether to include recursive subtrees in the result</param>
```
shiftkey: Remove this line

Zoltu: Fixed.

Octokit/Helpers/ApiUrls.cs
Zoltu: Fixed.

shiftkey:
Thanks @Zoltu!

PR #674, opened 2015-01-11
Adds truncated attribute to TreeResponse.

Zoltu:
See https://developer.github.com/v3/git/trees/#get-a-tree for details on the truncated attribute.

I am operating under the assumption that if a property is missing from the json object then the property will be set to its default value (false in this case).

I'm not sure how to test this.  If someone can point me at some existing tests that validate the json deserializer behaves correctly in certain scenarios (like when the property is missing) I can add some test coverage.  Though, if using an off-the-shelf deserializer then perhaps that wouldn't be testing anything meaningful in the library.

### DIFF
--- a/Octokit.Tests/SimpleJsonSerializerTests.cs
+++ b/Octokit.Tests/SimpleJsonSerializerTests.cs
@@ -203,6 +203,19 @@ namespace Octokit.Tests
 
                 Assert.Equal("blah", result.Links);
             }
+
+            [Fact]
+            public void DefaultsMissingParameters()
+            {
+                const string json = @"{""private"":true}";
+
+                var sample = new SimpleJsonSerializer().Deserialize<Sample>(json);
+
+                Assert.Equal(0, sample.Id);
+                Assert.Equal(null, sample.FirstName);
+                Assert.False(sample.IsSomething);
+                Assert.True(sample.Private);
+            }
         }
 
         public class Sample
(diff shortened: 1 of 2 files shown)


shiftkey:
> If someone can point me at some existing tests that validate the json deserializer behaves correctly in certain scenarios (like when the property is missing) I can add some test coverage.

We're using SimpleJSON for the serializer - we have some tests [here](https://github.com/octokit/octokit.net/blob/master/Octokit.Tests/SimpleJsonSerializerTests.cs) for the related features that we need.

I believe the `truncated` property should always be present when deserializing the JSON, so unless you can think of some [integration tests](https://github.com/octokit/octokit.net/blob/master/Octokit.Tests.Integration/Clients/TreeClientTests.cs) to add I wouldn't stress about it too much.

Zoltu:
I added a json parser test just to prove what happens if an attribute is missing.

I don't know of a public repo returns a truncated tree so that is difficult to integration test.  Even with the recursive flag set to 1, getting the full tree for octokit still doesn't truncate.  Also, such an integration test would be brittle if GitHub changes the truncation level.  Given all of that, I'm going to forgo an integration test unless someone really wants it and can recommend a repo to test against or a way to force truncation.

shiftkey:
> I don't know of a public repo returns a truncated tree so that is difficult to integration test.

Don't worry about it. Happy to leave this as-is.

shiftkey:
Thanks!

PR #676, opened 2015-01-17
Removes doc comment typos.

Zoltu:
Not sure how those 's' characters got there, but they would show up in the VS tooltip.


Haacked:
Thanks!

Issue #677, opened 2015-01-18
Add default constructors to response models.


Zoltu:
As a user, I would like to be able to easily test my code that utilizes Octokit.  One of the things that I may want to do is mock library calls (particularly calls that normally would hit network/disk) to return mocked objects.  This allows me to unit test my business logic without needing to deeply understand the internal workings of Octokit.

A recent change (https://github.com/octokit/octokit.net/commit/21ca33817d96725391c7705f62d2eeb4de298531) altered many of the models to be read-only.  The problem here is that if I mock a call like `IReferencesClient.Get(...)`, I cannot easily construct a `Reference` to return from that mocked function.  In theory, I could derive a class from Reference and create a constructor that uses the protected setters to populate the objects, but I would have to do that for every model.  I can't use a traditional mocking library because the getters are not virtual.

Suggestions (in order of least work to most work):

1. Make the properties virtual.  This will at least allow me to leverage a mocking library.
1. Provide a constructor that can populate the object.
1. Don't directly use the models, instead use interfaces everywhere.  The models as they are now would just become private implementations of those interfaces.

Haacked:
These models are basically data objects so I don't like 1 or 3 for that reason. They aren't meant to have any behavior.

I like 2. Each object should have a default constructor (so we can deserialize them) and a constructor that fully populates it.

Would you be interested in submitting a PR to introduce those ctors? :smile:

Zoltu:
I'll get a PR put together for this.

PR #679, opened 2015-01-18
Creates constructors for all Models.Response.

Zoltu:
Resolves https://github.com/octokit/octokit.net/issues/677.

There are a lot of code analysis warnings about URIs as Strings.  Let me know what the preferred way to resolve this is and I'll apply the fix across the board in this PR.

Also, this doesn't build currently due to the usage of obsolete properties.  An example is User.cs which has an Obsolete GravatarId.  The constructor assigns it, which causes an obsolete usage warning at compile time.  What is the preferred way to deal with this situation?

### DIFF
--- a/Octokit.Tests.Integration/Clients/ReleasesClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/ReleasesClientTests.cs
@@ -146,10 +146,7 @@ public class ReleasesClientTests
 
             var stream = Helper.LoadFixture("hello-world.txt");
 
-            var newAsset = new ReleaseAssetUpload
-            {
-                ContentType = "text/plain", FileName = "hello-world.txt", RawData = stream
-            };
+            var newAsset = new ReleaseAssetUpload("hello-world.txt", "text/plain", stream, null);
 
             var result = await _releaseClient.UploadAsset(release, newAsset);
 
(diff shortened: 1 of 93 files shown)


Haacked:
> There are a lot of code analysis warnings about URIs as Strings. Let me know what the preferred way to resolve this is and I'll apply the fix across the board in this PR.

My inclination is to pass in `Uri` and call `ToString()` on them when setting the property. These constructors aren't used in normal usage. They're primarily there for testing and in the rare case someone needs to override a behavior and explicitly create these types. In those cases, why not require a URI. @shiftkey any thoughts?

I think we should get rid of the obsolete properties in this PR. It's about time. :smile:

shiftkey:
> In those cases, why not require a URI. @shiftkey any thoughts?

The rule is basically:

> A type declares a method with a string parameter whose name contains "uri", "Uri", "urn", "Urn", "url", or "Url" and the type does not declare a corresponding overload that takes a System.Uri parameter.

My first instinct was "Could we mute this rule?" And yes, adding this element to `Octokit.ruleset` kills off these warnings: `<Rule Id="CA1054" Action="None" />`

So that's what I'd favour for this - as long as the `ctor`s are `protected`, whatever.  

> I think we should get rid of the obsolete properties in this PR. 

:thumbsup:

Haacked:
> So that's what I'd favour for this - as long as the ctors are protected, whatever.

Most of the ctors are `public`. The point of them is to allow people to create these instances without having to subclass them.

Zoltu:
I removed the obsolete properties.

For the warning about URL as a string, I went with the rule suggested by @shiftkey since there wasn't an agreement in this thread and it was by far the easier solution.

I moved all of the constructors to above the properties per http://www.stylecop.com/docs/SA1201.html since there doesn't appear to be a hard and fast standard and a good portion of the response models that already had constructors followed this style.

I created default constructors for all of the response models since it appears SimpleJson needs them to work correctly (tests failed without).

Finally, I changed a couple (maybe it ended up only being one) response model from public setter to private setter to match all of the others.

Haacked:
This is really great! It's tedious work and very much appreciated!


selfie-0

Haacked:
BTW, now that we have proper ctors, we don't need all the model properties to have `protected` setters. Probably best to make them `private`.

Issue #655, opened 2015-01-02
Consistent way to create and teardown test repositories


shiftkey:
We have a consistent way to create a test repository for the purpose of integration tests:

```
var repoName = Helper.MakeNameWithTimestamp("source-repo");
var repository = await client.Repository.Create(new NewRepository { Name = repoName, AutoInit = true });
```

And a couple of different ways to cleanup this repository:

```
finally
{
    if (repository != null)
       client.Repository.Delete(repository.Owner.Login, repository.Name);
}
```
OR
```
finally
{
    Helper.DeleteRepo(createdRepository);
}
```
OR 
```
finally
{
    Helper.DeleteRepo("owner", "name");
}
```

We also have a couple of places where we use `IDisposable` to signal to xUnit to run some disposal code outside the scope of the test.

So basically, everything is terrible.

There's a couple of constraints to be wary of here:

 - the repository should be deleted when the test completes (or an exception occurs)
 - `IDisposable.Dispose` isn't `await`-able :cry:
 - The [Delete Repository](https://developer.github.com/v3/repos/#delete-a-repository) API is a network call, we should probably wait for it to complete.

Ideally, I'd like some sort of wrapper which we can use:

```
using (var repository = gitClient.CreateTestRepository("source-repo"))
{


}
```

Where `CreateTestRepository` is an extension method that takes care of generating the unique name for the repository and calling the necessary APIs. There might be some ickyness inside `Dispose` to workaround the blocking/non-blocking problem, but hopefully this makes it easier to scaffold new integration tests...

Haacked:
I like the `using` approach. :+1:

willsb:
What's the current status of this one? Are any bits of [this closed PR](https://github.com/octokit/octokit.net/pull/680) relevant? May I start working on it?

Haacked:
Nobody's working on this currently. Feel free.

devkhan:
Should be closed?

PR #680, opened 2015-01-19
[WIP] for issue #655 Delete repository

rajeshgupthar:
Introduced some plumbing code to be able to create disposable repositories while doing integration testing.
Do let me know if I am on the right track, before I get too far :)


Octokit.Tests.Integration/GithubClientExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+{
+    public static class GithubClientExtensions
+    {
+        public static DisposableRepository2 CreateTestRepository(this IGitHubClient client, NewRepository newRepository)
+        {
+            return ((TestRepositoriesClient)client.Repository).CreateTestRepository(newRepository);
+        }
+    }
+
+    public class DisposableRepository2 : Repository, IDisposable
```
Haacked: Why is this named `DisposableRepository2` and not just `DisposableRepository`?

Octokit.Tests.Integration/GithubClientExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+    {
+        public TestRepositoriesClient(IApiConnection apiConnection) : base (apiConnection)
+        {
+
+        }
+
+        public DisposableRepository2 CreateTestRepository(NewRepository newRepo)
+        {
+            var repoTask = base.Create(newRepo);
+            repoTask.Wait();
```
Haacked: Not a fan of calling `Wait` here. I'd rather this method return `Task<DisposableRepository>`. Then users can just `await` the call to this.

Octokit.Tests.Integration/GithubClientExtensions.cs
rajeshgupthar: I agree, it should be DisposableRepository. I had earlier created another class by the name DisposableRepository hence this class got its name as DisposableRepository2.

Octokit.Tests.Integration/GithubClientExtensions.cs
rajeshgupthar: But this is the method that will be used within the using statement, which is not awaitable? Did I go wrong somewhere?

Haacked: > But this is the method that will be used within the using statement, which is not awaitable?

Says who?! :smile:

It's perfectly fine to have an `await` within a using block. http://stackoverflow.com/questions/16566547/do-using-statements-and-await-keywords-play-nicely-in-c-sharp

rajeshgupthar: Oops, my bad. Implemented the methods as Task<DisposableRepository> as suggested.

Octokit.Tests.Integration/GithubClientExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Octokit.Tests.Integration
+{
+    public static class GithubClientExtensions
+    {
+        public async static Task<DisposableRepository> CreateDisposableRepository(this IGitHubClient client, NewRepository newRepository)
+        {
+            return await ((TestRepositoriesClient)client.Repository).CreateDisposableRepository(newRepository);
```
shiftkey: Rather than introducing extra abstractions for GitHubClient and RepositoryClient, you could do this inside `CreateDisposableRepository`:

```
var repository = await client.Repository.Create(newRepository);
return DisposableRepository.InitFromRepository(repository);
```

Octokit.Tests.Integration/GithubClientExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+        }
+    }
+
+    public class DisposableRepository : Repository, IDisposable
```
shiftkey: I'd love to not need this class, but instead have tests access the repository directly, e.g:

```
public class DisposableRepository : IDisposable
{
    public Repository Repository { get; private set; }
    public static DisposableRepository Init(Repository repository)
    { 
         var disp = new DisposableRepository() { Repository = repository };
         return disp;
    }
    /* IDisposable code here */
}
```

But that means we'd need to change the tests to read that inner property. I'm somewhat neutral on which approach is better, but the reflection code in here concerns me a bit.

shiftkey:
@rajeshgupthar thanks for starting on this. I had some hacky code lying around to tackle this same problem, so I added some thoughts about the approach I went for versus where you went into the mix.

I think you're on the right track :thumbsup:

rajeshgupthar:
Many test classes have the setup and tear down in constructor and the Dispose method. I have left alone them as it is and just fixed the try/finally in RepositoriesClientTests.cs. I hope this should be fine.
The doubt that I have is, if all tests disposes off the created repository properly, then how come there are some repositories left out after running the integration tests?

rajeshgupthar:
@shiftkey What concerns you with the reflection code in DisposableRepository class? Do you find it buggy? or Am i breaking any rules/conventions by having it there?
Apart from this, should I consider this pull request as closed?

Octokit.Tests.Integration/Helpers/DisposableRepository.cs
```diff
@@ -0,0 +1,27 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Octokit.Tests.Integration.Helpers
+{
+    public class DisposableRepository : Repository, IDisposable
+    {
+        public static DisposableRepository InitFromRepository(Repository repository)
```
shiftkey: Rename this to `Initialize` - I think there's enough detail from the API surface to indicate we're working with repositories here.

shiftkey:
Apologies for not getting back to this earlier:

> What concerns you with the reflection code in DisposableRepository class? Do you find it buggy? or Am i breaking any rules/conventions by having it there?

In general, I favour composition over inheritance when it comes to situations where you're working with test doubles. It's purely a personal preference in this case - if you were to instantiate the `Repository` object and keep it around as a property on the `DisposableRepository` class, you'd achieve the same effect.

I'll let @haacked chime in here about how he feels about this PR.

Haacked:
I share a similar concern. I worry it could end up being fragile. For example, are we sure setting all those repositories matches the exact state a real `Repository` would be in when initialized from a path? Can we guarantee that for all time?

I like the idea of a `RepositoryUnitOfWork` or `RepositoryContext` class (for lack of a better name) or even a `RepositoryScope` class. That class is disposable and exposes a vanilla `Repository` property. Pretty much what @shiftkey suggested.

shiftkey:
After further reflection on this, and @kristianhald showing me some tricks using xUnit test fixtures to abstract this away in #776, I need to gather my thoughts and update the original issue again.

Thanks for the work you've put in here @rajeshgupthar, and apologies for not being clear on what I wanted...

PR #681, opened 2015-01-20
Changed Gist Comment Client Id param from int to string

thedillonb:
Gists & their comments all have string Ids. Changed the GistCommentClient to take a string instead of an int.

e.g: https://api.github.com/gists/edaab9be9eaa9bf1ee5f/comments


shiftkey:
Nice spot!

shiftkey:
We're already planning other breaking changes for this release, so I'm happy to take this in and add it to the release notes.

Issue #686, opened 2015-01-27
List all public repositories


shiftkey:
Reference: https://developer.github.com/v3/repos/#list-all-public-repositories 

- [x] `GetAllPublic()`
- [x]  `GetAllPublic(int since)`

rms81:
Hi,

Do you mind that I try to implement this one?

shiftkey:
![](http://i.imgur.com/fwRaskw.jpg)

rms81:
Pull request created.

alfhenrik:
Does the `GetAllPublic(int since)` still need implementing, or is there still something to figure out in regards to Pagination?

shiftkey:
@alfhenrik the endpoint which takes a `since` doesn't exist, so I've left this open.

Pagination is also different for the public repository APIs:

```
Link: <https://api.github.com/repositories?since=364>; rel="next"
```

Instead of the default:

```
Link: <https://api.github.com/resource?page=2>; rel="next",
      <https://api.github.com/resource?page=5>; rel="last"
```

But we can deal with that in #760 or :boot: to a subsequent PR.

shiftkey:
Reopening this one as the work done in #774 neglected to catch a scenario with pagination that was specific to this API:

 - the `Link` has a `since` parameter older than the specified id
 - when invoking the next link, `GetAll` will overwrite the parameter with the original id
 - forever paging :sob:

If you revert 26a8bf0e818ad8a620987e13d5ed7b1b8bacfa01 you'll get this PR back at least.

shiftkey:
Shipped in 0.11.0

PR #691, opened 2015-02-01
Fix for issue #686 List all public repositories

rms81:
Added the functionality and tests. Not sure if it can be used as is though.
The request takes to long to return. About 10 minutes for about 65K repositories in my tests.
Maybe this could be done passing the pagination ability to the caller of the method.

### DIFF
--- a/Octokit.Tests.Integration/Clients/RepositoriesClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/RepositoriesClientTests.cs
@@ -534,9 +534,10 @@ public class RepositoriesClientTests
         }
     }
 
+    
     public class TheGetAllPublicMethod
     {
-        [IntegrationTest]
+        [IntegrationTest(Skip = "Takes too long to run.")]
         public async Task ReturnsAllPublicRepositories()
         {
             var github = Helper.GetAuthenticatedClient();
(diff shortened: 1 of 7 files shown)


Octokit.Tests.Integration/Clients/RepositoriesClientTests.cs
```diff
@@ -534,6 +534,19 @@ public class TheGetMethod
         }
     }
 
+    public class TheGetAllPublicMethod
+    {
+        [IntegrationTest]
```
shiftkey: If you could throw a `Skip` on this test, it'd be greatly appreciated - this test is going to run for a *long, long, long* time

shiftkey:
>  Maybe this could be done passing the pagination ability to the caller of the method.

This is something I need to think about again. I should open a discussion somewhere on this...

shiftkey:
This looks great! :metal: If you can mute that slow test I'm happy to merge this in!

Octokit.Tests.Integration/Clients/RepositoriesClientTests.cs
rms81: Of course. I will do that.

rms81:
Done

PR #701, opened 2015-02-11
removed WatchedCount property from Repository, resolves #699

### LINKED ISSUE #699: WatchersCount on repository incorrect
It looks like the WatchersCount property is set incorrectly on the repository object, it's always the same as the StargazersCount property. For example, it is showing as 605 for this repo when it should be 80. Will submit a PR for this when I have time :).

DaveWM:


Haacked:
I approve this message. We need to make sure we account for it in the release notes.

shiftkey:
Once I get #702 out the door, I'll pull in a bunch of these fixes and do a follow-up release.

PR #702, opened 2015-02-16
0.7.0 - All Torn Down

shiftkey:
- [x] version bump
- [x] release notes
- ~~[ ] :thumbsup: and merged #712~~ :boot:
- ~~[ ] update release notes to add #712~~ :boot:
- [x] drop pre-release workarounds in build scripts
- [x] update timestamp
- [x] :shipit:
- [x] tag it and merge it

### DIFF
--- a/Octokit/Http/JsonHttpPipeline.cs
+++ b/Octokit/Http/JsonHttpPipeline.cs
@@ -34,7 +34,7 @@ namespace Octokit.Internal
             if (!request.Headers.ContainsKey("Accept"))
             {
 
-                request.Headers["Accept"] = String.Join(",", new[]{v3ApiVersion, moondragonPreviewApiVersion}) ;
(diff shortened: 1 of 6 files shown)


Octokit/Http/JsonHttpPipeline.cs
```diff
@@ -30,7 +33,8 @@ public void SerializeRequest(IRequest request)
 
             if (!request.Headers.ContainsKey("Accept"))
             {
-                request.Headers["Accept"] = "application/vnd.github.v3+json; charset=utf-8";
+
+                request.Headers["Accept"] = String.Join(",", moondragonPreviewApiVersion, v3ApiVersion);
```
shiftkey: @joshvera my testing of this (and a part of the HTTP spec I neglected to recall) indicates that the first match of an `Accept` value is the one the server will use. In this case, we'd need to put the preview API first.

PR #705, opened 2015-02-17
Add codeseach params

### LINKED ISSUE #418: SearchCode should mandate a repository to query on
We need to constrain our SearchCode query to a specific repo:

```
[Fact]
public async Task SearchForFunctionInCode()
{
    var request = new SearchCodeRequest("addClass");
    var repos = await _gitHubClient.Search.SearchCode(request);

    Assert.NotEmpty(repos.Items);
}
```

https://api.github.com/search/code?q=addClass+repo:jquery/jquery returns `200`
https://api.github.com/search/code?q=addClass returns `422`

kfrancis:
Fixes #418

### DIFF
--- a/Octokit/Models/Request/SearchCodeRequest.cs
+++ b/Octokit/Models/Request/SearchCodeRequest.cs
@@ -18,10 +18,13 @@ namespace Octokit
     {
         public SearchCodeRequest(string term) : base(term) { }
 
-        public SearchCodeRequest(string term, string repo)
+        public SearchCodeRequest(string term, string owner, string name)
             : this(term)
         {
-            this.Repo = repo;
+            Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
+            Ensure.ArgumentNotNullOrEmptyString(name, "name");
+
+            this.Repo = string.Format(CultureInfo.InvariantCulture, "{0}/{1}", owner, name);
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


Octokit/Models/Request/SearchCodeRequest.cs
```diff
@@ -18,6 +18,12 @@ public class SearchCodeRequest : BaseSearchRequest
     {
         public SearchCodeRequest(string term) : base(term) { }
 
+        public SearchCodeRequest(string term, string repo)
```
Haacked: This is a little confusing. I assume `repo` here should actually be the repo name and repo owner. For example, `octokit/octokit.net` as opposed to just `octokit.net`.

I think we should probably use the pattern we use elsewhere and have separate parameters for the owner and the name and then concatenate it ourselves.

```csharp
public SearchCodeRequest(string term, string owner, string name)...
```

/cc @shiftkey for thoughts in case I'm missing something.

shiftkey: Yes, this should include the `owner` property.

For example, https://api.github.com/search/code?q=addClass+repo:jquery returns a validation error. But https://api.github.com/search/code?q=addClass+repo:jquery/jquery returns an array of results.

kfrancis:
Thanks for the feedback @Haacked @shiftkey

Haacked:
selfie-1

PR #712, opened 2015-02-23
Added User.Type property

shiftkey:
I need to put this in for a couple of upcoming API changes, and I think it was added since the Account/User/Organization APIs were first ported over.

~~We have an `AccountType` enum which we use in the search APIs - I didn't want to couple this to that, but perhaps I need to.~~ done and done

- [ ] rename the properties
- [ ] ensure response isn't set (Nullable?) if value doesn't exist in JSON dictionary

### DIFF
--- a/Octokit.Tests.Integration/Clients/UsersClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/UsersClientTests.cs
@@ -18,7 +18,20 @@ public class UsersClientTests
             var user = await github.User.Get("tclem");
 
             Assert.Equal("GitHub", user.Company);
+            Assert.Equal(UserType.User, user.Type);
         }
+
+        [IntegrationTest]
+        public async Task ReturnsSpecifiedOrganization()
+        {
+            var github = Helper.GetAuthenticatedClient();
+
+            var user = await github.User.Get("octokit");
+
+            Assert.Null(user.Company);
+            Assert.Equal(UserType.Organization, user.Type);
+        }
+
 
         [IntegrationTest]
         public async Task ReturnsSpecifiedUserUsingAwaitableCredentialProvider()
@@ -50,6 +63,7 @@ public class UsersClientTests
             var user = await github.User.Current();
 
             Assert.Equal(Helper.UserName, user.Login);
+            Assert.Equal(UserType.User, user.Type);
         }
     }
 
--- a/Octokit.Tests.Integration/Clients/UsersClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/UsersClientTests.cs
@@ -32,7 +32,6 @@ public class UsersClientTests
             Assert.Equal(UserType.Organization, user.Type);
         }
 
-
         [IntegrationTest]
         public async Task ReturnsSpecifiedUserUsingAwaitableCredentialProvider()
         {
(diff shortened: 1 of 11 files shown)


Octokit/Models/Response/UserType.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace Octokit
+{
+    public enum UserType
```
shiftkey: I'm not so sold on this name. `AccountType` is used for a certain request, but `Org` doesn't map to `Organization`.

shiftkey:
Ready for review

Octokit/Models/Response/UserType.cs
Haacked: I don't understand. Are you saying we can't use `AccountType` here because the API endpoint that needs `UserType` returns "Organization" and the API endpoint that needs `AccountType` returns "Org"? Ugh.

Octokit.Tests.Integration/Clients/UsersClientTests.cs
```diff
@@ -18,9 +18,22 @@ public class TheGetMethod
...
+        {
+            var github = Helper.GetAuthenticatedClient();
+
+            var user = await github.User.Get("octokit");
+
+            Assert.Null(user.Company);
+            Assert.Equal(UserType.Organization, user.Type);
+        }
+
+
```
Haacked: Extra newline.

Octokit/Models/Response/UserType.cs
shiftkey: Sorry, I should have been clearer here:

 - [This link](https://help.github.com/articles/searching-users/#search-for-users-or-organizations) indicates when searching based on user you use `type:org` in the request parameters.
 - [This link](https://developer.github.com/v3/orgs/#get-an-organization) shows the response you get back has `"type": "Organization"` in the response.

shiftkey:
Did a quick check of the docs to see if there would be any problems with old GHE installs. [11.10.340](https://developer.github.com/enterprise/11.10.340/v3/users/#get-a-single-user) has this value, as does [2.0](https://developer.github.com/enterprise/2.0/v3/users/#get-a-single-user).

Should I worry about a fallback value nonetheless?

Octokit/Models/Response/UserType.cs
Haacked: /khanify api inconsistency

I think `AccountType` is the best name. So here's an idea. Make the search one `AccountSearchType` or `AccountRequestType` and keep this one `AccountType`.

What do you think? My reasoning is I bet people will see this one much more often than they'll use the search api so I want to give this one the nice name.

joshvera:
@shiftkey I think it's alright if we don't. Will this crash if given a JSON dictionary without this key?

Octokit/Models/Response/UserType.cs
shiftkey: :ok_hand:

shiftkey:
@joshvera I can add a test for this. I'm pretty sure (because of how enums work) is that it'll default to `User` (the first value), so that was why I was thinking about making it nullable.

joshvera:
@shiftkey I see, I don’t think we should consider that valid.

shiftkey:
As this is introducing *further* breaking changes (and we're not so urgent on getting this out the door), I'll retarget this against `master` and not hold up shipping the release.

PR #714, opened 2015-02-24
Added Account.Type support

shiftkey:
Supersedes #712

- [x] renamed existing `AccountType` to `AccountSearchType`
- [x] added nullable `Account.Type` property
- [x] added integration tests for user and repository APIs
- [x] update `ctors` to set new field


Haacked:
Flawless victory!

PR #717, opened 2015-02-24
Delete the missing files from the csproj for Octokit-Monotouch

thedillonb:
Octokit-Monotouch won't build with the missing files

### DIFF
--- a/Octokit/Octokit-Monotouch.csproj
+++ b/Octokit/Octokit-Monotouch.csproj
@@ -216,7 +216,6 @@
     <Compile Include="Models\Response\ReleaseAssetUpload.cs" />
     <Compile Include="Models\Request\ReleaseUpdate.cs" />
     <Compile Include="Models\Response\Repository.cs" />
-    <Compile Include="Models\Response\Signature.cs" />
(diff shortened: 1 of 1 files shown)


thedillonb:
With the portable version out, is there even a need for the Monotouch/Monoandroid versions? Attempting to track down my issue with the `Invalid Format` exception I accidently opened the `XamarinStudio` solution. But I'm not sure that it's even needed anymore?

shiftkey:
@thedillonb It's been a while since I opened Xamarin Studio (expired subscriptions also don't help), so I'm not quite sure whether I can/should drop it.

Haacked:
God I would love it if we could drop it. I am not set up to test it and it would be one lest thing to worry about. @thedillonb At this point, I'm ready to trust your call on this. I assume you've tested it out a bit?

shiftkey:
I'm happy to accept this PR, but I'll also open a discussion about whether we need to support separate MonoTouch/MonoDroid projects these days...

shiftkey:
> God I would love it if we could drop it. 

https://github.com/octokit/octokit.net/issues/719

PR #718, opened 2015-02-24
actually fix regression with updating issue labels

### LINKED ISSUE #715: IssueUpdate should allow `null` to be sent
Seems to have been a regression in v0.7

See note https://github.com/octokit/octokit.net/pull/561#issuecomment-75678180

shiftkey:
Fixes #715 

When you update an issue, the value in `Labels` is important. From the [docs](https://developer.github.com/v3/issues/#edit-an-issue):

> Labels to associate with this issue. Pass one or more Labels to *replace* the set of Labels on this Issue. Send an empty array (`[ ]`) to clear all Labels from the Issue. 

There's three scenarios we need to support here, and I think this is the best compromise:

 - [x] you start with `.ToUpdate()` to setup your update request - this gives you a `null` collection of branches, which means you're not changing anything server-side.
 - [x] calling `AddLabel` manages the internal state of the list - and give you a collection which will replace whatever is currently there
 - [x] calling `ClearLabels` should give you an empty collection - if this is used, it'll drop whatever labels you have defined on the server.
 - [x] write some docs

How does that feel @thedillonb @haacked?

### DIFF
--- a/Octokit.Tests.Integration/Clients/IssuesClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/IssuesClientTests.cs
@@ -273,6 +273,55 @@ public class IssuesClientTests : IDisposable
         Assert.Null(updatedIssue.Milestone);
     }
 
+    [IntegrationTest]
+    public async Task DoesNotChangeLabelsByDefault()
+    {
+        var owner = _repository.Owner.Login;
+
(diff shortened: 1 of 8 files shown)


thedillonb:
@shiftkey, sounds good to me!

shiftkey:
Ready for review

Haacked:
Nice!

PR #711, opened 2015-02-19
Support the correct spelling of event types

janovesk:
Let's use the official spelling of [event types](https://developer.github.com/v3/issues/events/) to avoid blowing up on deserialization of events. 

Fixes https://github.com/octokit/octokit.net/issues/707

PR #727, opened 2015-02-25
Support the correct spelling of event types, take two.

janovesk:
[Pull request #711](https://github.com/octokit/octokit.net/pull/711) was created under the assumption that the Parameter attribute worked on both serialization and deserialization of json. Unfortunately that turned out not to be the case. It's only supported under serialization. 

Let's try again! :)


shiftkey:
cc @haacked as he's more familiar with the serialization rules than I am

Haacked:
selfie-0

PR #730, opened 2015-02-27
added support for ref value in get contents api

goalie7960:
as far as i could tell, it was impossible to specify the ref parameter on the get contents api method https://developer.github.com/v3/repos/contents/#get-content

### DIFF
--- a/Octokit.Reactive/Clients/IObservableRepositoryContentsClient.cs
+++ b/Octokit.Reactive/Clients/IObservableRepositoryContentsClient.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Reactive;
 
 namespace Octokit.Reactive
@@ -76,6 +77,22 @@ namespace Octokit.Reactive
         /// A collection of <see cref="RepositoryContent"/> representing the content at the specified path
         /// </returns>
         IObservable<RepositoryContent> GetAllContents(string owner, string name, string path);
+
+        /// <summary>
+        /// Returns the contents of a file or directory in a repository.
+        /// </summary>
+        /// <remarks>
+        /// If given a path to a single file, this method returns a collection containing only that file.
+        /// See the <a href="https://developer.github.com/v3/repos/contents/#get-contents">API documentation</a> for more information.
+        /// </remarks>
+        /// <param name="owner">The owner of the repository</param>
+        /// <param name="name">The name of the repository</param>
+        /// <param name="path">The content path</param>
+        /// <param name="reference">The name of the commit/branch/tag. Default: the repository’s default branch (usually master)</param>
+        /// <returns>
+        /// A collection of <see cref="RepositoryContent"/> representing the content at the specified path
+        /// </returns>
+        IObservable<IReadOnlyList<RepositoryContent>> GetContents(string owner, string name, string path, string reference);
 
         /// <summary>
         /// Creates a commit that creates a new file in a repository.
--- a/Octokit.Reactive/Clients/IObservableRepositoryContentsClient.cs
+++ b/Octokit.Reactive/Clients/IObservableRepositoryContentsClient.cs
@@ -92,7 +92,7 @@ namespace Octokit.Reactive
         /// <returns>
         /// A collection of <see cref="RepositoryContent"/> representing the content at the specified path
         /// </returns>
-        IObservable<IReadOnlyList<RepositoryContent>> GetContents(string owner, string name, string path, string reference);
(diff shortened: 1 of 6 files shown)


Octokit/Clients/RepositoryContentsClient.cs
```diff
@@ -36,6 +37,32 @@ public RepositoryContentsClient(IApiConnection apiConnection) : base(apiConnecti
...
+        /// See the <a href="https://developer.github.com/v3/repos/contents/#get-contents">API documentation</a> for more information.
+        /// </remarks>
+        /// <param name="owner">The owner of the repository</param>
+        /// <param name="name">The name of the repository</param>
+        /// <param name="path">The content path</param>
+        /// <param name="reference">The name of the commit/branch/tag. Default: the repository�s default branch (usually master)</param>
+        /// <returns>
+        /// A collection of <see cref="RepositoryContent"/> representing the content at the specified path
+        /// </returns>
+        public async Task<IReadOnlyList<RepositoryContent>> GetContents(string owner, string name, string path, string reference)
```
shiftkey: Because we maintain an set of `IObservable<T>` APIs that mirror this, the build server is currently complaining.

To fix this, you'll need to add the new method to `IObservableRepositoryContentsClient` and implement it over there too.

shiftkey:
Would it be possible to add some tests for this? I'm not sure how valuable it'd be to target references that might be transient, but look at the integration tests in `RepositoryContentsClientTests` for some inspiration.

Octokit.Reactive/Clients/IObservableRepositoryContentsClient.cs
```diff
@@ -78,6 +79,22 @@ public interface IObservableRepositoryContentsClient
...
+        /// See the <a href="https://developer.github.com/v3/repos/contents/#get-contents">API documentation</a> for more information.
+        /// </remarks>
+        /// <param name="owner">The owner of the repository</param>
+        /// <param name="name">The name of the repository</param>
+        /// <param name="path">The content path</param>
+        /// <param name="reference">The name of the commit/branch/tag. Default: the repository’s default branch (usually master)</param>
+        /// <returns>
+        /// A collection of <see cref="RepositoryContent"/> representing the content at the specified path
+        /// </returns>
+        IObservable<IReadOnlyList<RepositoryContent>> GetContents(string owner, string name, string path, string reference);
```
shiftkey: This should be `IObservable<RepositoryContent>`

Octokit.Reactive/Clients/ObservableRepositoryContentsClient.cs
```diff
@@ -124,6 +125,30 @@ public IObservable<RepositoryContent> GetAllContents(string owner, string name,
...
+        /// A collection of <see cref="RepositoryContent"/> representing the content at the specified path
+        /// </returns>
+        public IObservable<IReadOnlyList<RepositoryContent>> GetContents(string owner, string name, string path, string reference)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
+            Ensure.ArgumentNotNullOrEmptyString(name, "name");
+            Ensure.ArgumentNotNullOrEmptyString(path, "path");
+            Ensure.ArgumentNotNullOrEmptyString(reference, "reference");
+
+            return _client.Repository.Content.GetContents(owner, name, path, reference).ToObservable();
```
shiftkey: Check the method above and steal the `GetAndFlattenAllPages` code to unwrap this list

goalie7960:
going to close this, will reopen when i'm ready

Octokit/Clients/IRepositoryContentsClient.cs
```diff
@@ -24,6 +25,22 @@ public interface IRepositoryContentsClient
...
+        /// See the <a href="https://developer.github.com/v3/repos/contents/#get-contents">API documentation</a> for more information.
+        /// </remarks>
+        /// <param name="owner">The owner of the repository</param>
+        /// <param name="name">The name of the repository</param>
+        /// <param name="path">The content path</param>
+        /// <param name="reference">The name of the commit/branch/tag. Default: the repository�s default branch (usually master)</param>
+        /// <returns>
+        /// A collection of <see cref="RepositoryContent"/> representing the content at the specified path
+        /// </returns>
+        Task<IReadOnlyList<RepositoryContent>> GetContents(string owner, string name, string path, string reference);
```
shiftkey: In #771 we added a rule to the test suite to ensure all methods which returned a `IReadOnlyList<T>` were named with the prefix `GetAll` - if you update this name and the related observable method name you should be :sparkles:

Haacked:
selfie-0

PR #734, opened 2015-03-01
0.7.2 - Can't Handle My Scale

### LINKED ISSUE #733: Repository.GetAllForCurrent is getting all *not* for current
var githubClient = new GitHubClient(new ProductHeaderValue("ListRepoTest"));
    githubClient.Credentials = new Credentials("[TOKEN]");
    var repos = await githubClient.Repository.GetAllForCurrent();

This seems to return all the repos not just the ones for the "current" user. Is that because I'm using a token instead of user/password? If so, how do I provide user/password credentials when I'm also using 2fa?

shiftkey:
Fixes #733 

@distantcam there's one particular *popular* repository you have access to which is causing the excessive paging you're seeing.

I'm removing that feature (as it's part of the preview APIs) and will bring this up with the team in the morning.

- [x] release notes
- [x] integration tests
- [x] :shipit:

### DIFF
--- a/Octokit.Tests/Http/ConnectionTests.cs
+++ b/Octokit.Tests/Http/ConnectionTests.cs
@@ -439,7 +439,7 @@ namespace Octokit.Tests.Http
                 httpClient.Received().Send(Arg.Is<IRequest>(req =>
                     req.BaseAddress == _exampleUri &&
                     req.Body == body &&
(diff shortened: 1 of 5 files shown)


shiftkey:
And this is live https://www.nuget.org/packages/Octokit/0.7.2

distantcam:
Yep, solved my problem thanks.

PR #739, opened 2015-03-03
Added obsolete tag to member in file. 

### LINKED ISSUE #736: Deprecate Repository.SubscriberCount
As per https://github.com/octokit/octokit.net/issues/735#issuecomment-76650035 we should mark this field as `[Obsolete]` and indicate to use the `Watching` client.

basildk:
Added obsolete tag to the SubscribersCount field
Not fully confident in understanding the codebase, so might need som more work. 
If so please let me know.

Fixes #736


Haacked:
Flawless Victory! :sparkles:

selfie-1

Issue #625, opened 2014-12-13
Events pagination


Odonno:
I can see on the GitHub API that the Events API (https://developer.github.com/v3/activity/events/) can be based on pagination (only page property, no per_page property).

Currently, the octokit framework let us get events with a simply collection (no paging system).

Could you fix it ?

dampir:
@ryangribble, @shiftkey could this one be closed as solved because Octokit.NET in one little step from release with pagination support?

shiftkey:
Fixed in https://github.com/octokit/octokit.net/pull/1240 and will be available in the next Octokit release...

Issue #259, opened 2013-12-11
First Class Support for Pagination


shiftkey:
So we have the ability to specify the `page` and `per_page` parameters in the GitHub API, like this:

`$ curl https://api.github.com/user/repos?page=2&per_page=100`

But we don't expose this in Octokit.net

I'd like to take a stab at introducing this for the `RepositoriesClient` and `ObservableRepositoriesClient` as this is likely to be something that other clients are going to be interested in.

I've been pondering on the API, and given we use the action `GetAll` to signify "get all repositories" I think specifying overloads (not optional parameters due to [various](http://haacked.com/archive/2010/08/09/versioning-issues-with-optional-arguments.aspx/) [reasons](http://haacked.com/archive/2010/08/11/more-optional-versioning-fun.aspx/)) is a better approach to this, for those who want to opt-in to the behaviour.

cc @niik @Haacked

M-Zuber:
Completed with [this release](https://github.com/octokit/octokit.net/releases/tag/v0.20.0)

PR #740, opened 2015-03-04
[WIP] implement first-class paging support

shiftkey:
:cactus: :cactus: :cactus: :cactus: :cactus: :cactus: :cactus: :cactus: :cactus: **WORK IN PROGRESS** :cactus: :cactus: :cactus: :cactus: :cactus: :cactus: :cactus: :cactus: :cactus: 
:cactus: :cactus: :cactus: :cactus: :cactus: :cactus: :cactus: :cactus: :cactus:  **DO NOT MERGE** :cactus: :cactus: :cactus::cactus: :cactus: :cactus: :cactus: :cactus: :cactus:  

I'd been noodling around with how to tackle paging in Octokit, and here's my first attempt at a solution.

#### The Problem

By default, if there's new data to fetch, the Octokit client will go and fetch it. Thanks to the [pagination](https://developer.github.com/v3/#pagination) APIs, this is rather easy to do. For my use case, that's been Good Enough. But I'm at the point where I'd like to do things like:

 - adjust the number of results per page
 - stop at a certain page number
 - skip certain pages
 - opt-in to preview API behaviour by overriding accepts header

Some endpoints support other sorts of filtering (for example, [issues search](https://developer.github.com/v3/issues/#list-issues-for-a-repository) has a `since` parameter. I'm not worrying about those here - this approach should work for all endpoints that return arrays of results.

This isn't something I've focused on given I spent most of my days working with desktop apps, but I know other platforms would love having this level of control to conserve data.

And we have #259 and #625 as open issues, so why not do try and resolve this?

#### The Approach

You can await any type in C# with the presence of `async`/`await` - and it's pretty easy once you know how to get started. For backwards-compatibility, I didn't want to force you to update your code - if you didn't opt-in to this behaviour, you should get the existing behaviour.

So I started with this ideal API:

```csharp
var repositories = await client.Repository.GetAllForCurrent()
    .WithPageCount(10)
    .SelectPages(10);
 ```

Instead of returning `Task<IReadOnlyList<T>>` here I'm returning `ILazyRequest<T>` - and it's this type that has the extra options necessary. These is how one can tweak the internal state of the request before it executes.

To get the `await` *magic* to work, you need to define a public static method in a static type that takes your type as an extension method and returns a `TaskAwaiter<T>`.

```csharp
public static class CustomAwaiter
{
    public static TaskAwaiter<IReadOnlyList<T>> GetAwaiter<T>(this ILazyRequest<T> request)
    {
        Ensure.ArgumentNotNull(request, "request");

        return request.ToTask();
    }
}
```

This lives under the `Octokit` namespace, so it should Just Work with existing users.

`ToTask()` just defers to the existing behaviour, but this gives me the hook necessary to inject the additional parameters before executing the request.

There's also some enhancements necessary to `IApiPagination` to allow clients to opt-out to paging early. I should be able to extract this dependency from `ApiConnection` by the time this is done, so I'm fine with changing that behaviour around.

This is the API I've settled on:

```csharp
var repositories = await github.Repository.GetAllForCurrent()
    .WithOptions(new ApiOptions { PageSize = 10, PageCount = 1 });
```

#### Limitations

~~I need to think about how to do this with the Observable-based APIs - currently we just cheat and transform the Task-based API, but because we surface an  `IObservable<T>` we'd need to do something similar here. Plus there are extension methods like `Skip` and `Take` available that we should be respecting (and not fetching all the records in the first place). Perhaps this is when the two libraries diverge, implementation-wise :cry:~~ 

**EDIT:** we actually use a different implementation inside the `GetAll`s for the observable clients - `GetAndFlattenAllPages` - so I'm gonna put this question aside for another day.

#### TODO

 - [x] unbreak all the tests
 - [x] resolve muted fx:cop: issues
 - [x] `ILazyRequest` -> `IDeferredRequest`
 - [ ] finish `RepositoriesClient` migration
 - [x] experiment with content responses in Issue Comments
 - ~~[ ] ponder on how to do this with `Octokit.Reactive`~~ :boot: punt to another time

### DIFF
--- a/Octokit/Http/CustomAwaiter.cs
+++ b/Octokit/Http/CustomAwaiter.cs
@@ -1,0 +1,15 @@
+﻿using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+
+namespace Octokit
+{
+    public static class CustomAwaiter
+    {
+        public static TaskAwaiter<IReadOnlyList<T>> GetAwaiter<T>(this ILazyRequest<T> request)
+        {
+            Ensure.ArgumentNotNull(request, "request");
+
+            return request.ToTask();
+        }
+    }
+}
--- a/Octokit/Http/CustomAwaiter.cs
+++ b/Octokit/Http/CustomAwaiter.cs
@@ -9,7 +9,7 @@ namespace Octokit
         {
             Ensure.ArgumentNotNull(request, "request");
 
-            return request.ToTask();
+            return request.ToTask().GetAwaiter();
         }
     }
 }
--- a/Octokit/Http/CustomAwaiter.cs
+++ b/Octokit/Http/CustomAwaiter.cs
@@ -5,7 +5,7 @@ namespace Octokit
 {
     public static class CustomAwaiter
     {
-        public static TaskAwaiter<IReadOnlyList<T>> GetAwaiter<T>(this ILazyRequest<T> request)
+        public static TaskAwaiter<IReadOnlyList<T>> GetAwaiter<T>(this IDeferredRequest<T> request)
         {
             Ensure.ArgumentNotNull(request, "request");
 
--- a/Octokit/Http/IDeferredRequest.cs
+++ b/Octokit/Http/IDeferredRequest.cs
@@ -11,9 +11,9 @@ namespace Octokit
         public string Accepts { get; set; }
     }
 
-    public interface ILazyRequest<T>
+    public interface IDeferredRequest<T>
     {
-        ILazyRequest<T> WithOptions(ApiOptions options);
+        IDeferredRequest<T> WithOptions(ApiOptions options);
 
         Task<IReadOnlyList<T>> ToTask();
     }
--- a/CustomDictionary.xml
+++ b/CustomDictionary.xml
@@ -21,6 +21,7 @@
       <Word>Mergeable</Word>
       <Word>Symlink</Word>
       <Word>Submodule</Word>
+      <Word>Awaiter</Word>
     </Recognized>
   </Words>
   <Acronyms>
--- a/Octokit.Tests.Conventions/SyncObservableClients.cs
+++ b/Octokit.Tests.Conventions/SyncObservableClients.cs
@@ -82,6 +82,7 @@ namespace Octokit.Tests.Conventions
                 case TypeCategory.GenericTask:
                     // single item - Task<TResult> => IObservable<TResult>
                 case TypeCategory.ReadOnlyList:
+                case TypeCategory.LazyReadOnlyList:
                     // list - Task<IReadOnlyList<TResult>> => IObservable<TResult>
                     return typeof(IObservable<>).MakeGenericType(typeInfo.Type);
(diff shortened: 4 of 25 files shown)


shiftkey:
I'll pause here before going further and see if people have feedback on this approach.

Octokit/Http/IDeferredRequest.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Threading.Tasks;
+
+namespace Octokit
+{
+    public class ApiOptions
+    {
+        public int? StartPage { get; set; }
+        public int? PageCount { get; set; }
+        public int? PageSize { get; set; }
+        public string Accepts { get; set; }
```
shiftkey: This might also be how we support different media types in the response without hard-coding :shit: everywhere. Added a task to explore this further.

cc @ammeep @thedillonb

shiftkey:
Went back and forth with @JakeGinnivan about this API stuff, and he pointed out that this change will make the task returned from `DeferredRequest` cold. And that's going against the [Task-based Asynchronous Pattern](http://www.microsoft.com/en-us/download/confirmation.aspx?id=19957) book which recommends starting the task as early as possible.

> "All tasks returned from TAP methods must be “hot.” **If a TAP method internally uses a Task’s constructor to instantiate the task to be returned, the TAP method must call Start on the Task object prior to returning it.** Consumers of a TAP method may safely assume that the returned task is “hot,” and should not attempt to call Start on any Task returned from a TAP method. Calling Start on a “hot” task will result in an InvalidOperationException (this check is handled automatically by the Task class)."

I don't think this is impossible to do within our current infrastructure, but right now I need to be mindful of this.

Octokit/Http/CustomAwaiter.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+
+namespace Octokit
+{
+    public static class CustomAwaiter
+    {
+        public static TaskAwaiter<IReadOnlyList<T>> GetAwaiter<T>(this IDeferredRequest<T> request)
```
shiftkey: So I could :fire: this awaiter and make the user call `.ToTask()` for these methods.

It's less magical, but takes away the hot/cold issue.

JakeGinnivan: Still a breaking change. The API itself has changed from hot to cold and existing code now needs to .ToTask()

Right?

shiftkey: Yeah, I'm not sure I can magic up a type here that's both Hot and also modifiable using WithOptions.

So I'm weighing up varying degrees of breaking changes to see how they feel.

shiftkey: > The API itself has changed from hot to cold and existing code now needs to .ToTask()

Yes, these API calls will no longer magically work with `await`.

shiftkey:
Been thinking on different approaches. Closing this one for now.

PR #742, opened 2015-03-06
Add filter parameters for RepositoryClient.GetAllForCurrent

shiftkey:
See https://developer.github.com/v3/repos/#list-your-repositories for the details

cc @joshvera @Haacked


joshvera:
:ship:

shiftkey:
@joshvera just going to give this a once-over to ensure it matches up with our other conventions before merging

Issue #738, opened 2015-03-03
missing unsubscribed value from EventInfo enum


erichexter:
Requested value 'unsubscribed' was not found. For an Issue Event.  When calling the following api: client.Issue.Events.GetForRepository(user, repository);

Here is the enum, that is missing the value.
https://github.com/octokit/octokit.net/blob/master/Octokit/Models/Response/EventInfo.cs#L75

Looking at the docs, I don't see unsubscribed as a documented event, but I do see a subscribed event. 
https://developer.github.com/v3/issues/events/

Exception Stack Trace Below.
Description: An unhandled exception occurred during the execution of the current web request. Please review the stack trace for more information about the error and where it originated in the code. 

Exception Details: System.ArgumentException: Requested value 'unsubscribed' was not found.

Source Error: 

Line 93:                         // remove '-' from values coming in to be able to enum utf-8
Line 94:                         stringValue = RemoveHyphenAndUnderscore(stringValue);
Line 95:                         return Enum.Parse(type, stringValue, ignoreCase: true);
Line 96:                     }

shiftkey:
@erichexter perhaps this is something missing from the docs - is this a public repository I could run a test against to confirm this behaviour?

erichexter:
It is a public repository. I got the error running the command on
https://github.com/waffleio/waffle.io


Eric Hexter

blog | http://Hex.LosTechies.com
info | http://www.linkedin.com/in/erichexter

shiftkey:
@erichexter :heart:

shiftkey:
Weird, I couldn't trigger the issue using this test harness (against b3859cf6eecc510d484d7f846e855fd07952a0b0):

```
[IntegrationTest]
public async Task ReturnsACollection()
{
    var github = Helper.GetAuthenticatedClient();
    var events = await github.Activity.Events.GetAllForRepository("waffleio", "waffle.io");
    Assert.NotEmpty(events);
} 
```

@erichexter am I missing something?

erichexter:
So I was using client.Issue not client.Activity 
``` 
client.Issue.Events.GetForRepository(user, repository);
```

shiftkey:
Found it: https://api.github.com/repos/waffleio/waffle.io/issues/events/142230057

darrencamp:
I'm going to have a go at fixing this one.

shiftkey:
@darrencamp thanks!

I'll follow up with the team whether this should be documented, but for now I think we should just Fix The Bug.

[...]

PR #751, opened 2015-03-12
Fix for issue #738

testdarrenc:


shiftkey:
Could you add the `IssueEventTests` file to the `Octokit.Tests-NetCore45` project?

shiftkey:
There's also an integration test you can write around this bug:

```
[IntegrationTest]
public async Task CanDeserializeUnsubscribeEvent()
{
    var client = Helper.GetAuthenticatedClient();
    var issue = await client.Issue.Events.Get("waffleio","waffle.io", 142230057);
    Assert.Equal(EventInfoState.Unsubscribed, issue.Event);
}
```

shiftkey:
Flawless Victory!

PR #758, opened 2015-03-18
Replace qed with AppVeyor in readme

akoeplinger:
AFAIK qed isn't used in this repo anymore.
I also added a link to the AppVeyor account to the build badge which is much more useful than getting the image.


Haacked:
:sparkles: Thanks!

PR #759, opened 2015-03-18
Nuke GetReadme and GetReadmeHtml

### LINKED ISSUE #750: cleanup obsolete methods in IRepositoryClient
`GetReadme` and `GetReadmeHtml` have been marked as obsolete for a while.

Someone can go through and get credit for finally removing them

:fire::fire::fire::fire::fire:

khellang:
Closes #750

### DIFF
--- a/Octokit.Tests.Integration/Clients/RepositoriesClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/RepositoriesClientTests.cs
@@ -575,33 +575,6 @@ public class RepositoriesClientTests
         }
     }
 
-    public class TheGetReadmeMethod
-    {
-        [IntegrationTest]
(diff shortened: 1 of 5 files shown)


Octokit.Tests.Integration/Clients/RepositoriesClientTests.cs
```diff
@@ -575,33 +575,6 @@ public class TheGetAllForOrgMethod
         }
     }
 
-    public class TheGetReadmeMethod
```
khellang: These tests are also present in `RepositoryContentsClientTests` :confused:

shiftkey: That sounds about right, but they're calling the right API there at least...

PR #760, opened 2015-03-19
Pagination Support

shiftkey:
The goal here is to have overloads flowing through Octokit which let the consumer specify the pagination options available.

The current properties supported are:

 - `PerPage` - number of results to return for each page
 - `StartPage` - skip pages by specify this value
 - `PageCount` - number of pages to return

Previous attempts at this PR have meant updating everything, but for now I've just updated `IReleasesClient` as it has one usage. I've added some integration tests which caught an issue with the internals that I need to look at (the `HACK:` commit).

 - [x] rewrite one method to support an `ApiOptions` overload
 - [x] port over existing behaviour for internals of pagination
 - [x] write convention test to ensure all `GetAll*` methods have an overload
 - [x] port a usage over and add integration tests
 - [x] add some new unit tests around this behaviour
 - [x] extract duplicated behaviour for pagination into a separate function
 - [x] clean up implementation so there's less null hacks

### Ship List

 - [x] finish all the xml-docs stuff
 - [x] documentation on how pagination works
 - [x] @haacked is happy
 - [x] someone else is happy
 - [ ] script to generate issues for API migration steps

### DIFF
--- a/Octokit.Tests.Conventions/Exception/ApiOptionsMissingException.cs
+++ b/Octokit.Tests.Conventions/Exception/ApiOptionsMissingException.cs
@@ -21,10 +21,10 @@ namespace Octokit.Tests.Conventions
 
         static string FormatMethod(MethodInfo m)
         {
-            var parametersFormatteds = m.GetParameters()
+            var formattedParameters = m.GetParameters()
                 .Select(p => string.Format("{0} {1}", p.ParameterType.Name, p.Name));
 
-            var parameterList = string.Join(", ", parametersFormatteds);
+            var parameterList = string.Join(", ", formattedParameters);
 
             return string.Format(" - {0}({1})", m.Name, parameterList);
         }
--- a/Octokit.Tests.Conventions/Exception/ApiOptionsMissingException.cs
+++ b/Octokit.Tests.Conventions/Exception/ApiOptionsMissingException.cs
@@ -1,0 +1,32 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+
+namespace Octokit.Tests.Conventions
+{
+    public class ApiOptionsMissingException : Exception
+    {
+        public ApiOptionsMissingException(Type type, IEnumerable<MethodInfo> methods)
+            : base(CreateMessage(type, methods)) { }
+
+        static string CreateMessage(Type type, IEnumerable<MethodInfo> methods)
+        {
+            var methodsFormatted = String.Join("\r\n", methods.Select(FormatMethod));
+            return "Methods found on type {0} require an overload which accepts an parameter of type ApiOptions:\r\n{1}"
+                      .FormatWithNewLine(
+                          type.Name,
+                          methodsFormatted);
+        }
+
+        static string FormatMethod(MethodInfo m)
+        {
+            var parametersFormatteds = m.GetParameters()
+                .Select(p => string.Format("{0} {1}", p.ParameterType.Name, p.Name));
+
+            var parameterList = string.Join(", ", parametersFormatteds);
+
+            return string.Format(" - {0}({1})", m.Name, parameterList);
+        }
+    }
+}
--- a/Octokit.Tests.Conventions/PaginationTests.cs
+++ b/Octokit.Tests.Conventions/PaginationTests.cs
@@ -27,6 +27,26 @@ namespace Octokit.Tests.Conventions
             }
         }
 
+        [Theory]
+        [MemberData("GetClientInterfaces")]
+        public void CheckPaginationGetAllMethodNames(Type clientInterface)
+        {
+            var methodsOrdered = clientInterface.GetMethodsOrdered();
+
+            var methodsThatCanPaginate = methodsOrdered
+                .Where(x => x.ReturnType.GetTypeInfo().TypeCategory == TypeCategory.ReadOnlyList)
+                .Where(x => x.Name.StartsWith("Get"));
+
+            var invalidMethods = methodsThatCanPaginate
+                .Where(x => !x.Name.StartsWith("GetAll"))
+                .ToList();
+
+            if (invalidMethods.Any())
+            {
+                throw new PaginationGetAllMethodNameMismatchException(clientInterface, invalidMethods);
+            }
+        }
+
         static MethodInfo MethodHasAppropriateOverload(MethodInfo method, MethodInfo[] methodsOrdered)
         {
             var parameters = method.GetParametersOrdered();
@@ -69,7 +89,10 @@ namespace Octokit.Tests.Conventions
 
         public static IEnumerable<object[]> GetClientInterfaces()
         {
-            return typeof(IEventsClient).Assembly.ExportedTypes.Where(TypeExtensions.IsClientInterface).Select(type => new[] { type });
+            return typeof(IEventsClient).Assembly.ExportedTypes
+                .Where(TypeExtensions.IsClientInterface)
+                .Where(type => type != typeof(IStatisticsClient))
+                .Select(type => new[] { type });
         }
     }
 }
--- a/Octokit.Tests.Conventions/PaginationTests.cs
+++ b/Octokit.Tests.Conventions/PaginationTests.cs
@@ -1,0 +1,75 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using Xunit;
+
+namespace Octokit.Tests.Conventions
+{
+    public class PaginationTests
+    {
+        [Theory(Skip = "Disable this to run it and find all the places where things break")]
+        [MemberData("GetClientInterfaces")]
+        public void CheckObservableClients(Type clientInterface)
+        {
+            var methodsOrdered = clientInterface.GetMethodsOrdered();
+
+            var methodsWhichCanPaginate = methodsOrdered
+                .Where(x => x.Name.StartsWith("GetAll"));
+
+            var invalidMethods = methodsWhichCanPaginate
+                .Where(method => MethodHasAppropriateOverload(method, methodsOrdered) == null)
+                .ToList();
+
+            if (invalidMethods.Any())
+            {
+                throw new ApiOptionsMissingException(clientInterface, invalidMethods);
+            }
+        }
+
+        static MethodInfo MethodHasAppropriateOverload(MethodInfo method, MethodInfo[] methodsOrdered)
+        {
+            var parameters = method.GetParametersOrdered();
+            var name = method.Name;
+            return methodsOrdered
+                .Where(x => x.Name == name)
+                .FirstOrDefault(x => MethodHasOverloadForApiOptions(x, parameters));
+        }
+
+        static bool MethodHasOverloadForApiOptions(MethodInfo methodInfo, ParameterInfo[] expected)
+        {
+            var actual = methodInfo.GetParameters();
+
+            if (actual.Length != expected.Length + 1)
+            {
+                return false;
+            }
+
+            for (var i = 0; i < expected.Length; i++)
+            {
+                var a = actual.ElementAt(i);
+                var e = expected.ElementAt(i);
+
+                if (a.Name != e.Name)
+                {
+                    return false;
+                }
+
+                if (a.ParameterType != e.ParameterType)
+                {
+                    return false;
+                }
+            }
+
+            var lastParameter = actual.Last();
+
+            return lastParameter.Name == "options"
+                   && lastParameter.ParameterType == typeof(ApiOptions);
+        }
+
+        public static IEnumerable<object[]> GetClientInterfaces()
+        {
+            return typeof(IEventsClient).Assembly.ExportedTypes.Where(TypeExtensions.IsClientInterface).Select(type => new[] { type });
+        }
+    }
+}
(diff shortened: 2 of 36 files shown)


alfhenrik:
While writing some tests for `ObservableRepositoriesClient.GetAllPublic(int since)` I ran into some issues...in `ConnectionExtensions.GetPages` the `resp.HttpResponse.ApiInfo` object is `null`, the same test but calling `ObservableRepositoriesClient.GetAllPublic()` works just fine...is this something that you are aware of (couldn't find any raised issues on the matter)? If not, I can dive into some debugging to try and find the culprit :smile:

shiftkey:
@alfhenrik could it be that the mock setup isn't returning the right object any more? Aside from that I'm not quite sure...

alfhenrik:
No idea :grimacing: will take a look later on tonight and see if I can figure out what is going wrong.

alfhenrik:
Well...colour me :flushed: turns out the issue was the person typing on the keyboard. I hadn't setup the mock object properly. :sob: It's working now that I've got it setup correctly.

naveensrinivasan:
@shiftkey  Are we planning to bring this into the main branch?

shiftkey:
@naveensrinivasan probably not - the branch history makes it look like I was hacking it in (which I was). I'll see if I can rebase it on top of the current `master`, otherwise I might gradually roll it out (use Roslyn to suggest code fixes, rather than fail the build?)

naveensrinivasan:
@shiftkey  Thanks! Paging would be a feature would cause me to go directly to the API because it does not exists.

shiftkey:
@naveensrinivasan yep, totally understandable. I had an idea to make this easier to roll out, now that we're on VS2015. I'll see if I can bash that out so that this can be rolled out incrementally (and others can chime in).

naveensrinivasan:
@shiftkey  That would be awesome! I would certainly try and contribute to it.

dail8859:
Would love to see something to provide pagination, especially with etags not being supported. Using this to make unauthenticated requests is almost pointless since you can hit your rate limit so quickly. :(

shiftkey:
After #1105 came up again I realised I wanted to revive this PR. So I've dropped every instance of updating the API here for just one endpoint, so the diff is less noisy. There's a broken test here I need to fix, but the diff is mostly ready for review. 

Would love some feedback on this, as I think it's something we could roll out incrementally once we have the core pieces in place.

shiftkey:
Alright, pausing here feels good.

I'll annotate some of the interesting bits to help with the review process.

Octokit/Helpers/Pagination.cs
```diff
@@ -0,0 +1,76 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+
+namespace Octokit
+{
+    internal static class Pagination
```
shiftkey: The contents of this class have fallen out of the refactoring to get pagination into the app.

I found myself duplicating this for the `Task` and `Observable` APIs, so this ended up as the simplest abstraction that could do both.

Perhaps I'll go further with the refactoring to get this feeling more natural, but for now it does the job.

Octokit/Models/Request/ApiOptions.cs
```diff
@@ -0,0 +1,62 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace Octokit
+{
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
+    public class ApiOptions
```
shiftkey: I'm somewhat concerned about this becoming a "bucket" to put all the stuff we don't really know how to put elsewhere, but ultimately this is what the user should pass through to the library to tweak the defaults.

Octokit/Http/ReadOnlyPagedCollection.cs
```diff
@@ -28,7 +28,14 @@ public ReadOnlyPagedCollection(IApiResponse<List<T>> response, Func<Uri, Task<IA
             var nextPageUrl = _info.GetNextPageUrl();
             if (nextPageUrl == null) return null;
 
-            var response = await _nextPageFunc(nextPageUrl).ConfigureAwait(false);
+            var maybeTask = _nextPageFunc(nextPageUrl);
+
+            if (maybeTask == null)
```
shiftkey: I feel so awful about this escape hatch. But making `_nextPageFunc()` currently returns a task to indicate it has results. There's no real equivalent here to indicate we should bail out, so I'm going down this path.

Thankfully tests caught some stupid things I'd assumed, like `null.ConfigureAwait(false)` is a dumb idea.

Octokit.Tests.Conventions/PaginationTests.cs
```diff
@@ -0,0 +1,98 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using Xunit;
+
+namespace Octokit.Tests.Conventions
+{
+    public class PaginationTests
+    {
+        [Theory(Skip = "Disable this to run it and find all the places where things break")]
```
shiftkey: If I get a :thumbsup: and land this PR, I can then run this test and extract tasks to get things update accordingly.

For now, it's not so noisy...

shiftkey: Also, that should say `Enable` instead of `Disable`.

Octokit.Tests.Integration/Clients/ReleasesClientTests.cs
```diff
@@ -75,6 +75,83 @@ public void Dispose()
...
+            var skipStartOptions = new ApiOptions
+            {
+                PageSize = 5,
+                PageCount = 1,
+                StartPage = 2
+            };
+
+            var secondPage = await _releaseClient.GetAll(owner, name, skipStartOptions);
+
+            Assert.NotEqual(firstPage[0].Id, secondPage[0].Id);
```
shiftkey: We haven't implemented `IEqualityComparer` on the response models, so I'm being lazy here.

If anyone else cares about this I'll take it to a separate discussion.

Octokit/Http/ReadOnlyPagedCollection.cs
ryangribble: Dont feel too bad :grinning: It seems like an acceptable compromise in order to implement the "bail out" without having to totally change the entire way Octokit.net has looped through all the paged responses until now...

Octokit.Tests/Clients/ReleasesClientTests.cs
```diff
@@ -21,7 +20,8 @@ public void RequestsCorrectUrl()
 
                 client.Received().GetAll<Release>(Arg.Is<Uri>(u => u.ToString() == "repos/fake/repo/releases"),
                     null,
-                    "application/vnd.github.v3");
+                    AcceptHeaders.StableVersion,
```
ryangribble: Just a note on usage of ```AcceptHeaders``` helper from unit test land...

When I implemented the ```AcceptHeaders``` helper class, it was purposely only used in implementation code, and left any unit tests still needing to "hardcode" the expected headers value.  This was in order to follow the principle that the tests shouldn't just echo something from the implementation side lest there ever be a bug in the implementation side and the test just passes blindly.  Just mentioning it incase you accidentally used it here.  This would be the first case of it's usage in test land so thought it worth pointing out.

I guess I modelled it similarly to how ```RequestsCorrectUrl``` unit tests always hardcode the expected url rather than using the ```ApiUrls``` helper class like the implementation side does.  Doing a quick check now, it seems this rule has been broken for ```ApiUrls``` a couple of times, although only about 6 usages in unit tests, out of a total of 419

Octokit/Clients/AuthorizationsClient.cs
```diff
@@ -36,7 +36,7 @@ public AuthorizationsClient(IApiConnection apiConnection) : base(apiConnection)
         /// <returns>A list of <see cref="Authorization"/>s.</returns>
         public Task<IReadOnlyList<Authorization>> GetAll()
         {
-            return ApiConnection.GetAll<Authorization>(ApiUrls.Authorizations(), null);
+            return ApiConnection.GetAll<Authorization>(ApiUrls.Authorizations(), (IDictionary<string,string>)null);
```
ryangribble: This is pretty :hankey: but I guess managing the overloads of the classes like ```ApiConnection``` gets pretty tricky with yet another option, and no way for the compiler to know when you call ```GetAll(string, null)``` which of the 2 parameter overloads you intended.  Though it would be nice to see whether the overloads can be consolidated to avoid this (or is the intention that things like the accepts header of timeout would shift into this new ```ApiOptions``` parameter?  Even so I guess to maintain backwards compat, most of the overloads would have to stay (or can we just cop a breaking change, given something like implementing pagination support is pretty major, and you arent actually out of beta yet)?

ryangribble:
Well im not sure if i count as your "someone else" being relatively new to contributing on octokit.net, however I've reviewed the changes and have made a few comments inline.

In general as an end user this is great and gives all the expected/desired flexibility in setting page size, number of pages and start page number.

As a contributor I think the implementation makes sense and is done in a way that seems fitting with the style of octokit.net once you dig down into those lower ```ApiConnection``` type areas.

The only annoyance/concern is around the fact that with so many overloads on ```ApiConnection```, it looks like it's going to possibly be a bit trickier when writing unit tests where the ```NSubstitute``` mocks start to need things like ```(Dictionary<string, string>) null``` and so on.  Not sure if there's any way to avoid that though.

From the perspective of "someone else" I give it double thumbs :+1: :+1: 

LGTM

Octokit.Tests.Integration/Clients/ReleasesClientTests.cs
ryangribble: Ive hit this quite a few times when writing tests as well.  Would be great to be able to do ```Equality``` on the response models (and request models for that matter).  If you start a separate discussion I'd be interested to see what the plans are, however for this PR I'd say the test above is fine

Octokit.Tests/Clients/ReleasesClientTests.cs
shiftkey: @ryangribble good point - yeah, I'll undo this change

Octokit/Clients/AuthorizationsClient.cs
shiftkey: I shouldn't *need* to pass `null` in here. I'll investigate further and see what other options I have here.

shiftkey:
> The only annoyance/concern is around the fact that with so many overloads on `ApiConnection`, it looks like it's going to possibly be a bit trickier when writing unit tests where the `NSubstitute` mocks start to need things like (`Dictionary<string, string>`) null and so on. Not sure if there's any way to avoid that though.

If we structure the overloads right (that is, funnel the various `GetAll` overloads into one implementation), perhaps we'll see negligible changes to the tests, e.g.

```csharp
public Task<IReadOnlyList<Release>> GetAll(string owner, string name)
{
    Ensure.ArgumentNotNullOrEmptyString(name, "repository");

    // call the overload which requires ApiOptions
    return GetAll(owner, name, ApiOptions.None);
}
```

And if the actual implementation ends up invoking the `IConnection.Get<T>` with the maximum parameter count, we might have something resembling consistency...

```csharp
var client = Substitute.For<IApiConnection>();

client.Received().GetAll<Release>(Arg.Is<Uri>(u => u.ToString() == "repos/fake/repo/releases"),
    null,
    "application/vnd.github.v3",
    Args.ApiOptions);
```

Octokit.Tests.Integration/Clients/ReleasesClientTests.cs
shiftkey: :boom: https://github.com/octokit/octokit.net/issues/1109

shiftkey:
@ryangribble you've done plenty to contribute to Octokit in recent times, so your feedback here is totally welcome! :heart:

shiftkey:
@ryangribble feeling pretty good about the direction of this looking at the last couple of tweaks in e93dc53e8c234f0c4f44b35bbc9f7de0648b5346

ryangribble:
Looks good!  Takes care of my quibble around unit test parameter checking voodoo

less nulls in one's life are always a good thing :+1:

Octokit.Tests.Conventions/Exception/ApiOptionsMissingException.cs
```diff
@@ -0,0 +1,32 @@
...
+            var methodsFormatted = String.Join("\r\n", methods.Select(FormatMethod));
+            return "Methods found on type {0} require an overload which accepts an parameter of type ApiOptions:\r\n{1}"
+                      .FormatWithNewLine(
+                          type.Name,
+                          methodsFormatted);
+        }
+
+        static string FormatMethod(MethodInfo m)
+        {
+            var parametersFormatteds = m.GetParameters()
```
kiwipom: `formattedParameters` ?

shiftkey:
Just a heads up, this build isn't triggering our AppVeyor CI checks - chasing up with the support team about what's happening.

Octokit/Helpers/Pagination.cs
BigAlInTheHouse: +namespace

shiftkey: @BigAlInTheHouse I'm not sure I follow

shiftkey:
@haacked are you able to have a look over this and see how you feel about this?

Haacked:
:heart_eyes: :+1:

shiftkey:
I'll prep my script to generate the tasks for this, and then merge this in (probably tomorrow) so we can get this rolled out across the entire library in time for the next release...

shiftkey:
And the milestone is opened https://github.com/octokit/octokit.net/milestones/Pagination%20Support so I'm going to merge this in *finally*...

PR #763, opened 2015-03-19
Repository name is required when creating a repo

Haacked:
I was trying to create a repository and I wasn't sure which parameters were required. Following our philosophy of exposing required parameters in the constructor, I change the `NewRepository` object to take in a repository name and to make that property readonly.


Haacked:
:yum:

shiftkey:
Just waiting on that greeeeeen.

PR #766, opened 2015-03-21
Update Followers to follow GetAll* convention

### LINKED ISSUE #764: IFollowersClient doesn't follow naming conventions
We use an unofficial convention of `GetAll*` to indicate when a method returns a collection of results, however these two methods aren't named in this way:

 - `IFollowersClient.GetFollowingForCurrent`
 - `IFollowersClient.GetFollowing`

These should be renamed.

alfhenrik:
Fixes #764


shiftkey:
Thanks!

Issue #698, opened 2015-02-09
Forking a repository to current user


kbilsted:
how do i fork a repository? the #313 supposedly fixed this, but I'm not able to find the api that has been implemented. Neither from the nuget package nor when i download the source for browsning.

johnduhart:
That PR was never completed. If you want you can pick up where I left off at #495

kbilsted:
I would require serious amounts of support to pull that off. I can't even open the csproj under windows 7 which I'm currently using.

Also it's a bit unclear to me what the outstanding issues are and why the PR was closed in the first place

shiftkey:
@kbilsted you'll need VS2013 to open the solution locally

kbilsted:
I have VS2013, it opens a window telling me windows 8.1 is required.

shiftkey:
> I have VS2013, it opens a window telling me windows 8.1 is required.

That's just for some specific projects which use NetCore. Don't worry about it!

kbilsted:
Sure I can ignore that, but when I then build from within VS all hell breaks loose. Hundreds of warnings and errors.

Haacked:
Hmm, try unloading that project.

kristianhald:
I'll look at the pull request to determine if I can make the necessary changes so that it can be merged.

DavidStrickland0:
Anyone get anywhere on this one? Is pr 495 still the best spot to start from?

shiftkey:
@DavidStrickland0 #776 is almost wrapped up and ready to merge, so forking a repository should go out in the next release!

shiftkey:
I'll drop a note in here when it's been released, but #776 has been merged into `master`

DavidStrickland0:
Appreciate the work. Thanks

shiftkey:
@DavidStrickland0 @kbilsted this is now available in v0.12 on NuGet.

kbilsted:
Cheers

PR #495, opened 2014-05-20
 [WIP] Implement Repository Hooks & Repository Forks

johnduhart:
This replaces #313. I've rebased it with master, and cleaned up a few things that were missing on the original PR. This is what's left

- [ ] Write integration tests
 - [ ] For Hooks API
 - [ ] For Forks API

Anything else?

PR #776, opened 2015-03-23
Fork/Hook issue #698

kristianhald:
Hi,

First time doing a pull request, so if I have done anything different than expected, then I would be happy to know.

I have tried continuing on the pr for the issue #495, where it was stated that it was missing integration tests.

Therefore, I would like to have some input on the changes made and if I have missed anything, that needs to be changed.

I have done the following changes to the code:
- [x] Updated the previous code to match current convention rules and changes made in the interfaces
- [x] Implemented integration tests for the forks api
- [x] Implemented integration tests for the hooks api
- [x] Implemented hooks ping command
- [x] Added 'sort' parameter to forks list command
- [x] Merged master into the branch to get the code up to date as the original pull request was old

Tasks missing:
- [x] Waiting for response from GitHub on error reported, when creating a hook and stating it should be 'active' = false, then it is created as 'active' = true. Minor error, but I have a test checking it and I do not like having a test that creates active web hooks (I can change the assert, but as I would like some input on the code, there is some time)


kristianhald:
The pr is ready for review, if possible.

shiftkey:
@kristianhald thanks for reviving this PR, I'll walk through it in the coming day or so and let you know how I feel about it...

kristianhald:
Sounds good. Thank you very much. If you find something I will find time during the week to correct it, so that we can finish it up.

shiftkey:
Just had a look at the merge conflict, and it seems to be related to files being added to both csproj files. You'll pick it up if you try to merge master into your branch.

Octokit.Reactive/Clients/IObservableRepositoriesClient.cs
```diff
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
```
shiftkey: :fire: unnecessary `using` statement

Octokit.Reactive/Clients/IObservableRepositoryForksClient.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using System.Diagnostics.CodeAnalysis;
```
shiftkey: :fire: unnecessary `using` statement.

Octokit.Tests.Integration/Clients/RepositoryHooksClientTests.cs
```diff
@@ -0,0 +1,187 @@
+﻿using System.Linq;
+using System.Threading.Tasks;
+using Octokit.Tests.Integration.fixtures;
+using Xunit;
+
+namespace Octokit.Tests.Integration.Clients
+{
+    public class RepositoryHooksClientTests
+    {
+        [Collection("Repositories Hooks Collection")]
```
shiftkey: :grey_question: we've turned off parallelization here as part of running the integration tests, so these guards aren't necessary. Do you think they're useful to have anyway when running these hooks tests?

Octokit.Tests.Integration/fixtures/RepositoriesHooksCollection.cs
```diff
@@ -0,0 +1,9 @@
+﻿using Xunit;
+
+namespace Octokit.Tests.Integration.fixtures
+{
+    [CollectionDefinition("Repositories Hooks Collection")]
```
shiftkey: Ah, now I'm seeing the other valuable part!

Octokit.Tests.Integration/fixtures/RepositoriesHooksCollection.cs
```diff
@@ -0,0 +1,9 @@
...
+    public class RepositoriesHooksCollection : ICollectionFixture<RepositoriesHooksFixture>
```
shiftkey: cc @haacked as this might be a better way for us to simplify the setup and teardown of repositories in tests that we were throwing around in #655

Octokit.Tests/Clients/RepositoryHooksClientTest.cs
```diff
@@ -0,0 +1,200 @@
...
+
+                connection.Received().GetAll<RepositoryHook>(Arg.Is<Uri>(u => u.ToString() == "repos/fake/repo/hooks"));
+            }
+
+            [Fact]
+            public async Task EnsuresNonNullArguments()
+            {
+                var client = new RepositoriesClient(Substitute.For<IApiConnection>());
+
+                await AssertEx.Throws<ArgumentNullException>(async () => await client.Hooks.GetAll(null, "name"));
```
shiftkey: Swap these out in the new tests for the `Assert.ThrowsAsync<>` that ship with xUnit 2.0 - it'll mean the inner `async/await` goes away as well...

shiftkey: Opened https://github.com/octokit/octokit.net/issues/782 to address this in our other tests.

Octokit/Clients/RepositoriesClient.cs
```diff
@@ -263,6 +265,25 @@ public Task<IReadOnlyList<Repository>> GetAllForOrg(string organization)
         public ICommitStatusClient CommitStatus { get; private set; }
 
         /// <summary>
+        /// A client for GitHub's Repository Hooks API.
+        /// </summary>
+        /// <remarks>See <a href="http://developer.github.com/v3/repos/hooks/">Hooks API documentation</a> for more information.</remarks>
+        public IRepositoryHooksClient Hooks
```
shiftkey: :lipstick: make the contents of the property definition a oneliner

Octokit/Clients/RepositoriesClient.cs
```diff
@@ -263,6 +265,25 @@ public Task<IReadOnlyList<Repository>> GetAllForOrg(string organization)
...
+        {
+            get; private set;
+        }
+
+        /// <summary>
+        /// A client for GitHub's Repository Forks API.
+        /// </summary>
+        /// <remarks>See <a href="http://developer.github.com/v3/repos/forks/">Forks API documentation</a> for more information.</remarks>        
+        public IRepositoryForksClient Forks
```
shiftkey: :lipstick: make the contents of the property definition a oneliner

Octokit/Models/Request/RepositoryForksListRequest.cs
```diff
@@ -0,0 +1,39 @@
...
+            get
+            {
+                return String.Format(CultureInfo.InvariantCulture, "Sort: {0}", Sort);
+            }
+        }
+    }
+
+    public enum Sort
+    {
+        [Parameter(Value = "newest")]
```
shiftkey: Inside our serializer we'll take the C# values and lowercase them, so these overrides aren't really necessary.

Where they do become helpful is where you want to show a completely different value.

Feel free to drop them.

Octokit/Models/Request/RepositoryHookTestRequest.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Octokit
+{
+    public class RepositoryHookTestRequest
```
shiftkey: If we don't need to send any data with the payload for a POST, can we live without this class?

Octokit/Models/Request/RepositoryHooksPingRequest.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Octokit
+{
+    public class RepositoryHooksPingRequest
```
shiftkey: As above, if we've got an empty class here I think we can just do away with it for now.

Octokit/Clients/RepositoryHooksClient.cs
```diff
@@ -0,0 +1,115 @@
...
+        /// <remarks>See <a href="http://developer.github.com/v3/repos/hooks/#test-a-hook">API documentation</a> for more information. 
+        /// This will trigger the hook with the latest push to the current repository if the hook is subscribed to push events. If the hook 
+        /// is not subscribed to push events, the server will respond with 204 but no test POST will be generated.</remarks>
+        /// <returns></returns>
+        public Task Test(string owner, string repositoryName, int hookId)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
+            Ensure.ArgumentNotNullOrEmptyString(repositoryName, "repositoryName");
+
+            return ApiConnection.Post<RepositoryHookTestRequest>(ApiUrls.RepositoryHookTest(owner, repositoryName, hookId), new RepositoryHookTestRequest());
```
shiftkey: Ok, now that I understand the reason for the empty classes let me see if I can suggest an alternative:

Given both these methods should return `204 No Content`, should we be inspecting the response code? With many of our Delete actions we use `Task<HttpStatusCode>` to check that nothing unexpected happened (see `WatchedClient.UnwatchRepo` for an example).

We don't have an equivalent overload for `Post` right now, but given you have two scenarios here which seem to fit it I think it's worth incorporating.

Thoughts?

Octokit/Models/Request/EditRepositoryHook.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Diagnostics;
+using System.Globalization;
+
+namespace Octokit
+{
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
+    public class EditRepositoryHook
+    {
+        [Parameter(Key = "config")]
+        public dynamic Config { get; set; }
```
shiftkey: Could we just use `object` here?

shiftkey: Or could we make it a collection type that helps the user understand what sort of payload they're sending?

Octokit/Models/Request/NewRepositoryHook.cs
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        [Parameter(Key = "name")]
+        public string Name { get; set; }
+
+        /// <summary>
+        /// Is a key value structure determined by the web hook being created
+        /// </summary>
+        [Parameter(Key = "config")]
+        public dynamic Config { get; set; }
```
shiftkey: Same question as above - I think this could be a dictionary of string pairs

shiftkey:
@kristianhald thanks for this (and also thanks to @AndyCross and @johnduhart for getting this started). A few questions about the design and some places to clean up, but this is some really fantastic work already!

Octokit.Tests.Integration/fixtures/RepositoriesHooksCollection.cs
kristianhald: Yeah. There were two reasons for why I want to use collection fixtures. Never tried them before, but I do like them and would like to have them in other test frameworks. :smile:

The first and foremost reason is that for all/alot of my tests I only want to create my repositories and other readonly states once for all tests, so to minimize the times I am hitting the github Api. If I run all the tests in Octokit I will be throttled by github.

The second reason is that it helps encapsulate the creation and disposal logic, though it is a bit indirect and requires people to know them.

I looked at the code and not having read it for 2-3 weeks, I have some suggestions which I would like to run by you guys.

I define the name of the collection as a constant string in the collection class:
```C#
    [CollectionDefinition(CollectionDefinitionName)]
    public class RepositoriesHooksCollection : ICollectionFixture<RepositoriesHooksFixture>
    {
      public const string CollectionDefinitionName = "Repositories Hooks Collection";
    }
```

Which I then use in the 'CollectionDefinition' attribute. On each test class instead of writing the string name of the collection, I use the constant string defined in the collection class:
```C#
        [Collection(RepositoriesHooksCollection.CollectionDefinitionName)]
        public class TheGetAllMethod
        {
            readonly RepositoriesHooksFixture _fixture;

            public TheGetAllMethod(RepositoriesHooksFixture fixture)
            {
```
That should make it a bit less brittle than before.

Furthermore the `RepositoriesHooksCollection` class is very small and very closely tied to the `RepositoriesHooksFixture` class. Would it be an idea to move the collection class into the same file as the fixture class. I am not normally a big fan of having two class inside the same file.

Octokit.Reactive/Clients/IObservableRepositoriesClient.cs
kristianhald: :pensive: Oops. Apologies. Has been removed.

Octokit.Reactive/Clients/IObservableRepositoryForksClient.cs
kristianhald: :pensive: Oops. Apologies. Has been removed.

Octokit.Tests/Clients/RepositoryHooksClientTest.cs
kristianhald: Replaced the assertions in the forks and hooks unit tests.

Octokit/Clients/RepositoriesClient.cs
kristianhald: :one: lined

Octokit/Clients/RepositoriesClient.cs
kristianhald: :one: lined

Octokit/Models/Request/RepositoryForksListRequest.cs
kristianhald: Sounds good. They have been removed.

Octokit/Clients/RepositoryHooksClient.cs
kristianhald: Sounds like a good way to go. Especially if the pattern is already being used for one of the other http methods.

I have implemented a suggestion of the solution. 

When I am done with the rest of the review comments I will push them to the pr branch.

Octokit/Models/Request/NewRepositoryHook.cs
kristianhald: Changed from the dynamic type to Dictionary<string, string>.
I had to place the config in the constructor of the 'EditRepositoryHook' with a default argumentless constructor also.

The reason being the code analysis requires the dictionary not to have a public 'set' and I needed the Config property to be default 'null', because else the edit call would overwrite the config settings even if the user did not provide any.

kristianhald:
Hi @shiftkey thanks for the comments on the code and the design. I have cleaned it up and added the Post method. 
I have added to each of your comments a short description of the change I have made in the code and I have tried keeping the changes in their own commits, so that you can compare the changes.

The 'Changed await AssertEx.Throws' to 'Assert.ThrowsAsync' is wrong. I accidentally also removed the 'await'. They have been added again to the tests in a later commit.

I ran a bit out of time here at the last moment, so I hope everything is ok. Please write back in regards to any changes that needs to be added.

kristianhald:
hi @shiftkey, Just wanted to make certain that the code is ok and not missing anything. I did the merge and solved the conflicts. I think that the merge from master must be done close to the merge to master. If the code changes are ok, I will do another merge.

kbilsted:
What is the status of this PR? I really need the features it adds.

shiftkey:
@kristianhald apologies for the delay - I've just gotten back from BUILD so am digging myself out of inbox hell. I'll add this to the list of things to review today. There's a merge conflict with the branch, but we can address that separate to the changes here.

kbilsted:
@shiftkey   Great stuff. As long as I know this is on your radar, I can wait a few more cycles :-)

Octokit.Tests/Clients/RepositoryForksClientTests.cs
```diff
@@ -0,0 +1,84 @@
...
+                    Arg.Is<Uri>(u => u.ToString() == "repos/fake/repo/forks"),
+                    Arg.Is<Dictionary<string, string>>(d => d["sort"] == "stargazers"));
+            }
+
+            [Fact]
+            public async Task EnsuresNonNullArguments()
+            {
+                var client = new RepositoriesClient(Substitute.For<IApiConnection>());
+
+                await Assert.ThrowsAsync<ArgumentNullException>(async () => await client.Forks.GetAll(null, "name", null));
```
shiftkey: :lipstick: the inner async/await pair can be scrubbed from these files, i.e. `() => client.Forks....`

shiftkey:
@kristianhald looking good, just a bit of cleanup for the tests and merging master in and I'm happy to take this.

Octokit.Tests.Integration/Clients/RepositoryHooksClientTests.cs
```diff
@@ -0,0 +1,222 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Octokit.Tests.Helpers;
```
shiftkey: :fire: unnecessary using statement

Octokit.Tests/Clients/RepositoryForksClientTests.cs
```diff
@@ -0,0 +1,84 @@
...
+
+                connection.Received().Post<Repository>(Arg.Is<Uri>(u => u.ToString() == "repos/fake/repo/forks"), newRepositoryFork);
+            }
+
+            [Fact]
+            public async Task EnsuresNonNullArguments()
+            {
+                var client = new RepositoriesClient(Substitute.For<IApiConnection>());
+
+                await Assert.ThrowsAsync<ArgumentNullException>(async () => await client.Forks.Create(null, "name", new NewRepositoryFork()));
```
shiftkey: :lipstick: drop the inner async/await pair for these

kristianhald:
Thanks for the review @shiftkey. I will look into the issues tonight after work.

Octokit.Tests/Clients/RepositoryForksClientTests.cs
kristianhald: Gone @ :gun: point

Octokit.Tests.Integration/Clients/RepositoryHooksClientTests.cs
kristianhald: :pensive: Has been removed.

Octokit.Tests/Clients/RepositoryForksClientTests.cs
kristianhald: Gone @ :gun: point. They wanted too much to :zzz:
Did the same changes in 'RepositoryHooksClientTests'.

kristianhald:
@shiftkey, review changes have been made, including a merge handling the conflicts.
I hope everything is well else you know how to get a hold of me :email:

Octokit.Tests.Integration/Clients/RepositoryForksClientTests.cs
```diff
@@ -0,0 +1,76 @@
...
+                var github = Helper.GetAuthenticatedClient();
+
+                var forkCreated = await github.Repository.Forks.Create("octokit", "octokit.net", new NewRepositoryFork());
+
+                Assert.NotNull(forkCreated);
+                Assert.Equal(String.Format("{0}/octokit.net", Helper.Credentials.Login), forkCreated.FullName);
+                Assert.Equal(true, forkCreated.Fork);
+            }
+
+            [IntegrationTest]
```
shiftkey: Change this attribute to `[OrganizationTest]` so that it only runs if you have the right environment variables set :metal:

shiftkey:
Just that one test attribute and this is good to merge!

shiftkey:
Thanks for the amazing work @kristianhald

kristianhald:
Thanks and thanks for the merge. :ok_hand:

PR #777, opened 2015-04-03
v0.9.0 - It's the Vibe

shiftkey:
A few breaking changes in here, so why not push this out as soon as possible?

- [x] release notes
- [x] version bump
- [x] tests pass
- [x] :shipit:


shiftkey:
This is up on NuGet :metal:

PR #780, opened 2015-04-16
Add new org api parameters.

shana:
Type and Affiliation/Visibility are exclusive, if one is set the other
can't be, so the enum properties have to be nullable. The default
behaviour when no parameters are passed is whatever the new default
behaviour is on the server for the new api (currently, it is
Visibility = All and Affiliation = Owner | Collaborator).


shiftkey:
Some CI failures in the CI process.

- I *think* the serialization should work without the `_` in `Organization_Member`
- `TypeInfo` is the new `Type` - likely some extension method needs to be added for PCL-based platforms.

Do we need to send any preview flags as part of updating this API?

Haacked:
I'm going to push a new PR with these changes and some of my fixes on top of them.

PR #783, opened 2015-04-20
Add more info to TwoFactorChallengeFailedException

Haacked:
I was running into an issue where i wanted more information from the
TwoFactorChallengeFailedException. It turns out, the exception could
easily provide both the TwoFactorType AND the authentication code
provided. So :boom:!

### DIFF
--- a/Octokit/Exceptions/TwoFactorAuthorizationException.cs
+++ b/Octokit/Exceptions/TwoFactorAuthorizationException.cs
@@ -61,7 +61,7 @@ namespace Octokit
         /// </summary>
         public TwoFactorType TwoFactorType { get; private set; }
 
-        #if !NETFX_CORE
+#if !NETFX_CORE
         /// <summary>
         /// Constructs an instance of TwoFactorRequiredException.
         /// </summary>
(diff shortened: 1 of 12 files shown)


Octokit/Exceptions/TwoFactorChallengeFailedException.cs
```diff
@@ -37,9 +39,16 @@ public override string Message
...
+
+        static TwoFactorType ParseTwoFactorType(ApiException exception)
+        {
+            return exception == null ? TwoFactorType.None : Connection.ParseTwoFactorType(exception.HttpResponse);
+        }
+
 #if !NETFX_CORE
         /// <summary>
-        /// Constructs an instance of TwoFactorChallengeFailedException
+        /// Constructs an instance of TwoFactorRequiredException.
```
shiftkey: :memo: this should remain as `TwoFactorChallengeFailedException`

Octokit/Exceptions/TwoFactorAuthorizationException.cs
```diff
@@ -0,0 +1,115 @@
...
+
+            TwoFactorType = twoFactorType;
+        }
+
+        /// <summary>
+        /// Expected 2FA response type
+        /// </summary>
+        public TwoFactorType TwoFactorType { get; private set; }
+
+        #if !NETFX_CORE
```
shiftkey: :lipstick: indenting

Octokit/Exceptions/TwoFactorChallengeFailedException.cs
Haacked: Nice catch!

Haacked:
:spaghetti:

PR #788, opened 2015-04-22
bring back GetAllPublic repositories

shiftkey:
As part of shipping 0.10 I disabled `GetAllPublic` because I couldn't get to the bottom of why it was infinite-looping. With a bit of sleep, however, I found a workaround.

As part of generating the complete URL for the request, Octokit will take values in it's request collection and overwrite whatever's in the URL. This is fine for the first request, but it was also kicking in for subsequent requests - and getting stuck.

I'm taking away this nice thing for just this endpoint, and providing just the URL it needs to start things off.

The test is still taking too long, so I've disabled it anyway until I can get a chance to finish off #760.

cc @alfhenrik

### DIFF
--- a/Octokit.Reactive/Clients/ObservableRepositoriesClient.cs
+++ b/Octokit.Reactive/Clients/ObservableRepositoriesClient.cs
@@ -103,6 +103,21 @@ namespace Octokit.Reactive
         }
 
         /// <summary>
+        /// Retrieves every public <see cref="Repository"/> since the last repository seen.
+        /// </summary>
+        /// <remarks>
+        /// The default page size on GitHub.com is 30.
+        /// </remarks>
+        /// <param name="request">Search parameters of the last repository seen</param>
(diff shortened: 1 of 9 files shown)


alfhenrik:
:v: :ok_hand:

Octokit.Reactive/Clients/ObservableRepositoriesClient.cs
```diff
@@ -103,6 +103,23 @@ public IObservable<Repository> GetAllPublic()
...
+        /// <remarks>
+        /// The default page size on GitHub.com is 30.
+        /// </remarks>
+        /// <param name="request">Search parameters of the last repository seen</param>
+        /// <returns>A <see cref="IReadOnlyPagedCollection{Repository}"/> of <see cref="Repository"/>.</returns>
+        public IObservable<Repository> GetAllPublic(PublicRepositoryRequest request)
+        {
+            Ensure.ArgumentNotNull(request, "request");
+
+            var url = ApiUrls.AllPublicRepositories(request.Since);
```
Haacked: This is wrong. Please do it another way (TESTING)

PR #789, opened 2015-04-26
Renamed PullRequestFile's *Uri properties to *Url

thedillonb:
Properties not serialized correctly due to invalid names.


Haacked:
:sparkles: Thanks!

Issue #784, opened 2015-04-21
Getting zipball with Octokit is returning me null or invalid zip file


NikolayIT:
I have the following code:

    // ... client = new Client(...);
    // some authentication logic...
    // some other queries to GitHub that work correctly
    var url = "https://api.github.com/repos/SomeOrg/SomePrivateRepo/zipball";
    var response = await this.client.Connection.Get<byte[]>(new Uri(url),
                                                            new Dictionary<string, string>(),
                                                            null);
    var data = response.Body;
    var responseData = response.HttpResponse.Body;

Three problems:

1. `data` is `null`
2. `responseData` is of type `string`
3. when I try `Encoding.ASCII.GetBytes(response.HttpResponse.Body.ToString());` I get invalid zip file

![untitled](https://cloud.githubusercontent.com/assets/3106986/7248361/cb813c2e-e81a-11e4-854e-6cf47d55d995.png)

Am I doing something wrong or is it a bug?

shiftkey:
Fixed in #792

PR #792, opened 2015-04-27
Allow to download zip-attachments

csware:
Fixes issue #784.

### DIFF
--- a/Octokit/Http/HttpClientAdapter.cs
+++ b/Octokit/Http/HttpClientAdapter.cs
@@ -76,8 +76,8 @@ namespace Octokit.Internal
                 {
                     contentType = GetContentMediaType(responseMessage.Content);
 
-                    // We added support for downloading images. Let's constrain this appropriately.
-                    if (contentType == null || !contentType.StartsWith("image/"))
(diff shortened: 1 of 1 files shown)


shiftkey:
Thanks!

Octokit/Http/HttpClientAdapter.cs
```diff
@@ -76,8 +76,8 @@ public HttpClientAdapter(IWebProxy webProxy)
                 {
                     contentType = GetContentMediaType(responseMessage.Content);
 
-                    // We added support for downloading images. Let's constrain this appropriately.
-                    if (contentType == null || !contentType.StartsWith("image/"))
+                    // We added support for downloading images and zip-files. Let's constrain this appropriately.
+                    if (contentType == null || (!contentType.StartsWith("image/") && !contentType.StartsWith("application/")))
```
shiftkey: Just a heads-up - this change also traps `application/json` output which breaks serialization (I should have caught this). I've updated it for the latest release to be more strict: https://github.com/octokit/octokit.net/blob/master/Octokit/Http/HttpClientAdapter.cs#L86

Issue #781, opened 2015-04-19
[RFC] HttpClientAdapter and Octokit - Future Plans


shiftkey:
### The Setup

You've probably never looked at `HttpClientAdapter` before. That's because it's buried deep inside Octokit. But that's not been a deliberate thing - it's just the central place for actually doing the leg work of creating, sending and receiving HTTP requests.

If you wanted to create it by hand in Octokit today, this is the code you need:

```csharp
var connection = new Connection(new ProductHeaderValue("ha-ha-business"), new HttpClientAdapter());
var client = new GitHubClient(connection);
```

But anyway, let me start with a few notes:

 - ~~[Repository Redirects](https://developer.github.com/changes/2015-04-17-preview-repository-redirects/) is currently being tested in the GitHub API - I'd love to support this but it's currently not possible without reworking `HttpClientAdapter`.~~ :ship:ed
 - Credentials in Octokit.net are very basic (you can only set it once, and it's not tied to a specific domain).
 - I have some experiments using Octokit that require swapping out internals, so I'm acutely feeling this pain. And #729 was raised a while ago in a similar vein.

### The Confrontation

As there's many intertwined concepts in here, I'm gonna try and just start from the beginning:

#### HttpClient and You

Initially we used `HttpClientAdapter` to abstract away all the `HttpClient`-ness. And that's fine. But `IHttpClient` is just one method (very similar to `HttpClientAdapter`), it has it's own abstraction for `IRequest/IResponse` and the conversion to `HttpRequestMessage/HttpResponseMessage` is ~~a closely guarded secret~~ just kept internal.

If we extract the message building logic out of `HttpClientAdapter` you're just left with the simple wrapper over `HttpClient`. And that's okay, maybe we don't need to :fire: it off right away. But I think that's some low-hanging fruit to do to appease the SRP gods.

#### Use `HttpClientHandler` Right

The other thing that `HttpClientAdapter` handles is the setup of `HttpClient`. In hindsight, there is a better place for much of this code - a custom implementation of `HttpClientHandler`! This could be passed in as a `ctor` parameter, removing yet another responsibility of `HttpClientAdapter`.

There's one issue I have with this approach, and it's the global `Timeout` property on `HttpClient` - there doesn't seem to be an equivalent at the handler level. I dislike leaning on this crutch but it's been useful when uploading releases, so I'm cautious about migrating away from this.

A thought I had was to combine the received cancellation token with a timeout-based cancellation token, like this:

```csharp
[...]

PR #796, opened 2015-05-08
deprecate httpclient timeout in favour of cancellation token

shiftkey:
This is the first step down the path I talked about in #781 - a minor internal change to `HttpClientAdapter` to make it's `HttpClient` reusable

 - use a `CancellationToken` based on the specified `IRequest.Timeout` value
 - deprecate the `BaseAddress` and `Timeout` values when constructing the `HttpClient`
 - update impacted tests


Haacked:
Looks good to me! :sparkles:

Issue #782, opened 2015-04-19
Obsolete AssertEx.Throws test helper 


shiftkey:
This was written before the support for testing asynchronous tasks was added in xUnit 2.0 (because we pre-dated it).

Here's a rough task list:

 - mark `AssertEx.Throws` as obsolete
 - :eyes: and :fire: all usages - they should be `Assert.ThrowsAsync` now, and it'll drop the need for an inner `async/await` block.
 - :fire: the `AssertEx.Throws` helper
 - dance party

M-Zuber:
Shouldn't `Assert.TaskThrows` be `Assert.ThrowsAsync`?

shiftkey:
@M-Zuber yuuup, I typed it up in a rush. I'll update it.

M-Zuber:
I am having some issues with getting this to work. Should I open a PR to discuss or have it here?

shiftkey:
Fixed in #799

PR #799, opened 2015-05-11
Obsolete assert ex.throws

M-Zuber:
Clean up the tests in order to comply with #782 

- [x] `OctoKit.Tests`
  - [x] `Clients`
  - [x] `Helpers`
  - [x] `Http`
  - [x] `Reactive`
- [x] `Octokit.Tests.Integration`
  - [x] `Clients`

### DIFF
--- a/Octokit.Tests.Integration/Clients/AuthorizationClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/AuthorizationClientTests.cs
@@ -119,7 +119,7 @@ namespace Octokit.Tests.Integration.Clients
             Assert.Equal(created.Token, applicationAuthorization.Token);
 
             await client.Authorization.Delete(created.Id);
-            AssertEx.Throws<NotFoundException>(async () => await client.Authorization.Get(created.Id));
+            Assert.ThrowsAsync<NotFoundException>(() => client.Authorization.Get(created.Id));
         }
 
         [ApplicationTest]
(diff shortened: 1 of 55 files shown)


M-Zuber:
As can be seen in some of the commits, a number of expected exception types had to be changed. Please let me know if there is an issue with this.

shiftkey:
> As can be seen in some of the commits, a number of expected exception types had to be changed. 

This is likely due to the fact we don't assert the exception type in that helper method I wrote. Oops.

M-Zuber:
@shiftkey for the `Reactive` tests when I remove the inner `async / await`, it complains that it needs an explicit conversion to `Task`.
Which is the better option:
```c#
await Assert.ThrowsAsync<ArgumentNullException>(() => (Task)client.Create(null, "name", new NewBlob()));
//or leave the inner async/await
await Assert.ThrowsAsync<ArgumentNullException>(async () => await client.Create(null, "name", new NewBlob()));

```

shiftkey:
@M-Zuber try this:

```csharp
await Assert.ThrowsAsync<ArgumentNullException>(() => client.Create(null, "name", new NewBlob()).ToTask());
```

M-Zuber:
@shiftkey If the final `AppVeyor` builds succeeds then this is pretty much done.

One thing that I did not do is places where the call was not awaited where left that way after updating the method call.

M-Zuber:
@shiftkey

shiftkey:
Thanks @M-Zuber!

PR #805, opened 2015-05-18
Add SHA and make message optional for pull request merge

### LINKED ISSUE #804: Add SHA property to MergePullRequest
The [API](https://developer.github.com/v3/pulls/#merge-a-pull-request-merge-button) caters for specifying the SHA that pull request head must match to allow merge, but the [MergePullRequest model](https://github.com/octokit/octokit.net/blob/master/Octokit/Models/Request/MergePullRequest.cs) does not, we should add this so that it can be passed in when trying to merge a PR.

alfhenrik:
Fixes #804

### DIFF
--- a/Octokit/Models/Request/MergePullRequest.cs
+++ b/Octokit/Models/Request/MergePullRequest.cs
@@ -11,23 +11,21 @@ namespace Octokit
     [DebuggerDisplay("{DebuggerDisplay,nq}")]
     public class MergePullRequest
     {
-        public MergePullRequest(string message)
-        {
-            Ensure.ArgumentNotNull(message, "message");
-
-            Message = message;
-        }
-
         /// <summary>
         /// The message that will be used for the merge commit (optional)
         /// </summary>
-        public string Message { get; private set; }
+        public string Message { get; set; }
+
+        /// <summary>
+        /// The SHA that pull request head must match to allow merge (optional)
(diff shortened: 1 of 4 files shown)


alfhenrik:
:shipit: :trollface: :grinning:

Octokit/Models/Request/MergePullRequest.cs
```diff
@@ -11,23 +11,21 @@ namespace Octokit
...
-            Ensure.ArgumentNotNull(message, "message");
-
-            Message = message;
-        }
-
         /// <summary>
         /// The message that will be used for the merge commit (optional)
         /// </summary>
-        public string Message { get; private set; }
+        public string Message { get; set; }
```
Haacked: I'm currently fixing this, but according to the docs here: https://developer.github.com/v3/pulls/#merge-a-pull-request-merge-button

This is _not_ optional and it should be `CommitMessage`. Did you happen to see something that suggests otherwise?

alfhenrik: I would've thought I *did* see something that suggested that the `Message

Haacked: Well the _response_ is `message` but the _request_ is `commit_message`. The docs are actually a bit unclear on this so I'll ask @pengwynn for clarification.

PR #808, opened 2015-05-22
support redirect requests natively

### LINKED ISSUE #748: Get Archive Link for Repository Contents
Reference: https://developer.github.com/v3/repos/contents/#get-archive-link

This should belong under `IRepositoryContentsClient` but the workflow for turning a `302` into actual content might need a bit of thought.

### LINKED ISSUE #798: support repository redirects natively
Currently in preview, but would be nice to have support for this sooner rather than later

https://developer.github.com/changes/2015-04-17-preview-repository-redirects/

cc @pavel-b-novikov

### LINKED ISSUE #812: Using "Personal Access Tokens"
I have been developing an ASP.NET site that uses Octokit but kept falling foul of the API limits. To help with this I decided to create a "Personal Access Token" in GitHub as I understand that ups your API usage allowance.

The code I now use is 

```c#
        client = new GitHubClient(new ProductHeaderValue("OMHToken"));
        client.Credentials = new Credentials("<<my token here>>");
```

This seemed to have no effect, am I using the token in the correct place?

shiftkey:
Fixes #798 
Fixes #748 
Fixes #812

Building on the awesome start that @darrelmiller started over in https://github.com/darrelmiller/octokit.net/commit/e5d7d6e5f0990202cbd9677843620485c046229a

- [x] cry forever, then get up and ponder on what to do with all the suppressed warnings about incorrect disposal
- [x] write integration tests for redirecting repositories:
  - `https://github.com/robconery/massive` -> `https://github.com/fbouma/massive`
- [x] write some tests around creating data and handling the `307`
  - setup a public repository to test this
- [x] `RepositoryContentsClient.GetArchiveLink` depends on not following the redirect, prolly need to address this
- [x] documentation
- [x] set the timeout for downloading the archive above the default

@jasonrudolph anything I've missed here?


darrelmiller:
I just noticed the RedirectHandler CopyResult uses a .Result.  That's because I pulled the code from a .net 4.0 project.  It should probably be fixed to use async/await.  I'd do it but I'm not sure how to add a commit to a PR.

jasonrudolph:
Thanks for digging into this, @shiftkey. :zap:

> @jasonrudolph anything I've missed here?

My C# leaves much to be desired, but @kdaigle and I sketched out this plaintext spec that might help answer your question:

> * **MUST** follow redirect requests if HTTP status code is 301, 302, or 307. (These are the redirect codes actively used by the GitHub API.)
> * **MUST** redirect a 30x status code if the HTTP method is HEAD, OPTIONS, GET, POST, PUT, PATCH, or DELETE.
> * **MUST** use the original request's HTTP method when following a redirect where the HTTP status code is 307.
> * **SHOULD** use the original request's HTTP method when following a redirect where the HTTP status code is 301 or 302.
> * **MUST** use the original request's authentication credentials when following a redirect where the original host matches the redirect host.
> * **MUST NOT** use the original request's authentication credentials when following a redirect where the original host does not match the redirect host.
> * **SHOULD** only follow 3 redirects.

Does that help?

darrelmiller:
1) I'm not doing redirect counting to limit redirects.  I can fix that.
2) I'm not checking host before copying auth header.  I can fix that.
3) Currently I change the method on a 301 or 302 because it is allowed.  Changing the behaviour to not change the method is easy enough.

darrelmiller:
I have done all the fixes to the RedirectHandler and added a bunch of tests for it.  However, I have absolutely no @!#$@#$ clue how to push the changes up to anywhere on Github.
I pulled down the pr/808 branch and made the changes there, but am unable to push that branch with the fixes to my fork.

shiftkey:
@darrelmiller if you're working on your fork and it's in a branch I know about, I can pull in your changes into this PR...

darrelmiller:
@shiftkey I don't really know how.  But I succeeded in pushing my changes to my fork.  https://github.com/darrelmiller/octokit.net/commit/cf9c84506454013a007bc10db53a6bd9dd5a5a75

darrelmiller:
Couple of comments:  

The 301 and 302 behaviour is now different to what HttpWebRequest does by default.  But the new behaviour is what was originally intended for 301/302. (Possibly the most useful diagram I have ever seen on this subject is here https://tools.ietf.org/html/rfc7238#section-1)

Keeping the auth header around based on host is consistent with what the .Net CredentialCache does.  However, based on my interpretation of the HTTP specs, it should really be based on the realm.  Which would require a 401 dance to find out the new realm.  I would file this under "nice idea in theory but no-one does it".  The pedant in me felt the need to mention it.

jasonrudolph:
> The 301 and 302 behaviour is now different to what HttpWebRequest does by default.  But the new behaviour is what was originally intended for 301/302. (Possibly the most useful diagram I have ever seen on this subject is here https://tools.ietf.org/html/rfc7238#section-1)

@darrelmiller: Thanks for linking to that diagram. Wonderfully succinct and useful. :zap:

shiftkey:
@darrel_miller your latest changes are applied onto this PR - thanks for that :sparkling_heart:.

i'll look at those failing tests, and putting together some additional tests for the GitHub API.

Octokit/Http/HttpClientAdapter.cs
```diff
@@ -109,6 +103,8 @@ protected virtual HttpRequestMessage BuildRequestMessage(IRequest request)
             {
                 var fullUri = new Uri(request.BaseAddress, request.Endpoint);
                 requestMessage = new HttpRequestMessage(request.Method, fullUri);
+
+                requestMessage.Properties["AllowAutoRedirect"] = request.AllowAutoRedirect;
```
shiftkey: @darrelmiller I'm almost inclined to obsolete `request.AllowAutoRedirect` and simplify things - as we need to control this behaviour in our very specific way. Any objections?

shiftkey:
Just added a passing test against the new redirects - I'll add some more later this week, but this is feeling rather close...

Octokit/Http/HttpClientAdapter.cs
darrelmiller: It felt like a nice feature to have, but I'm struggling to think of a scenario where I would actually need it.

shiftkey: It was one of the first things I'd removed when refactoring earlier attempts. Good enough for me :fire::fire::fire::fire::fire:

darrelmiller:
:+1:

shiftkey:
Added in a test for issue creation and finding a repository. Ready for review.

shiftkey:
Just added one more little task before I'm happy with this.

shiftkey:
Extracted #817 rather than drag this out any longer.

Ready for review :dog:

shiftkey:
@haacked would love some :eyes: on this before I merge it and continue down the path of extracting more HTTP-friendly abstractions...

darrelmiller:
Just one observation that would probably have been much more useful before we started all this.

If you are actually using the CredentialCache with credentials associated to the correct origin server, then when you redirect to a URL with the same host then the credentials are carried over.  Using a credential cache avoids setting creds in a default header and takes care of applying the correct credentials on every request.   

Maybe it shouldn't be the responsibility of redirection code to be carrying over the authorization header.

shiftkey:
> If you are actually using the CredentialCache with credentials associated to the correct origin server, then when you redirect to a URL with the same host then the credentials are carried over.

Yep, I'd like to address this at some point - we should mimic this behaviour in our abstraction and move it into `HttpClient`.

darrelmiller:
Funny you should mention that, but I just finished writing an AuthenticationService/HttpCredentialCache that plugs into the message handler pipeline.  It's not on Github yet though.

shiftkey:
@haacked \*bump for code review plz\*

Octokit/Http/HttpClientAdapter.cs
```diff
@@ -152,5 +149,102 @@ static string GetContentMediaType(HttpContent httpContent)
...
+            var response = await base.SendAsync(request, cancellationToken);
+
+            // Can't redirect without somewhere to redirect too.  Throw?
+            if (response.Headers.Location == null) return response;
+
+            // Don't redirect if we exceed max number of redirects
+            var redirectCount = 0;
+            if (request.Properties.Keys.Contains(RedirectCountKey))
+            {
+                redirectCount = (int)request.Properties[RedirectCountKey];
```
Haacked: Couldn't this fail if there's no value here? Or do we always seed this with a value?

Haacked:
Looks good to me. Just one question.

Octokit/Http/HttpClientAdapter.cs
darrelmiller: I think this would only fail if someone pre-filled the property with a value that cannot be cast to an int.  A value is assigned to the property when it is initially created, right after this If ends.

jasonrudolph:
@darrelmiller @shiftkey: :clap: :star2:

shiftkey:
I'll run through the integration tests and see if I can put together a :ship: on Monday (my time).

PR #765, opened 2015-03-21
Repository Get Archive Link

alfhenrik:
Partial fix for #748 

- [x] Get link to the archive
- ~~[ ] Get the **actual** content of the archive.... :santa: :gift:~~ Will send a separate PR for this one

PR #810, opened 2015-05-31
finish support for downloading the code archive

### LINKED ISSUE #748: Get Archive Link for Repository Contents

shiftkey:
This was started in #765 but due to #808 we either need to:

 - implement a workaround for users to opt-out of redirects, leave the feature incomplete
 - just finish the feature, make retrieving the link unnecessary

I've favoured the latter, and marked the existing `GetArchiveLink` methods as obsolete.

Doing this as a separate PR into the redirect PR so we can discuss whether users want to access the link without initiating the download.

Fixes #748 

cc @alfhenrik

### DIFF
--- a/Octokit.Reactive/Clients/IObservableRepositoryContentsClient.cs
+++ b/Octokit.Reactive/Clients/IObservableRepositoryContentsClient.cs
@@ -34,7 +34,16 @@ namespace Octokit.Reactive
         /// <param name="owner">The owner of the repository</param>
         /// <param name="name">The name of the repository</param>
         /// <returns></returns>
+        [Obsolete("Use GetArchive to download the archive instead")]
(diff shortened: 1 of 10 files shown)


alfhenrik:
Yep, sounds good to me. 
Reading the API docs, the purpose of that endpoint is to redirect to the actual content so we should probably adhere to that I guess :smile:

And if someone is crying out for functionality to get *just* the link then we can reconsider and add that back in.

shiftkey:
Thanks @alfhenrik for the sanity check!

Issue #732, opened 2015-02-28
Search in repositories without term


alisabzevari:
I have tested api with this request:
https://api.github.com/search/repositories?q=language:assembly&sort=stars&order=desc
It returns the correct result. I wanted to get all repositories with assembly language.

The problem is Octokit.net does not allow creating SearchRepositoriesRequest with empty term in constructor.

Haacked:
Ah. Ok. Mind sending a pull request? We should just make that ctor empty and allow setting the term property.

alisabzevari:
I tried to fix the issue but I found something else:

This is not a valid request:
https://api.github.com/search/code?q=in:file+repo:jquery/jquery
In the error message it says "Searches that use qualifiers only are not allowed." but this request is valid:
https://api.github.com/search/code?q=language:js+repo:jquery/jquery
I think there is a misconception in API.

alexandrugyori:
I added a pull request adding no-args constructor since I needed the feature.

raffaeler:
Can someone tell me why I still can't search without a term?
Am I doing something wrong?
*  throws: `new SearchCodeRequest(null, owner, repositoryName)`
* can't specify the owner/repository with the default ctor

Thanks!

PR #828, opened 2015-06-29
Added no-args constructor to the SearchRepositoriesRequest. Fixing issue 732.

alexandrugyori:
Fixed issue #732, to be able to create requests with no search term.
Added no-args constructor to BaseSearchRequest also.
Added a test to check things work as expected.


Haacked:
:sparkles:

PR #835, opened 2015-07-14
search across multiple repositories

### LINKED ISSUE #834: search across multiple repositories
it's currently impossible to do something like this:

> https://api.github.com/search/issues?q=repo:aspnet/dnx+repo:aspnet/logging+type:pr+state:open&sort=created&order=desc&per_page=100

with Octokit. The syntax is valid, but we constrain things currently to one repository.

cc @damianedwards

shiftkey:
Fixes #834 

This is a breaking API change to support being able to search across repositories. We currently have an `ctor` overload that accepts an `owner` and `name` for the repository:

```csharp
var request = new SearchIssuesRequest("windows", "aspnet", "dnx");
```

The change here is adding a public property `Repos` which accepts a collection of `owner/name` strings (i've added some validation to ensure these are formatted correctly for the server):

```csharp
var request = new SearchIssuesRequest("windows");
request.Repos = new Collection<string> {
    "aspnet/dnx",
    "aspnet/dnvm"
};
```

This replaces the old `Repo` property, which was constrained to one string:

```csharp
var request = new SearchIssuesRequest("windows");
request.Repo = "aspnet/dnx";
```

Rather than carry around both `Repo` and `Repos` for a period of time (leading to possible confusion), I'm just going to :fire: the property here - I think many people would prefer to just use the `Repos` property anyway...

- [x] confirm regex rules match our repository rules
- [x] make changes in `SearchIssueRequest`
- [x] make changes in `SearchCodeRequest`
- [x] a more-friendly abstraction than just a collection?
- [x] document how to search issues (with some examples)
- [x] document how to search pull requests (with some examples)
- ~~[ ] document an example of searching code~~
- [x] wtf ci?

[Rendered docs](https://github.com/octokit/octokit.net/blob/refine-search-api/docs/search.md)

cc @damianedwards @M-Zuber

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -77,6 +77,9 @@ tools/xunit.runner.console
 *.ncrunch*
 *.GhostDoc.xml
 
+# FAKE temporary files
+.fake/
+
 # New VS Test Runner creates arbitrary folders with PDBs
 *.pdb
 pingme.txt
--- a/Octokit.Reactive/Octokit.Reactive.csproj
+++ b/Octokit.Reactive/Octokit.Reactive.csproj
@@ -39,7 +39,8 @@
     <Prefer32Bit>false</Prefer32Bit>
     <RunCodeAnalysis>true</RunCodeAnalysis>
     <CodeAnalysisRuleSet>..\Octokit.ruleset</CodeAnalysisRuleSet>
-    <DocumentationFile>bin\Release\Net45\Octokit.Reactive.XML</DocumentationFile>
+    <DocumentationFile>
+    </DocumentationFile>
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="System" />
--- a/Octokit.Tests.Integration/Clients/SearchClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/SearchClientTests.cs
@@ -63,7 +63,8 @@ public class SearchClientTests
     [Fact]
     public async Task SearchForOpenIssues()
     {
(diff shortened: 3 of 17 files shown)


shiftkey:
@damianedwards @M-Zuber writing up a walkthrough of the search options available - [how does this read?](https://github.com/octokit/octokit.net/blob/refine-search-api/docs/search.md)

M-Zuber:
It reads quite nicely. 
The examples are very nice to have, a bonus would be a simple run down of the properties available even if they don't have examples.  (I would be willing to try and do a PR once this in)

khellang:
What about creating a specialized object for holding the repos? This way you could use a custom collection initializer, forcing the user to specify both owner and repo name.

shiftkey:
@khellang good point - if I'm breaking the API here I'll see if I can do better than just a collection

M-Zuber:
Preaching to the choir, but I fully agree with @khellang

khellang:
Sent a PR at #842

Haacked:
This is looking pretty good. Anything needing specific review?

shiftkey:
@haacked not really, I should write an example for `SearchCodeRequest.cs` because that's impacted by this change. But I think everything in the code is :sparkles:. I'll add a note.

Haacked:
@shiftkey should I merge this?

shiftkey:
@haacked sure, I just like putting docs in front of me - because otherwise they'll never get written :trolleybus:

Haacked:
:sparkles: all around.

PR #842, opened 2015-07-19
Changed repos to a specialized collection

khellang:
I didn't remove the validation, since it still gives some value in validating the owner and repo names separately.

### DIFF
--- a/Octokit/Models/Request/SearchIssuesRequest.cs
+++ b/Octokit/Models/Request/SearchIssuesRequest.cs
@@ -2,6 +2,7 @@
 using System.Diagnostics;
 using Octokit.Internal;
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using System.Globalization;
@@ -20,7 +21,7 @@ namespace Octokit
         /// </summary>
         public SearchIssuesRequest()
         {
-            Repos = new Collection<string>();
+            Repos = new RepositoryCollection();
         }
 
         /// <summary>
@@ -29,7 +30,7 @@ namespace Octokit
         /// <param name="term">The term to filter on</param>
         public SearchIssuesRequest(string term) : base(term)
         {
-            Repos = new Collection<string>();
+            Repos = new RepositoryCollection();
         }
 
         [Obsolete("this will be deprecated in a future version")]
@@ -39,9 +40,7 @@ namespace Octokit
             Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
             Ensure.ArgumentNotNullOrEmptyString(name, "name");
 
-            var repo = string.Format(CultureInfo.InvariantCulture, "{0}/{1}", owner, name);
-
-            Repos.Add(repo);
+            Repos.Add(owner, name);
         }
 
         /// <summary>
@@ -196,7 +195,7 @@ namespace Octokit
         public string User { get; set; }
 
         [SuppressMessage("Microsoft.Usage", "CA2227:CollectionPropertiesShouldBeReadOnly")]
-        public Collection<string> Repos { get; set; }
+        public RepositoryCollection Repos { get; set; }
 
         public override IReadOnlyList<string> MergedQualifiers()
         {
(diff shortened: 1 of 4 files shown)


khellang:
What's up with the CI build?

shiftkey:
> What's up with the CI build?

It's been bumped to MSBuild 14.0 which is triggering a neat PCL/FxCop issue. Reported [here](http://help.appveyor.com/discussions/problems/2522-windows-server-2012-environment-now-using-msbuild-2014-by-default)

khellang:
Something like this? http://kristian.hellang.com/building-csharp-6-code-with-appveyor/

Octokit/Models/Request/SearchIssuesRequest.cs
```diff
@@ -337,4 +336,53 @@ public enum IssueTypeQualifier
...
+        {
+            return _repositories.GetEnumerator();
+        }
+
+        IEnumerator IEnumerable.GetEnumerator()
+        {
+            return GetEnumerator();
+        }
+
+        public void Add(string owner, string name)
```
shiftkey: Fun fact - you can also have `Add(string nameWithOwner)` here and both sorts will happily work with the collection initializer syntax.

@haacked should we support both? I still like the succinctness of writing them inline rather than splitting them out...

shiftkey:
@khellang nah, I wanted to stick with v12 for the moment.

I fought a bunch with the toolchain today, couldn't find a way out. It's a minor thing (I don't *need* to run FxCop on the PCL build as the code is shared) but an annoyance nonetheless.

Octokit/Models/Request/SearchIssuesRequest.cs
khellang: I actually had both in there, but decided to delete the overload. I guess it doesn't hurt to have it in, since we still do the same, full, validation :smile:

khellang: I added a commit with the single string overloads, you can just strip them out if you decide you don't want both.

khellang:
I'm not really sure if this builds, because I get all sorts of errors in VS 2015. I had hoped the CI build would tell me, but no luck there either :stuck_out_tongue:

shiftkey:
> I had hoped the CI build would tell me, but no luck there either

i got you bb. Once @haacked has had a chance to chime in I'll test things on VS2015 proper.

Octokit/Models/Request/SearchIssuesRequest.cs
```diff
@@ -337,4 +336,74 @@ public enum IssueTypeQualifier
         [Parameter(Value = "issue")]
         Issue
     }
+
+    public class RepositoryCollection : IEnumerable<string>
```
Haacked: Since the repository list will always be a finite collection, should this be an `IReadOnlyCollection<string>`? And could it not simply inherit a concrete class at that point?

shiftkey: :thumbsup: to more specialized collection

Not sure if the PCL library might get in the way of using a concrete class here though...

Haacked: Well there's List<T> that'll work. :stuck_out_tongue: Also, it looks like this is supposed to be a mutable collection right now, so maybe List<T> would make more sense than `ReadOnlyCollection<T>`. Or we could go the immutable route and have `Add` always return a new collection.

I just think having this implement `IEnumerable<T>` is not right. At minimum, `ICollection<T>` would make more sense.

khellang: The only reason it implements `IEnumerable<string>` is because it's needed for collection initializers to work. I had originally implemented `ICollection<string>`, but landed on `IEnumerable` in the original implementation because it didn't allow simple strings.

Haacked:
:+1: on this. Is our CI having an issue?

shiftkey:
@haacked yeah, it's a VS2015 issue (Code Analysis yay) which we got accidentally upgraded to a few days ago [details here](http://help.appveyor.com/discussions/problems/2522-windows-server-2012-environment-now-using-msbuild-2014-by-default).

I'll address it in #835 which is the base branch for this PR.

Issue #811, opened 2015-06-08
Rate Information


cdaapi:
Is it possible to access the rate information via the .NET library as per https://developer.github.com/v3/rate_limit/

shiftkey:
Not currently, but it's been raised previously: #504

Red-Folder:
Is this not different to #504?  #504 asked to get the rate limits per API call (included in the headers)

This request asked for the rate limits as part of the MiscellaneousClient - a discrete API request.

I'm currently having a play at both.  Hopefully produce a PR for the MiscellaneousClient some point next week.

shiftkey:
@Red-Folder you're right. I was thinking about it from the "i need this info" side rather than the fact you can get it in either place.

> I'm currently having a play at both. Hopefully produce a PR for the MiscellaneousClient some point next week.

I think this is going to be easier to knock out than the details in #504, but I suspect we'll need both eventually. Thanks for looking into it!

Red-Folder:
Functionality added in commit #848 

Believe this can now be closed

PR #848, opened 2015-07-24
Added GetRateLimits to MiscellaneousClient

Red-Folder:
Added support for the rate_limit API (https://developer.github.com/v3/rate_limit/).  Note that this is the Miscellaneous API call - not the rate details returned as part of the headers for every API call.  This functionality is referenced in #811 

I'm interested in feedback (thus marked as WIP).  Taken a couple of design decisions that may not fit with the overall solution:

1. Created a separate ResourceRateLimit model rather than re-purpose the existing RateLimit.  I've chosen to keep them separate as they have such different source (API vs Header) and could be changed by GitHub independently.
2. Created a structure made up of MiscRateLimt/ ResourcesRateLimits/ ResourcesRateLimit using the Poco Json deserialization - fitting the API structure.  Possibly better to perform custom deserialization (similar to ActivityPayload?) to return an array of a simple object (resource name, limit, remaining, reset).  Note entirely sure of the names either.
3. I've included the depreciating "Rate" for completeness.  Given that it will depreciate, then maybe better to simply not get anyone into bad habits.  Would be removed anyway if I go for custom deserialization.
4. To utilize Poco Json deserialization I've created a reset (long) and resetAsDateTimeOffset.  Better way to do?  Again, if I go down custom deserialization then I can just use the reset as DateTimeOffset.

Feedback always welcomed.


Red-Folder:
Further thoughts on point 1;
If I re-purpose the existing RateLimit model then any consumer code doesn't need to handle the (logically) same data twice

Octokit/Models/Response/MiscRateLimits.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Diagnostics;
+using System.Globalization;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Octokit
+{
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
+    public class MiscRateLimits
```
Haacked: We try to avoid abbreviations. So I'd call this `MiscellaneousRateLimits`

Octokit/Models/Response/ResourceRateLimit.cs
```diff
@@ -0,0 +1,55 @@
+﻿using System;
+using System.Diagnostics;
+using System.Globalization;
+
+using Octokit.Helpers;
+
+namespace Octokit
+{
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
+    public class ResourceRateLimit
```
Haacked: I'd rather we just use `RateLimit` and not this class. The name of this class is too similar to `ResourceRateLimits`. The data we show in this response and in the headers are the same and not likely to change (as we try and avoid breaking changes).

Octokit/Models/Response/ResourceRateLimits.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System;
+using System.Diagnostics;
+using System.Globalization;
+
+using Octokit.Helpers;
+
+namespace Octokit
+{
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
+    public class ResourceRateLimits
```
Haacked: Per my [previous suggestion](https://github.com/octokit/octokit.net/pull/848/files#r35466147), I think this should be `ResourceRateLimit` and the property types should be `RateLimit`

Haacked:
> If I re-purpose the existing RateLimit model then any consumer code doesn't need to handle the (logically) same data twice

Yeah, I don't think we're likely to change `RateLimit` in any way that would change them in one place, but not the other. As far as I can tell, they're the same type in the back-end. So I think re-using `RateLimit` and not introducing a new type is the way to go. :smile:

Thanks for the contribution!

Octokit/Models/Response/ResourceRateLimit.cs
Red-Folder: Changed now to use RateLimit

Octokit/Models/Response/MiscRateLimits.cs
Red-Folder: Changed as suggested

Octokit/Models/Response/ResourceRateLimits.cs
Red-Folder: Done

Red-Folder:
@Haacked AppVeyor isn't building, is this because the original ResourceRateLimit.cs comment is still open, or is it something wrong with the code?

Originally I though it was the code - that I needed to put some #if !NETFX_CORE blocks round attributes.  But I've made changes to that an it doesn't seem to make any change.

shiftkey:
@Red-Folder it looks like there's a merge conflict with this branch - AppVeyor should go green after that's been addressed.

Octokit/Octokit-Mono.csproj
```diff
@@ -396,7 +396,12 @@
     <Compile Include="Exceptions\TwoFactorAuthorizationException.cs" />
     <Compile Include="Http\HttpMessageHandlerFactory.cs" />
     <Compile Include="Models\Response\TeamMembership.cs" />
+<<<<<<< HEAD
```
shiftkey: Looks like these files need to be unified...

Red-Folder: Looking into it now

Red-Folder:
Ok, green now.  Now I leave it alone and get some sleep.

Lessons learnt this evening:
1. Do a fetch & merge from upstream BEFORE pushing
2. How to use Meld
3. That I should double check everything is as I expect in the commit
4. Run the automated tests (even after I make the smallest possible change that in no possible way could break the solution)
5. I should have gone to bed about 2 hours ago.  I'm too old to up past midnight anymore

On a more serious note;
* Appreciate the feedback & assistance given - please keep it coming.  
* I still need to run the integration tests (for some reason I cant make run yet - keep getting bad credentials)

shiftkey:
@Red-Folder given this is a new feature (and the test is written) I'm happy to take it in as-is. 

Don't stress too much about running the integration tests, it's certainly the toughest part of contributing (and something I need to revisit).

Octokit.Tests/Clients/MiscellaneousClientTests.cs
```diff
@@ -58,6 +59,72 @@ public class TheGetEmojisMethod
...
+                    )
+                );
+                var connection = Substitute.For<IConnection>();
+                connection.Get<MiscellaneousRateLimit>(Args.Uri, null, null).Returns(Task.FromResult(response));
+                var client = new MiscellaneousClient(connection);
+
+                var result = await client.GetRateLimits();
+
+                // Test the high level object
+                Assert.NotNull(result);
```
shiftkey: While I love the explicitness here, I think letting this fall through to the Assert.Equals is totally fine to simplify the tests...

Octokit/Http/RateLimit.cs
```diff
@@ -2,17 +2,22 @@
...
 using Octokit.Helpers;
+using System.Diagnostics;
+using System.Globalization;
+using Octokit.Internal;
 
 namespace Octokit
 {
 #if !NETFX_CORE
     [Serializable]
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
```
shiftkey: I think this can live outside the `#if`

Octokit/Http/RateLimit.cs
```diff
@@ -66,5 +94,18 @@ public virtual void GetObjectData(SerializationInfo info, StreamingContext conte
             info.AddValue("Reset", Reset.Ticks);
         }
 #endif
+
+#if !NETFX_CORE
+        internal string DebuggerDisplay
+#else
+        public string DebuggerDisplay
```
shiftkey: Does this need to be `public`?

Octokit/Models/Response/ResourceRateLimit.cs
shiftkey: I'm not sure where we ended up with this - is it `ResourceRateLimit` or `RateLimit`?

Octokit.Tests/Clients/MiscellaneousClientTests.cs
Red-Folder: Just to clarify, is this the NotNull check?  What about the other NotNull tests further down - lose them as well to make more readable?

Octokit/Http/RateLimit.cs
Red-Folder: Will do.  I changed all this when I thought it was causing the appveyor problems (there was reasoning behind my thinking but I won't bore you)

Octokit/Http/RateLimit.cs
Red-Folder: As above - introduced (incorrectly) as trying to fix appveyor.  Will change.

Octokit.Tests/Clients/MiscellaneousClientTests.cs
Red-Folder: I assumed the above was yes & yes

Octokit/Models/Response/ResourceRateLimit.cs
Red-Folder: Three models in use:

1. Originally MiscRateLimit -> Changed to MiscellaneousRateLimit to avoid abbreviation
2. Originally ResourceRateLimits -> Changes to ResourceRateLimit 
3. Originally ResourceRateLimit -> Changes to use existing RateLimit

Hopefully that makes sense

Haacked: :+1:

Haacked:
Thanks for your contribution!

selfie-0

Red-Folder:
Is it wrong that after 20+ years in IT, I'm seeing a 6 second vine from Phil Hacck as my most glowing accreditation?

Surely there must be a "thumbs up" accreditation section somewhere in Linkedin .... off to go look

Haacked:
> Surely there must be a "thumbs up" accreditation section somewhere in Linkedin .... off to go look

LOL! It's a GitHub special accreditation. Much better than LinkedIn.

Issue #504, opened 2014-06-03
Request: Include remaing rate-limit in responses


saguiitay:
For example, in the SearchCodeResult (and all other Result objects), include a remaining rate limits.

This will allow an extremely simple to implement throttling of my requests

shiftkey:
> I'm currently using the former in some test code - but doesn't feel correct to add to a PR.

I think 1) is going to be a more maintainable solution in the long-term. The other solution means deserializing becomes a two-step operation (read the content, then read the headers). Would love to hear your concerns.

Red-Folder:
Gonna have a go at option 1 and produce a PR for review (hopefully next 24 hours)

Basically just going to provide a property on the client which calls through to connection which calls through the HTTPClientAdaptor (which will store the last ApiInfo during the Send).

Putting the chained calls all the way through the stack avoids having to keep a reference to the HttpClientAdaptor - which makes it easier for the user.

shiftkey:
As this requires significant thought - and likely significant changes to the API responses - as well as there exists the ability to get this data using `IGitHubClient.GetLastApiInfo()` I'm happy to close this out for now.

I've opened https://github.com/octokit/octokit.net/issues/1142 to add some documentation for this feature before I forget.

PR #855, opened 2015-07-31
Access to the last ApiInfo object

Red-Folder:
As discussed in issue #504 

Allows the user to get the contents of the last ApiInfo (so last Api call).  Provided as a property of GitHubClient which calls a property of Connection then a property of HttpClientAdapter.

The most obvious use for it is to check the current rate limit without making a separate API call.

Two concerns for me:

* Can't put a test in for the HttpClientAdapter.  The underlying HttpClient is concrete and the send method is not virtual so can be substituted.  Could abstract the HttpClient (interface and wrapper) - this would help to put more testing into for the HttpClientAdapter (currently quite light).
* Must be a better design than having to implement a property through every layer

Thanks

### DIFF
--- a/Octokit/Http/ApiInfo.cs
+++ b/Octokit/Http/ApiInfo.cs
@@ -51,5 +51,25 @@ namespace Octokit
         /// Information about the API rate limit
         /// </summary>
         public RateLimit RateLimit { get; private set; }
+
+        /// <summary>
+        /// Allows you to clone ApiInfo 
+        /// </summary>
+        /// <returns>A clone of <seealso cref="ApiInfo"/></returns>
+        public ApiInfo Clone()
+        {
+            // Seem to have to do this to pass a whole bunch of tests (for example Octokit.Tests.Clients.EventsClientTests.DeserializesCommitCommentEventCorrectly)
+            // I believe this has something to do with the Mocking framework.
+            if (this == null || this.Links == null || this.OauthScopes == null || this.RateLimit == null || this.Etag == null)
+                return null;
+
+            return new ApiInfo(this.Links.Clone(),
+                                this.OauthScopes.Clone(),
+                                this.AcceptedOauthScopes.Clone(),
+                                new String(this.Etag.ToCharArray()),
+                                this.RateLimit.Clone());
+        }
+
+
     }
 }
--- a/Octokit/Http/ApiInfo.cs
+++ b/Octokit/Http/ApiInfo.cs
@@ -60,7 +60,7 @@ namespace Octokit
         {
             // Seem to have to do this to pass a whole bunch of tests (for example Octokit.Tests.Clients.EventsClientTests.DeserializesCommitCommentEventCorrectly)
             // I believe this has something to do with the Mocking framework.
-            if (this == null || this.Links == null || this.OauthScopes == null || this.RateLimit == null || this.Etag == null)
+            if (this.Links == null || this.OauthScopes == null || this.RateLimit == null || this.Etag == null)
                 return null;
 
             return new ApiInfo(this.Links.Clone(),
--- a/Octokit/Http/Connection.cs
+++ b/Octokit/Http/Connection.cs
@@ -136,6 +136,12 @@ namespace Octokit
             _jsonPipeline = new JsonHttpPipeline();
         }
 
+        /// <summary>
+        /// Gets the latest API Info - this will be null if no API calls have been made
+        /// </summary>
+        /// <returns><seealso cref="ApiInfo"/> representing the information returned as part of an Api call</returns>
+        public ApiInfo LastApiInfo { get { return _httpClient.LastApiInfo; } }
+
         public Task<IApiResponse<T>> Get<T>(Uri uri, IDictionary<string, string> parameters, string accepts)
         {
             Ensure.ArgumentNotNull(uri, "uri");
--- a/Octokit/Http/Connection.cs
+++ b/Octokit/Http/Connection.cs
@@ -140,25 +140,15 @@ namespace Octokit
         /// Gets the latest API Info - this will be null if no API calls have been made
         /// </summary>
         /// <returns><seealso cref="ApiInfo"/> representing the information returned as part of an Api call</returns>
-        public ApiInfo LastApiInfo
-        { 
-            get
-            {
-                lock (LastApiInfoLocker)
-                {
-                    return _lastApiInfo;
-                }
-            }
-            private set
-            {
-                lock (LastApiInfoLocker)
-                {
-                    _lastApiInfo = value;
-                }
-            }
+        public ApiInfo GetLastApiInfo()
+        {
+            // We've choosen to not wrap the _lastApiInfo in a lock.  Originally the code was returning a reference - so there was a danger of
+            // on thread writing to the object while another was reading.  Now we are cloning the ApiInfo on request - thus removing the need (or overhead)
+            // of putting locks in place.
+            // See https://github.com/octokit/octokit.net/pull/855#discussion_r36774884
+            return _lastApiInfo == null ? null : _lastApiInfo.Clone();
         }
         private ApiInfo _lastApiInfo;
-        private readonly object LastApiInfoLocker = new object();
 
         public Task<IApiResponse<T>> Get<T>(Uri uri, IDictionary<string, string> parameters, string accepts)
         {
@@ -550,7 +540,8 @@ namespace Octokit
             var response = await _httpClient.Send(request, cancellationToken).ConfigureAwait(false);
             if (response != null)
             {
-                LastApiInfo = response.ApiInfo;
+                // Use the clone method to avoid keeping hold of the original (just in case it effect the lifetime of the whole response
+                _lastApiInfo = response.ApiInfo.Clone();
             }
             HandleErrors(response);
             return response;
--- a/Octokit/Http/Connection.cs
+++ b/Octokit/Http/Connection.cs
@@ -140,7 +140,25 @@ namespace Octokit
         /// Gets the latest API Info - this will be null if no API calls have been made
         /// </summary>
         /// <returns><seealso cref="ApiInfo"/> representing the information returned as part of an Api call</returns>
-        public ApiInfo LastApiInfo { get; private set; }
+        public ApiInfo LastApiInfo
+        { 
+            get
+            {
+                lock (LastApiInfoLocker)
+                {
+                    return _lastApiInfo;
+                }
+            }
+            private set
+            {
+                lock (LastApiInfoLocker)
+                {
+                    _lastApiInfo = value;
+                }
+            }
+        }
+        private ApiInfo _lastApiInfo;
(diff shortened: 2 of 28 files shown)


Octokit/Http/IApiInfo.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Octokit
+{
+    /// <summary>
+    /// Provides a property for the Last recorded API infomation
+    /// </summary>
+    public interface IApiInfo
```
khellang: `IApiInforProvider`? This sounds like an interface for the `ApiInfo` itself.

Octokit/Http/HttpClientAdapter.cs
```diff
@@ -45,7 +51,11 @@ public HttpClientAdapter(Func<HttpMessageHandler> getHandler)
                 // Make the request
                 var responseMessage = await _http.SendAsync(requestMessage, HttpCompletionOption.ResponseContentRead, cancellationTokenForRequest)
                                                 .ConfigureAwait(false);
-                return await BuildResponse(responseMessage).ConfigureAwait(false);
+                var response = await BuildResponse(responseMessage).ConfigureAwait(false);
+
+                LastApiInfo = response.ApiInfo;
```
khellang: This does not have to be set here and IMO it's not a HTTP concern at all. All requests go trough [`Connection.RunRequest`](https://github.com/octokit/octokit.net/blob/a7bc09e76725cf053c4546e681def2c550796a64/Octokit/Http/Connection.cs#L522-L529) so I think it would be better to move this up a level.

Octokit/Http/IHttpClient.cs
```diff
@@ -10,7 +10,7 @@ namespace Octokit.Internal
     /// <remarks>
     /// Most folks won't ever need to swap this out. But if you're trying to run this on Windows Phone, you might.
     /// </remarks>
-    public interface IHttpClient : IDisposable
+    public interface IHttpClient : IDisposable, IApiInfo
```
khellang: This does not feel right. API info is not a HTTP concern (see previous comment).

Octokit/Http/HttpClientAdapter.cs
khellang: Also, what about threading? Should there be a lock introduced here?

khellang:
@Red-Folder I sent you a PR at https://github.com/Red-Folder/octokit.net/pull/1, I still haven't addressed the threading issue, though.

Red-Folder:
@khellang Great stuff.  I'll pull in over the weekend

Red-Folder:
Can anyone recommend a good Api call to use for the integration test?

I'm current using the Rate Limit call - but this doesn't include links, oauth scope, accepted oauth scope, & etag,in the ApiInfo.  I'd like an Api call which will populate all of these for testing

shiftkey:
@Red-Folder I'd suggest a couple of calls:

 - `ETag` can be found by creating a new repository - that should be easy enough
 - `Links` are a bit trickier (because most of our existing APIs are going to follow them as part of pagination). You might be able to check for the last value after the pagination is done - so something like "list your repos" might work for you.
 - `X-OAuth-Scopes` and `X-Accepted-OAuth-Scopes` require the caller to be configured with token authentication - that depends on the `OCTOKIT_OAUTHTOKEN` environment variable being set, but the caller might not have this setup.

Octokit/Http/Connection.cs
```diff
@@ -136,6 +136,30 @@ public Connection(ProductHeaderValue productInformation, Uri baseAddress, ICrede
...
+            {
+                lock (LastApiInfoLocker)
+                {
+                    _lastApiInfo = value;
+                }
+            }
+        }
+        private ApiInfo _lastApiInfo;
+        private readonly object LastApiInfoLocker = new object();
+
```
Red-Folder: @khellang I've added this for the thread safety.  Thoughts welcome

Octokit/Http/Connection.cs
```diff
@@ -524,6 +548,10 @@ public Credentials Credentials
             request.Headers.Add("User-Agent", UserAgent);
             await _authenticator.Apply(request).ConfigureAwait(false);
             var response = await _httpClient.Send(request, cancellationToken).ConfigureAwait(false);
+            if (response != null)
+            {
+                LastApiInfo = response.ApiInfo;
+            }
```
Red-Folder: @khellang I moved the setting of the LastApiInfo before HandleErrors so that the LastApiInfo is still set even if an exception is thrown from the HandleErrors

Red-Folder:
@shiftkey Added further integration tests now - seems to cover all bases if a little repetitive.  I've added a new test attribute for Personal Access Token - the OAuth/ AcceptedOAuth test uses that attribute.

Red-Folder:
Hi guys, any further feedback on this on?

I'm in danger of losing my dev environment in the next few days so would like to get this PR boxed off.

Thanks

Octokit/Http/Connection.cs
Haacked: Hmm, do we actually need locks here? If someone is reading the value _just_ as another thread is writing the value, it sort of doesn't matter who wins. After all, the read value will only be _slightly_ outdated.

In other words, I don't see any problems with race conditions in this case as this isn't data that's timing sensitive. I don't think torn reads are a problem because these are 32 bit references.

This value is going to be set on every request, I'd rather avoid a lock here unless it's absolutely needed.

Red-Folder: @khellang @Haacked I'm easy either way - with locking/ without

What's the preferred method of resolving coding decisions?  A well argued debate between respectful & collaborate peers?  Rocks/ Paper/ Scissors?  A fight to the death?

(is it wrong to hope for the last one)

Haacked: I've been taking the Benevolent Dictator role. :stuck_out_tongue:

I'd like to hear from @khellang if there are specific concerns with removing the locks. If not, let's remove them.

Red-Folder: (Off topic)

Please tell me someone out there is photoshoping the below - replacing with @Haacked and the Octopus (not saying which way round)

image

khellang: I think it's fine to remove the locks, it's probably a bit to much. I just brought it up because there's now some shared, mutable state going on.

Haacked: :thumbsup:

Let's remove the locks, but make sure we comment it and explain why so if someone else comes along, they understand why there's shared mutable state.

Also, one idea we should do is to set `LastApiInfo` to a copy of the `ApiInfo` to reduce the sharing of mutable state. In fact, I think `LastApiInfo` should be a method `GetLastApiInfo` and it should always return a copy of the last `ApiInfo`

That way, Octokit is in control of the lifetime of that object and a consumer can't inadvertently keep the object around longer than expected. Thoughts?

khellang: SGTM :+1:

Octokit/Http/ApiInfo.cs
```diff
@@ -51,5 +51,25 @@ public class ApiInfo
...
+        /// <summary>
+        /// Allows you to clone ApiInfo 
+        /// </summary>
+        /// <returns>A clone of <seealso cref="ApiInfo"/></returns>
+        public ApiInfo Clone()
+        {
+            // Seem to have to do this to pass a whole bunch of tests (for example Octokit.Tests.Clients.EventsClientTests.DeserializesCommitCommentEventCorrectly)
+            // I believe this has something to do with the Mocking framework.
+            if (this == null || this.Links == null || this.OauthScopes == null || this.RateLimit == null || this.Etag == null)
+                return null;
```
Red-Folder: Extra points for explaining why I needed to add this null check.  For some reason bundles of tests fail without it (comment and see what happens).  I think it is something to do with how the mocking framework is handling the object - but can't understand why it touches it.

Red-Folder:
Could someone have a look over my cloning logic.  Seems to work - but I could do with a fresh set of eyes on it.

Note that I'm having to do new String(originalString.ToCharArray()) as nothing else (string.Copy for example) seems to work with portable version.

Octokit/Http/ApiInfo.cs
```diff
@@ -51,5 +51,25 @@ public class ApiInfo
...
+
+        /// <summary>
+        /// Allows you to clone ApiInfo 
+        /// </summary>
+        /// <returns>A clone of <seealso cref="ApiInfo"/></returns>
+        public ApiInfo Clone()
+        {
+            // Seem to have to do this to pass a whole bunch of tests (for example Octokit.Tests.Clients.EventsClientTests.DeserializesCommitCommentEventCorrectly)
+            // I believe this has something to do with the Mocking framework.
+            if (this == null || this.Links == null || this.OauthScopes == null || this.RateLimit == null || this.Etag == null)
```
Haacked: When would `this` be null though?

Octokit/Http/RateLimit.cs
```diff
@@ -99,5 +99,19 @@ internal string DebuggerDisplay
...
+        /// </summary>
+        /// <returns>A clone of <seealso cref="RateLimit"/></returns>
+        public RateLimit Clone()
+        {
+            var clone = new RateLimit();
+            clone.Limit = this.Limit;
+            clone.Remaining = this.Remaining;
+            clone.ResetAsUtcEpochSeconds = this.ResetAsUtcEpochSeconds;
+
+            return clone;
```
Haacked: As a minor code style, I think this should use object initializer syntax.

```csharp
return new RateLimit
{
    Limit = this.Limit,
    Remaining = this.Remaining,
    ResetAsUtcEpochSeconds = this.ResetAsUtcEpochSeconds
}
``

Haacked:
Looks good. The title suggests this is a WIP. Is that still the case or is this good to merge?

Octokit/Http/ApiInfo.cs
Red-Folder: Fair point.

I really not sure why I need this line at all - but something blew up within the mocking framework so I added everything and kitchen sink.

Maybe went a tad overboard ;)

Red-Folder:
@Haacked Sorry Phil, I'd not been chasing this.  I'd lost my dev environment & between jobs (must find new job before wife makes me paint more fences).

Dev environment back up and running.  Will make the two minor tweaks you mention above then should be good to merge.  Probably today or tomorrow

Haacked:
@Red-Folder No problem! Thanks for continuing to work on this and good luck finding a new job!

Red-Folder:
Pops head above the parapet ... "Looks good to go"

(Prepares to duck back down quickly)

Haacked:
Thanks!
selfie-0

Red-Folder:
Ah ... excellent.  A second thumbs up to add to the collection.

I've added a whole section on my bio page for them ;)

http://www.red-folder.com/Home/MyBio (then scroll down).  One day I'll setup the anchors, make the site size correctly, finish it, etc, etc, blah, blah

PR #856, opened 2015-07-31
fix the TeamClient.AddMembership(..) call

davidalpert:
While exploring some organizational automation with `octokit.net` I discovered that the `TeamClient.AddMembership(..)` call passes a `null` value into a parameter of `ApiConnection.Put(..)` that fails on an `Octokit.Ensure.ArgumentNotNull(..)` call.

The GitHub API 3.0 Team docs state that the correct way to call the _Add Team Member_ endpoint is to essentially pass an empty body:

> Note that you’ll need to set Content-Length to zero when calling out to this endpoint. For more information, see “HTTP verbs.”
> -  [Add Team Member](https://developer.github.com/v3/orgs/teams/#add-team-member) API docs

My proposed fix (that I will submit shortly in this PR) is to have `ApiConnection.Put(..)` check for a `null` body and respond by setting the `Content-Length` header to zero as instructed, if that is possible, rather than throwing an `ArgumentNullExceptionValue`.

#### Todo
- [x] vet the design approach;
- [x] configure the underlying `Connection` object to add the `Content-Length` header when the `body` is `null`
- [x] removed the `Content-Length` header as that generated a `System.InvalidOperationException: Misused header name.`
- [x] confirmed worked for my use case, used octokit to add users to a team;

### DIFF
--- a/Octokit.Tests/Clients/TeamsClientTests.cs
+++ b/Octokit.Tests/Clients/TeamsClientTests.cs
@@ -165,7 +165,7 @@ namespace Octokit.Tests.Clients
 
                 connection.Received().Put<Dictionary<string, string>>(
                     Arg.Is<Uri>(u => u.ToString() == "teams/1/memberships/user"),
-                    Arg.Is<string>(u => u == null));
+                    Arg.Is<object>(u => u == ApiConnection.EmptyBody));
             }
 
             [Fact]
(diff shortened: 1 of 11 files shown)


Haacked:
Rather than accepting a null value, I wonder if we should have a special value. For example, we could have a static `ApiConnection.EmptyBody` property that's just an empty `object` and use that for cases when you're supposed to pass in an empty body. I try and avoid overloading `null` like this because it's hard to determine if you passed `null` as a mistake or intentionally.

Also, what happens if you just pass in `new object()` right now as the `data` argument. Does that do the right thing today?

davidalpert:
@Haacked tried your suggestion of an Empty static instead of `null`; moved it to it's own class to avoid coupling `Connection` to depend on `ApiConnection` while `ApiConnection` depends on `IConnection`.

Take a look. All tests pass. Ran `.\build FixProjects` to add `ResponseBody.cs` to all projects. And it fixed my use case. :+1:

shiftkey:
@davidalpert thanks for finding this and fixing it!

thumbsupfamily

PR #857, opened 2015-07-31
Add multiple dates to DateRange filter

### LINKED ISSUE #841: DateRange syntax doesn't let you specify two dates
Example:

```
created:"2012-04-30 .. 2012-07-04"
```

We could have a second `ctor` which takes two `DateTime` parameters...

ChrisMissal:
resolves #841

This seemed really straightforward, but if there's more to it, let me know!


shiftkey:
Looks good to me!

thumbs-up-steve

PR #860, opened 2015-08-05
Tidying up some code

Haacked:
Just tidying up some code


shiftkey:
Adding this to the VS2015 milestone as it requires the end user to have access to VS2015 to build it...

Haacked:
> it requires the end user to have access to VS2015 to build it...

It does? I tried to avoid using any new C# features. What did I do wrong?

shiftkey:
@haacked apologies, I mixed this up with another PR - I thought I saw some `nameof` usage in here.

Haacked:
@shiftkey nope. You might have seen my github/VisualStudio PR. Note that there's no way this would be green if I used C# 6 features, right?

khellang:
Also, hopefully, Visual Studio would complain since `LangVersion` is set to `5` :wink:

shiftkey:
>  Note that there's no way this would be green if I used C# 6 features, right?

CI is running against MSBuild v14, so compiler support for these new features is available. But yes, as @khellang pointed out we've locked the projects to C# 5...

khellang:
Have you thought about when you'd like to cross over to the greener C# 6 side?

shiftkey:
> Have you thought about when you'd like to cross over to the greener C# 6 side?

Not until after the VS2015 migration is done at the earliest...

Haacked:
@shiftkey can this be merged?

Octokit/Http/RateLimit.cs
```diff
@@ -52,15 +53,15 @@ public RateLimit(int limit, int remaining, long reset)
...
-        [ParameterAttribute(Key = "ignoreThisField")]
+        [Parameter(Key = "ignoreThisField")]
         public DateTimeOffset Reset { get; private set; }
 
         /// <summary>
         /// The date and time at which the current rate limit window resets - in UTC epoch seconds
         /// </summary>
-        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Performance", "CA1811:AvoidUncalledPrivateCode")]
-        [ParameterAttribute(Key = "reset")]
-        public long ResetAsUtcEpochSeconds { get { return Reset.ToUnixTime(); } private set { Reset = value.FromUnixTime(); } }
```
shiftkey: Yeah, I don't know why we'd need this `set`. Nice spot!

SolutionInfo.cs
```diff
@@ -1,11 +1,12 @@
 ﻿// <auto-generated/>
+
 using System.Reflection;
 using System.Runtime.InteropServices;
 
-[assembly: AssemblyProductAttribute("Octokit")]
-[assembly: AssemblyVersionAttribute("0.14.0")]
-[assembly: AssemblyFileVersionAttribute("0.14.0")]
-[assembly: ComVisibleAttribute(false)]
+[assembly: AssemblyProduct("Octokit")]
```
shiftkey: I wonder if this will be reverted the next time I run it through the FAKE build script...

PR #862, opened 2015-08-07
Handle empty content better

### LINKED ISSUE #836: Trying to get contributors for a repo with no commits results in a "Queued Operations expect status codes of Accepted or OK" error
What I did:
Called `Statistics.GetContributors` for a repository with no commits in it

What happened:
I got an exception that "Queued Operations expect status codes of Accepted or OK" (because GitHub sent back a 204 No Content response)

What I expected:
To get an empty list back

Haacked:
Fixes #836

If we get a 204 No Content response to GetQueuedContent, we should return an empty collection rather than throw an exception.


Haacked:
@shiftkey you ok with this?

PR #864, opened 2015-08-10
add filename for code search

fffej:
I couldn't find a helpful `FileName` property for code search (see https://help.github.com/articles/searching-code/#search-by-filename) , so I added it!

### DIFF
--- a/Octokit.Tests.Integration/Clients/SearchClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/SearchClientTests.cs
@@ -44,6 +44,19 @@ public class SearchClientTests
     }
 
     [Fact]
+    public async Task SearchForFileNameInCode()
+    {
+        var request = new SearchCodeRequest("swag")
+        {
+            FileName = "readme.md"
+        };
+
+        var repos = await _gitHubClient.Search.SearchCode(request);
+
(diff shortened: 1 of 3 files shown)


shiftkey:
This is really great! Thanks for picking this up!

If you feel like scoring some extra credit it'd be great to add an integration test so we can use the real API. Otherwise just let me know and I'll add it to the backlog.

fffej:
I had a quick look at the tests, but didn't see an obvious way to add a good test.

The closest example to crib seems to be `SearchForfunctionInCode`, but this just makes a request and asserts the response is not empty.  I didn't want to add another one as I'm not sure it asserts anything useful?  Perhaps I'm misunderstanding the tests?

shiftkey:
> I didn't want to add another one as I'm not sure it asserts anything useful?

To be honest, that's basically what I'd love to see here. Something simple to catch regressions which reflects the usage of this API. 

>  Perhaps I'm misunderstanding the tests?

The integration tests are there to catch anything unexpected changing between the API we expect (i.e. the library) and the API we have (i.e. the server). It's something I always run as part of doing each release, so if we can keep it in sync I'll be a happy maintainer.

fffej:
How's that for size?

Haacked:
crossing-thumbs-up

PR #871, opened 2015-08-26
Added PreviousFileName property to GitHubCommitFile

CorinaCiocanea:
This PreviousFileName property is returned from the GitHub API when a file was modified. Check https://api.github.com/repos/octokit/octokit.net/commits/997e955f for an example of returning a commit containing file renames. 
Let me know what you think about the change and if there are any modifications I should make related to it.


Haacked:
Flawless Victory

selfie-1

Issue #875, opened 2015-08-30
Doc comments on Issue.Url and Issue.HtmlUrl are wrong/missing


Eilon:
This one is wrong (it says milestone instead of issue): https://github.com/octokit/octokit.net/blob/master/Octokit/Models/Response/Issue.cs#L34-L37

And this one is missing: https://github.com/octokit/octokit.net/blob/master/Octokit/Models/Response/Issue.cs#L39

mikeparker:
If the related PR is merged can this issue be closed? @Eilon

shiftkey:
:shipit:

PR #876, opened 2015-08-30
#875 fixing doc comments on Issue type

Eilon:
@Haacked hello.


Eilon:
https://github.com/octokit/octokit.net/issues/875

Haacked:
Hmm, the build failure is unrelated to this code change. Also, that test works on my machine. /cc @shiftkey. I'll merge this change and we can look into that test failure separately.

Thanks @Eilon!

Eilon:
Thanks!

PR #878, opened 2015-09-04
Use non-deprecated auth endpoint 

niik:
Octokit currently uses an undocumented endpoint for creating app authorizations.

```
PUT /authorizations/clients/:client_id/:fingerprint
```

This makes Octokit send the fingerprint as part of the URL which only works as long as the fingerprint doesn't contain any characters that has special meaning in urls (most notably `/`). Seeing how fingerprints are an opaque string we can't expect them not to contain reserved characters.

This PR switches the endpoint to the [documented one](https://developer.github.com/v3/oauth_authorizations/#get-or-create-an-authorization-for-a-specific-app) which accepts the fingerprint as part of the JSON body where we can use any characters we want.

cc @Haacked @shiftkey


shiftkey:
:thumbsup: I've been meaning to obsolete this endpoint

Haacked:
selfie-1

PR #881, opened 2015-09-06
Add docs link into README

mikeparker:
The readme doesn't mention the docs folder, and it should.

I wasn't sure exactly where to put the link so if anyone thinks it should be bumped down a bit (perhaps in the "Getting Started" section?) let me know and i'll move it.


shiftkey:
@mikeparker thanks for reminding me about this.

>  if anyone thinks it should be bumped down a bit 

Let's put it in *after* `Getting Started`

Haacked:
I'll fix it up.

PR #889, opened 2015-09-10
Fixes for missing files in the Mono-Android project 

naveensrinivasan:
The mono android projects were referencing to files that were missing which was causing the build to fail. This is similar to #888


shiftkey:
:thumbsup: thanks!

PR #897, opened 2015-09-17
893 readonlypagecollection null body

### LINKED ISSUE #893: Retrieving contributors for an empty repo throws an ANE
In this case the response of the body is `null` and the call to https://github.com/octokit/octokit.net/blob/c6c332f5de4038be275d7e99649b8eb6ec668cf2/Octokit/Http/ApiConnection.cs#L467-L469 and then https://github.com/octokit/octokit.net/blob/c6c332f5de4038be275d7e99649b8eb6ec668cf2/Octokit/Http/ReadOnlyPagedCollection.cs#L13-L14 passes that `null` into the `ReadOnlyCollection` constructor.

adamralph:
fixes #893

### DIFF
--- a/Octokit/Http/ReadOnlyPagedCollection.cs
+++ b/Octokit/Http/ReadOnlyPagedCollection.cs
@@ -11,7 +11,7 @@ namespace Octokit.Internal
         readonly Func<Uri, Task<IApiResponse<List<T>>>> _nextPageFunc;
 
         public ReadOnlyPagedCollection(IApiResponse<List<T>> response, Func<Uri, Task<IApiResponse<List<T>>>> nextPageFunc)
-            : base(response != null ? response.Body : new List<T>())
+            : base(response != null ? response.Body ?? new List<T>() : new List<T>())
         {
             Ensure.ArgumentNotNull(response, "response");
             Ensure.ArgumentNotNull(nextPageFunc, "nextPageFunc");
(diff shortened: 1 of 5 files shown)


Octokit/Http/ReadOnlyPagedCollection.cs
```diff
@@ -11,7 +11,7 @@ public class ReadOnlyPagedCollection<T> : ReadOnlyCollection<T>, IReadOnlyPagedC
         readonly Func<Uri, Task<IApiResponse<List<T>>>> _nextPageFunc;
 
         public ReadOnlyPagedCollection(IApiResponse<List<T>> response, Func<Uri, Task<IApiResponse<List<T>>>> nextPageFunc)
-            : base(response != null ? response.Body : new List<T>())
+            : base(response != null ? response.Body ?? new List<T>() : new List<T>())
```
M-Zuber: Now I really understand the power of the `?.` operator

adamralph: Indeed! :unamused:

Haacked: We could probably upgrade our compiler to use C# 6 if someone wants to take a crack at it. As long as we don't break our other platforms.

Haacked:
selfie-0

Haacked:
This is now [live on NuGet](https://www.nuget.org/packages/octokit) as of version `0.16.0`!

khellang:
:clap:

Issue #885, opened 2015-09-09
Issue response should return EventsUrl


naveensrinivasan:
https://developer.github.com/v3/issues/#list-issues

```
{
    "id": 1,
    "url": "https://api.github.com/repos/octocat/Hello-World/issues/1347",
    "labels_url": "https://api.github.com/repos/octocat/Hello-World/issues/1347/labels{/name}",
    "comments_url": "https://api.github.com/repos/octocat/Hello-World/issues/1347/comments",
    "events_url": "https://api.github.com/repos/octocat/Hello-World/issues/1347/events",
    "html_url": "https://github.com/octocat/Hello-World/issues/1347",
```
The Issue response should return the Event's URL. This is similar to #883

PR #901, opened 2015-09-18
Add Events URL to the Issue class.

alfhenrik:
Fixes octokit/octokit.net#885


khellang:
:shipit:

Haacked:
Flawless victory!


selfie-0

PR #902, opened 2015-09-19
Build fix for Xamarin Studio Solution

naveensrinivasan:
With this fix we should be able to compile the Xamarin Studio Solution from the shell.


Haacked:
Thanks!

selfie-0

PR #906, opened 2015-09-21
Updated the readme with reactive octokit.

naveensrinivasan:
Updated the readme to include the nuget reference to Octokit.Reactive


Haacked:
:+1:

Issue #861, opened 2015-08-07
Consider making RepositoryContent.EncodedContent public


yufeih:
Please consider making `RepositoryContent.EncodedContent` public or provide an alternative to get the raw bytes of a file, this is useful when working with binary files.

naveensrinivasan:
I guess this could be security reason so that you don't extract a malware.

https://github.com/octokit/octokit.net/blob/e6e7c3d393160fabe4cc48dd6c410db2a19b7e72/Octokit/Models/Response/RepositoryContent.cs#L34-48

@Haacked

Haacked:
I'm ok with either making it public or maybe adding a method to read it as a binary stream/array.

yufeih:
Thanks, you guys rocks.

PR #907, opened 2015-09-22
Making Encodedcontent public #861

naveensrinivasan:
Making `EncodedContent` public to get the raw bytes of a file. #861

### DIFF
--- a/Octokit/Models/Response/RepositoryContent.cs
+++ b/Octokit/Models/Response/RepositoryContent.cs
@@ -31,7 +31,7 @@ namespace Octokit
         /// The Base64 encoded content if this is a file. Otherwise it's null.
         /// </summary>
         [Parameter(Key = "content")]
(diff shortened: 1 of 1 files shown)


Haacked:
That looks like a legit failure. We have some unit tests that check that models follow our conventions.

alfhenrik:
Yea, needs to have the `setter` as protected... :)

naveensrinivasan:
Thanks fixed that!. It is failing again in the `StopsMakingNewRequestsWhenTakeIsFulfilled`

Haacked:
> Thanks fixed that!. It is failing again in the `StopsMakingNewRequestsWhenTakeIsFulfilled`

I restarted the build. I haven't had time to look into why that test fails sometimes. It's worrying.

Haacked:
Looks good to me!

PR #908, opened 2015-09-22
Clarify why convention tests are failing

khellang:
Related to #907

The failing test yields:

>    Expected: True
      Actual:   False

Now it will include the offending members, which IMO is much more helpful:

> Model type 'Octokit.RepositoryContent' contains the following mutable properties: 
      EncodedContent

### DIFF
--- a/Octokit.Tests.Conventions/Exception/InvalidDebuggerDisplayAttributeValueException.cs
+++ b/Octokit.Tests.Conventions/Exception/InvalidDebuggerDisplayAttributeValueException.cs
@@ -1,0 +1,18 @@
+using System;
+
+namespace Octokit.Tests.Conventions
+{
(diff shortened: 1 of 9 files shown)


khellang:
Ugh. That didn't work so well...

> [00:00:00] Build started
[00:00:02] git config --global core.autocrlf input
[00:00:02] git clone -q https://github.com/octokit/octokit.net.git C:\projects\octokit-net
[00:02:32] fatal: early EOF
[00:02:32] fatal: The remote end hung up unexpectedly
[00:02:32] fatal: index-pack failed
[00:02:32] error: RPC failed; result=18, HTTP code = 200
[00:02:32] Command exited with code 128

Haacked:
Thanks!


selfie-0

PR #915, opened 2015-09-28
Add a bunch of XML doc comments

Haacked:
Our build status has a :poop: load of warning about undocumented public properties/methods/etc.

We could turn it off, or given that this is a library, we could actually document everything.

This ends up being a big task, so this PR just documents all the Request models. I also fixed a few bugs I found while doing this.

### DIFF
--- a/Octokit/Models/Request/NotificationsRequest.cs
+++ b/Octokit/Models/Request/NotificationsRequest.cs
@@ -21,19 +21,20 @@ namespace Octokit
         public bool Participating { get; set; }
 
         /// <summary>
-        /// Filters out any notifications updated before the given time. Default: DateTimeOffset.MinValue
+        /// Only show notifications updated after the given time. Defaults to the everything if unspecified.
         /// </summary>
-        public DateTimeOffset Since { get; set; }
+        public DateTimeOffset? Since { get; set; }
 
         /// <summary>
-        /// Construct a <see cref="NotificationsRequest"/> object
+        /// Only show notifications updated before the given time.
         /// </summary>
-        public NotificationsRequest()
-        {
-            All = false;
-            Participating = false;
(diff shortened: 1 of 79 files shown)


Octokit/Models/Request/NotificationsRequest.cs
```diff
@@ -21,19 +21,20 @@ public class NotificationsRequest : RequestParameters
         public bool Participating { get; set; }
 
         /// <summary>
-        /// Filters out any notifications updated before the given time. Default: DateTimeOffset.MinValue
+        /// Only show notifications updated after the given time. Defaults to the everything if unspecified.
```
Haacked: Hey @pengwynn, I mentioned this in the chat room, but figured a GitHub comment message was less transient.

According [to the docs](https://developer.github.com/v3/activity/notifications/#list-your-notifications) if you don't specify a `since` it defaults to `Time.now` which doesn't make sense to me as there can't be any notifications since now at the time you make this request.

It makes more sense that `Time.now` is the default for `before`, not `since`. Are the docs wrong?

Haacked:
Since @shiftkey is on vacation and these are mostly XML comment changes, I'm going to YOLO merge this. Feel free to review later!

Octokit/Models/Request/PublicRepositoryRequest.cs
```diff
@@ -14,6 +21,12 @@ public PublicRepositoryRequest(int since)
             Since = since;
         }
 
+        /// <summary>
+        /// Gets or sets the integer ID of the last Repository that you’ve seen.
+        /// </summary>
+        /// <value>
+        /// The since.
```
shiftkey: l o l

shiftkey:
Thanks for digging into this. Hopefully now I'm back I can help out with improving things around the build side.

Issue #779, opened 2015-04-07
GitHubCommit.Author/Committer should be a simple object, not Author type 


shiftkey:
Currently when you retrieve a commit from `{user}/{repo]/commits` you get this payload for the `Author` and `Committer`:

```json
      "author": {
        "name": "Gabriel Weyer",
        "email": "gabrielweyer@gmail.com",
        "date": "2014-05-04T00:46:35Z"
      },
      "committer": {
        "name": "Gabriel Weyer",
        "email": "gabrielweyer@gmail.com",
        "date": "2014-05-04T00:46:35Z"
      },
```

This doesn't align with what we have defined in `GitHubCommit.Author` and `GitHubCommit.Committer` - we should introduce a specify type which contains just these properties.

khellang:
Sounds like... `CommitCommitter` :trollface:

willsb:
I really wanna get my feet wet on OSS, and this bug seems like a good start. Can I call it mine and fix it? 

Also, I can't think of a name other than CommitCommitter lol

shiftkey:
@willsb I think `CommitEntity` is a better term :grinning: - and you're more than welcome to make a start on it

willsb:
@shiftkey Okay, I'll take a look at it this weekend. Just commenting here to let everyone aware :)

shiftkey:
![cantwait](https://cloud.githubusercontent.com/assets/359239/7882424/a3a87370-064c-11e5-97c7-f5ef7f216279.gif)

Haacked:
> Currently when you retrieve a commit from `{user}/{repo]/commits` you get this payload for the `Author` and `Committer`:

@shiftkey that's not exactly correct.

If you look at https://developer.github.com/v3/repos/commits/#get-a-single-commit you'll see the payload looks more like (bunch of stuff cut out):

```js
{
  "commit": {
    "author": {
      "name": "Monalisa Octocat",
      "email": "support@github.com",
      "date": "2011-04-14T16:00:49Z"
    },
    "committer": {
      "name": "Monalisa Octocat",
      "email": "support@github.com",
      "date": "2011-04-14T16:00:49Z"
    },
    "author": {
      "login": "octocat",
      "id": 1,...
    }
    "committer": {
      "login": "octocat",
      "id": 1,...
    }
}
```

So note that `GitHubCommit` has `Author` and `Committer` properties. But it also has an inner `Commit` property that also has an `Author` and `Committer`.

[...]

PR #916, opened 2015-09-28
Consolidate committer info

Haacked:
Replace `SignatureResponse` and `CommitEntity` with `Committer`

A recent PR added `CommitEntity` but we already had `SignatureResponse` expressly for this purpose.

So this commit renames `SignatureResponse` to `Committer` and removes `CommitEntity` and replaces it with `Committer`.

I also had to retarget the NetCore45 app so it builds in VS 2015.


bslliw:
My bad, didn't know about this one. But why was GitHubCommit changed back to use Author instead of Committer? (that was the initial issue)

willsb:
My bad, didn't know about this one. But why was GitHubCommit changed back to use Author instead of Committer? (that was the initial issue)

Haacked:
> My bad, didn't know about this one. But why was `GitHubCommit` changed back to use `Author` instead of `Committer`? (that was the initial issue)

Because according [to the docs here](https://developer.github.com/v3/repos/commits/#get-a-single-commit), the `author` and `committer` for a `GitHubCommit` correspond to a GitHub account (aka `Author`).

The `commit` property of `GitHubCommit` also has `author` and `committer` but those correspond to `Committer`.

I think the confusion stems from the fact that the API returns GitHub objects as well as Git objects and Octokit.net has confused the two in some places.

A `Committer` is the minimal signature that a Git commit contains `{'name', 'email', 'date'}`. GitHub will try to associate a committer (and author) of a Git commit to a GitHub account. It can't always do that and now that I think about it, I'm not sure what it responds when it can't.

Haacked:
I commented on [the original issue](https://github.com/octokit/octokit.net/issues/779#issuecomment-143803506)

Issue #914, opened 2015-09-27
Add helper class for creating web hooks


Haacked:
The [`NewRepositoryHook`](https://github.com/octokit/octokit.net/blob/master/Octokit/Models/Request/NewRepositoryHook.cs) class is used to create new repository hooks. This class takes in a dictionary of config values.

According to the [documentation for creating repository hooks](https://developer.github.com/v3/repos/hooks/#create-a-hook), if you want to create web hooks, there's some required and optional values that should be in that dictionary. We should provide a strongly typed class specifically for web hooks.

My thought is to add `NewRepositoryWebHook` that derives from `NewRepositoryHook` and adds these properties. It then merges those properties with the supplied `config` dictionary.

alfhenrik:
I'll take this one :)

PR #917, opened 2015-09-28
Add helper class for creating web hooks

alfhenrik:
Introducing `NewRepositoryWebHook`, a new strongly typed class that derives from the existing `NewRepositoryHook` class but adds properties and constructor parameters for config settings that should be passed to the API when creating a new web hook. These settings are merged with the config settings that are specific to the web service the web hook will be interacting with (as per [https://api.github.com/hooks](https://api.github.com/hooks)).

Fixes octokit/octokit.net#914

### DIFF
--- a/Octokit/Models/Request/NewRepositoryWebHook.cs
+++ b/Octokit/Models/Request/NewRepositoryWebHook.cs
@@ -1,5 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 
@@ -145,16 +144,4 @@ namespace Octokit
         Json
     }
 
-    public class WebHookConfigComparer : IEqualityComparer<KeyValuePair<string, string>>
-    {
-        public bool Equals(KeyValuePair<string, string> x, KeyValuePair<string, string> y)
-        {
-            return x.Key == y.Key;
-        }
-
-        public int GetHashCode(KeyValuePair<string, string> obj)
-        {
-            return obj.Key.GetHashCode();
-        }
-    }
 }
--- a/Octokit/Models/Request/NewRepositoryWebHook.cs
+++ b/Octokit/Models/Request/NewRepositoryWebHook.cs
@@ -62,7 +62,7 @@ namespace Octokit
         /// <param name="url">
         /// A required string defining the URL to which the payloads will be delivered.
         /// </param>
-        public NewRepositoryWebHook(string name, IReadOnlyDictionary<string, string> config, string url) 
+        public NewRepositoryWebHook(string name, IReadOnlyDictionary<string, string> config, string url)
             : base(name, config)
         {
             Ensure.ArgumentNotNullOrEmptyString(url, "url");
@@ -88,7 +88,7 @@ namespace Octokit
         /// The content type.
         /// </value>
         public WebHookContentType ContentType { get; set; }
-        
+
         /// <summary>
         /// Gets the secret used as the key for the HMAC hex digest
         /// of the body passed with the HTTP requests as an X-Hub-Signature header.
@@ -97,7 +97,7 @@ namespace Octokit
         /// The secret.
         /// </value>
         public string Secret { get; set; }
-        
+
         /// <summary>
         /// Gets whether the SSL certificate of the host will be verified when 
         /// delivering payloads. The default is `false`.
@@ -110,14 +110,29 @@ namespace Octokit
 
         public override NewRepositoryHook ToRequest()
         {
-            var config = Config.Union(new Dictionary<string, string>
+            var webHookConfig = GetWebHookConfig();
+            if (Config.Any(c => webHookConfig.ContainsKey(c.Key)))
             {
-                    { "url", Url },
-                    { "content_type", ContentType.ToParameter() },
-                    { "secret", Secret },
-                    { "insecure_ssl", InsecureSsl.ToString() }
-            }).ToDictionary(k => k.Key, v => v.Value);
+                var invalidConfigs = Config.Where(c => webHookConfig.ContainsKey(c.Key)).Select(c => c.Key);
+                throw new RepositoryWebHookConfigException(invalidConfigs);
+            }
+
+            var config = webHookConfig
+                .Union(Config, new WebHookConfigComparer())
+                .ToDictionary(k => k.Key, v => v.Value);
+
             return new NewRepositoryHook(Name, config);
+        }
+
+        Dictionary<string, string> GetWebHookConfig()
+        {
+            return new Dictionary<string, string>
+            {
+                { "url", Url },
+                { "content_type", ContentType.ToParameter() },
+                { "secret", Secret },
+                { "insecure_ssl", InsecureSsl.ToString() }
+            };
         }
     }
 
@@ -128,5 +143,18 @@ namespace Octokit
     {
         Form,
         Json
+    }
+
+    public class WebHookConfigComparer : IEqualityComparer<KeyValuePair<string, string>>
+    {
+        public bool Equals(KeyValuePair<string, string> x, KeyValuePair<string, string> y)
+        {
+            return x.Key == y.Key;
+        }
+
+        public int GetHashCode(KeyValuePair<string, string> obj)
+        {
+            return obj.Key.GetHashCode();
+        }
     }
 }
--- a/Octokit/Models/Request/NewRepositoryWebHook.cs
+++ b/Octokit/Models/Request/NewRepositoryWebHook.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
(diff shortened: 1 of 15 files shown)


Haacked:
> Currently there are 2 constructors, one that takes the required (name, config and url) parameters and defaults the optional (contentType, secret and insecureSsl) parameters and one that takes all parameters.

Funny you should mention this. I just submitted a PR that includes some guidelines around our model objects. https://github.com/octokit/octokit.net/blob/consolidate-committer-info/Octokit/Models/README.md

For request objects (such as this), I think optional parameters should simply be read/write properties and required parameters are read only and go in the constructor. Does that make sense?

alfhenrik:
:+1: will update PR tonight

alfhenrik:
`NewRepositoryWebHook` now conforms to the model request guidelines of **required** properties are added as `ctor` parameters with read-only properties and optional properties are added as read-write properties.

Also added a method, `ToRequest()`, that merges the passed in `config` object with the webhook specific config properties that is used when passing the object to the API.

Octokit/Models/Request/NewRepositoryWebHook.cs
```diff
@@ -0,0 +1,132 @@
...
+        /// Gets the secret used as the key for the HMAC hex digest
+        /// of the body passed with the HTTP requests as an X-Hub-Signature header.
+        /// </summary>
+        /// <value>
+        /// The secret.
+        /// </value>
+        public string Secret { get; set; }
+        
+        /// <summary>
+        /// Gets wether the SSL certificate of the host will be verified when
```
Haacked: /wether/whether

Octokit/Models/Request/NewRepositoryWebHook.cs
```diff
@@ -0,0 +1,132 @@
...
+    /// To create a webhook, the following fields are required by the config:
+    /// <list type="bullet">
+    /// <item>
+    ///   <term>url</term>
+    ///   <description>A required string defining the URL to which the payloads will be delivered.</description>
+    /// </item>
+    /// <item>
+    ///   <term>content_type</term>
+    ///   <description>
+    ///     An optional string defining the media type used to serialize the payloads.Supported values include json and
```
Haacked: Please add a space after the period before the word "Supported"

Octokit/Models/Request/NewRepositoryWebHook.cs
```diff
@@ -0,0 +1,132 @@
...
+    ///     form.The default is form.
```
Haacked: Space after the period.

Octokit/Models/Request/NewRepositoryWebHook.cs
```diff
@@ -0,0 +1,132 @@
...
+    /// </list>
+    /// <para>
+    /// API: https://developer.github.com/v3/repos/hooks/#create-a-hook
+    /// </para>
+    /// </remarks>   
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
+    public class NewRepositoryWebHook : NewRepositoryHook
+    {
+        /// <summary>
+        /// Initializes a new instance of the<see cref="NewRepositoryWebHook"/> class.
```
Haacked: Space after "the"

Octokit/Models/Request/NewRepositoryWebHook.cs
```diff
@@ -0,0 +1,132 @@
...
+    ///   <description>
+    ///     An optional string that’s passed with the HTTP requests as an X-Hub-Signature header. The value of this
+    ///     header is computed as the HMAC hex digest of the body, using the secret as the key.
+    ///   </description>
+    /// </item>
+    /// <item>
+    ///   <term>insecure_ssl:</term>
+    ///   <description>
+    ///     An optional string that determines whether the SSL certificate of the host for url will be verified when 
+    ///     delivering payloads.Supported values include "0" (verification is performed) and "1" (verification is not
```
Haacked: Space after "payloads."

Octokit/Models/Request/NewRepositoryWebHook.cs
```diff
@@ -0,0 +1,132 @@
...
+    /// </para>
+    /// </remarks>   
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
+    public class NewRepositoryWebHook : NewRepositoryHook
+    {
+        /// <summary>
+        /// Initializes a new instance of the<see cref="NewRepositoryWebHook"/> class.   
+        /// Using default values for ContentType, Secret and InsecureSsl.     
+        /// </summary>
+        /// <param name="name">Use web for a webhook or use the name of a valid service. (See /hooks for the list of valid service names.)
```
Haacked: Let's quote "web" here.

Octokit/Models/Request/NewRepositoryWebHook.cs
```diff
@@ -0,0 +1,132 @@
...
+        /// Use "web" for a webhook or use the name of a valid service. (See
```
Haacked: This line looks like it doesn't belong.

Octokit/Models/Request/NewRepositoryWebHook.cs
```diff
@@ -0,0 +1,132 @@
...
+        /// </summary>
+        /// <param name="name">Use web for a webhook or use the name of a valid service. (See /hooks for the list of valid service names.)
+        /// Use "web" for a webhook or use the name of a valid service. (See 
+        /// <see href="https://api.github.com/hooks">https://api.github.com/hooks</see> for the list of valid service
+        /// names.)
+        /// </param>
+        /// <param name="config">
+        /// Key/value pairs to provide settings for this hook. These settings vary between the services and are
+        /// defined in the github-services repository. Booleans are stored internally as “1” for true, and “0” for
+        /// false. Any JSON true/false values will be converted automatically.
```
Haacked: Remove "JSON" here.

Octokit/Models/Request/NewRepositoryWebHook.cs
```diff
@@ -0,0 +1,132 @@
...
+        /// </summary>
+        /// <value>
+        ///  <c>true</c> if SSL certificate verification is not performed; 
+        /// otherwise, <c>false</c>.
+        /// </value>
+        public bool InsecureSsl { get; set; }
+
+        public override NewRepositoryHook ToRequest()
+        {
+            var config = Config.Union(new Dictionary<string, string>
```
Haacked: What happens if `Config` contains a duplicate key? We should test that scenario and define what we expect to happen.

alfhenrik: So right now, it will fail with the following error `An item with the same key has already been added.` :bomb: This is of course due to the way the equality comparison on a key value pair checks the object, not just the key.

So we can introduce a custom equality comparer

```csharp
public class WebHookConfigComparer : IEqualityComparer<KeyValuePair<string, string>>
    {
        public bool Equals(KeyValuePair<string, string> x, KeyValuePair<string, string> y)
        {
            return x.Key == y.Key;
        }

        public int GetHashCode(KeyValuePair<string, string> obj)
        {
            return obj.Key.GetHashCode();
        }
    }
```

and use that like this

```csharp
var allConfigs = Config.Union(new Dictionary<string, string>
            {
                { "url", Url },
                { "content_type", ContentType.ToParameter() },
                { "secret", Secret },
                { "insecure_ssl", InsecureSsl.ToString() }
            }, new WebHookConfigComparer());
```

This way we will only include distinct key/value pairs based on the key itself, the value that would be used is the one that was passed in the `config` parameter in the `ctor`. Personally I'm leaning to that we should be using the values of the `NewRepositoryWebHook` properties to override the values in the `config` parameter...although, maybe we should throw a helpful error message if they pass web hook specific config values in the `config` parameter...thoughts?

Haacked: Yeah, I kind of lean towards throwing an exception. Quietly overriding values could be surprising. Also, if you choose to use the `NewRepositoryWebHook` we should alert you to the fact if you're doing it wrong.

alfhenrik: Yea, I'm not too keen of unexpected behaviour of a framework :smile: so I'm happy to return an error message if duplicate keys exists.

alfhenrik:
We now throw an exception with a more helpful message if the user has added any webhook specific config values to the `config` parameter in the `ctor`.

alfhenrik:
I think this is now done, just awaiting the :white_check_mark: of approval

shiftkey:
Looks like there's a csproj merge conflict here :cry:

alfhenrik:
crap, totally missed your comment... :disappointed: will try and fix it up as soon as possible.

alfhenrik:
OK, merge conflicts sorted, now just waiting for AppVeyor

shiftkey:
I don't think that second build is ever going to complete, so I'm gonna merge this in.

Thanks again @alfhenrik!

alfhenrik:
:+1:

Issue #904, opened 2015-09-19
Investigate and fix the `StopsMakingNewRequestsWhenTakeIsFulfilled` test


haacked:
The test [StopsMakingNewRequestsWhenTakeIsFulfilled](https://github.com/octokit/octokit.net/blob/e6e7c3d393160fabe4cc48dd6c410db2a19b7e72/Octokit.Tests/Reactive/ObservableRepositoriesClientTests.cs#L93-L158) fails consistently now. We need to figure out why and fix the code or the test.

alfhenrik:
I'm unable to replicate this error locally, even running the same commands that AppVeyor runs (i.e. the git commands to get the code) yields a successful unit test run :cry: Could it perhaps be something environmental to AppVeyor...

naveensrinivasan:
Even I couldn't replicate it locally.

alfhenrik:
I'm installing VS2013 on my Server 2012 R2 dev machine and will test it on there as well

alfhenrik:
Everything works fine on my 2012R2 dev machine as well...so I'm quite stump as to why it would be failing on AppVeyor...

chenjiaming93:
it only appears on AppVeyor... Even though I am committing the same content, I cannot duplicate it

PR #918, opened 2015-09-30
Adds overloads to GetArchive for adding custom timeouts 

### LINKED ISSUE #817: control request timeout when downloading archive
In #808 I set 1 hour as the request timeout when handling the redirect.

If you're going to want something different (for example, set it shorter to fail fast), you're going to want to specify overloads for `GetArchive`. I need to focus on more core concerns right now, so here's an opportunity for someone else to jump in and help out.

willsb:
This (hopefully) fixes #817


Octokit.Reactive/Clients/IObservableRepositoryContentsClient.cs
```diff
@@ -97,6 +97,30 @@ public interface IObservableRepositoryContentsClient
...
+        /// Get an archive of a given repository's contents, in a specific format
+        /// </summary>
+        /// <remarks>https://developer.github.com/v3/repos/contents/#get-archive-link</remarks>
+        /// <param name="owner">The owner of the repository</param>
+        /// <param name="name">The name of the repository</param>
+        /// <param name="archiveFormat">The format of the archive. Can be either tarball or zipball</param>
+        /// <param name="reference">A valid Git reference.</param>
+        /// <param name="timeout"> Timeout in minutes </param>
+        /// <returns>The binary contents of the archive</returns>
+        IObservable<byte[]> GetArchive(string owner, string name, ArchiveFormat archiveFormat, string reference, int timeout);
```
Haacked: Honestly, I'd rather not have this method and only have the `TimeSpan` I think it promotes better practices. Also, if we really want this overload, we can add it as an extension method to the interface.

Octokit.Reactive/Clients/ObservableRepositoryContentsClient.cs
```diff
@@ -166,6 +166,36 @@ public IObservable<RepositoryContent> GetAllContents(string owner, string name,
...
+        /// </summary>
+        /// <remarks>https://developer.github.com/v3/repos/contents/#get-archive-link</remarks>
+        /// <param name="owner">The owner of the repository</param>
+        /// <param name="name">The name of the repository</param>
+        /// <param name="archiveFormat">The format of the archive. Can be either tarball or zipball</param>
+        /// <param name="reference">A valid Git reference.</param>
+        /// <param name="timeout"> Time span until timeout </param>
+        /// <returns>The binary contents of the archive</returns>
+        public IObservable<byte[]> GetArchive(string owner, string name, ArchiveFormat archiveFormat, string reference, TimeSpan timeout)
+        {
```
Haacked: We should probably validate that the TimeSpan is not negative as a negative value is almost certainly a programmer error. I'd throw an `ArgumentOutOfRange` exception.

Octokit/Clients/RepositoryContentsClient.cs
```diff
@@ -187,14 +187,44 @@ public Task<string> GetArchiveLink(string owner, string name, ArchiveFormat arch
...
+        /// <param name="owner">The owner of the repository</param>
+        /// <param name="name">The name of the repository</param>
+        /// <param name="archiveFormat">The format of the archive. Can be either tarball or zipball</param>
+        /// <param name="reference">A valid Git reference.</param>
+        /// <param name="timeout"> Time span until timeout </param>
+        /// <returns>The binary contents of the archive</returns>
+        public async Task<byte[]> GetArchive(string owner, string name, ArchiveFormat archiveFormat, string reference, TimeSpan timeout)
         {
             Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
             Ensure.ArgumentNotNullOrEmptyString(name, "name");
```
Haacked: Validate the `TimeSpan` argument here too please.

willsb:
hmm, it seems like the build failed due to #904

Octokit.Reactive/Clients/ObservableRepositoryContentsClient.cs
shiftkey: :thumbsup:

shiftkey:
Apologies for the delay, and thanks for working on this!

PR #919, opened 2015-09-30
add System to required framework assemblies for net45

adamralph:
Currently, installing Octokit into a project which does not have a reference to `System` fails like so:

image

I always strip out all references from a project and only add back in what is required (it's surprising how often something like a simple business logic lib doesn't even need a reference to `System`).

Anyhow, this nasty looking error can be avoided by declaring `System` as a required framework assembly reference.

I've no idea if this is also required for `netcore45`. I tried it out in a sample `netcore45` project but the whole notion of referencing framework assemblies explicitly doesn't even seem to exist there.

### DIFF
--- a/Octokit.nuspec
+++ b/Octokit.nuspec
@@ -26,6 +26,7 @@
       </group>
     </dependencies>
     <frameworkAssemblies>
+      <frameworkAssembly assemblyName="System" targetFramework="net45" />
       <frameworkAssembly assemblyName="System.Net.Http" targetFramework="net45, netcore45" />
     </frameworkAssemblies>
   </metadata>


khellang:
:shipit:

Haacked:
> I've no idea if this is also required for netcore45. I tried it out in a sample netcore45 project but the whole notion of referencing framework assemblies explicitly doesn't even seem to exist there.

Hey @Eilon! Got any suggestions about this?

Eilon:
Sorry not sure about `netcore45` as I've never really use that. But this change as-is certainly seems to fix a real issue.

Eilon:
I talked to @anurse and he wasn't super sure. There is a concept of reference assemblies for `netcore45` (Windows Phone, effectively), but either way a typical Windows Phone app effectively references *everything* on the system anyway (the dependencies are trimmed later).

So as such, I think this change is good as-is.

Haacked:
selfie-3

Eilon:
@Haacked I strongly suspect that GIF will get reused now :smile:

alfhenrik:
Hahah, that's pretty funny :grinning:

Issue #505, opened 2014-06-04
LINQPad samples


shiftkey:
http://www.linqpad.net/nugetsamples.aspx

cc @distantcam

shiftkey:
Taking this off my radar for the foreseeable future.

naveensrinivasan:
@Haacked  I wanted to check if this something that we want to do now.

IMHO having samples would greatly improve people using this for their needs. I think linqpad is a great tool for doing that without much of ceremony. I use it very often to try out new API's. 

Based on this I forked  @shiftkey https://github.com/naveensrinivasan/octokit-linqpad-samples and updated with some more samples.  

I updated the octokit's Nuspec locally with the linqpad-samples 
![linqpad-samples](https://cloud.githubusercontent.com/assets/172697/10124817/33d7626e-652d-11e5-9819-f4a14c5ba7bd.JPG)

and when I use this nuget package it comes up with the samples in linqpad
![samples](https://cloud.githubusercontent.com/assets/172697/10124822/94352984-652d-11e5-9a00-f64a13e64fc5.JPG)

![samples-treeview](https://cloud.githubusercontent.com/assets/172697/10124828/c7097f40-652d-11e5-82c1-fe54cd60bef8.JPG)

If it is interesting, then I could a do a PR with 
* the samples, 
* updated the build to include the linqpad samples and
* also include a test to make sure any changes don't break the samples.

Haacked:
Should this go in the main Octokit package or perhaps a special samples package? For example, we may want to create Octokit.Linqpad and Octokit.Reactive.Linqpad.

naveensrinivasan:
I agree it could more of like `Octokit.Samples` where we could use one of the options as Linqpad and we could use the same samples for XS/VS project sample.

alfhenrik:
:+1: for it to be it's own package

M-Zuber:
I would also prefer to have a separate package

shiftkey:
> Should this go in the main Octokit package or perhaps a special samples package?

The general guidance from the [Linqpad site](https://www.linqpad.net/nugetsamples.aspx) is to contain them in the main package:

> 2. Update your NuGet packing script to copy your samples into a folder called linqpad-samples.
> 3. Add the linqpad-samples tag to your NuGet submission to ensure free access.

I gather this means the user will have them automagically after installing (rather than digging around for a second package).

Also, I'm not sure what that last step means.

naveensrinivasan:
[...]

PR #921, opened 2015-10-02
Adds octokit samples

naveensrinivasan:
Implement LINQPad samples in NuGet packages. #505 

http://www.linqpad.net/nugetsamples.aspx

- [x] Implement LINQPad samples 
- [x] Update build to include the sample as nuget package
- [x]  ~~Write a test~~  Change the `build.fsx` to do a  [lprun](https://www.linqpad.net/lprun.aspx) `-compileonly` check to make sure new check-in's don't break the samples.


naveensrinivasan:
The build failed due to #904

naveensrinivasan:
The build failed because of `StopsMakingNewRequestsWhenTakeIsFulfilled`

PR #923, opened 2015-10-04
Gitignore exception

### LINKED ISSUE #423: Validation checks on NewRepository.GitignoreTemplate
One of our integration tests was failing with this response:

```
{
  "message":"Validation Failed",
  "documentation_url":"http://developer.github.com/v3/repos/#create",
  "errors":
  [{
    "resource":"Repository",
    "code":"custom",
    "field":"gitignore_template",
    "message":"gitignore_template is an unknown gitignore template."
  }]
}
```

This is because we were passing `visualstudio`, rather than `VisualStudio`.

We should surface this as a proper exception message in the future...

willsb:
Fixes #423

### DIFF
--- a/Octokit/Clients/RepositoriesClient.cs
+++ b/Octokit/Clients/RepositoriesClient.cs
@@ -117,10 +117,7 @@ namespace Octokit
                     throw new PrivateRepositoryQuotaExceededException(e);
                 }
                 
-                if (String.Equals(
-                    "gitignore_template is an unknown gitignore template.",
-                    errorMessage,
-                    StringComparison.OrdinalIgnoreCase))
+                if (errorMessage.EndsWith("is an unknown gitignore template.", StringComparison.OrdinalIgnoreCase))
                 {
                     throw new InvalidGitIgnoreTemplateException(e);
                 }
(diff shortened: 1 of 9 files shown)


Haacked:
Ok, the fact that it always says `gitignore_template` at the beginning of the string appears to be a bug in our API. Let's not rely on the bug behavior.

willsb:
Should I wait until the bug is fixed to continue with this one or use the `EndsWith` alternative?

Haacked:
Use the `EndsWith`. It'll be future proof. Hopefully.

Octokit/Clients/RepositoriesClient.cs
```diff
@@ -116,6 +116,12 @@ public Task<Repository> Create(string organizationLogin, NewRepository newReposi
                 {
                     throw new PrivateRepositoryQuotaExceededException(e);
                 }
+                
+                if (errorMessage != null && errorMessage.EndsWith("is an unknown gitignore template.", StringComparison.OrdinalIgnoreCase))
```
shiftkey: Could we clean this up so it looks like the `String.Equals` call above?

shiftkey: I might have been too eager with this, given you need `EndsWith` to verify this. Disregard.

shiftkey:
Thanks for this - I merged in the `csproj` conflict so this is :ship:ed!

Issue #867, opened 2015-08-23
Merged is always false in Pull Request list


lucamorelli:
I'm reading the list of the Pull Request list of a repository looking for all the merge, but I noticed the property **Merged** is always false, although the pull request in the site is merged and also in the request object the property **MergedAt** is not null.
Another thing: How can I find the final commit generated at the end of the merge?

shiftkey:
> I'm reading the list of the Pull Request list of a repository looking for all the merge, but I noticed the property **Merged** is always false, 

I think this is a serialization issue - there's no `merged` field sent by the server, so it should be marked as obsolete in favour of using `MergedAt`. See an earlier discussion on this (at least I think it's the right thing I was ranting about) https://github.com/octokit/octokit.net/issues/502

> Another thing: How can I find the final commit generated at the end of the merge?

The `MergeCommitSha` property should be set when the pull request has been merged, but a quick scan of the codebase indicates that it doesn't match the expected Git commit. 

@pengwynn I'm looking at https://github.com/octokit/octokit.net/pull/857 and I see a comment about https://github.com/octokit/octokit.net/commit/1776ca1086a861b81bd499aaf8a0dd9da76ca4ce being the merge commit.

<img width="749" alt="screen shot 2015-08-24 at 2 24 52 pm" src="https://cloud.githubusercontent.com/assets/359239/9433073/205e9670-4a6c-11e5-93f8-a669c5473e52.png">

However when I retrieve it from the API I get a different value: https://api.github.com/repos/octokit/octokit.net/pulls/587

`merge_commit_sha=e27e3c3c9fafe3b69b238b478981587837fd62f0` appears to be a server-side commit (possibly the merge after it was last pushed).

There's another issue which I'm poking into about how a different value appears when this PR is listed in `/pulls?state=closed` which I still need to dig into, but let me know if I'm missing something obvious...

chenjiaming93:
See https://developer.github.com/v3/pulls/#get-a-single-pull-request
"The merge_commit_sha attribute holds the SHA of the test merge commit; however, this attribute is deprecated and is scheduled for removal in the next version of the API. The Boolean mergeable attribute will remain to indicate whether the pull request can be automatically merged."  
I think we need to set Merged to MergedAt.HasValue? or any other ideas?

Haacked:
I commented on the PR.

shiftkey:
The `MergedAt` change is now available in v0.18. I think we're good here.

[...]

PR #936, opened 2015-10-10
Pullrequestmerged

chenjiaming93:
#867.
Set Merged to be Merged||MergedAt.HasValue. Merged is not always FALSE anymore.

### DIFF
--- a/Octokit/Models/Response/PullRequest.cs
+++ b/Octokit/Models/Response/PullRequest.cs
@@ -14,7 +14,7 @@ namespace Octokit
             Number = number;
         }
 
-        public PullRequest(Uri url, Uri htmlUrl, Uri diffUrl, Uri patchUrl, Uri issueUrl, Uri statusesUrl, int number, ItemState state, string title, string body, DateTimeOffset createdAt, DateTimeOffset updatedAt, DateTimeOffset? closedAt, DateTimeOffset? mergedAt, GitReference head, GitReference @base, User user, string mergeCommitSha, bool merged, bool? mergeable, User mergedBy, int comments, int commits, int additions, int deletions, int changedFiles)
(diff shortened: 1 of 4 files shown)


Octokit/Models/Response/PullRequest.cs
```diff
@@ -34,7 +34,7 @@ public PullRequest(Uri url, Uri htmlUrl, Uri diffUrl, Uri patchUrl, Uri issueUrl
             Base = @base;
             User = user;
             MergeCommitSha = mergeCommitSha;
-            Merged = merged;
+            Merged = merged || MergedAt.HasValue;
```
Haacked: Let's completely remove the `merged` constructor parameter. `Merged` will be a calculated property. So this should be `Merged = MergedAt.HasValue`.

Octokit/Models/Response/PullRequest.cs
```diff
@@ -34,7 +34,7 @@ public PullRequest(Uri url, Uri htmlUrl, Uri diffUrl, Uri patchUrl, Uri issueUrl
             Base = @base;
             User = user;
             MergeCommitSha = mergeCommitSha;
```
Haacked: Lets remove this property and constructor parameter since it's deprecated.

chenjiaming93: well. Another quuestion is merge-commit-sha. It is suggested that we use Mergeable instead of merge-commit-sha. Shall we also remove that? 
https://developer.github.com/v3/pulls/#get-a-single-pull-request

Haacked: Yes, that's what I'm referring to. We should remove `MergeCommitSha`

ryangribble:
Howdy folks, I was just digging into this issue myself (merged always being false on all pull request, even those that have been merged) and came across this PR as well as 2 issues and some other PRs etc.  It seems in the various shuffling that change Merged field to be a calculated field hasnt actually made it through

Is the best course of action for someone to submit a new PR or can this current one be updated and merged in?  If you want me to do it let me know.  In the meantime I can obviously use the workaround of pr.MergedAt.HasValue

Haacked:
This PR appears to be abandoned. I'm opening up a new one #970 to supersede this one.

PR #924, opened 2015-10-04
ye old build script story

shiftkey:
First off, sorry for letting the deployment instructions get out of date and of course something would come up as soon as I go on holiday :cry:

Fixes #899 

Here's me brain-dumping and scripting stuff out, so it won't happen again:

- [x] `script/bootstrap` to setup your repository right
- [x] `script/build` to just build and test
- [x] `script/package` to ensure you can correctly generate packages
- [x] refactor pre-requisite installation into common function for use in both `build` and `package`
- [x] resolve missing dependency for SourceLink (found on multiple environments)
- [x] update and centralize docs
- [ ] the normalizing of the repo seems to fail - why?

There's lots of potential headaches here with running the integration tests (exhausting API limits, hitting anti-spam measures, etc) so that's a lower priority for me right now.

I'll also leave automating the releases in #905 to @naveensrinivasan

PR #940, opened 2015-10-11
Created build.sh

naveensrinivasan:
Included build.sh to build form non-windows.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+if test "$OS" = "Windows_NT"
+then
+  # use .Net
+
+"./tools/nuget/nuget.exe" "install" "xunit.runner.console" "-OutputDirectory" "tools" "-ExcludeVersion" "-version" "2.0.0"
+"./tools/nuget/nuget.exe" "install" "FAKE.Core" "-OutputDirectory" "tools" "-ExcludeVersion" "-version" "4.4.2"
+"./tools/nuget/nuget.exe" "install" "SourceLink.Fake" "-OutputDirectory" "tools" "-ExcludeVersion" "-version" "1.1.0"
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+if test "$OS" = "Windows_NT"
+then
+  # use .Net
+
+"./tools/nuget/nuget.exe" "install" "xunit.runner.console" "-OutputDirectory" "tools" "-ExcludeVersion" "-version" "2.0.0"
+"./tools/nuget/nuget.exe" "install" "FAKE.Core" "-OutputDirectory" "tools" "-ExcludeVersion" "-version" "4.4.2"
+"./tools/nuget/nuget.exe" "install" "SourceLink.Fake" "-OutputDirectory" "tools" "-ExcludeVersion" "-version" "1.1.0"
+packages/FAKE/tools/FAKE.exe $@ --fsiargs -d:MONO build.fsx
```
shiftkey: Do we need to pass in the `MONO` flag here?

shiftkey:
I'll take this in and then rework my stuff in #924 to be x-plat friendly. Just that one question.

build.sh
naveensrinivasan: It is for conditional compilation to `MONO`. I don't think so.

naveensrinivasan:
@shiftkey If you are going to pull this then can I  close this PR?

shiftkey:
@naveensrinivasan sorry, I should have been clearer. As this is new code (around an area we don't currently support) I'm happy to take this in and then iterate on it.

build.sh
shiftkey: :thumbsup:

naveensrinivasan:
Thanks. I was thinking we should start supporting mono, its just the question of making the builds ready for it. The code supports it. Once the build scripts are ready for non-windows we could even enable builds on https://travis-ci.org/

PR #943, opened 2015-10-16
Fixes for Downloading ReleaseAsset zip File 

### LINKED ISSUE #854: Downloading ReleaseAsset zip File - NullReferenceException
I am trying to download a zip file that is an asset in a release. 

```c#
var request = client.Release.GetAll("username", "repo");
var releases = await request;
var latest = releases[0];
var assets = await client.Release.GetAllAssets("username", "repo", latest.Id);
var myAsset_zipFile = assets[0];

var resp= await client.Connection.Get<byte[]>(new Uri(myAsset_zipFile.Url),
          new Dictionary<string, string>(),
          null); //NullReference Here

var data = resp.Body;
var responseData = resp.HttpResponse.Body;
System.IO.File.WriteAllBytes("myZip.zip", (byte[])responseData);
```

I am getting a NullReferenceException in SimpleJson.cs in the DeserializeObject function at the line that reads obj = ConstructorCache\[type\](); when I call client.Connection.Get.

If I have trying to get this to work without success for some time and have tried several different methods. 

It is worth mentioning that I am trying to download from a private repo.

Potentially Related #792

naveensrinivasan:
This commit  addresses the http response which wasn't handling response `content-type` `application/octet-stream` for binary items.

Fixes #854


Octokit/Http/HttpClientAdapter.cs
```diff
@@ -70,16 +70,22 @@ static CancellationToken GetCancellationTokenForRequest(IRequest request, Cancel
 
             object responseBody = null;
             string contentType = null;
+
+            // We added support for downloading images,zip-files and application/octet-stream. 
+            // Let's constrain this appropriately.
+            var binaryContentTypes = new[] {"application/zip"
+                ,"application/x-gzip",
+                "application/octet-stream"};
```
naveensrinivasan: I don't know if the response content type could be anything else other than `application/octet-stream`. From the test's that I have done with API  for `zip` and `txt` the response content type has only been `application/octet-stream`. Do we need the other types?

shiftkey: I wouldn't worry about other types here...

Octokit/Http/HttpClientAdapter.cs
```diff
@@ -70,16 +70,22 @@ static CancellationToken GetCancellationTokenForRequest(IRequest request, Cancel
...
                 {
                     contentType = GetContentMediaType(responseMessage.Content);
 
-                    // We added support for downloading images and zip-files. Let's constrain this appropriately.
-                    if (contentType != null && (contentType.StartsWith("image/")
-                        || contentType.Equals("application/zip", StringComparison.OrdinalIgnoreCase)
-                        || contentType.Equals("application/x-gzip", StringComparison.OrdinalIgnoreCase)))
+                    if (contentType != null && (contentType.StartsWith("image/") ||
+                        binaryContentTypes
+                        .Any(item => item.Equals(contentType,StringComparison.OrdinalIgnoreCase))))
```
shiftkey: :lipstick: bring this up to the previous line, and a space after the `,`

Octokit/Http/HttpClientAdapter.cs
```diff
@@ -70,16 +70,22 @@ static CancellationToken GetCancellationTokenForRequest(IRequest request, Cancel
 
             object responseBody = null;
             string contentType = null;
+
+            // We added support for downloading images,zip-files and application/octet-stream. 
+            // Let's constrain this appropriately.
+            var binaryContentTypes = new[] {"application/zip"
```
shiftkey: :lipstick: align all of these with a trailing `,` on each line (where relevant)

shiftkey:
I think this is good to go. @Haacked?

Haacked:
Awww yeah!

PR #955, opened 2015-10-31
Prepping searching code doc

hahmed:
Searching tha code docs.

### DIFF
--- a/docs/search.md
+++ b/docs/search.md
@@ -184,7 +184,61 @@ var request = new SearchRepositoriesRequest("mvc client side framework")
 
 ## Search Code
 
-**TODO**
+To search for code using Octokit you need to create a `SearchCodeRequest` and populate it with the search criteria.
+
+```csharp
+// Initialize a new instance of the SearchCodeRequest class with a search term
+var request = new SearchCodeRequest("auth");
+
+// Or we can restrict the search to a specific repo
+var request = new SearchCodeRequest("auth", "dhh", "rails");
+
+var result = await githubClient.Search.SearchCode(request);
+```
+
+Now we can further filter our search.
+
+```csharp
+var request = new SearchCodeRequest("auth")
+{
+    // maybe we want to restrict the search to the file only
+    In = new[] { CodeInQualifier.File },
+
+    // or we can search the file and the path too
+    In = new[] { CodeInQualifier.File, CodeInQualifier.Path },
+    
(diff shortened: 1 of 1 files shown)


docs/search.md
```diff
@@ -184,7 +184,61 @@ var request = new SearchRepositoriesRequest("mvc client side framework")
...
+Now we can further filter our search.
+
+```csharp
+var request = new SearchCodeRequest("auth")
+{
+    // maybe we want to restrict the search to the file only
+    In = new[] { CodeInQualifier.File },
+
+    // or we can search the file and the path too
+    In = new[] { CodeInQualifier.File, CodeInQualifier.Path },
```
naveensrinivasan: I would prefer code that would compile. This one will fail to compile because there is duplication of `In`

naveensrinivasan:
```
var request = new SearchCodeRequest("auth")
{
	// maybe we want to restrict the search to the file only
	In = new[] { CodeInQualifier.File },


	// how about we find a file based on a certain language
	Language = Language.JavaScript,

	// do we want to search forks too?
	Forks = true,

	// find files that are above 1000 bytes
	Size = Range.GreaterThan(1000),

	// we may want to restrict the search to the path of a file
	Path = "app/assets",

	// we may want to restrict the file based on file extension
	Extension = "json",

	// restrict search to a specific file name
	FileName = "app.json",

	// search within a users or orgs repo
	User = "dhh"
};
```
The above sample does not produce a result. My two cents we should have a sample that produces some result.

hahmed:
Granted it does not compile, however documentation is generally letting the user know how to use the api in a more digestible way. I see the docs as *complementing* the code, tests or even to an extent intellisense.

There have been too many times where I do not have a clue about a library, I simply want to pick up an example that just works. I do agree that maybe after each feature we have a working example? 

cc/ @shiftkey

shiftkey:
> My two cents we should have a sample that produces some result.

I think this is a nice goal, but perhaps we can address this elsewhere - with more interactive or detailed examples? For these sections, we should be focusing on understanding the API options available.

docs/search.md
shiftkey: :thumbsup: I think we should keep this second line around instead of both

hahmed: Done.

shiftkey:
Thanks!

PR #956, opened 2015-11-03
VS2015 migration

shiftkey:
- [x] can build under VS2015 locally
- [x] can build on AppVeyor
- [x] `netcore45` -> `netcore451` target it any more
- [x] test and verify standard version can build and test this
- [x] docs
- [x] CI back to green
- [x] run integration tests
- [x] test preview package in a Store app

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -83,3 +83,6 @@ tools/xunit.runner.console
 # New VS Test Runner creates arbitrary folders with PDBs
 *.pdb
 pingme.txt
+
+# it's 2015, no need to keep this around when migrating projects
+Backup/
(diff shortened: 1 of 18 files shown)


Haacked:
Nice! I know @Eilon was just asking me if we could switch Octokit.net to VS 2015. I think he even offered to help. 😛

shiftkey:
@Haacked it looks like the xml-doc warnings are causing the build to fail incorrectly. I *really* don't want to go through and triage it all, but maybe I have to?

Haacked:
I would set the doc warnings as warnings and not errors and assign it to me to fix later.

shiftkey:
@Haacked good point, I'll do that now and see if the build is happy with that...

Eilon:
Looks :shipit: to me!

naveensrinivasan:
:+1:

Issue #800, opened 2015-05-11
 client.Repository.Content.GetAllContents for root not currently possible.


DavidStrickland0:
Task<IReadOnlyList<RepositoryContent>> GetAllContents(string owner, string name, string path);
Will currently only work with subdirectories ie:
`client.Repository.Content.GetAllContents("octokit", "octokit.net", "Octokit")` returns 18 items.
However
`client.Repository.Content.GetAllContents("octokit", "octokit.net", "")` Throws a ArgumentException("String cannot be empty", name)

For a quick fixed I hacked GetAllContents as follows locally but a better option is probably needed.

```cs
public async Task<IReadOnlyList<RepositoryContent>> GetAllContents(string owner, string name, string path)
        {
            Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
            Ensure.ArgumentNotNullOrEmptyString(name, "name");
            //Ensure.ArgumentNotNullOrEmptyString(path, "path");
            System.Uri url;
            if (string.IsNullOrEmpty(path))
            {
                url = ApiUrls.RepositoryContent(owner, name);
            }
            else
            {
                url = ApiUrls.RepositoryContent(owner, name, path);
            }
            return await ApiConnection.GetAll<RepositoryContent>(url);
        }
```

SplitInfinity:
I am able to get the contents of a repository's root directory by passing in "/" as the path. I am unsure if this is due to changes made after this issue was opened.

ostruk:
:+1: 

"/" only works if you're retrieving master branch. If you're retrieving some other branch it defaults to master branch, because the generated url is actually invalid: "contents//?ref=branch-name". Remove the check for empty string when specifying path.

naveensrinivasan:
You could get all the contents of the branch using the overload `client.Repository.Content.GetAllContents`  which takes the `reference` as the last parameter.


I am able to get all the contents of the branch and master including the sub-directories. Here is a sample.

```
async Task Main(string[] args)
{
	var owner = "octokit";
	var reponame = "octokit.net";

	GitHubClient client = new GitHubClient(new Octokit.ProductHeaderValue("Octokit.samples"));
	client.Credentials = new Credentials(Util.GetPassword("github"));
	var branchcontents = await client.Repository.Content.GetAllContents(owner,reponame,"/Octokit.Reactive/Properties","codeformatter");
	var mastercontents = await client.Repository.Content.GetAllContents(owner,reponame,"/Octokit.Reactive/Properties");
[...]

PR #959, opened 2015-11-04
Fix for client.Repository.Content.GetAllContents for root #800

naveensrinivasan:
The client.Repository.Content.GetAllContents #800  does not allow getting root folders for Branches. This fix would allow the users to pass null/empty strings for the path which will address the issue.


shiftkey:
While dropping the constraint means that you can get to the root of the repository, I think this behaviour is still unclear. Would it be smarter to have an overload that doesn't take a path?

```csharp
public async Task<IReadOnlyList<RepositoryContent>> GetAllContents(string owner, string name)
```

naveensrinivasan:
So without the path the function would only return the root folder which also takes the `ref` in the argument?

If that is the case then the function would have to be named something like `GetAllContentsOfRoot`
```
public async Task<IReadOnlyList<RepositoryContent>> GetAllContentsOfRoot(string owner, string name,string ref)

```

shiftkey:
@naveensrinivasan you're right, I hadn't thought about the refs side. While we are adding in a lot of overloads here, I think they all end up going through one method - so we should be fine.

naveensrinivasan:
@shiftkey 
``` csharp
public async Task<IReadOnlyList<RepositoryContent>> GetAllContents(string owner, string name, string reference)
```
Cannot have this because there is already another overload with the same parameter types

``` csharp
public async Task<IReadOnlyList<RepositoryContent>> GetAllContents(string owner, string name, string path)
```

What do you want to do?

Haacked:
What if we name these `GetAllContentsByRef` and `GetAllContentsByPath`?

naveensrinivasan:
Sounds good to me!

naveensrinivasan:
Implemented changes. The method names are `GetAllContents` and `GetAllContentsByRef`. I reused the  `GetAllContents` name because in one of the overloads we don't pass the `path` and it does not make sense to have a function name as  `GetAllContentsByPath` and don't have the path in argument.

Octokit.Reactive/Clients/IObservableRepositoryContentsClient.cs
```diff
@@ -124,7 +134,19 @@ public interface IObservableRepositoryContentsClient
         /// <returns>
         /// A collection of <see cref="RepositoryContent"/> representing the content at the specified path
         /// </returns>
-        IObservable<RepositoryContent> GetAllContents(string owner, string name, string path, string reference);
+        IObservable<RepositoryContent> GetAllContentsByRef(string owner, string name, string path, string reference);
```
Haacked: I would reorder `path` and `reference`. The reason is the other overload to this method takes in `owner`, `name`, `reference`. If the user is calling that overload, then decides, no, they want to call this one, they only need to tack on the `path` parameter. They don't need to re-order parameters. Based on my experience, this is a common source of API call errors.

Octokit.Reactive/Clients/ObservableRepositoryContentsClient.cs
```diff
@@ -166,6 +166,25 @@ public IObservable<RepositoryContent> GetAllContents(string owner, string name,
...
+        /// </returns>
+        public IObservable<RepositoryContent> GetAllContents(string owner, string name)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
+            Ensure.ArgumentNotNullOrEmptyString(name, "name");
+
+            return _client
+                .Connection
+                .GetAndFlattenAllPages<RepositoryContent>(ApiUrls.RepositoryContent(owner, name, string.Empty));
+
```
Haacked: Remove extraneous newline here.

Haacked:
:bread: Back to you.

naveensrinivasan:
Done with the changes.

Haacked:
Hmm, @naveensrinivasan, the branch still has conflicts that need to be resolved.

shiftkey:
Looks like some extra changes are in here. I've extracted the relevant commits to #1064 to bring this up to date.

Octokit.Reactive/Clients/IObservableRepositoryContentsClient.cs
tyadav492: path parameter is not accepting any special char like '#' and space. I am trying to retrieve files form directory and i am getting not found exception and under same root directory it is getting all file which does not have any special char.
I have two sub directories : 1) C# Code  and 2) TestApp. i am able to access TestApp but not "C# Code" as it is throwing not found exception. suggest me any solution for this as i am not able to rename the directory sue to some reason.

Issue #963, opened 2015-11-17
Uploading release assets in parallel results in TaskCanceledException


eiriktsarpalis:
I'm maintaining a project which requires a large number of files to be uploaded with each release. Roughly 30 files of 8MB each. My current build script uploads the files in sequence, which can take an annoying amount of time. I've been investigating whether I could use octokit to upload assets in parallel, but this has resulted in the upload tasks returning random TaskCanceledExceptions. Reproduction is simple really, in F#

``` fsharp
let client : GitHubClient = ...
let draftRelease : Release = ...
let uploadFile (fileName : string) : Task<ReleaseAsset> =
    let fi = FileInfo(fileName)
    let fs= File.OpenRead(fi.FullName)
    let assetUpload = new ReleaseAssetUpload(fi.Name,"application/octet-stream",fs,Nullable<TimeSpan>())
    client.Release.UploadAsset(draftRelease, assetUpload)

let globalTask = files |> Array.map uploadFile |> Task.WhenAll
globalTask.Result
```

naveensrinivasan:
The reason for the `TaskCanceledExceptions` is because the  default [HttpClient] (https://msdn.microsoft.com/en-us/library/system.net.httpwebrequest.timeout(v=vs.100).aspx) `Timeout` is 100 seconds. 

I tried changing in the sample from `Nullable<TimeSpan>()`  to `new TimeSpan(0, 10, 0)` to see if that would fix. But the [HttpClientAdapter] (https://github.com/octokit/octokit.net/blob/f0111aa72239da06c6268ae0c47562957f6603bc/Octokit/Http/HttpClientAdapter.cs) that is implemented does not set the timeout passed in the request to the `HttpClient`. The `Timeout` is being used only used in the `CancellationTokenSource` https://github.com/octokit/octokit.net/blob/f0111aa72239da06c6268ae0c47562957f6603bc/Octokit/Http/HttpClientAdapter.cs#L53-65

This is a bug. The `Timeout` property should set to the `TimeSpan` set in the Request. 

I'll do a PR for this. 

By the time if you want a workaround you could have a `HttpClient` class that implements `IHttpClient` and pass that to the `Connection` class like this.

```csharp
	var client = new Connection(new Octokit.ProductHeaderValue("uploadrelease"),new YourOwnHttpClient())
```

eiriktsarpalis:
Do you have any updates on this fix? Thanks.

shiftkey:
[...]

PR #807, opened 2015-05-19
added a tailored CodeFormatter to Octokit

shiftkey:
This is wired up in the build script so you can run it at your leisure:

`.\build FormatCode`

The source code for the tool: https://github.com/shiftkey/codeformatter/pull/1

I disabled a couple of rules from the upstream codeformatter repo which go against our coding style:

 - new line above first `using` statement
 - specify the visibility of a field/member explicitly
 - private fields naming style

The net result is that things are actually pretty good, and we get some cleanup of whitespace for free.

- [x] is the unicode (c) correct inside `AssemblyInfo.cs`?
-  ~~[x] update `AssemblyInfo.cs` contents~~ :boot:ed
- [x] braces for empty constructors - what rule is this?
-  ~~[ ] remove copyright rename rule~~ :boot:ed
- [x] wait for #799 to land and then rebase this 
- [x] docs docs docs
- [x] hold for VS2015 support
- [x] run over codebase once #808 is merged

PR #965, opened 2015-11-24
Provided a httpclient timeout option

naveensrinivasan:
The `HttpClient` does not take the `Timeout` argument which by default is 100 seconds. Implemented a change where the user has an option to provide the timeout. This is to address the #963.

### DIFF
--- a/Octokit.Tests.Integration/Helper.cs
+++ b/Octokit.Tests.Integration/Helper.cs
@@ -131,6 +131,7 @@ namespace Octokit.Tests.Integration
         {
             return new GitHubClient(new ProductHeaderValue("OctokitTests"));
         }
+        
         public static IGitHubClient GetAnonymousClient(TimeSpan httpTimeout)
         {
             return new GitHubClient(new ProductHeaderValue("OctokitTests"),httpTimeout);
--- a/Octokit.Tests.Integration/Helper.cs
+++ b/Octokit.Tests.Integration/Helper.cs
@@ -118,6 +118,7 @@ namespace Octokit.Tests.Integration
             };
         }
 
+
         public static GitHubClient GetAuthenticatedApplicationClient()
         {
             return new GitHubClient(new ProductHeaderValue("OctokitTests"))
@@ -129,6 +130,10 @@ namespace Octokit.Tests.Integration
         public static IGitHubClient GetAnonymousClient()
         {
             return new GitHubClient(new ProductHeaderValue("OctokitTests"));
+        }
+        public static IGitHubClient GetAnonymousClient(TimeSpan httpTimeout)
+        {
+            return new GitHubClient(new ProductHeaderValue("OctokitTests"),httpTimeout);
         }
 
         public static IGitHubClient GetBadCredentialsClient()
--- a/Octokit/GitHubClient.cs
+++ b/Octokit/GitHubClient.cs
@@ -27,7 +27,6 @@ namespace Octokit
         {
         }
 
-
         /// <summary>
         /// Create a new instance of the GitHub API v3 client pointing to 
         /// https://api.github.com/
--- a/Octokit/GitHubClient.cs
+++ b/Octokit/GitHubClient.cs
@@ -27,6 +27,23 @@ namespace Octokit
         {
         }
 
+
+        /// <summary>
+        /// Create a new instance of the GitHub API v3 client pointing to 
+        /// https://api.github.com/
+        /// </summary>
+        /// <param name="productInformation">
+        /// The name (and optionally version) of the product using this library. This is sent to the server as part of
+        /// the user agent for analytics purposes.
+        /// </param>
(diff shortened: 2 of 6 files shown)


shiftkey:
Yeah, this is something I'd like to make available to consumers - I've just been grappling with how to get there due to how we're currently using `HttpClient` - see #781 for all the details.

Your solution highlights the pain we're currently in with respect to overloads to `GitHubClient`, and most of the `ctor` parameters can actually be used for `HttpClient`. I don't feel great about layering more on there, but perhaps that's a reasonable short-term fix.

It's been almost six months since I last thought about this, so I'll go back to this and revisit that topic.

naveensrinivasan:
Wasn't aware of that.

Octokit.Tests.Integration/Helper.cs
```diff
@@ -130,6 +131,10 @@ public static IGitHubClient GetAnonymousClient()
         {
             return new GitHubClient(new ProductHeaderValue("OctokitTests"));
         }
+        public static IGitHubClient GetAnonymousClient(TimeSpan httpTimeout)
```
Haacked: Please separate methods with newlines.

Octokit/GitHubClient.cs
```diff
@@ -27,6 +27,23 @@ public GitHubClient(ProductHeaderValue productInformation)
...
+        /// https://api.github.com/
+        /// </summary>
+        /// <param name="productInformation">
+        /// The name (and optionally version) of the product using this library. This is sent to the server as part of
+        /// the user agent for analytics purposes.
+        /// </param>
+        /// <param name="httpTimeout">
+        /// The number of milliseconds to wait before the request times out. The default HTTP timeout is 100000.
+        /// </param>
+        public GitHubClient(ProductHeaderValue productInformation,TimeSpan httpTimeout)
```
Haacked: Add a space after the comma.

Octokit/GitHubClient.cs
```diff
@@ -40,6 +57,22 @@ public GitHubClient(ProductHeaderValue productInformation, ICredentialStore cred
             : this(new Connection(productInformation, credentialStore))
         {
         }
+        /// <summary>
```
Haacked: Newline between previous method and start of this one.

Octokit/GitHubClient.cs
```diff
@@ -56,6 +89,25 @@ public GitHubClient(ProductHeaderValue productInformation, Uri baseAddress)
         {
         }
 
+
```
Haacked: No need for two newlines though. :stuck_out_tongue:

Octokit/GitHubClient.cs
```diff
@@ -27,6 +27,23 @@ public GitHubClient(ProductHeaderValue productInformation)
...
+        /// <summary>
+        /// Create a new instance of the GitHub API v3 client pointing to 
+        /// https://api.github.com/
+        /// </summary>
+        /// <param name="productInformation">
+        /// The name (and optionally version) of the product using this library. This is sent to the server as part of
+        /// the user agent for analytics purposes.
+        /// </param>
+        /// <param name="httpTimeout">
+        /// The number of milliseconds to wait before the request times out. The default HTTP timeout is 100000.
```
Haacked: This comment isn't accurate. The parameter is a `TimeSpan`, not an integer representing milliseconds.

naveensrinivasan:
Fixed the line breaks and spaces. @shiftkey Will the [CodeFormatter] (https://github.com/octokit/octokit.net/pull/807)  address these line breaks and spaces?

shiftkey:
@naveensrinivasan @Haacked yes, if you're on VS2015 you should be able to run it from the command line: `.\build FormatCode`

If that's not working As Advertised™, I can investigate further.

naveensrinivasan:
When I branched the VS2015  branch wasn't available. Will do that going forward. Thanks!

shiftkey:
@naveensrinivasan first off, thanks for working on this. I should have made my reservations about adding another `ctor` parameter clearer, as I'd been trying to think of a better solution to this problem but was sidetracked as usual.

I'm not enthusiastic about adding in yet another `ctor` parameter to `GitHubClient` - mostly due to the current complexity of how timeouts work with the HTTP stack, which you've called out elsewhere. The other concern I have here is about supporting this long-term - at some point I'd love to support a smaller set of `ctors` on `GitHubClient`, but that's a debate that has yet to happen so I'm just speaking for myself right now.

So, while I feel awful for closing this PR, I'm trying to make amends by opening a discussion about making the whole HTTP stack more friendly to Octokit consumers #984. I've also opened the first PR which takes us down that path #985 so we can talk about tangible things. I'd love your feedback on either (or both!) topic, as I don't think there's as much work involved as I previously expected - we just need to sit down and work through it...

naveensrinivasan:
@shiftkey  Not an issue. :+1:

Issue #967, opened 2015-12-01
Gists with Unicode Cause Serialization Issues


ferventcoder:
I was/am working with Robert S from support. He suggested we write up an issue for you although I am not sure it is an issue in Octokit. I've captured some output request/responses from what Octokit sends to Github with Fiddler: https://gist.github.com/ferventcoder/83fa0e5b93202537aa25

The JSON validates in a JSON-lint validator. ~~In fact the same request will later pass without issue.~~ (I should validate this one for sure, we've been seeing more now that we are getting closer to the end of our queue). It just feels that there may be something wrong with the underlying API that is a bit flaky.

We've come to lovingly refer to our Gist creation error emails (gathered and sent as one email every 15 minutes) as our heartbeat that the service is still running.

That said, we create a lot of gists: https://gist.github.com/choco-bot

ferventcoder:
I just realized I didn't share the error output:

~~~
There are 1 exceptions of 'ApiException'.Octokit.ApiException: Problems parsing JSON
   at Octokit.Connection.HandleErrors(IResponse response) in c:\dev\octokit.net\Octokit\Http\Connection.cs:line 569
   at Octokit.Connection.<RunRequest>d__2b.MoveNext() in c:\dev\octokit.net\Octokit\Http\Connection.cs:line 547
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Octokit.Connection.<Run>d__27`1.MoveNext() in c:\dev\octokit.net\Octokit\Http\Connection.cs:line 531
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Octokit.ApiConnection.<Post>d__12`1.MoveNext() in c:\dev\octokit.net\Octokit\Http\ApiConnection.cs:line 195
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #972, opened 2015-12-06
Json serialization for Unicode

naveensrinivasan:
This will fix the Unicode Json serialization bug https://github.com/octokit/octokit.net/issues/967.

When I was looking at the code I realized there was already a fix with a PR https://github.com/facebook-csharp-sdk/simple-json/pull/66 .Reused the same code for this fix.

### DIFF
--- a/Octokit.Tests/SimpleJsonSerializerTests.cs
+++ b/Octokit.Tests/SimpleJsonSerializerTests.cs
@@ -69,21 +69,28 @@ namespace Octokit.Tests
             [Fact]
             public void HandleUnicodeCharacters()
             {
+                const string backspace = "\b";
+                const string tab = "\t";
+
                 var sb = new StringBuilder();
                 sb.Append("My name has Unicode characters");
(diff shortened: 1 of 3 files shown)


ferventcoder:
Sweet. :)

Haacked:
I'm hesitant to take this because `SimpleJson.cs` is a dependency we pull in from NuGet. Have you tried the latest version to see if it already has a fix?

If not, could you submit a PR upstream to https://github.com/facebook-csharp-sdk/simple-json?

I worry that the next time we upgrade, we'll break this. At least the unit test will catch it though. :smile:

naveensrinivasan:
@Haacked  I looked for that first before making the change. https://github.com/facebook-csharp-sdk/simple-json hasn't been updated in the last 20 months. There is already a PR with this fix for the main repo which is still pending from April. That's the reason for making a change in this repo.

I understand your concerns. What do you want to do?

khellang:
Ping @prabirshrestha and ask what he's been up to? :wink:

Haacked:
Ok, as long as the PR is upstream, I'll take this one.

PR #973, opened 2015-12-06
Generate nuget packages on appveyor 

naveensrinivasan:
Generate nuget packages on appveyor is something that will help consuming the latest master build without waiting for a release 

For example in this  https://github.com/octokit/octokit.net/issues/967#issuecomment-162253541 the consumer of the package had to go through the hazzle of building nuget locally to test this. 

I was enable to enable the nuget packages on my appveyor builds 
https://ci.appveyor.com/nuget/octokit-net-naveen

As part of this PR the main appveyor build settings page has to be updated with a meaningful nuget name which I think is owned by @Haacked 

Here is mine
https://ci.appveyor.com/project/naveensrinivasan/octokit-net/settings/nuget

image

After updating that we could include that in the README.md so that anyone can consume the latest builds.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,9 +1,10 @@
 init:
   - git config --global core.autocrlf input
 build_script:
-  - cmd: build.cmd CreatePackages
+  - cmd: build.cmd BuildApp
   - cmd: build.cmd UnitTests
   - cmd: build.cmd ConventionTests
+  - cmd: build.cmd CreatePackages
 test: off
 nuget:
   account_feed: true
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,10 +1,9 @@
 init:
   - git config --global core.autocrlf input
 build_script:
-  - cmd: build.cmd BuildApp
+  - cmd: build.cmd CreatePackages
   - cmd: build.cmd UnitTests
   - cmd: build.cmd ConventionTests
-  - cmd: build.cmd CreatePackages
 test: off
 nuget:
   account_feed: true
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -1,7 +1,13 @@
...
 build_script:
-  - cmd: build.cmd BuildApp
+  - cmd: build.cmd CreatePackages
   - cmd: build.cmd UnitTests
   - cmd: build.cmd ConventionTests
 test: off
+nuget:
+  account_feed: true
+  project_feed: true
+  disable_publish_on_pr: true
```
shiftkey: I'm not 100% clear on this value - if this is disabled, it would mean packages aren't created for PRs - which is the inspiration for this change. Or am I mis-understanding something?

naveensrinivasan:
IMO the nuget package using appevyor should be from the main instread for all the PR's `disable_publish_on_pr: true` this would make the consumers point an use the latest stable version of the main instead of  waiting for the release.

Haacked:
> IMO the nuget package using appevyor should be from the main instread for all the PR's disable_publish_on_pr: true this would make the consumers point an use the latest stable version of the main instead of waiting for the release.

:thumbsup:

Haacked:
I changed the feed to https://ci.appveyor.com/nuget/octokit-net.

Haacked:
Thanks @naveensrinivasan!

naveensrinivasan:
@Haacked  Do you want to publish on PR's  `disable_publish_on_pr: true` ?

Haacked:
> Do you want to publish on PR's disable_publish_on_pr: true ?

Nope.

Haacked:
Although, I can see the benefit of that setting for testing. Until someone asks for it, I'll leave everything the way it is.

PR #974, opened 2015-12-06
ToString for ApiException should contain response payload

### LINKED ISSUE #945: when exception raised, ToString should contain response payload
This will help for debugging scenarios where an intermittent error occurs.

Originally raised in #944

asizikov:
Overrided `ToString` methods for `ApiException` and `ApiError`
Fixes #945 

I'm not really sure about format though. Right now `ToString` produces the following output: 

```
Octokit.ApiException: Validation Failed
Octokit.ApiErrorMessage: Validation Failed
Documentation url: 
Errors:
Message: key is already in use
Code: custom
Field: key
Resource: PublicKey
```

- [x] agree on message format
- [x] update current message format if needed

### DIFF
--- a/Octokit/Models/Response/ApiError.cs
+++ b/Octokit/Models/Response/ApiError.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Text;
 
 namespace Octokit
 {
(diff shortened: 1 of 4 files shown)


asizikov:
@shiftkey could you please share your vision about the format of message?

Haacked:
I think we should just show the Response payload. It'll omit the fields that don't exist.

asizikov:
@Haacked so I just should store raw unparsed payload in the `ApiException`? Ok, will update the PR tomorrow. Thanks for quick reply!

Haacked:
@asizikov Yeah, if that's easy. @shiftkey what do you think?

shiftkey:
@Haacked @asizikov  make it so!

asizikov:
Hmm, the build seems to be broken. I guess it's not related to my changes.

Anyway, here is the new output sample: 

```
Octokit.ApiException: Validation Failed
{"errors":[{"code":"custom","field":"key","message":"key is already in use","resource":"PublicKey"}],"message":"Validation Failed"}
```

shiftkey:
@asizikov yeah, that's #983

asizikov:
oh, cool, you're aware of it :+1:

shiftkey:
@asizikov if you merge `master` into your branch you should have a green build again

Octokit/Models/Response/ApiError.cs
```diff
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Text;
```
shiftkey: :lipstick: this namespace doesn't seem necessary

asizikov: yup, it doesn't. 
Deleted.

asizikov:
Aaand it's green now!

shiftkey:
@asizikov thanks!

PR #975, opened 2015-12-07
Implemented GetLatestRelease API.

### LINKED ISSUE #966: Get latest release?
I currently don't see anything to *cleanly* get the [latest release](https://developer.github.com/v3/repos/releases/#get-the-latest-release)...unless I've just been overlooking something obvious, which is possible.

chenjiaming93:
trying to fix
#966

### DIFF
--- a/Octokit/Helpers/ApiUrls.cs
+++ b/Octokit/Helpers/ApiUrls.cs
@@ -140,6 +140,17 @@ namespace Octokit
         }
 
         /// <summary>
+        /// Returns the <see cref="Uri"/> that returns the latest release for the specified repository
+        /// </summary>
+        /// <param name="owner">The owner of the repository</param>
+        /// <param name="name">The name of the repository</param>
+        /// <returns></returns>
+        public static Uri LatestReleases(string owner, string name)
+        {
+            return "repos/{0}/{1}/releases/latest".FormatUri(owner, name);
+        }
+
+        /// <summary>
         /// Returns the <see cref="Uri"/> that returns all the assets for the specified release for the specified repository.
         /// </summary>
         /// <param name="owner">The owner of the repository</param>
--- a/Octokit/Helpers/ApiUrls.cs
+++ b/Octokit/Helpers/ApiUrls.cs
@@ -145,7 +145,7 @@ namespace Octokit
         /// <param name="owner">The owner of the repository</param>
         /// <param name="name">The name of the repository</param>
         /// <returns></returns>
-        public static Uri LatestReleases(string owner, string name)
+        public static Uri LatestRelease(string owner, string name)
         {
             return "repos/{0}/{1}/releases/latest".FormatUri(owner, name);
         }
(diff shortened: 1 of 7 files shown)


Octokit/Helpers/ApiUrls.cs
```diff
@@ -140,6 +140,17 @@ public static Uri Releases(string owner, string name, int id)
         }
 
         /// <summary>
+        /// Returns the <see cref="Uri"/> that returns the latest release for the specified repository
+        /// </summary>
+        /// <param name="owner">The owner of the repository</param>
+        /// <param name="name">The name of the repository</param>
+        /// <returns></returns>
+        public static Uri LatestReleases(string owner, string name)
```
shiftkey: :lipstick: perhaps make this singular i.e. `LatestRelease`

shiftkey:
Looking good so far @chenjiaming93! 

A couple of things I'd like to verify as part of integration tests:

 - that we're able to deserialize a release correctly using this endpoint
 - what happens when no releases exist for a project?

chenjiaming93:
I have added tests. This one is similar to GetRelease API. Just replaced release id with "latest".  In the tests for getrelease we are only checking if we can receive the get request at that endpoint and throw proper exceptions when missing arguments. If there is no latest release, are we expected to get a Null JSON or we a  500 error?

shiftkey:
@chenjiaming93 we have some existing tests which use the real GitHub API - let's add a couple of tests for this endpoint there: https://github.com/octokit/octokit.net/blob/master/Octokit.Tests.Integration/Clients/ReleasesClientTests.cs

A couple of tests I think we can do to complete this feature:

 - "get the latest release for Octokit" - we get a non-null `Release` back
 - "get the latest release for a new repository" -> get a helpful error back

>  If there is no latest release, are we expected to get a Null JSON or we a 500 error?

I'm honestly not sure - let's find out by writing the test!

I think the end result should be a custom exception that we throw to the user. There's a detailed example [here](https://github.com/octokit/octokit.net/blob/f0111aa72239da06c6268ae0c47562957f6603bc/Octokit/Clients/RepositoriesClient.cs#L75-L127) for how we handle when creating a repository fails - I think you'll have to do something similar and `try-catch` the call and then check if the error is something we're familiar with.

chenjiaming93:
@shiftkey  I see. I will add integration tests soon.

khalidabuhakmeh:
Any update on this issue?

chenjiaming93:
@khalidabuhakmeh  I am busy with my final exams this week. I will add integration tests this weekend.

Haacked:
Happy New Year! @chenjiaming93 were you still planning to add integration tests?

chenjiaming93:
@Haacked  I'm just back to school. Will start working tmr

shiftkey:
I've moved the integration tests out to #1086 rather than hold this up any longer.

Thanks for the contribution @chenjiaming93!

Issue #983, opened 2015-12-08
Update Linqpad examples reference to local dll instead of nuget repo


naveensrinivasan:
The linqpad references are pointing to nuget packages. It should be pointing to local dll to avoid build failures on release. https://github.com/octokit/octokit.net/releases/tag/v0.17.0

```
/tools/LINQPad\lprun.exe  -compileonly C:\projects\octokit-net\samples\linqpad-samples\6-create-repository.linq 
C:\projects\octokit-net\samples\linqpad-samples\6-create-repository.linq(62,17): error CS0246: The type or namespace name 'SignatureResponse' could not be found (are you missing a using directive or an assembly reference?)
Running build failed.
Error:
System.Exception: lprun.exe returned with a non-zero exit code for C:\projects\octokit-net\samples\linqpad-samples\6-create-repository.linq
   at Microsoft.FSharp.Core.Operators.FailWith[T](String message)
   at Microsoft.FSharp.Collections.SeqModule.Iterate[T](FSharpFunc`2 action, IEnumerable`1 source)
   at FSI_0001.Build.clo@145-25.Invoke(Unit _arg7) in C:\projects\octokit-net\build.fsx:line 148
   at Fake.TargetHelper.runSingleTarget(TargetTemplate`1 target) in C:\code\fake\src\app\FakeLib\TargetHelper.fs:line 483
```

shiftkey:
Had a look at this and found it to be a larger yak than previously anticipated. Problems to address:

 - `lprun` doesn't have an easy way to install certain packages - it's probably using `%APPDATA%\NuGet\NuGet.Config`
 - `nuget.exe` has a way to manage these sources from the CLI so we could script it, but perhaps we're not allowed to do this on AppVeyor. I gather this will be the same thing for TravisCI.
 - if we can do this, we should clean up after it done
 - what about if the build fails - does FAKE have a good way to run cleanup scripts?

naveensrinivasan:
I am going to fix the specific example to get it working and open another issue to address this for every check-in.

shiftkey:
@naveensrinivasan :thumbsup:

shiftkey:
A thought based on re-reading [the docs](https://www.linqpad.net/lprun.aspx) just now:

> LINQPad also supports the ref directive for referencing additional assemblies. You can specify the assembly either via its filename (including the directory if it's outside the .NET Framework) or its fully qualified name (if it's in the GAC): 

[...]

PR #987, opened 2015-12-09
Linqpad example fixes for SignatureResponse

naveensrinivasan:
Linqpad example fixes had reference to `SignatureResponse` which is now changed to `Committer` #983. This only fixes the specific file that is causing the build to fail.


shiftkey:
Thanks for this!

PR #988, opened 2015-12-10
cleaning up this old, unloved project

shiftkey:
I use [this script](https://github.com/shiftkey/octokit-spike-scriptcs), and I'm not sure if anyone else is. If people want to bring that in, I can totally do that (or some sort of use-Octokit-to-cleanup meta experiment).

nothing-to-see-here


Haacked:
I didn't even know this was here or what it did. :stuck_out_tongue:

PR #991, opened 2015-12-10
Document product header value

Haacked:
Fixed a parameter name that was misleading and documented the methods and properties.

### DIFF
--- a/Octokit/Http/ProductHeaderValue.cs
+++ b/Octokit/Http/ProductHeaderValue.cs
@@ -9,8 +9,8 @@
         {
         }
 
-        public ProductHeaderValue(string name, string value)
-            : this(new System.Net.Http.Headers.ProductHeaderValue(name, value))
+        public ProductHeaderValue(string name, string version)
+            : this(new System.Net.Http.Headers.ProductHeaderValue(name, version))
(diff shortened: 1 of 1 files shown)


alfhenrik:
LGTM :smiley:

Octokit/Http/ProductHeaderValue.cs
```diff
@@ -19,11 +36,17 @@ public ProductHeaderValue(string name, string value)
...
+        /// <summary>
+        /// The name of the product that's using Octokit
+        /// </summary>
         public string Name
         {
             get { return _productHeaderValue.Name; }
         }
 
+        /// <summary>
+        /// Gets the version of the product.
```
shiftkey: :lipstick: `The version of the product` (i.e. drop the `Gets`)?

shiftkey:
One little nit, everything else seems great!

Octokit/Http/ProductHeaderValue.cs
Haacked: Ha! Ghost Doc. :stuck_out_tongue:

Haacked:
:kimono:

PR #996, opened 2015-12-13
Add Protected Branches API support

ryangribble:
This Work-In-Progress PR adds support for the protected branches API, which is [currently in PREVIEW](https://developer.github.com/changes/2015-11-11-protected-branches-api/)

## Notes
This is our first contribution to octokit.net so just let us know if anything isn't as desired etc... I've tried to comply with existing similar implementations, coding/comment styles, unit and integration tests and so on, so hopefully it isn't too far off the mark

Since the protected branches API is in preview, I wasn't sure if it's OK to be added, but we need this support for tooling we are writing and would rather contribute to upstream than use a local fork, unless we have to!  I did notice a couple of other preview features in the codebase.  I also saw that octokit.rb has implemented the protected branches functions, so hopefully this is OK

## Implementation Details
At a high level the change involves:
- Create new models for Branch Protection details
- Add Protection element to existing Branch class
- Existing GetBranch() and GetAllBranches() calls specify new preview accepts header so this data is returned
- Added a new EditBranch() call, similar to other Edit() methods, that takes a BranchUpdate object specifying the protection settings/fields to change.  This is Patched into the Branch object/URI as per other similar examples
- Modified existing tests for GetBranch() and GetAllBranches() calls to handle the accepts parameter now being specified
- Added unit and integration tests for the branch protection features, which are all passing

## To Do List
- [X] ~~Haven't yet looked at what is required on the Octokit.Reactive project~~ Added new function, not sure if anything more to do?

### DIFF
--- a/Octokit.Tests.Conventions/ModelTests.cs
+++ b/Octokit.Tests.Conventions/ModelTests.cs
@@ -85,6 +85,12 @@ namespace Octokit.Tests.Conventions
                         continue;
                     }
 
+                    // Let's skip collections that only have a private Setter
+                    if (property.SetMethod != null && property.SetMethod.IsPrivate)
+                    {
+                        continue;
+                    }
+
                     mutableCollectionProperties.Add(property);
                 }
             }
--- a/Octokit.Tests.Conventions/ModelTests.cs
+++ b/Octokit.Tests.Conventions/ModelTests.cs
@@ -85,12 +85,6 @@ namespace Octokit.Tests.Conventions
                         continue;
                     }
 
-                    // Let's skip collections that only have a private Setter
-                    if (property.SetMethod != null && property.SetMethod.IsPrivate)
-                    {
-                        continue;
-                    }
-
                     mutableCollectionProperties.Add(property);
                 }
             }
--- a/Octokit/Clients/RepositoriesClient.cs
+++ b/Octokit/Clients/RepositoriesClient.cs
@@ -505,5 +505,46 @@ namespace Octokit
             const string previewAcceptsHeader = "application/vnd.github.loki-preview+json";
             return ApiConnection.Get<Branch>(ApiUrls.RepoBranch(owner, repositoryName, branchName), null, previewAcceptsHeader);
         }
+
+        /// <summary>
+        /// Protects the specified branch with the values given in <paramref name="update"/>
+        /// </summary>
+        /// <param name="owner">The owner of the repository</param>
+        /// <param name="repositoryName">The name of the repository</param>
+        /// <param name="branchName">The name of the branch</param>
+        /// <param name="update"></param>
+        /// <returns>The updated <see cref="T:Octokit.Branch"/></returns>
+        public Task<Branch> ProtectBranch(string owner, string repositoryName, string branchName, BranchUpdate update)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
+            Ensure.ArgumentNotNullOrEmptyString(repositoryName, "repositoryName");
+            Ensure.ArgumentNotNullOrEmptyString(branchName, "branchName");
+            Ensure.ArgumentNotNull(update, "update");
+
+            update.Protection.Enabled = true;
+
+            const string previewAcceptsHeader = "application/vnd.github.loki-preview+json";
+            return ApiConnection.Patch<Branch>(ApiUrls.RepoBranch(owner, repositoryName, branchName), update, previewAcceptsHeader);
+        }
+
+        /// <summary>
+        /// Unprotects the specified branch
+        /// </summary>
+        /// <param name="owner">The owner of the repository</param>
+        /// <param name="repositoryName">The name of the repository</param>
+        /// <param name="branchName">The name of the branch</param>
+        /// <returns>The updated <see cref="T:Octokit.Branch"/></returns>
+        public Task<Branch> UnprotectBranch(string owner, string repositoryName, string branchName)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
+            Ensure.ArgumentNotNullOrEmptyString(repositoryName, "repositoryName");
+            Ensure.ArgumentNotNullOrEmptyString(branchName, "branchName");
+
+            var update = new BranchUpdate();
+            update.Protection.Enabled = false;
+
+            const string previewAcceptsHeader = "application/vnd.github.loki-preview+json";
+            return ApiConnection.Patch<Branch>(ApiUrls.RepoBranch(owner, repositoryName, branchName), update, previewAcceptsHeader);
+        }
     }
 }
--- a/Octokit/Clients/RepositoriesClient.cs
+++ b/Octokit/Clients/RepositoriesClient.cs
@@ -176,8 +176,7 @@ namespace Octokit
             Ensure.ArgumentNotNullOrEmptyString(branch, "branchName");
             Ensure.ArgumentNotNull(update, "update");
 
-            const string previewAcceptsHeader = "application/vnd.github.loki-preview+json";
-            return ApiConnection.Patch<Branch>(ApiUrls.RepoBranch(owner, name, branch), update, previewAcceptsHeader);
+            return ApiConnection.Patch<Branch>(ApiUrls.RepoBranch(owner, name, branch), update, AcceptHeaders.ProtectedBranchesApiPreview);
         }
 
         /// <summary>
@@ -408,8 +407,7 @@ namespace Octokit
             Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
             Ensure.ArgumentNotNullOrEmptyString(name, "name");
 
-            const string previewAcceptsHeader = "application/vnd.github.loki-preview+json";
-            return ApiConnection.GetAll<Branch>(ApiUrls.RepoBranches(owner, name), null, previewAcceptsHeader);
+            return ApiConnection.GetAll<Branch>(ApiUrls.RepoBranches(owner, name), null, AcceptHeaders.ProtectedBranchesApiPreview);
         }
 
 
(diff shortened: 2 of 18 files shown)


shiftkey:
@ryangribble welcome aboard, just had a quick look at this and it looks really good! Thanks for the hard work kicking this off!

> Added new function, not sure if anything more to do?

The `Octokit.Reactive` project is intended to be API-equivalent to the `Octokit` project, but returning `IObservable<T>` instead of `Task<T>`. I think what you've done is just enough, however CI isn't kicking in so I need to look at that...

Octokit/Models/Response/BranchProtection.cs
```diff
@@ -0,0 +1,112 @@
...
+    /// Protection details for a <see cref="Branch"/>.
+    /// Note: this is a PREVIEW api: https://developer.github.com/changes/2015-11-11-protected-branches-api/
+    /// </summary>
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
+    public class BranchProtection
+    {
+        /// <summary>
+        /// Should this branch be protected or not
+        /// </summary>
+        public bool Enabled { get; set; }
```
shiftkey: As this is a response object, this will currently fail our checks for public setters.

For serialization purposes, `protected set` will work fine - but we also define `ctors` (like in `Branch` above) for scenarios where you may need to set it up in a certain way.

You can see these failures by running this from the repo root:
 
> .\build BuildApp
> .\build ConventionTests

Octokit/Models/Response/BranchProtection.cs
```diff
@@ -0,0 +1,112 @@
...
+        }
+    }
+
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
+    public class RequiredStatusChecks
+    {
+        /// <summary>
+        /// Who required status checks apply to
+        /// </summary>
+        public EnforcementLevel EnforcementLevel { get; set; }
```
shiftkey: :arrow_up: as above

Octokit/Models/Response/BranchProtection.cs
```diff
@@ -0,0 +1,112 @@
...
+
+        /// <summary>
+        /// The list of status checks to require in order to merge into this <see cref="Branch"/>
+        /// </summary>
+        public ICollection<string> Contexts { get; private set; }
```
shiftkey: I'm also seeing this property fail the convention test but this makes zero sense. Perhaps we're not liking `private set` here?

shiftkey:
Integration tests are passing on my end :thumbsup:

@Haacked how do you feel about the API design here? Anything stand out for feedback?

ryangribble:
No worries Ill fix up those protected set and parameterised ctor's shortly

Regarding API design at first (and you can see on my earlier commits) I was following the octokit.rb implementation of having ProtectBranch() and UnprotectBranch() calls, however this didnt really suit the design of octokit.net which is more focused on having an Edit() method for an entity, and passing in the xxxUpdate model object with the changes... so I re-implemented as a single EditBranch()  function

Octokit/Models/Response/BranchProtection.cs
ryangribble: FYI this one is implemented similar to the IssueUpdate object with it's list of Labels.  Implemented as a List<string> with public get and private set, then AddLabel() and ClearLabels() functions that manipulate the local private list.

I guess the difference is that IssueUpdate is a Request model object, whilst in this case the RequiredStatusChecks are in the Response section?  I can make it protected rather than private I guess

shiftkey: > I guess the difference is that IssueUpdate is a Request model object, whilst in this case the RequiredStatusChecks are in the Response section?

That's exactly right - there's some `PATCH` specific behaviour around labels, and that's the solution we ended up with.

For response object, I wouldn't worry about the `Add*` and `Clear*` methods - a `ctor` which accepts a collection to use is going to be easier for consumers to work with around testing...

ryangribble: Oops, can't actually set it to protected or a compile error occurs
I made a tweak to the convention tests to allow lists that have a private setter, let me know if that's OK

Octokit/Models/Response/BranchProtection.cs
ryangribble: Done: https://github.com/TattsGroup/octokit.net/commit/37df149b3ec70b38dfcef858171a239551fa07e1

Octokit/Models/Response/BranchProtection.cs
ryangribble: Done: https://github.com/TattsGroup/octokit.net/commit/37df149b3ec70b38dfcef858171a239551fa07e1

shiftkey:
@ryangribble Pro Tip™: don't forget to add your work account to your [GitHub profile](https://github.com/settings/emails), so those last commits are linked to your profile. Or just amend the author info and `push -f` to your branch.

Octokit/Models/Response/BranchProtection.cs
ryangribble: Just saw the comment about ```Add*``` and ```Clear*``` methods so they have gone the way of the dodo now :)

ryangribble:
@shiftkey good catch thanks, email should be fixed now

ryangribble:
LOL so I was just testing this against our GitHub Enterprise on premises installation, and it looks like the preview API isnt actually in GitHub Enterprise version 2.4.2 (eventhough the actual Branch Protection features are).  Will need to wait for Enterprise to include the API feature before we can actually use this stuff in our tooling :smile:

Octokit.Tests.Conventions/ModelTests.cs
```diff
@@ -85,6 +85,12 @@ public void ResponseModelsHaveReadOnlyCollections(Type modelType)
                         continue;
                     }
 
+                    // Let's skip collections that only have a private Setter
+                    if (property.SetMethod != null && property.SetMethod.IsPrivate)
```
khellang: Why should collection properties with private setters be mutable?

Octokit/Models/Response/BranchProtection.cs
khellang: This fails because the collection itself is mutable. It should be changed to an `IReadOnlyCollection`. The proposed fix makes no sense. There's already a different test that checks for setter access. :smile:

Octokit.Tests.Conventions/ModelTests.cs
ryangribble: this change makes it so collection properties with private setters are NOT added to the mutable list.  Otherwise convention tests were failing saying that these properties werent obeying the Response object rules

khellang: But this is totally out of scope for this test, and it actually makes it incorrect. This test checks that "Response Models Have Read Only Collections". Making it skip properties with private setters also makes it allow mutable collections, which defeats the test in the first place. There's already [a test that checks for non-public setters](https://github.com/octokit/octokit.net/blob/master/Octokit.Tests.Conventions/ModelTests.cs#L42-L62).

shiftkey: Sorry to jump in here before this goes further off the rails, but thanks @khellang for reminding me about the intent of this test.

`ICollection<T>` has the mutable `Add`, `Remove` and `Clear` - so even if it's a read-only property, it can be manipulated arbitrarily after the object is created. We're trying to discourage that, so we use `IReadOnlyList<T>` on our collections.

An example of that is `CommitActivity` - https://github.com/octokit/octokit.net/blob/master/Octokit/Models/Response/CommitActivity.cs - if we could update `RequiredStatusChecks.Contexts` to be a more specific interface than `ICollection<T>` and then drop this change we should be able to get back to a green build...

ryangribble: Roger that, I'm the first to admit I haven't quite dealt with this many convention tests etc before, I was definitely hacking my way to success 😀

I'll see what I can do with the Contexts collection and ditch this incorrect convention test

ryangribble:
RequiredStatusChecks.Context is now an IReadonlyList, implemented as per CommitActivity example
Removed convention test hack

Also rebased on latest master

ryangribble:
Unsure about the AppVeyor test failure.  All unit tests are passing for me locally?

shiftkey:
@ryangribble that test seems to be rather flaky :disappointed:

ryangribble:
Looks like the rerun tests all passed etc 

Is there anything further needed from me?

Octokit/Clients/RepositoriesClient.cs
```diff
@@ -502,7 +521,8 @@ public Task<Branch> GetBranch(string owner, string repositoryName, string branch
             Ensure.ArgumentNotNullOrEmptyString(repositoryName, "repositoryName");
             Ensure.ArgumentNotNullOrEmptyString(branchName, "branchName");
 
-            return ApiConnection.Get<Branch>(ApiUrls.RepoBranch(owner, repositoryName, branchName));
+            const string previewAcceptsHeader = "application/vnd.github.loki-preview+json";
```
naveensrinivasan: This `const string` is duplicated in 3 methods. Could we declare it at the class level and use it?

shiftkey: :thumbsup:

ryangribble: I was following an example I saw in another part of the code base where the preview header was defined each time, but I can change it to a class level. 

Should the unit tests use the same define or their own class level define or specify string  each time?

shiftkey: The licenses API? Yeah, there's two usages in `MiscellaneousClient` which could use a bit of :lipstick: to tidy up at some stage.

Let's define a field on the test class which gets reused here, and then look at the other usages separately.

ryangribble:
I really dont like the idea of referring to the test projects from the implementation classes (if that's what you meant) so what Ive done is created a AcceptHeaders helper class with the preview header defined, and changed the imlpementation to use that.  

The tests are left as is, as I believe it's good practice to have each test specify what it's success criteria are, rather than use some commonly defined thing (that could be changed later unaware of the implication to tests that were using it).  

Ive also got a separate branch and will follow this up with a new PR (after this one is accepted) where Ive gone and cleaned up all uses of Accepts headers (eg the licenses API as well as the standard json and html payloads etc) to use the new helper class, and remove the hardcoded strings from multiple places in implementation classes.  Let's get this PR merged first though, then I'll follow up with the general Accepts Headers tidy ups for all usages...

Also rebased on latest master

shiftkey:
@ryangribble that's fine, happy to merge this in after the Travis build goes green and iterate on this further...

shiftkey:
Thanks @ryangribble!

PR #970, opened 2015-12-02
Remove pull request merged property

Haacked:
This continues the work started in PR #936. That PR was abandoned so I'm creating a new branch that people can PR into to complete this work.

Closes #936

PR #997, opened 2015-12-13
Complete PullRequest Merged field changes

ryangribble:
This replaces #970 (which replaced #936)

As well as the original changes that removed the Merged property, Ive also removed MergeCommitSha as mentioned by @Haacked in #936 

Ive actually added the ```bool Merged ``` property back, with only a getter that returns ```MergedAt.HasValue```  

Happy with this?  Otherwise I could remove this property entirely and all consumers would have to do their own check for ```MergedAt.HasValue```


shiftkey:
As the webhooks weren't configured, I pulled this down and ran the relevant tests :thumbsup: :tada:

Thanks for closing this out @ryangribble!

PR #998, opened 2015-12-13
Minor cleanup - duplicate (unused) file

ryangribble:
It looks like Commitrer.cs has been moved to /Models/Common, and this old copy in Models/Response was left behind.  Identical contents to the one in /Models/Common and not referenced by any csproj files etc, so I think it should be safe to remove


shiftkey:
Yep, definitely not needed any more! Nice spot!

PR #999, opened 2015-12-13
let's do CI right

shiftkey:
This is some parity work as part of moving our OSS projects over to some dedicated resources on AppVeyor. I wanted to do this later in the week but it looks like webhooks weren't being delivered over the weekend, so let's fix this:

- [x] moved this webhook to the `github-windows` org on AppVeyor
- [x] verify builds are configured correctly
- [x] verify PR is triggered
- [x] confirm NuGet feed in README is correct
- [x] filter out rogue `tools` package from being published
- [ ] :fire: old build configuration

cc @haacked

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # Octokit - GitHub API Client Library for .NET 
-[![Build Status](https://ci.appveyor.com/api/projects/status/github/octokit/octokit.net?branch=master)](https://ci.appveyor.com/project/Haacked15676/octokit-net) [![Join the chat at https://gitter.im/octokit/octokit.net](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/octokit/octokit.net?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
+[![Build status](https://ci.appveyor.com/api/projects/status/cego2g42yw26th26/branch/master?svg=true)](https://ci.appveyor.com/project/github-windows/octokit-net/branch/master)) [![Join the chat at https://gitter.im/octokit/octokit.net](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/octokit/octokit.net?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 
 ![logo](octokit-dotnet_2.png)
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,4 +10,5 @@ nuget:
   account_feed: true
   project_feed: true
 artifacts:
-- path: '**\octokit*.nupkg'
+- path: 'packaging\octokit*.nupkg'
+  name: OctokitPackages
--- a/build.fsx
+++ b/build.fsx
@@ -222,7 +222,6 @@ Target "CreatePackages" DoNothing
 "Clean"
    ==> "AssemblyInfo"
    ==> "CheckProjects"
-   ==> "ValidateLINQPadSamples"
    ==> "BuildApp"
 
 "Clean"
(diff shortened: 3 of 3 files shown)


Haacked:
> but it looks like webhooks weren't being delivered over the weekend

I had set the octokit/octokit.net webhook to only send `push` because someone suggested that solved the problem for them including pull requests and avoids the triple build issue. But it just occurred to me that might only work for pull requests pushed to this repository and not for pull requests pushed to a _fork_ of this repository. So I changed the webhook just not to also send pull requests.

shiftkey:
@haacked ah, okay. I traced the triple commit status issue to a legacy commit status, so now we should only see two statuses when PR is created against this repository. 

The multiple builds issue is actually due to two things occurring - a build for the tip of the PR branch, and a build for the merged commit (the PR branch merged into master, if it does cleanly merge). I'm not sure of a way to switch that off, but I think the default webhook settings are now correct. Make sense?

shiftkey:
@haacked I'll let you remove the old project configuration from AppVeyor and :thumbsup: this PR.

Haacked:
@shiftkey so did you change octokit/octokit.net's webhook to point to the new one?

Haacked:
It looks like it. :sparkles: Nice work!

shiftkey:
> so did you change octokit/octokit.net's webhook to point to the new one?

ryangribble:
> The multiple builds issue is actually due to two things occurring - a build for the tip of the PR branch, and a build for the merged commit (the PR branch merged into master, if it does cleanly merge). I'm not sure of a way to switch that off, but I think the default webhook settings are now correct. Make sense?

Hey guys, interesting situation because we are running into similar things ourselves with our on-prem GithUb enterprise and Teamcity.  Hopefully not telling everyone what they already know but each pull request has 2 "branches" /refs/pull/999/head (which is the head of the PR) and /refs/pull/999/merge (which is a potential merge commit of the PR with the target branch, eg master).  So when doing CI builds ideally you want to build the "merge" one and not the "head" one, since it doesnt really matter if the PR itself builds, what we care about is whether master builds once the PR is merged

But if you are using the commit status API to set build status on the PR/commits, then when the status is pushed, that changes the commit hash on the /merge branch which then fires off another build etc.  Unfortunately we havent figured out a neat way of dealing with this as yet so we are stuck building /head of PR's when we really want to be building /merge

Is this similar to the "multiple builds" issue you are referring to with octokit and appveyor?  If there are any solutions Id be all ears.  One thing is we could send the webhooks to an intermediary service that would analyze the github/PR etc and then figure out whether it was only a commit status update or a "real change" and only relay the build to TeamCity if it was a "real change".  But that's getting pretty nasty :)

shiftkey:
@ryangribble yeah, I'm not *too* fussed about it, but I do like verifying the branch tip as well as the merge commit. But yes, it's exactly the same issue - here's how AppVeyor displays it:

screen shot 2015-12-14 at 10 10 25 am

Haacked:
@ryangribble Interesting. I think there are two potential ideal solutions.

1. If GitHub gave more granularity in Webhooks when sending pushes to build. You could contact https://github.com/c and suggest that.
2. If AppVeyor gave us more options to control which branch of a pull request gets built. This is similar to your intermediary service suggestion. @FeodorFitsner is pretty responsive and might be willing to consider that change. :smile:

FeodorFitsner:
You see two builds if both feature branch and PR are in the same repo. With forked repos this works nice as you get PR build only in the master repo.

Regarding commit status - to set PR status the status should be set on branch commit, not a virtual "merge" commit - at least AppVeyor does that right now, last time I tried setting status on merge commit it didn't change PR status.

I don't know yet what would be the best strategy to implement prevention of branch commit build if that commit participates in PR(s). When you do a push to feature branch, usually "push" event comes first (guaranteed?), then "sync" events for all affected PRs. How do we know that commit is part of PR(s), not a regular "working" commit? Check [all commits of all PRs](https://developer.github.com/v3/pulls/#list-commits-on-a-pull-request)? This could be slow.

PR #1004, opened 2015-12-14
Add support to run integration tests against GitHub Enterprise

ryangribble:
We are running GitHub Enterprise on premises so I was looking for a way to be able to run some of the integration tests against that.

Following the existing patterns in the Helper class, I have made it so that if environment variable "OCTOKIT_GITHUBENTERPRISEURL" is set, this is used when setting up the GitHubClient returned by various helper methods...  If this environment variable is not present, the default GitHub URL is used

I am not fully across all the differences between Enterprise and github.com but certainly I just found out that the protected branches API (preview) is not yet in Enterprise 2.4.2.  So this does mean that some tests would fail when running on GitHub Enterprise, however it still seems like a useful capability to be able to target a GHE instance if/when required.  Since it's a non invasive change that shouldnt affect any other usages, I'm positing that it could be included (even if in an undocumented/unsupported capacity)

Thoughts?


shiftkey:
I like this and really have nothing further to add. :tada: 

Gonna let this sit for a bit and see if others have feedback.

shiftkey:
@ryangribble did you want to include this in the integration test script

https://github.com/octokit/octokit.net/blob/master/script/configure-integration-tests.ps1

I think it should work as-is as an optional variable, and the alias could be something like "GitHub Enterprise server"

ryangribble:
OK, added it to the helper script as an optional variable

shiftkey:
Thanks!

PR #1007, opened 2015-12-14
Fixes #1001 CannotBeMergedDueNotInMergeableState

### LINKED ISSUE #1001: Review CannotBeMergedDueNotInMergeableState test
When this test was first written, it was designed to be in a state which would raise an exception if the consumer tried to merge it. That's now not the case.

TODO:

 - tweak the commits on the branch to see that it is back in an unmergeable state
 - verify the test now fails with the appropriate exception

gabrielweyer:
Fixes #1001 

I made two changes:

 1. I updated `master` head with the new commit so that the PR is not mergeable (not sure if it's the best way to do this.
 2. The `Merge` method wasn't awaiting the API call so the exception couldn't be caught, I made the method `async` and I'm now awaiting the `PUT` request.

### DIFF
--- a/Octokit/Clients/PullRequestsClient.cs
+++ b/Octokit/Clients/PullRequestsClient.cs
@@ -110,7 +110,7 @@ namespace Octokit
         /// <param name="number">The pull request number</param>
         /// <param name="mergePullRequest">A <see cref="MergePullRequest"/> instance describing a pull request merge</param>
         /// <returns>An <see cref="PullRequestMerge"/> result which indicates the merge result</returns>
-        public Task<PullRequestMerge> Merge(string owner, string name, int number, MergePullRequest mergePullRequest)
+        public async Task<PullRequestMerge> Merge(string owner, string name, int number, MergePullRequest mergePullRequest)
         {
             Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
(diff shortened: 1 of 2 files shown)


Octokit/Clients/PullRequestsClient.cs
```diff
@@ -118,7 +118,7 @@ public Task<PullRequestMerge> Merge(string owner, string name, int number, Merge
 
             try
             {
-                return ApiConnection.Put<PullRequestMerge>(ApiUrls.MergePullRequest(owner, name, number), mergePullRequest);
+                return await ApiConnection.Put<PullRequestMerge>(ApiUrls.MergePullRequest(owner, name, number), mergePullRequest);
```
khellang: `ConfigureAwait(false)`?

gabrielweyer: :+1: pushed another commit fixing this.

Haacked:
> I updated master head with the new commit so that the PR is not mergeable (not sure if it's the best way to do this.

Seems reasonable to me.

The funny part is that this branch is also not mergeable and I got confused for a second about what you were making not mergeable. :stuck_out_tongue: Please fix up this branch so we can merge it in. Thanks!

shiftkey:
Thanks @gabrielweyer!

PR #995, opened 2015-12-12
Targeting mono

naveensrinivasan:
Targeting mono.

- [x] Build App 
- [x] Run Unit Tests  
- [x] Successfully build it on travis-ci
- [x] Run test's on travis-ci

PR #1012, opened 2015-12-16
`String` changed to `string`

M-Zuber:
keeping it all the same

### DIFF
--- a/Octokit.Reactive/Clients/ObservableRepositoryContentsClient.cs
+++ b/Octokit.Reactive/Clients/ObservableRepositoryContentsClient.cs
@@ -91,7 +91,7 @@ namespace Octokit.Reactive
         [Obsolete("Use GetArchive to download the archive instead")]
         public IObservable<string> GetArchiveLink(string owner, string name, ArchiveFormat archiveFormat)
         {
-            return GetArchiveLink(owner, name, archiveFormat, String.Empty);
+            return GetArchiveLink(owner, name, archiveFormat, string.Empty);
         }
 
         /// <summary>
--- a/Octokit.Tests.Conventions/Exception/InterfaceHasAdditionalMethodsException.cs
+++ b/Octokit.Tests.Conventions/Exception/InterfaceHasAdditionalMethodsException.cs
@@ -21,7 +21,7 @@ namespace Octokit.Tests.Conventions
 
         static string CreateMessage(Type type, IEnumerable<string> methods)
         {
-            var methodsFormatted = String.Join("\r\n", methods.Select(m => String.Format(" - {0}", m)));
+            var methodsFormatted = string.Join("\r\n", methods.Select(m => string.Format(" - {0}", m)));
             return "Methods found on type {0} which should be removed:\r\n{1}"
                       .FormatWithNewLine(
                           type.Name,
--- a/Octokit.Tests.Conventions/Exception/InterfaceMethodsMismatchException.cs
+++ b/Octokit.Tests.Conventions/Exception/InterfaceMethodsMismatchException.cs
@@ -21,14 +21,14 @@ namespace Octokit.Tests.Conventions
 
         static string Format(ParameterInfo parameterInfo)
         {
-            return String.Format("{0} {1}", parameterInfo.ParameterType.Name, parameterInfo.Name);
+            return string.Format("{0} {1}", parameterInfo.ParameterType.Name, parameterInfo.Name);
         }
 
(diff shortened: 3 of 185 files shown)


khellang:
:shipit: This needs a rebase, though. @shiftkey how's the C#6/project.json stuff going? Most of these should be converted to interpolated strings at some point :smile:

M-Zuber:
I'll do the rebase when my Internet is not from my phone :grin:
Does that mean  making a new PR? Haven't rebased in a while

khellang:
I think the steps should be something along the lines of
 - `git checkout uppercase-is-not-always-right` // make sure you're on the right branch
 - `git remote add upstream git@github.com:octokit/octokit.net.git` // if you haven't already, add a new origin to the upstream repository
 - `git pull --rebase upstream master` // fetch the latest changes from upstream master and rebase on top of it
 - `git push -f origin uppercase-is-not-always-right` // push the changes :sparkles:

Haacked:
Oh boy. So our convention is something that we can't currently capture with R# rules today (afaik).

* Do use `String` when calling static methods of `System.String`.

ex: `String.Format(...)`

* Do use `string` when specifying a type such as the return type of a method, an argument, a field, a property.

ex: `void DoStuff(string foo);`

M-Zuber:
NP. 
You guys where not around in gitter and @shiftkey seemed to not know this twitter. 
 
I will close this and reopen a new PR that adds to the documentation. What is the target file? Contributing.md?

M-Zuber:
What about `new string`?

Haacked:
> What about `new string`?

Do we do that anywhere? :stuck_out_tongue:

The general principle I have in my head is that when we're using `System.String` like a primitive, we use the C# shorthand syntax. Hence `string` for return types, etc.

When we use it like a type, I like it to look like a type. Hence `String.Format`, `String.Empty`, etc.

Hence I'd prefer `new String()` or `new Int32()` over `new string()` and `new int()`. Anyone know if we can configure R# or VS rules to follow that convention? Or am I being too picky?

khellang:
> Anyone know if we can configure R# or VS rules to follow that convention? Or am I being too picky?

Nope. It's either the keyword or the CLR type. Keyword is default :smile:

Haacked:
I wrote a [blog post about the convention](http://haacked.com/archive/2015/12/16/to-string-or-not/). I'm seriously considering changing my mind given that the [corefx team has their guidelines](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md) and to be quite honest, I hate fighting defaults and I hate arguing every single convention.

I'd rather have a one-time argument where we decide on a set of conventions someone else maintains and when people complain about any given convention, we tell them to take it up with the dotnet team. :stuck_out_tongue:

M-Zuber:
While I have come around to @Haacked's way of thinking in this regard,
having a set of guidelines that is not our fault can be very useful.

One point that has not been mentioned is people new to C#. It may be confusing to them as in a lot of cases the existence of the BCL types is not even known to them. Sticking to the keywords may be less confusing.

I can reopen the PR if you so desire

Haacked:
> I can reopen the PR if you so desire

Sure. I'd love to hear @shiftkey's thoughts on this.

M-Zuber:
[Here is what happened on twitter today](https://twitter.com/mordzuber/status/677083920014442496)

shiftkey:
Okay, so here's my thoughts on this:

 - `String.Method` definitely feels more Type-y, and I do empathise with the mental "jump" that occurs if you come across a `string.Method`
 - For other types that have aliases in C# (`long`, `int`, etc) I don't really recall having the same hassle - in fact, I kinda squirm if i see `Int64` or `Int32` mentioned explicitly.
 - The using statement arguments are kinda trivial for me - there's likely other things in the `System` namespace which you'll need, aside from `String`.

For the sake of consistency, I'm :thumbsup: on moving it all to `string`.

shiftkey:
@khellang 

> how's the C#6/project.json stuff going? 

Given we're all on VS2015 now, this is possibly something worth entertaining (although there might be work to bring the Mono side up to parity once #995 lands).

Haacked:
selfie-0

M-Zuber:
:joy: That sums up the thread so perfectly. 
Thank you for the patience.

M-Zuber:
@shiftkey can I open a WIP for the string.format work?

shiftkey:
@M-Zuber I'll open an issue after merging the Mono PR

PR #1013, opened 2015-12-16
WIP: Create inheritance hierarchy that matches backend

Haacked:
The backend API returns varying types of accounts with different sets of
properties. This is the beginning of attempting to model that a bit more
closely.

### DIFF
--- a/Octokit/Models/Response/Account.cs
+++ b/Octokit/Models/Response/Account.cs
@@ -5,13 +5,37 @@ namespace Octokit
     /// <summary>
     /// Base class for a GitHub account, most often either a <see cref="User"/> or <see cref="Organization"/>.
     /// </summary>
-    public abstract class Account
+    public abstract class Account : AccountSimple
     {
         protected Account() { }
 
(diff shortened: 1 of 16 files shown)


Haacked:
This doesn't change any of the API return types yet. I just wanted to take a baby step here. The next step will probably be to define interfaces.

shiftkey:
@Haacked while we're being a bit pedantic with types - how attached are you to the `Simple` suffix? I think the name is good, I just keep expecting to see it as a prefix, e.g. `SimpleAccount`, `SimpleOrganization`.

I do like `UserWithIdentity` - which is also a suffix - so perhaps I'm just in a "naming things is hard" phase of the review process...

Haacked:
The reason I like the suffix is it makes it easier to find the set of related types when using intellisense or looking at the directory structure. For example, if you start typing `Account` you'll see all the types starting with that. I'm not too attached though.

shiftkey:
@haacked :thumbsup:

ryangribble:
Im wanting to close out old PRs if they aren't ever going to see the light of day.

Are we still wanting to head in this direction?  Eg creating things like AccountSimple, UserSimple, PullRequestSimple and so on?

I do see the benefit in terms of not misleading users by showing fields existing on objects that arent populated unless you've called the "load single object" API call...  But I do wonder how much extra work is involved to split all the classes and figure out which fields belong in the Simple vs the Full objects, adding all the new overloads and obsoleting the current ones, rewriting tests etc.  Im wondering if the benefit is worth that initial and ongoing work?

Haacked:
I don't have time to follow-up on this anymore and you're right, it's not worth the effort.

PR #948, opened 2015-10-28
[WIP] Documentation Is A Nice Thing To Have™

shiftkey:
This will be the feature branch for whatever documentation we think needs to live in Octokit.

Currently we're just pulling in everything under the `docs` folder, but I'm sure you all have ideas or suggestions.

You can view the current state of this branch here: https://octokitnet.readthedocs.org/en/documentation/

Fixes #463

PR #1014, opened 2015-12-17
Add link to corefx coding conventions

M-Zuber:
I am working in gitter to understand what to put in a seperate file of custom coding conventions for the project.
But this should be here either way

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -133,7 +133,7 @@ being addressed.
 Some things that will increase the chance that your pull request is accepted.
 
 * Follow existing code conventions. Most of what we do follows [standard .NET
-  <!--conventions](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md) except in a few places. We include a ReSharper team settings file.-->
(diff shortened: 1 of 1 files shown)


shiftkey:
Possibly related - should we reference the CodeFormatter project added in #807?

If so, I need to update [this repo](https://github.com/shiftkey/codeformatter) to reflect it's intent...

M-Zuber:
Oh yes,  that should definitely be mentioned. 
The way I am leaning is too link from contributing. md to a CustomCodingConventions.md file.

M-Zuber:
But that will have to wait until I am not on my phone. 
(also editing a comment)

shiftkey:
@M-Zuber this feels like it could be part of the documentation we're putting together in #948. Do you agree?

M-Zuber:
This specific pull request or the new file? 
New file yes,  this not (as I see it) 

As a side note: responding through email takes an awful long time to show up...

shiftkey:
Let's take this in and then expand on it in some documentation.

Thanks @M-Zuber!

Issue #1015, opened 2015-12-18
Move INotificationsClient under IActivitiesClient


shiftkey:
Currently it's a top-level client:

```csharp
...
var result = await client.Notifications.GetAllForCurrent();
```

But the [API documentation](https://developer.github.com/v3/activity/notifications/) has it located under `Activities`:

<img width="363" alt="screen shot 2015-12-18 at 11 17 10 am" src="https://cloud.githubusercontent.com/assets/359239/11886357/38144112-a579-11e5-93a7-0e10bb7c944e.png">

We should do the same - but mark the top-level property as `[Obsolete]` for the next release to help with transitioning.

```csharp
var result = await client.Activity.Notifications.GetAllForCurrent();
```

Haacked:
:+1:

M-Zuber:
I'd like to try and take this

shiftkey:
Fixed in #1019

PR #1018, opened 2015-12-20
updating various dependencies

shiftkey:
I found that we were tracking multiple versions of some dependencies under `packages`. This PR updates us to the latest version of the various dependencies we need:

- `Microsoft.Bcl`, `Microsoft.Bcl.Build` and `Microsoft.Net.Http`
- `NSubstitute` - tests which were throwing an exception needed to indicate the expected return type
- `xUnit`

Let's see if I've broken anything:

 - [x] verify AppVeyor build is green
 - [x] verify Travis build is green
 - [x] run integration tests locally
 - [x] tests can be run in VS2015

PR #1019, opened 2015-12-20
Move INotificationsClient to under IActivitiesClient

M-Zuber:
per #1015.

Top level client obsoleted.


Octokit.Reactive/IObservableGitHubClient.cs
```diff
@@ -16,6 +16,7 @@ public interface IObservableGitHubClient : IApiInfoProvider
         IObservableReleasesClient Release { get; }
         IObservableSshKeysClient SshKey { get; }
         IObservableUsersClient User { get; }
+        [System.Obsolete("The Notifications API belongs under the Activities client")]
```
shiftkey: :lipstick: Perhaps make this message a bit more concise:

> Notifications are now available under the Activities client. This will be removed in a future update.

Octokit/IGitHubClient.cs
```diff
@@ -116,6 +116,7 @@ public interface IGitHubClient : IApiInfoProvider
         /// <remarks>
         /// Refer to the API docmentation for more information: https://developer.github.com/v3/activity/notifications/
         /// </remarks>
+        [System.Obsolete("The Notifications API belongs under the Activities client")]
```
shiftkey: See other comment about obsolete message

shiftkey:
`packages/System.Net.Http.2.0.20710.0/System.Net.Http.2.0.20710.0.nupkg` should be removed from this PR, as it's covered by #1018

M-Zuber:
Fixed

shiftkey:
@M-Zuber the rogue `packages` entry is still present

M-Zuber:
.gif) 
I guess this is why `packages` is usually in the .gitignore  ¯\_(ツ)_/¯
I'll fix it when I'm in the office

M-Zuber:
Done and rebased

shiftkey:
```
curl: (7) Failed connect to storage.bos.xamarin.com:80; Operation timed out
The command "curl -sS http://storage.bos.xamarin.com/bot-provisioning/PortableReferenceAssemblies-2014-04-14.zip > /tmp/pcl-assemblies.zip" failed and exited with 7 during .
```

@naveensrinivasan have you seen this before?

naveensrinivasan:
@shiftkey  it is for the PortableReferenceAssemblies that we get from the xamarin. Looks like it is failing to get it failure because of  timeout. Did you try re-running the build on the same commit?

shiftkey:
@naveensrinivasan it was failing for me on multiple builds yesterday (both for OS X and Linux) but seems to have come good today. Thanks for the :eyes:.

shiftkey:
@M-Zuber thanks!

PR #1020, opened 2015-12-22
Add comment to make test account usage clearer

### LINKED ISSUE #615: make it clearer that you should be using a test account for integration tests
- Update `CONTRIBUTING.md`
 - Add comment in credential helper

RobPethick:
Fixes #615

### DIFF
--- a/Octokit.Tests.Integration/Helper.cs
+++ b/Octokit.Tests.Integration/Helper.cs
@@ -8,6 +8,7 @@ namespace Octokit.Tests.Integration
     {
         static readonly Lazy<Credentials> _credentialsThunk = new Lazy<Credentials>(() =>
         {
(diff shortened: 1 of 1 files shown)


Octokit.Tests.Integration/Helper.cs
```diff
@@ -8,6 +8,7 @@ public static class Helper
     {
         static readonly Lazy<Credentials> _credentialsThunk = new Lazy<Credentials>(() =>
         {
+            // These environment variables should be set to a test GitHub account using the powershell script configure-integration-tests.ps1
```
shiftkey: Should we make this a bit clearer and use xml-doc notation on the `Credentials` property in here?

shiftkey:
Thanks @RobPethick!

Issue #968, opened 2015-12-02
a doc to list api coverage?


chenjiaming93:
recently I am reading through GitHub API references and trying to report unimplemented or incomplete APIs. Do we have a document to list API coverage in Octokit.net?

shiftkey:
Literally just discussing this headache with @haacked. Contemplating scraping the website to get this list.

shiftkey:
Nevermind, looks like @naveensrinivasan might have already cracked this:

https://twitter.com/snaveen/status/674384982803136513

![](https://pbs.twimg.com/media/CVvlkXxVAAEmtT3.png)

naveensrinivasan:
https://gist.github.com/naveensrinivasan/58fa8eb57e61535f10db

shiftkey:
@naveensrinivasan this is so good! I'm kinda buried in work-related stuff this week but I'll try and make time to look into this...

naveensrinivasan:
Cool :+1:

chenjiaming93:
awesome

shiftkey:
Closing this out in favour of #1038 - will spin more work out from that.

PR #1038, opened 2015-12-23
[WIP] first pass at auditing the API

shiftkey:
This is the script I was using to review and identify gaps in the current API implementation.

It's not perfect (I'm not checking that clients actually implement the required routes) but I've already pulled out a number of tasks which fall into one of two buckets:

 - missing features
 - inconsistencies between the GitHub API and our implementation - mostly names of things at this stage

I had a lengthy discussion about doing more of this in Roslyn over at https://github.com/shiftkey/Octokit.CodeAnalysis/pull/1 but this felt like a easier first step to tackle the problem.


shiftkey:
And credit to @naveensrinivasan for starting this off in https://github.com/octokit/octokit.net/issues/968#issuecomment-163066719

naveensrinivasan:
@shiftkey Thanks. :metal:

shiftkey:
Not sure how much people want to review this, but it's relatively self-contained so I'm happy to let it sit over the break...

M-Zuber:
From a quick glance-over this looks very nice. :+1: 
Re the CodeAnalysis that you are bulding, I think this is a complement. One is for the developer, the other for maintainers. (Also existence of method vs implementation)

shiftkey:
> Re the CodeAnalysis that you are bulding, I think this is a complement.

I agree. I think there's some post-compilation stuff that we can verify easily via tests (especially when comparing the documented APIs to what's been implemented) and this has already been helpful in that regard.

shiftkey:
I'm putting the CodeAnalysis stuff aside for now to focus on other things.

Someone is more than welcome to pick it up in my place, and I'm happy to help shape it.

shiftkey:
Now that a bunch of open tasks have been resolved, I should merge in those changes and try running the tests again - and of course get some actual data from the audit so I can hopefully test further details.

Putting back to WIP

shiftkey:
Was able to rebase this branch and see it only raises 5 API errors, but I'm going to close this PR and revisit the overall process for incorporating this properly:

 - I'd like it to be outside the current suite of integration tests
 - The rules engine is not intuitive at all (hell, I barely understand what I've done here)
 - What to do about the violations that I've suppressed?

PR #1039, opened 2015-12-23
Added missing fields from API and removed fields that are not returned by API

gabrielweyer:
### Issue

Also has an `Id` and `Locked` [properties](https://developer.github.com/v3/issues/#get-a-single-issue). I tested the change in the `CanDeserializeIssue` method.

### IssueEventPayload

Doesn't have a property `Assignee` or `Label`. GitHub only fire an event when an issue is opened, closed or re-opened. GitHub does not fire an event when assigning someone to the issue, assigning a label or a milestone (I tested manually on my test repository).

### IssueEvents

Has a `CommitUrl` [property](https://developer.github.com/v3/issues/events/#get-a-single-event). I tested the change in the `CanDeserializeMergedEvent` method.

**Note**: the documentation doesn't have the `CommitUrl` property yet but it is being returned by the API (I used `Fiddler` as my proxy to inspect what was returned).

### DIFF
--- a/Octokit.Tests.Integration/Clients/IssuesClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/IssuesClientTests.cs
@@ -23,6 +23,22 @@ public class IssuesClientTests : IDisposable
     }
 
     [IntegrationTest]
+    public async Task CanDeserializeIssue()
+    {
+        const string title = "a test issue";
+        const string description = "A new unassigned issue";
+        var newIssue = new NewIssue(title) { Body = description };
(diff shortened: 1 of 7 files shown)


naveensrinivasan:
Do we want to use the property  `CommitUrl` that hasn't yet been documented?

shiftkey:
> Do we want to use the property `CommitUrl` that hasn't yet been documented?

I'm tentatively :thumbsup: on this, but I'll let @Haacked throw his 2c in as well before we close this out.

Haacked:
Looks good to me. Thanks!

rokonec:
GitHub webhooks are triggered on assignment and labeling and have indeed Label and Assignee property when applicable.
Please revert those changes.

PR #1040, opened 2015-12-23
IRepositoriesClient.RepoCollaborators -> IRepositoriesClient.Collaborator

### LINKED ISSUE #1036: IRepositoriesClient.RepoCollaborators -> IRepositoriesClient.Collaborator
The prefix is unnecessary, and is a pluralized property that we shouldn't really have:

https://developer.github.com/v3/repos/collaborators/

```csharp
var collaborators = await github.Repository.RepoCollaborators.GetAll("octokit", "octokit.net");
```

We should mark this property as obsolete and introduce a new property which is singular:

```csharp
var collaborators = await github.Repository.Collaborator.GetAll("octokit", "octokit.net");
```

M-Zuber:
Fixes #1036 

Updated.


shiftkey:
@M-Zuber I think you can make this change without renaming types - perhaps some extra changes are causing the build failures to creep in...

M-Zuber:
Okay, so the wanted change is just the properties on the `RepositoriesClient`s?
I'll start over in a new branch, if that's okay. A little simpler for me

shiftkey:
> Okay, so the wanted change is just the properties on the `RepositoriesClients`?

Correct. Let's keep it simple!

> I'll start over in a new branch, if that's okay. A little simpler for me

If you're fine with force-pushing this branch to blow away the current changes, I'm fine with that. Otherwise just open a new PR.

M-Zuber:
Force push is fun :smile: 
*I'll be back*

shiftkey:
> Force push is fun :smile:

Octokit.Reactive/Clients/ObservableRepositoriesClient.cs
```diff
@@ -22,7 +22,10 @@ public ObservableRepositoriesClient(IGitHubClient client)
             CommitStatus = new ObservableCommitStatusClient(client);
             Hooks = new ObservableRepositoryHooksClient(client);
             Forks = new ObservableRepositoryForksClient(client);
+#pragma warning disable CS0618 // Type or member is obsolete
```
M-Zuber: This ensures that current consumers won't get null

M-Zuber:
@shiftkey that gif more then accurately describes what I just went through


CI y u no pas???

Octokit/Clients/RepositoriesClient.cs
```diff
@@ -25,7 +25,10 @@ public RepositoriesClient(IApiConnection apiConnection) : base(apiConnection)
             CommitStatus = new CommitStatusClient(apiConnection);
             Hooks = new RepositoryHooksClient(apiConnection);
             Forks = new RepositoryForksClient(apiConnection);
+#pragma warning disable CS0618 // Type or member is obsolete
```
M-Zuber: See above

M-Zuber:
BTW I noticed that there seems to be no tests for this property. Is this on purpose?

Octokit.Reactive/Clients/ObservableRepositoriesClient.cs
shiftkey: :thumbsup:

shiftkey:
> BTW I noticed that there seems to be no tests for this property. Is this on purpose?

No, likely just something that was overlooked. Tracking that in #1045

shiftkey:
@M-Zuber thanks!

PR #1041, opened 2015-12-23
Move all HTTP headers into AcceptHeaders helper class

ryangribble:
As mentioned in #996 this is a follow up PR that tidies up all hardcoded HTTP headers used within Octokit.net, into the AcceptHeaders helper class that was added

- Replaced all uses of hardcoded HTTP header strings to declare/use entries in the AcceptHeaders helper class, like the protected branches API header in #996 does

- Left test methods alone (ie still hardcoding their expected accept strings) since we shouldn't share definitions between test and implementation code, probably shouldnt even share it between 1 test and another

This change was fairly precise/surgical in nature and all unit and integration tests are passing, so hopefully it's all good!


Octokit/Helpers/AcceptHeaders.cs
```diff
@@ -2,6 +2,14 @@
 {
     public static class AcceptHeaders
     {
+        public const string DefaultJson = "application/vnd.github.quicksilver-preview+json; charset=utf-8, application/vnd.github.v3+json; charset=utf-8";
+
+        public const string DefaultHtml = "application/vnd.github.html";
+
+        public const string ReleaseApi = "application/vnd.github.v3";
```
shiftkey: Let's name this something like `StableVersion` as its the replacement for the releases API now that's out of preview

Octokit/Helpers/AcceptHeaders.cs
```diff
@@ -2,6 +2,14 @@
 {
     public static class AcceptHeaders
     {
+        public const string DefaultJson = "application/vnd.github.quicksilver-preview+json; charset=utf-8, application/vnd.github.v3+json; charset=utf-8";
```
shiftkey: This is related to [repository redirects](https://developer.github.com/changes/2015-04-17-preview-repository-redirects/) and then the latest stable version - could we rename it to something like `RedirectsPreviewThenStableVersionJson` ?

Octokit/Helpers/AcceptHeaders.cs
shiftkey: It can probably be cleaned up so that the defaults are loaded in...

shiftkey:
Looking good, just a few little suggestions on Naming Things™

ryangribble:
No worries, naming things is the hardest part so always keen for feedback :stuck_out_tongue:

Haacked:
I like where this is headed! I agree with the naming suggestions @shiftkey proposed.

ryangribble:
Cool, I'll push some changes later this week when I'm back (currently overseas)

ryangribble:
Hey guys, Im back from holidays and have pushed up some renames as well as merged latest master into my branch.  Let me know what you think of the names

shiftkey:
@ryangribble looks good to me, thanks for the work!

Issue #1034, opened 2015-12-23
IRepositoriesClient.CommitStatus -> IRepositoriesClient.Status


shiftkey:
A naming thing: https://developer.github.com/v3/repos/statuses/

```csharp
var statuses = await github.Repository.CommitStatus.GetAll("octokit", "octokit.net", sha);
```

We should mark this property as obsolete and introduce a new property, named `Status`:

```csharp
var statuses = await github.Repository.Status.GetAll("octokit", "octokit.net", sha);
```

RobPethick:
I'd like to work on this one

M-Zuber:
Have fun, you might want to look #1042 for inspiration :grinning:

asizikov:
@shiftkey Am I right: this issue was already fixed by #1043

shiftkey:
@asizikov yes - another one that missed the "auto-close on PR" boat! Thanks for reminding me!

PR #1043, opened 2015-12-23
Update CommitStatus to Status on RepositoriesClient and ObservableRep…

RobPethick:
…ositoriesClient #1034

### DIFF
--- a/Octokit/Clients/IRepositoriesClient.cs
+++ b/Octokit/Clients/IRepositoriesClient.cs
@@ -6,7 +6,9 @@ using System.Collections.Generic;
 
 namespace Octokit
 {
-    /// <summary>
+  using System;
+
+  /// <summary>
     /// A client for GitHub's Repositories API.
     /// </summary>
     /// <remarks>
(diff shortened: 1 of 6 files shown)


Octokit/Clients/IRepositoriesClient.cs
```diff
@@ -6,7 +6,9 @@
 
 namespace Octokit
 {
-    /// <summary>
+  using System;
```
shiftkey: :lipstick: move this up outside the `namespace` with the other `using`s

Octokit/Clients/IRepositoriesClient.cs
```diff
@@ -6,7 +6,9 @@
...
+
+  /// <summary>
```
shiftkey: :lipstick: indenting

shiftkey:
A couple of little formatting things, everything else seems :gem:

RobPethick:
Thanks @shiftkey  updated :)

Haacked:
Nicely done! Thanks a lot! :+1:

PR #1051, opened 2015-12-28
Add integration tests for RepoCollab client

### LINKED ISSUE #1045: add some tests for Repository.Collaborator
Reported here: https://github.com/octokit/octokit.net/pull/1040#issuecomment-166945235

> BTW I noticed that there seems to be no tests for this property. Is this on purpose?

It'd be nice to get some integration tests in here - mostly to verify that we're able to retrieve and deserialize details about the collaborators - as a regression test.

M-Zuber:
I've added 2 tests and hope to add more. Just want to make sure that I am on the right track.
fixes #1045

[Review on Reviewable](https://reviewable.io/reviews/octokit/octokit.net/1051)

### DIFF
--- a/Octokit.Tests.Integration/Clients/RepositoryCollaboratorClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/RepositoryCollaboratorClientTests.cs
@@ -1,0 +1,56 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Octokit;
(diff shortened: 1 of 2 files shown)


M-Zuber:
I am not sure why the builds are failing :(

I did not add a separate test for the `Add` method as it is used in both tests already written.
Also there is no test for the delete method - should it be added?

shiftkey:
Just nudged the problem builds and it's now all good. Will keep an eye out for other flakiness.

PR #1054, opened 2016-01-01
Clarified "_issuesClient"

JakesCode:
Added some more info on where "_issuesClient" is obtained from following a recent Issue.


Haacked:
Thanks!

PR #1055, opened 2016-01-01
Update labels.md

JakesCode:
Added some content to help others.


Haacked:
This is a great start. Might be nice to show an example of creating an issue. Also, what happens if you call `Add("some-label");` where `some-label` is not an issue. I assume it fails.

Haacked:
I'll merge this in, but it would be great to expand on it. Thanks!

PR #1047, opened 2015-12-24
Move Release to IRepositoriesClient from IGitHubClient

RobPethick:
#1032

PR #1058, opened 2016-01-04
Dev put release on iRepository from iGitHub client

RobPethick:
Fixes #1047


RobPethick:
@shiftkey any idea why this is failing? It looks like it is complaining about building Tests-Portable but it seems to build locally just fine for me have I missed something?

naveensrinivasan:
@RobPethick @shiftkey  Nothing is wrong. It just timed out on linux and but was successful on OSX. We are good.

RobPethick:
@naveensrinivasan Ah I thought it might be something like that thanks for confirming

shiftkey:
I have zero issues with this PR. Thanks for the effort @RobPethick!

PR #1060, opened 2016-01-13
Adds support for alternate accept header to retrieve star creation timestamps

### LINKED ISSUE #1059: Support for star creation timestamps
[The API docs explain](https://developer.github.com/v3/activity/starring/#list-stargazers) that by passing a special `Accept` header (`Accept: application/vnd.github.v3.star+json`) you can have the stargazers list return star creation timestamps (same for getting the repositories starred by a user). This is useful if you're trying to see star behavior over time (I.e., to create a graph or sparkline).

Would there be any opposition to supporting this in Octokit? I'm assuming the best way to represent this would be similar to the result JSON by returning a wrapping object for each row of data that includes the starred at timestamp and the `User` as a property?

If this is a go, I can put a PR together...

daveaglick:
Resolves #1059

### DIFF
--- a/Octokit/Models/Response/RepositoryStar.cs
+++ b/Octokit/Models/Response/RepositoryStar.cs
@@ -1,0 +1,40 @@
+﻿using System;
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+using System.Globalization;
+
+namespace Octokit
+{
+    /// <summary>
+    /// Represents additional information about a star (such as creation time)
+    /// </summary>
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
+    public class RepositoryStar
+    {
+        public RepositoryStar() { }
+
+        public RepositoryStar(DateTimeOffset starredAt, Repository repo)
+        {
+            StarredAt = starredAt;
+            Repo = repo;
+        }
+
+        /// <summary>
+        /// The date the star was created.
+        /// </summary>
+        public DateTimeOffset StarredAt { get; protected set; }
+
+        /// <summary>
+        /// The repository associated with the star.
+        /// </summary>
+        public Repository Repo { get; protected set; }
+
+        internal string DebuggerDisplay
+        {
+            get
+            {
+                return string.Format(CultureInfo.InvariantCulture, Repo.DebuggerDisplay);
+            }
+        }
+    }
+}
--- a/Octokit/Models/Response/UserStar.cs
+++ b/Octokit/Models/Response/UserStar.cs
@@ -1,0 +1,40 @@
+﻿using System;
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+using System.Globalization;
+
+namespace Octokit
+{
+    /// <summary>
+    /// Represents additional information about a star (such as creation time)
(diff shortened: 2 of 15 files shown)


daveaglick:
Sorry, CI is coming up with errors I didn't catch on my local machine. Almost got it hammered out. I'll squash when I'm done and let you know when this is good for review.

daveaglick:
Okay, this should be good to go now. Travis is still reporting an error - looks like something in Octokit.Tests-Portable, but I don't know how to make heads or tails of it. Just let me know if it's something else I need to clean up.

shiftkey:
@daveaglick all good, there's some CI friction that I need to look into at some stage

Octokit/Models/Response/UserStar.cs
```diff
@@ -0,0 +1,40 @@
...
+
+namespace Octokit
+{
+    /// <summary>
+    /// Represents additional information about a star (such as creation time)
+    /// </summary>
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
+    public class UserStar
+    {
+        public UserStar() { }
```
shiftkey: Is this necessary?

Octokit/Models/Response/RepositoryStar.cs
```diff
@@ -0,0 +1,40 @@
...
+
+namespace Octokit
+{
+    /// <summary>
+    /// Represents additional information about a star (such as creation time)
+    /// </summary>
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
+    public class RepositoryStar
+    {
+        public RepositoryStar() { }
```
shiftkey: Is this necessary?

daveaglick: You mean the empty ctors or the whole wrapping classes?

shiftkey: The empty `ctors`. I ask because we've gone back and forth on how the response models should be structured in the past - previously the properties were `public set`-ers, but then we switched over to `protected set` to indicate which were required.

cc @haacked for feels

shiftkey: The main benefit for using `protected set` was around testability, but I'm not sure that's so important now in hindsight...

daveaglick: Ah, gotcha. I just assumed they were needed by the model deserializer. If it can instantiate models without the empty public ctor, I'm fine removing them.

shiftkey: Ugh, that's a good point - I had a look at some other response models and found they also have an empty `ctor` - hold off on making that change, I'll let @Haacked refresh my memory here...

Haacked: If memory serves me, we do need the default ctor for serialization.

shiftkey: :thumbsup:

shiftkey:
@daveaglick thanks for this, and apologies for not getting to it weeks ago :grin:

daveaglick:
:thumbsup: Thanks, no problem!

PR #1063, opened 2016-01-25
remove Get<T> extension method to make tests more clear

### LINKED ISSUE #1062: How do I test the things
When writing unit tests, you have to write something like the following:
```c#
var client = new RepositoryPagesClient(connection);
client.Get("fake", "repo");
connection.Received().Get<Page>(Arg.Is<Uri>(u => u.ToString() == "repos/fake/repo/pages"), null);
```

Without the null on the last line the test will fail on an `Ensure` check that the uri is null, but it is not clear why it is needed or what it does.

I am mostly confident that the problem is not really with `octokit` but where can i go to better understand the tools being used?

shiftkey:
Fixes #1062

As part of #1062 I found that an extension methods for `Get<T>(uri)` is interfering with some of our unit tests - as it's an extension method contributors can't verify it was invoked correctly, and when that doesn't work blindly add the second parameter to get the test to pass. 

This is done in a lot of tests, and it's a trivial extension, so why not just make it a feature of the interface?

I'm happy to leave the other methods in `ApiExtensions` as-is, they're not used as broadly.

cc @Haacked for feels and any potential quirks about switching from extension method to interface method.

### DIFF
--- a/Octokit.Tests/Clients/BlobClientTests.cs
+++ b/Octokit.Tests/Clients/BlobClientTests.cs
@@ -21,8 +21,7 @@ namespace Octokit.Tests.Clients
 
                 client.Get("fake", "repo", "123456ABCD");
 
-                connection.Received().Get<Blob>(Arg.Is<Uri>(u => u.ToString() == "repos/fake/repo/git/blobs/123456ABCD"),
-                    null);
+                connection.Received().Get<Blob>(Arg.Is<Uri>(u => u.ToString() == "repos/fake/repo/git/blobs/123456ABCD"));
             }
 
             [Fact]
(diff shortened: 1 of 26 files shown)


Haacked:
The only downside is that if you implement the interface, you only need to implement the one `Get` method and you know the extension method is implemented correctly with regard to that one method. Now, implementers have to implement both and implement them correctly.

Having said that, this really does clean up the tests and API and reduces pain and it's a really really simple method so I'm :+1:

shiftkey:
@Haacked yeah, agreed on the need to implement this explicitly. If we had multiple implementations I'd probably explore something else...

PR #1064, opened 2016-01-25
Fix for Repository.Content.GetAllContents working at the root of the repo

### LINKED ISSUE #800: client.Repository.Content.GetAllContents for root not currently possible.

shiftkey:
Fixes #800 

There's some changes in #959 unrelated to the problem reported in #800 - and an outstanding merge conflict - so I've pulled in just those commits, cleaned up the xml-docs issue, and pushed this to verify the fix.

 - [x] there's some `async/await` i think i can clean up here?
 - [x] pretty sure this is a breaking change - document this for the release?

cc @naveensrinivasan to ensure I didn't miss any commits


naveensrinivasan:
@shiftkey  Thanks for pulling this in.

shiftkey:
Ready for review

shiftkey:
Actually, given #959 went through a review by myself and @haacked I'm just going to merge this in so I can get it into the next release...

jamesqo:
I just ran into this error. For some reason, it looks like the argument validation was added back during https://github.com/octokit/octokit.net/pull/1348 and no one took notice.

PR #1072, opened 2016-01-31
Add member role filter to OrganizationMembersClient.GetAll() functions

ryangribble:
While working on implementing GitHub Enterprise APIs I came across the fact that the OrganizationMembers API implementation was missing the "Role" filter option (all, admin, member).  
https://developer.github.com/v3/orgs/members/#parameters

I also added integration tests for the GetAll method (previously there were none for this whole API) and all work fine in my integration tests... 

however I found that querying for Admin/Member role only returns correct results when the requester is a member of the organization.  When the requester is not in the organization, the member and admin filters just return all users in the org.  

I also found that with the existing "filter" attribute (2fa_disabled, all) only owners of an organization are allowed to use the ```2fa_disabled``` filter, and octokit throws an exception if the requester is not an org owner.

Im not sure what settings you guys run the integration tests under for octokit.net so I have set the integration tests that would have triggered these conditions, to SKIP for now.  If the integration tests on AppVeyor/Travis do run as an org member or org admin, let me know and I can un-skip the tests (i also need to know the name of the org to use in the queries... currently "octokit" is used).

### DIFF
--- a/Octokit.Reactive/Clients/IObservableOrganizationMembersClient.cs
+++ b/Octokit.Reactive/Clients/IObservableOrganizationMembersClient.cs
@@ -60,6 +60,53 @@ namespace Octokit.Reactive
         IObservable<User> GetAll(string org, string filter);
 
         /// <summary>
+        /// <para>
+        /// List all users who are members of an organization. A member is a user that
+        /// belongs to at least 1 team in the organization.
+        /// </para>
+        /// <para>
+        /// If the authenticated user is also an owner of this organization then both
+        /// concealed and public member will be returned.
+        /// </para>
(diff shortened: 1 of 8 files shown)


shiftkey:
@ryangribble wow, this is great! Thanks for catching this!

> Im not sure what settings you guys run the integration tests under for octokit.net so I have set the integration tests that would have triggered these conditions, to SKIP for now. 

This is fine, and while I use a separate account for testing I'm not sure what might work for it around organizations. I'm aiming to prepare a new release this week, so I'll look into that (probably needs an extensibility point :sad:) while I'm preparing and testing everything...

PR #1085, opened 2016-02-03
StatisticsClient should not clobber /api/v3/ in path

### LINKED ISSUE #822: StatisticsClient isn't Enterprise-friendly
The URLs created internally need to respect Enterprise environments, and the `/api/v3/` prefix seems to be skipped based on some testing...

shiftkey:
Fixes #822

All other routes - e.g. `ApiUrls` - are defined without a leading slash.

Except these. This fixes that.

 - [x] fix the tests - you're better than this

### DIFF
--- a/Octokit.Tests/Clients/StatisticsClientTests.cs
+++ b/Octokit.Tests/Clients/StatisticsClientTests.cs
@@ -24,7 +24,7 @@ namespace Octokit.Tests.Clients
             [Fact]
             public async Task RetrievesContributorsForCorrectUrl()
             {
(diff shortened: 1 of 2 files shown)


shiftkey:
Gotta rebase once #1087 is merged

shiftkey:
Ready for review

Haacked:
Nice catch!

Haacked:
We should do a Roslyn analyzer to not allow absolute `Uri` in Octokit.net since I can't think of any place where they would be allowed. We can always suppress in those cases where they're necessary.

(attempted nerd snipe)

shiftkey:
> (attempted nerd snipe)

PR #1073, opened 2016-01-31
Implement GitHub Enterprise License and Organization API's

ryangribble:
## Status
:rocket:Ready to merge:rocket:

## Enhance integration test support for GitHub Enterprise
https://github.com/tattsgroup/octokit.net/commit/535709c3689eaa6b6a4556757c11d0d50b0aae0e
Added a new EnterpriseHelper class for creating enterprise clients, using a new set of environment variables for this purpose.  Enhanced the powershell script to assist in setting up these GitHub Enterprise integration test variables if desired

## Implement Enterprise License Client 
https://developer.github.com/v3/enterprise/license/
Fixes #1023
Includes normal and reactive implementations plus unit and integration tests for both

## Implement Enterprise Organization Client 
https://developer.github.com/v3/enterprise/orgs/
Fixes #1025
Includes normal and reactive implementations plus unit and integration tests for both

## Tests
Obviously the integration tests require a GitHub enterprise instance so wont run as part of Octokit builds currently.  Integration tests are passing on my GHE instance

PR #1093, opened 2016-02-04
Add overloads to specify branch for ContentRequest

### LINKED ISSUE #958: Creating file on a specific branch.
In the [github api](https://developer.github.com/v3/repos/contents/#create-a-file) I can see that I should be able to create files on non-master branches by specifying an optional "branch" parameter.

Is there a way to do this in octokit.net?

Thanks!

M-Zuber:
Fixes #958

### DIFF
--- a/Octokit.Tests.Integration/Clients/RepositoryContentsClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/RepositoryContentsClientTests.cs
@@ -107,7 +107,7 @@ namespace Octokit.Tests.Integration.Clients
                     new DeleteFileRequest("Deleted file", fileSha));
 
                 await Assert.ThrowsAsync<NotFoundException>(
-                    async () => await fixture.GetAllContents(repository.Owner.Login, repository.Name, "somefile.txt"));
+                     () => fixture.GetAllContents(repository.Owner.Login, repository.Name, "somefile.txt"));
             }
         }
 
@@ -154,7 +154,7 @@ namespace Octokit.Tests.Integration.Clients
                     new DeleteFileRequest("Deleted file", fileSha, branchName));
 
                 await Assert.ThrowsAsync<NotFoundException>(
-                    async () => await fixture.GetAllContents(repository.Owner.Login, repository.Name, "somefile.txt"));
+                    () => fixture.GetAllContents(repository.Owner.Login, repository.Name, "somefile.txt"));
             }
         }
 
--- a/Octokit.Tests.Integration/Clients/RepositoryContentsClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/RepositoryContentsClientTests.cs
@@ -111,6 +111,53 @@ namespace Octokit.Tests.Integration.Clients
             }
         }
 
+        [IntegrationTest]
+        public async Task CrudTestWithNamedBranch()
+        {
+            var client = Helper.GetAuthenticatedClient();
+            var fixture = client.Repository.Content;
+            var repoName = Helper.MakeNameWithTimestamp("source-repo");
+            var branchName = "other-branch";
+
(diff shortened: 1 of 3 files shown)


Octokit/Models/Request/CreateFileRequest.cs
```diff
@@ -91,6 +116,18 @@ public CreateFileRequest(string message, string content) : base(message)
         }
 
         /// <summary>
+        /// Initializes a new instance of the <see cref="DeleteFileRequest"/> class.
```
ryangribble: Typo in the xml doc with the cref to the class name

ryangribble:
I noted one typo of a class name in XmlDoc comments

Also could I suggest updating the unit tests to cover off the addition of the new branch parameter, and also adding an integration test to prove it actually creates the file on the specified branch?

M-Zuber:
Thank you for catching the typo! 
I should be able to add unit tests after the weekend. 

Would you test anything in the unit tests other then that the correct url was called? 

The integration test can create, update, and then delete a file on the same branch. Sounds good to you?

ryangribble:
Yeah if you look at the "crud" integration test I'd probably do the same actions as that, but specifying a named branch 

For unit test since the url is the same anyway, you could check that the CreateFileRequest object passed through, has the correct path/message/content/branch present in it.  Eg  have look at ```PassesRequestObject()``` test [here] (https://github.com/octokit/octokit.net/pull/1073/files#diff-1570280d72efc6265b42171826428691)

M-Zuber:
I wasn't able to get the unit tests to work, hopefully the integration test will be enough

ryangribble:
Hey @M-Zuber that looks good... your integration test is passing for me here.

On the topic of unit tests, your change was mostly to a Request object and it doesnt look like there are many unit tests of those things in octokit.net.  

However if you are interested, I did just send you a pull request that implements unit tests on the CreateFile() DeleteFile() and UpdateFile() calls in RepositoryClient, including making sure your new branch field is correctly passed through to the underlying ApiClient.  What ive found with Octokit unit tests for API clients, is we typically check 
- calls the correct URL
- passes through the request object
- checks parameters for nulls and throws exceptions if paramter is required

Octokit.Tests.Integration/Clients/RepositoryContentsClientTests.cs
```diff
@@ -111,6 +111,53 @@ public class TheGetContentsMethod
...
+                fileSha = contents.First().Sha;
+
+                await fixture.DeleteFile(
+                    repository.Owner.Login,
+                    repository.Name,
+                    "somefile.txt",
+                    new DeleteFileRequest("Deleted file", fileSha, branchName));
+
+                await Assert.ThrowsAsync<NotFoundException>(
+                    async () => await fixture.GetAllContents(repository.Owner.Login, repository.Name, "somefile.txt"));
```
khellang: I don't think this state machine is necessary. `Assert.ThrowsAsync` expects a `Task`, which `GetAllContents` already returns :smile:

M-Zuber: Copy paste error :smile:
Will fix when I'm back by a machine.

M-Zuber:
@ryangribble thank you for the PR. I asked a question there.

Thwaitesy:
@ryangribble when do you think this will be merged?

ryangribble:
I don't have commit rights, am sitting on a few pending PR's myself 😀

shiftkey:
@Thwaitesy @ryangribble apologies for the delay, I'm aiming to get through these by the end of the week. Thankfully we just did a big release, so the next one should be a bit easier to manage...

shiftkey:
Code looks good, tests all pass, and it got a pretty great review too...

M-Zuber:
YAY!!  I got a Stevie!! :grinning:

Issue #1026, opened 2015-12-23
Implement Enterprise - Search Indexing client


shiftkey:
New in Enterprise 2.4

https://developer.github.com/v3/enterprise/search_indexing/

```
/enterprise/search_indexing/
 - POST /staff/indexing_jobs
```

ryangribble:
Im going to take this one  too

Question: Given the various options that are passed in to this API call as "target" parameter, what would be the preferred approach?  I was thinking of creating individual methods for the various options...  I haven't found a similar API elsewhere yet, though Im not familiar with the entire API.

```
Target						Description
:owner						A user or organization account.
:owner/:repository			A repository.
:owner/*					All of a user or organization's repositories.
:owner/:repository/issues	All the issues in a repository.
:owner/*/issues				All the issues in all of a user or organization's repositories.
:owner/:repository/code		All the source code in a repository.
:owner/*/code				All the source code in all of a user or organization's repositories.
```
So my proposal would be to create functions similar to:

```
Enterprise.SearchIndexing.QueueOwner(string owner)
Enterprise.SearchIndexing.QueueRepository(string owner, string repository)
Enterprise.SearchIndexing.QueueAllRepositoriesForOwner(string owner)
Enterprise.SearchIndexing.QueueAllIssuesForRepository(string owner, string repository)
Enterprise.SearchIndexing.QueueAllIssuesForOwner(string owner)
Enterprise.SearchIndexing.QueueAllCodeForRepository(string owner, string repository)
Enterprise.SearchIndexing.QueueAllCodeForOwner(string owner)
```

Sound OK?

ryangribble:
Or another approach would be to have a single method with overrides for owner, or owner+repo, and an enum to specify the scoping

```
Enterprise.SearchIndexing.QueueJob(string owner, IndexTypeOwner indexType)
Enterprise.SearchIndexing.QueueJob(string owner, string repository, IndexTypeRepository indexType)

enum IndexTypeOwner
{
Owner,
AllOwnedRepositories,
AllIssues,
AllCode
}

enum IndexTypeRepository
{
Repository,
AllIssues,
AllCode
}
```

:question:

shiftkey:
@ryangribble the first snippet could be distilled down to this:

```
Enterprise.SearchIndexing.Queue(string owner)
Enterprise.SearchIndexing.Queue(string owner, string repository)
Enterprise.SearchIndexing.QueueAll(string owner)
[...]

PR #1094, opened 2016-02-05
Fix Json deserializion of string containing hyphens to List<string> property

ryangribble:
### Issue
As discussed in [#1026](https://github.com/octokit/octokit.net/issues/1026#issuecomment-178856752), there is an issue with the Json parser in a specific edge case.

When deserialising a ```string``` property into a class where that property is declared as a ```List<string>``` (as is the case with the responses from the Enterprise SearchIndexing API), hyphens in the string are stripped during deserialization.

Json
```{ "message":"User \"ryan-gribble\" was added to the indexing queue" }```

Class
```public class MessageList { public IReadOnlyList<string> Message { get; private set; } }```

Result
```"User \"ryangribble\" was added to the indexing queue"```

Expected Result
```"User \"ryan-gribble\" was added to the indexing queue"```

### Resolution
[This line](https://github.com/octokit/octokit.net/blob/master/Octokit/Http/SimpleJsonSerializer.cs#L93) (introduced in #727) was identified as responsible for the removal of the hyphens in the input string.  Code inspection indicates the line appears to be unnecessary for the actual ```Enum``` and ```Nullable Enum``` cases mentioned when the change was made, as they already had code to remove hyphens in their if blocks.

Unit tests were added to assert expected behaviour when deserializing a string attribute containing hyphens (and underscores for good measure) into ```string``` and ```<IReadOnlyList<string>``` properties.

After confirming the latter test failed, the offending line was removed and the tests then passed.

All other unit and integration tests were then also passed, indicating that no other known/tested functionality are impacted by the removal.

### DIFF
--- a/Octokit/Http/SimpleJsonSerializer.cs
+++ b/Octokit/Http/SimpleJsonSerializer.cs
@@ -90,7 +90,6 @@ namespace Octokit.Internal
                 var jsonValue = value as JsonObject;
                 if (stringValue != null)
                 {
-                    stringValue = stringValue.Replace("-", "");
                     if (ReflectionUtils.GetTypeInfo(type).IsEnum)
                     {
                         // remove '-' from values coming in to be able to enum utf-8
--- a/Octokit.Tests/SimpleJsonSerializerTests.cs
+++ b/Octokit.Tests/SimpleJsonSerializerTests.cs
@@ -1,5 +1,6 @@
 ﻿using Octokit.Helpers;
 using Octokit.Internal;
+using System.Collections.Generic;
 using System.Linq;
 using System.Text;
 using Xunit;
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by ryangribble** [comment]
@shiftkey Linux build failure looks unrelated to me

File: Octokit/Http/SimpleJsonSerializer.cs [hunk 1]
```diff
@@ -90,7 +90,6 @@ public override object DeserializeObject(object value, Type type)
                 var jsonValue = value as JsonObject;
                 if (stringValue != null)
                 {
-                    stringValue = stringValue.Replace("-", "");
```
shiftkey: The only reservation I had was that this code was being used elsewhere and some enum values depended on this, but then I saw those places were calling `RemoveHyphenAndUnderscore` and saw it had us covered:

```
static string RemoveHyphenAndUnderscore(string stringValue)
{
    stringValue = stringValue.Replace("-", "");
    // remove '-' from values coming in to be able to enum EventInfoState names with underscores in them. Like "head_ref_deleted" 
    stringValue = stringValue.Replace("_", "");
    return stringValue;
}
```

shiftkey:
@ryangribble thanks for getting to the bottom of this, and for adding in those tests!

Issue #1022, opened 2015-12-23
Implement Enterprise - LDAP client


shiftkey:
New in GitHub Enterprise 2.4

https://developer.github.com/v3/enterprise/ldap/

```
/enterprise/ldap/
 - PATCH /admin/ldap/users/:username/mapping
 - POST /admin/ldap/users/:username/sync
 - PATCH /admin/ldap/teams/:team_id/mapping
 - POST /admin/ldap/teams/:team_id/sync
```

ryangribble:
I'll take this one next

PR #1095, opened 2016-02-05
Implement GitHub Enterprise Search Indexing API

### LINKED ISSUE #1026: Implement Enterprise - Search Indexing client

ryangribble:
## Status
:rocket: Ready to merge!

~~This is built on top of #1073 which is not yet merged.  Only the top 3 commits are actually relating to this PR.  Once #1073 is merged, I will rebase this PR.~~

### Implement Enterprise Search Indexing Client 
https://developer.github.com/v3/enterprise/search_indexing/
Fixes #1026

## Test Notes
The integration tests require a GitHub Enterprise instance so wont run as part of Octokit builds currently.  Integration tests have been run succesfully against a local GHE instance


ryangribble:
@shiftkey Linux build failure looks unrelated. It's the "type initializer for list threw exception" variant

ryangribble:
Could you guys kick off the travis build again when you have a sec?  Failure was unrelated

shiftkey:
@ryangribble :boot:ed and :eyes:ing

ryangribble:
Dont forget that #1073 needs to go first, then this one, then #1022  :grinning:

shiftkey:
@ryangribble thanks for the heads-up, I'll make sure to prioritize these reviews before any other Octokit stuff as I know merging csproj changes is all kinds of terrible right now :grin:

shiftkey:
@ryangribble #1073 has been merged :metal:

ryangribble:
Awesome, this has now been rebased on master and should be ready for review and merge

ryangribble:
@shiftkey builds passed, ready when you are :)

Haacked:
:sparkles: Flawless Victory! :sparkles:

This looks great! The only comment I have is that if we add new targets, we'll have to add more methods to this class. We could consider adding a method that takes in a raw `target` string, but I don't think we need to do that now. Maybe this API very rarely changes.

ryangribble:
Yeah we had a bit of discussion on the issue #1026 about which way to go and ended up with the multiple methods

Haacked:
> Yeah we had a bit of discussion on the issue #1026 about which way to go and ended up with the multiple methods.

Ah, cool. Yeah, I like it. I'm just thinking about adding one additional method as a failsafe. But I don't want to do that yet.

> Thanks for merging!

Thanks for contributing!

ryangribble:
> Ah, cool. Yeah, I like it. I'm just thinking about adding one additional method as a failsafe. But I don't want to do that yet.

What would the failsafe be, out of interest?

> Thanks for contributing!
It's a pleasure on such a weel organised dotnet OSS project.  My goal is to implement all the GitHub Enterprise API's currently unsupported.  I think im just down to the Setup Console now (which is pretty big) as well as some extra methods on User Administration that are enterprise only.  

eg normal API doc User Administration:
https://developer.github.com/v3/users/administration/

vs Enterprise User Administration
https://developer.github.com/enterprise/2.5/v3/users/administration/

Heaps more nice methods to create/delete users, impersonation tokens, etc.  Fun times :grinning:

Haacked:
> What would the failsafe be, out of interest?

A method that took a target request object (or string). That way, if Enterprise ever added a new target string syntax, you wouldn't have to wait for the next Octokit.net to start calling it.

ryangribble:
Nice idea, I'll add it in the next couple of days!

PR #1097, opened 2016-02-07
Code to test URL for an enterprise instance.

Haacked:
Based on code we use in GitHub Desktop for Windows.

 - [x] resolve merge conflict
 - [ ] add section to docs [here](http://octokitnet.readthedocs.org/en/latest/getting-started/#connect-to-github-enterprise) about probing?


shiftkey:
Ooooooh!



Gonna look this over and cc @shana for a bit of review :eyes: when she has a moment.

shiftkey:
@haacked only extra bit i can think of here is a bit of documentation on using this. Thoughts?

Haacked:
Where should I put the documentation? In the README?

shiftkey:
@haacked the content for the documentation site is under `docs` - and the hierarchy for http://octokitnet.readthedocs.org/en/latest/ is [here](https://github.com/octokit/octokit.net/blob/master/mkdocs.yml#L15-L39) - not sure where it fits best, open to suggestions...

ryangribble:
Potentially a specific section on "working with GitHub Enterprise" in the docs?

Im still not entirely sure what the intention of this probe method is for.  Is it for octokit to do smart stuff internally, or just providing an API where someone writing an application like GitHub Desktop could allow users to enter a repo URL and it would be able to probe if its Enterprise or not?

shiftkey:
>  Is it for octokit to do smart stuff internally, or just providing an API where **someone writing an application like GitHub Desktop could allow users to enter a repo URL and it would be able to probe if its Enterprise or not**?

It's the latter - basically a helper method for someone to verify the user-specified input is a valid GitHub Enterprise server.

Haacked:
It requests the one unathenticated endpoint in a GitHub Enterprise instance.

Octokit/Clients/Enterprise/EnterpriseProbe.cs
```diff
@@ -0,0 +1,97 @@
...
+                var response = await httpClient.Send(request);
+                return response == null
+                    ? EnterpriseProbeResult.Failed
+                    : (IsEnterpriseResponse(response)
+                        ? EnterpriseProbeResult.Ok
+                        : EnterpriseProbeResult.NotFound);
+            }
+            catch (Exception)
+            {
+                return EnterpriseProbeResult.Failed;
```
shana: What I found while poking at enterprise responses is that if it throws an exception, it might still be an enterprise server, it just might be configured to not let you poke without authentication, even if this entry point is supposed to be unauthenticated. In this case, the exception will have the response, so this can look for that (I do it like so [here](https://github.com/github/VisualStudio/blob/74ee1705e088f4250b5ba195046f48ce0ab2e9fa/src/GitHub.Exports/ExceptionExtensions.cs))

Octokit/Clients/Enterprise/EnterpriseProbe.cs
```diff
@@ -0,0 +1,97 @@
...
+                Method = HttpMethod.Get,
+                BaseAddress = enterpriseBaseUrl,
+                Endpoint = endPoint,
+                Timeout = TimeSpan.FromSeconds(3)
+            };
+            request.Headers.Add("User-Agent", productHeader.ToString());
+
+            try
+            {
+                var response = await httpClient.Send(request);
```
shana: Could just do 

```
.Catch(ex => (ex as ApiException)?.HttpResponse ?? null);
```

here so the `IsEnterpriseResponse` below can handle both successful and failed http requests in one go. For the purposes of the check, we don't really care if it's successful, we just care to poke at the headers to see who's responding, right?

shiftkey: Editor's note: we're not quite on C# 6 here :cry:

shiftkey: Perhaps we can catch an `ApiException` below and interrogate the headers there...

Haacked: Seems legit to me. That's what GHfVS does. @niik, any opinion?

shana: Ah good ol' C# 6. I guess

```
IResponse response = null;
try
{
    response = await httpClient.Send(request);
}
catch (ApiException ex)
{
    response = ex.HttpResponse;
}
catch
{
}

return response == null
    ? EnterpriseProbeResult.Failed
    : (IsEnterpriseResponse(response)
        ? EnterpriseProbeResult.Ok
        : EnterpriseProbeResult.NotFound);
```

should work equally well.

shiftkey: :thumbsup:

Octokit/Clients/Enterprise/EnterpriseProbe.cs
Haacked: But wouldn't a request for "github.com" also succeed in that case? For a general purpose method, we'll need to distinguish between requests for github.com and github enterprise.

Haacked: I think I'll just validate that the URL is not github.com

Haacked:
Hey @niik would like your input here. I relaxed the check based on some comments from @shana. I'd like to use this method in both Desktop and GHfVS so we're consistent.

Basically, we check the response for a couple headers even if the request gets thrown an `ApiException`.

shana:
@Haacked On GHfVS, this probe only runs if the url is not github.com, that check happens way before this. Unsure whether that would be a useful assumption here or not though :P

Octokit/Octokit-MonoAndroid.csproj
```diff
@@ -450,11 +450,16 @@
...
+<<<<<<< ee71652f35732a13d08dbe58574a9ec677f38c0b
     <Compile Include="Clients\Enterprise\EnterpriseSearchIndexingClient.cs" />
     <Compile Include="Clients\Enterprise\IEnterpriseSearchIndexingClient.cs" />
     <Compile Include="Models\Request\Enterprise\NewOrganization.cs" />
     <Compile Include="Models\Request\Enterprise\SearchIndexingTarget.cs" />
     <Compile Include="Models\Response\Enterprise\SearchIndexingResponse.cs" />
+=======
+    <Compile Include="Clients\Enterprise\EnterpriseProbe.cs" />
+    <Compile Include="Clients\Enterprise\IEnterpriseProbe.cs" />
+>>>>>>> Code to test URL for an enterprise instance.
```
shana: Looks like you got a merge failure here.

shiftkey:
Just added a couple of tasks to get this over the line. I'm happy with the current implementation.

shiftkey:
Merge conflict is sorted, but I'm still trying to find a simple code sample here to illustrate the issue - currently you need to create it manually, like this:

```
var probe = new EnterpriseProbe(productHeader, httpClient);
var result = await probe.Probe(new Uri("enterpriseBaseUrl"));
if (result != EnterpriseProbeResult.Ok)
{
    // not valid
}
```

Thoughts @haacked @shana on how we could simplify this?

shana:
@shiftkey You mean a simple usage example? I think it should probably show what the return codes mean (Ok vs Failed vs NotFound), so people know how to properly handle them.

shiftkey:
@shana sorry, I should have been more clear. I wanted to put this into the "Getting Started" page but there's some setup of the dependencies that `EnterpriseProbe` needs before I can use any code sample for the probe. Do I also do that here, as painful as it is to create these?

Or perhaps I could leverage the `GitHubClient` setup and move `EnterpriseProbe` into `GitHubClient`, like this:

 ```csharp
var github = new GitHubClient(...);
var result = await gitHub.Enterprise.Probe.Probe(new Uri("enterpriseBaseUrl"));
if (result != EnterpriseProbeResult.Ok)
{
    // not valid
}
```

But that's also kinda backward, as `GitHubClient` requires the base URL...

**Comment by shana** [comment]
@shiftkey yeah, that's awkward, EnterpriseProbe requires GitHubClient, not the other way around. In VS I have MEF for this, which basically means that the probe constructor gets the productHeader and httpClient info from MEF, and whoever calls probe gets an instance of EnterpriseProbe through its constructor. That basically boils down to the example you had - do a `new`, await the probe, check the result - but the `new` is split off and done earlier.

So I think the example you first posted is appropriate for a simple usage. I think it's more important to clearly show what the return codes mean - Ok, Failed and NotFound - so people know how to handle them. How the EnterpriseProbe instance gets created is really a secondary thing, methinks. We could also have a static Probe method on EnterpriseProbe for a quick one-off, for scenarios where that makes sense.

**Comment by Haacked** [comment]
`gitHub.Enterprise.Probe.Probe(new Uri("enterpriseBaseUrl"));`

That `Probe.Probe` looks awkward. It's not helped by the fact that `Probe` is both a noun and a verb. What do you think about changing the method name to `Examine` since that's what a probe does.

`gitHub.Enterprise.Probe.Examine(new Uri("enterpriseBaseUrl"));`

**Comment by Haacked** [comment]
Another thought, we could simply put the `Probe` method on `Enterprise`.

`github.Enterprise.Probe(new Uri(...));`

**Comment by Haacked** [comment]
I kind of like that latter approach better to be honest, but would like to hear others thoughts.

**Comment by shana** [comment]
Not to derail the whole thing, but I was looking at it and it feels like this shouldn't actually be a separate class specifically to poke enterprise, at least not as a public API.

Correct me if I'm wrong: The goal here is to identify whether the remote server is a github instance, and if it is, whether it's an enterprise instance, right? With that in mind, the solution should probably be a method that returns what type of server it is - dotcom, enterprise, other (not github) or unknown (unable to determine).  Since `GitHubClient` encapsulates server information, the method should probably live here, so something like

```
enum ServerType {
 DotCom,
 Enterprise,
 Other,
 Unknown (or NotFound?)
}

class GitHubClient {
 public ServerType GetServerType() {}
```

The `EnterpriseProbe` class can be kept as an internal implementation of the probe, and the `GetServerType()` method calls it if the uri is not dotcom.

What do you think? (I'm happy to patch up the PR if you agree on this approach, I know you're crazy busy these days)

**Comment by M-Zuber** [comment]
In agreement with @shana (Someone needs to make a thesaurus for things to say instead of +1 :wink: )

**Comment by shiftkey** [comment]
Yeah, I like the direction that @shana is proposing about internalizing it - you've already specified the URL in the `ctor`, so leave the internal plumbing to Octokit.

Other thoughts:

 - I'm unsure about having the probe returning `DotCom` - this feels a lot like our internal lingo, but as you could invoke this on an instance without the URL set, which would be DotCom behaviour.

 - What if this were a method off `GitHubClient.Enterprise`? that's originally why we wrote this probing code, and keeping it near there might be more contextual than having it off `GitHubClient`

**Comment by shana** [comment]
@shiftkey Not sure why the original code was written like that. In my code, calling EnterpriseProbe always happens together with determining what the server is - so I only call it if it's not a dotcom, and sometimes will check if it's an enterprise for authentication purposes. So enterprise probe is actually just half of the logic of identifying what the server is, and is pretty much always used in that way. By itself is not terribly useful, it needs to be combined with "is this github.com, yes/no? is this enterprise, yes/no?"

> I'm unsure about having the probe returning DotCom - this feels a lot like our internal lingo, but as you could invoke this on an instance without the URL set, which would be DotCom behaviour.

The url is always set, it's just defaulting to "github.com" :wink:

It's actually important to distinguish between dotcom and non-dotcom github servers, because authentication differs for those,the UI may be different depending on which one is it, and if it's neither, UI must be hidden in my case, so I need to know that as well. This is why I'm suggesting having a server identification method that tells me what type of server it is, rather than something that's enterprise-specific.

**Comment by Haacked** [comment]
@shana I like that approach.

**Comment by Haacked** [comment]
One thought that just occurred to me. The reason why it's an "Enterprise Probe" today is that there's no need to probe if the base URL is "github.com" or "api.github.com". We can just assume it's GitHub.com the website and not something else.

However, I could imagine someone wanting to probe for capabilities in the future, in which this API could be the starting point for that.

**Comment by shana** [comment]
> However, I could imagine someone wanting to probe for capabilities in the future, in which this API could be the starting point for that.

Yeah, I actually already do something like that for wikis, and it would be nice if instead of manually firing probes I could just ask for capabilities from the API.

**Comment by ryangribble** [comment]
I like the idea of making this more like a ```GetCapabilities``` client/request.

Also, this may be a dumb question but what if someone has a hostfile entry or DNS alias to point "external.github" at "github.com".  To me it seems like rather than assume that "only" github.com is "DotCom" you sort of want to probe/confirm even "DotCom" capability, rather than going only off the URL?

**Comment by shana** [comment]
@ryangribble If the url isn't "github.com", then we poke it to see whether it is a github instance, and if it is, we classify it as an enterprise instance. In your scenario this would be a false positive (or negative?), but in practice it doesn't change much, clients will still know that the server is a github instance that they can interact with.

Distinguishing between a github.com and a github enterprise instance is not enough to tell a client what the server can actually do - the only important bit of information they can get out of it is whether the server is a github server or not. The only reason we're actually distinguishing between github.com and an enterprise instance is because we know that enterprise instances can potentially support only a subset of authentication types if they're older, so we're using the url as a shortcut to only run the fallback code if it's a github server but it's not github.com (which we know supports all the things).

Again, in your scenario, if the url is aliased, a client would fallback to handling the server more carefully, so it's not a big deal if there's a false positive because of that.

Concrete server capabilities checks I think should probably be done in separate calls, since all of these things require poking the server and aren't always needed by a client. In this way, we end up with a "Is this a server we can talk to?" call and "Ah, it is! So I want to use this feature, is it available?" calls.

**Comment by Haacked** [comment]
Ok, trying to get this back on track, I'm cool with the original suggestion you had @shana. Especially since you offered to do it. :stuck_out_tongue:

**Comment by shana** [comment]
@Haacked haha, yes... should I just push to this branch?

**Comment by Haacked** [comment]
Sure!

**Comment by shiftkey** [comment]
Closing this one out as it's gone very stale.

I'll leave the branch around if someone wants to pick this up later.

**Comment by shana** [comment]
Damnit, sorry, I dropped the ball on this one 😞

File: Octokit/Clients/Enterprise/EnterpriseProbe.cs [hunk 4]
```diff
@@ -0,0 +1,101 @@
...
+                ? EnterpriseProbeResult.Failed
+                : (IsEnterpriseResponse(response)
+                    ? EnterpriseProbeResult.Ok
+                    : EnterpriseProbeResult.NotFound);
+
+        }
+
+        static bool IsEnterpriseResponse(IResponse response)
+        {
+            return response.Headers["Server"] == "GitHub.com"
```
Haacked: We've run into a case where we might not be able to rely on this assumption. :cry:

Octokit/Clients/Enterprise/EnterpriseProbe.cs
```diff
@@ -0,0 +1,101 @@
...
+                && response.Headers.ContainsKey("X-GitHub-Request-Id");
```
Haacked: However, relying on this is safe because if they strip this out, it would break GitHub Enterprise.

PR #1103, opened 2016-02-10
Create branch helper method

### LINKED ISSUE #1098: Creating a branch
Currently (and this is per the API) there is no comfortable way to create a branch.
You have to do something like the following:
```c#
var master = await client.Git.Reference.Get(Helper.UserName, repository.Name, "heads/master");
await client.Git.Reference.Create(Helper.UserName, repository.Name, new NewReference("refs/heads/" + branchName, master.Object.Sha));
```

Is there room to consider creating a wrapper for this?

If so, is there also a way to cut down on the calls? Or would it just be for the users convenience?

M-Zuber:
Fixes #1098.

### DIFF
--- a/Octokit.Tests.Integration/Helpers/ReferenceExtensionsTests.cs
+++ b/Octokit.Tests.Integration/Helpers/ReferenceExtensionsTests.cs
@@ -24,8 +24,8 @@ namespace Octokit.Tests.Integration.Helpers
 
                 var allBranchNames = (await client.Repository.GetAllBranches(context.RepositoryOwner, context.RepositoryName)).Select(b => b.Name);
 
-                Assert.Contains("patch-1", allBrancheNames);
-                Assert.Contains("patch-2", allBrancheNames);
+                Assert.Contains("patch-1", allBranchNames);
+                Assert.Contains("patch-2", allBranchNames);
             }
         }
     }
--- a/Octokit.Tests.Integration/Helpers/ReferenceExtensionsTests.cs
+++ b/Octokit.Tests.Integration/Helpers/ReferenceExtensionsTests.cs
@@ -1,0 +1,32 @@
+﻿using Octokit.Helpers;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Octokit.Tests.Integration.Helpers
+{
+    public class ReferenceExtensionsTests
+    {
+        [IntegrationTest]
+        public async Task CreateABranch()
+        {
+            var client = Helper.GetAuthenticatedClient();
+            var fixture = client.Git.Reference;
+
+            using (var context = await client.CreateRepositoryContext("public-repo"))
+            {
+                var branchFromMaster = await fixture.CreateBranch(context.RepositoryOwner, context.RepositoryName, "patch-1");
+
+                var branchFromPath = await fixture.CreateBranch(context.RepositoryOwner, context.RepositoryName, "patch-2", branchFromMaster);
+
+                var allBrancheNames = (await client.Repository.GetAllBranches(context.RepositoryOwner, context.RepositoryName)).Select(b => b.Name);
+
+                Assert.Contains("patch-1", allBrancheNames);
+                Assert.Contains("patch-2", allBrancheNames);
+            }
+        }
+    }
+}
--- a/Octokit.Tests.Integration/Helpers/ReferenceExtensionsTests.cs
+++ b/Octokit.Tests.Integration/Helpers/ReferenceExtensionsTests.cs
@@ -22,7 +22,7 @@ namespace Octokit.Tests.Integration.Helpers
 
                 var branchFromPath = await fixture.CreateBranch(context.RepositoryOwner, context.RepositoryName, "patch-2", branchFromMaster);
 
-                var allBrancheNames = (await client.Repository.GetAllBranches(context.RepositoryOwner, context.RepositoryName)).Select(b => b.Name);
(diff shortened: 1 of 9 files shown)


Octokit/Helpers/ReferenceExtensions.cs
```diff
@@ -0,0 +1,45 @@
+﻿using System.Threading.Tasks;
+
+namespace Octokit.Helpers
+{
+    /// <summary>
+    /// Represents operations to simplify workink with references
```
shiftkey: s/workink/working/

Octokit.Tests.Integration/Helpers/ReferenceExtensionsTests.cs
```diff
@@ -0,0 +1,32 @@
...
+            var client = Helper.GetAuthenticatedClient();
+            var fixture = client.Git.Reference;
+
+            using (var context = await client.CreateRepositoryContext("public-repo"))
+            {
+                var branchFromMaster = await fixture.CreateBranch(context.RepositoryOwner, context.RepositoryName, "patch-1");
+
+                var branchFromPath = await fixture.CreateBranch(context.RepositoryOwner, context.RepositoryName, "patch-2", branchFromMaster);
+
+                var allBrancheNames = (await client.Repository.GetAllBranches(context.RepositoryOwner, context.RepositoryName)).Select(b => b.Name);
```
shiftkey: `s/Branche/Branch/`

M-Zuber:
I'm on my phone so I was only able to fix the spelling. 
I'll assume the errors from ci are due to merge conflicts. 
Will deal with it when I'm in the office //@shiftkey

M-Zuber:
@Haacked is this good to go?

Octokit/Helpers/ReferenceExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// <param name="branchName">The new branch name</param>
+        /// <param name="baseReference">The <see cref="Reference" /> to base the branch from</param>
+        public static async Task<Reference> CreateBranch(this IReferencesClient referencesClient, string owner, string name, string branchName, Reference baseReference)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
+            Ensure.ArgumentNotNullOrEmptyString(name, "name");
+            Ensure.ArgumentNotNullOrEmptyString(branchName, "branchName");
+            Ensure.ArgumentNotNull(baseReference, "baseReference");
+
+            return await referencesClient.Create(owner, name, new NewReference("refs/heads/" + branchName, baseReference.Object.Sha));
```
Haacked: Should we validate that `branchName` doesn't already start with `refs/heads/`? I could imagine that might be a point of confusion in calling this overload and a nice exception message would be helpful. Then again, what if they really wanted the ref to be named `refs/heads/refs/heads`. I mean, that's probably silly sauce and we should just tell them to call the original `Create` method in that case.

Haacked:
Looking good! One minor question.

Octokit/Helpers/ReferenceExtensions.cs
M-Zuber: The parameter name is `branchName`. I would hope that should be enough, but I do not see a problem with adding that check.

Octokit/Helpers/ReferenceExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// <param name="owner">The owner of the repository.</param>
+        /// <param name="name">The name of the repository.</param>
+        /// <param name="branchName">The new branch name</param>
+        public static async Task<Reference> CreateBranch(this IReferencesClient referencesClient, string owner, string name, string branchName)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
+            Ensure.ArgumentNotNullOrEmptyString(name, "name");
+            Ensure.ArgumentNotNullOrEmptyString(branchName, "branchName");
+
+            var baseBranch = await referencesClient.Get(owner, name, "heads/master");
```
M-Zuber: Interesting to note that passing in `"ref/heads/master"` does not work here.

Octokit/Helpers/ReferenceExtensions.cs
```diff
@@ -0,0 +1,56 @@
...
+        public static async Task<Reference> CreateBranch(this IReferencesClient referencesClient, string owner, string name, string branchName, Reference baseReference)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
+            Ensure.ArgumentNotNullOrEmptyString(name, "name");
+            Ensure.ArgumentNotNullOrEmptyString(branchName, "branchName");
+            Ensure.ArgumentNotNull(baseReference, "baseReference");
+
+            if (branchName.StartsWith("refs/heads"))
+            {
+                throw new ArgumentException("The branchName seems to be misformed. Try using the Create method");
```
M-Zuber: Is this about what you had in mind @Haacked?

Haacked: Close! :smile:

I was thinking something more specific.

```csharp
throw new ArgumentException(String.Format(CultureInfo.InvariantCulture, "The specified branchName '{0}' appears to be a ref name and not a branch name because it starts with the string 'refs/heads'. Either specify just the branch name or use the Create method if you need to specify the full ref name", branchName));
```

M-Zuber:
@Haacked any chance we can wrap this up without more merge conflicts :wink:

Haacked:
Thanks!

M-Zuber:
Mad skills with the ball :grinning:

Issue #1106, opened 2016-02-13
UserKeysClient only implements 2 of the 5 calls for this API


ryangribble:
[```IUserKeysClient```](https://github.com/octokit/octokit.net/blob/master/Octokit/Clients/IUserKeysClient.cs) currently only implements 2 of the 5 "public keys" methods on the ```users/keys``` GitHub API endpoint

https://developer.github.com/v3/users/keys/

### Currently implemented:
List public keys for a user
```   GET /users/:username/keys```

List public keys for the authenticated user
```GET /user/keys```

### Not yet implemented:
Get a single public key
```GET /user/keys/:id```

Create a public key
```POST /user/keys```

Delete a public key
```DELETE /user/keys/:id```

**Comment by shiftkey** [comment]
I knew this code lived somewhere (ignore the second TODO comment):

```
// TODO: this should be under Users to align with the API docs
// TODO: this should be named PublicKeys to align with the API docs
/// <summary>
/// Access GitHub's Public Keys API.
/// </summary>
/// <remarks>
/// Refer to the API docmentation for more information: https://developer.github.com/v3/users/keys/
/// </remarks>
public ISshKeysClient SshKey { get; private set; }
```

So yes, the methods on `ISshKeysClient` should be added to `IUserKeysClient` - and `IGitHubClient.SshKey` should be marked as obsolete for removal as it doesn't align with the API.

ryangribble:
> I knew this code lived somewhere
Ah right, we even use that ```SshKeyClient``` at the moment but I was tunnel visioning on the ```User.Keys``` side of things and didnt think that they are the same thing!

> So yes, the methods on ISshKeysClient should be added to IUserKeysClient - and IGitHubClient.SshKey should be marked as obsolete for removal as it doesn't align with the API.
i'm in the process of doing this now, all the methods are transferred but I have some questions about the obsolescence stuff

I've marked ```IGitHubClient.SshKey``` obsolete but that doesn't actually flag several use cases...  So should we actually also mark ```GitHubClient.SshKey``` obsolete (that gets anybody using a ```GitHubClient``` rather than an ```IGitHubClient```).   Then we need of course the ```Observable``` and ```IObservable``` versions as well.  Do we go as far to make the actual class ```SshKeysClient``` or it's interface ```ISshKeysClient``` obsolete so that anybody who directly creates those clients would get the warning?  

[...]

PR #1068, opened 2016-01-29
Implement user Administrator Client

paladique:
Per, #1030. All implemented tests are passing.

PR #1108, opened 2016-02-14
Implement Additional Enterprise methods on User Administration Client 

ryangribble:
## Status
:rocket: Ready to merge

~~This is built on top of #1099 which is not yet merged.~~

~~Only the last 4 commits are actually relating to this PR.  I will rebase this as soon as #1099  is merged.~~

~~Here's a diff link showing the consolidated changes for what will be in just this PR:~~
~~https://github.com/TattsGroup/octokit.net/compare/8be8c41966b57de0d994120e733ff02c9d111438...enterprise-useradministration~~

## Implement Additional Enterprise methods on User Administration Client 
https://developer.github.com/enterprise/2.5/v3/users/administration/
Includes normal and reactive implementations plus unit and integration tests for both, including integration tests for the 4 previously implemented methods in #1068 

### Notes
Differences in the API docs between github.com and github enterprise are few and far between however one of the major ones I found was that on GitHub Enterprise the UserAdministration API has several additional methods over github.com - methods for Create/Rename/Delete user, Create/Delete Impersonation Tokens and List/Delete Public Keys
https://developer.github.com/v3/users/administration/
vs
https://developer.github.com/enterprise/2.5/v3/users/administration/

#### SuspendedAt property on User
I found that the SuspendedAt property is returned by the API (at least on GitHub Enterprise) and whether this value is null or not indicates whether a user is suspended.  This seemed useful to know (and also was required in order to assert the Integration tests for the Suspend/Unsuspend methods) so Ive added this field to the ```User``` response object (I also checked whether it should be put on the base ```Account``` object, but ```Organization``` can't be suspended so ```User``` seems to be the correct place for it

#### ImpersonationToken method
Im not exactly sure how to actually use these!  As a GHE admin through the site admin section on the webUI we can "impersonate" a user and login/interact as them, however through the API you have to specify scopes on this call, so Im not sure if this is the same thing or not.  The calls as implemented do work, but Im not sure what to do with the ImperstonationToken once it's retrieved~
I also wasn't sure whether there should be a new Response type for the call or if the existing ```Authorization``` type is appropriate.  The only field the response contains that ```Authorization``` doesn't have, is "token" however I believe that may be being deprecated anyway.  For now I've just used ```Authorization``` as the return type.

As an aside, it looks to me like there is an issue with the ```Authorization``` object in that the field named [```Application```](https://github.com/octokit/octokit.net/blob/master/Octokit/Models/Response/Authorization.cs#L46) is actually ```app``` in the [API response](https://developer.github.com/v3/oauth_authorizations/#get-a-single-authorization) and thus always null.  Ill raise an issue for this tomorrow...

#### Naming Things (tm)
Given that this API now had a mix of methods acting on both Users, ImpersonationTokens and PublicKeys I originally wanted to name the methods ```CreateUser``` ```RenameUser``` ```DeleteUser```.  However given that the ```Suspend``` ```Unsuspend``` ```Promote``` ```Demote``` methods were already implemented with that naming (ie, no "User" suffix) and went live in the latest release, I decided for consistency to not include "User" in the ```Create``` ```Rename``` and ```Delete``` either.

#### Added ```CreateUserContext``` helper
Similar to the ```RepositoryContext``` and ```TeamContext``` helpers, I added a helper that allows the integration tests to create a user to run the tests, then clean it up afterwards.

## Tests
The integration tests require a GitHub enterprise instance so wont run as part of Octokit builds currently.  The ```CreateUserContext``` helper is working a treat, after running stacks of tests my test GHE server doesn't have any extraneious users hanging around :grinning: 

I set a couple of tests to Skip at the moment due to #1106 which I plan to fix up in the near future and then unskip the 2 integration tests.

### DIFF
--- a/Octokit/Clients/UserAdministrationClient.cs
+++ b/Octokit/Clients/UserAdministrationClient.cs
@@ -1,13 +1,11 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
+﻿using System.Collections.Generic;
+using System.Net;
 using System.Threading.Tasks;
 
 namespace Octokit
 {
     /// <summary>
-    /// A client for GitHub's User Administration API.
+    /// A client for GitHub's User Administration API (GitHub Enterprise)
     /// </summary>
     /// <remarks>
     /// See the <a href="https://developer.github.com/v3/users/administration/">Administration API documentation</a> for more details.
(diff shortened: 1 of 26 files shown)


Haacked:
> This is built on top of #1108 which is not yet merged

Umm, this _is_ 1108. :stuck_out_tongue:

ryangribble:
Lol, too many PR's! Meant to say #1099

ryangribble:
Now #1099 is merged, this one has been rebased and squahsed and is ready for your attention @shiftkey @Haacked 

Thanks

Octokit/Clients/UserAdministrationClient.cs
```diff
@@ -1,13 +1,11 @@
...
-using System.Text;
+﻿using System.Collections.Generic;
+using System.Net;
 using System.Threading.Tasks;
 
 namespace Octokit
 {
     /// <summary>
-    /// A client for GitHub's User Administration API.
+    /// A client for GitHub's User Administration API (GitHub Enterprise)
```
Haacked: Because this client is only used for Enterprise, we should consider adding a check to the constructor that ensures that the `ApiConnection.Connection.BaseUrl` is an Enterprise URL and if not, throw an exception that indicates that this client only works for GH:E. That would require that we create this instance lazily of course.

Perhaps an idea for a separate PR.

PR #1112, opened 2016-02-17
Complete UserKeysClient implementation and deprecate SshKeysClient

### LINKED ISSUE #1106: UserKeysClient only implements 2 of the 5 calls for this API

ryangribble:
### Status
:rocket: Ready to merge

~~This is built on top of #1099 and #1108 which are not yet merged.  I will rebase this as soon as the others are merged~~

~~In the meantime for review purposes, here's a diff link showing the consolidated changes for what will be in just this PR:~~
~~https://github.com/TattsGroup/octokit.net/compare/ea7fc21074c5cd4ac0f5d9bc3af891c9d37a64fc...userkeys-client~~

### Complete ```UserKeysClient``` implementation
Resolves #1106 
Implemented the missing ```PublicKey``` methods on ```UserKeysClient``` which are actually implemented as ```SshKey``` methods in ```SshKeysClient```
https://developer.github.com/enterprise/2.5/v3/users/keys

### Start the process of deprecating ```SshKeysClient```
```SshkKeysClient``` and associated ```SshKey``` class are older implementations that no longer align with the GitHub API structure or terminology.

As such it was agreed in #1106 that these shall be deprecated.  This PR marks all of the ```SshKey``` related items obsolete, and directs users to the equivalent ```PublicKey``` and ```UserKeysClient``` implementations where appropriate.

Then, in a future release, these ```SshKey``` related classes can be deleted.

#### ```Update``` method is _really_ deprecated
As per the [API documentation](https://developer.github.com/enterprise/2.5/v3/users/keys/#update-a-public-key), the ```Update()``` method for ```PublicKey``` is no longer supported, so this method was not moved to ```UserKeysClient``` and instead the ```[Obsolete]``` property instructs the user to delete/create the key (as per the API doc).

#### ~~```ModelExtensions``` isn't being ported over to ```PublicKey```~~
~~The ```ModelExtensions``` class contained extension method for ```SshKey``` to parse out the data+name of the key into ```SshKeyInfo``` class.  After reviewing the code, these methods are not used ANYWHERE except in the actual unit tests for the extension methods themselves.  The methods and class have been marked obsolete, but I currently decided not to create equivalent extension methods and ```PublicKeyInfo``` class, for ```PublicKey``` implementation, given that there was no actual usage of this stuff anywhere than  in itself!  Can easily port this over though, if there is concern...~~
After discussion in #1106 these extension methods and tests have been implemented (as ```PublicKeyExtensions``` and ```PublicKeyInfo```)

#### Added ```CreatePublicKeyContext``` helper
Similar to the other context helpers Ive been adding recently, I added a helper that allows the integration tests to create a ```PublicKey``` to run the tests, then clean it up afterwards.  Currently it is just a hardcoded key data, which works fine as long as the cleanup code deletes it as the context is disposed.  If/when an appropriate key generator approach is found as per #1003 which has similar issues around ```DeployKey```s, this context helper could be enhanced to produce unique/random keys each time rather than the hardcoded key data, but for now it's working fine

## Tests
Previously no integration tests existed for ```SshKey``` functions but thanks to ```CreatePublicKeyContext``` helper we now have full integration tests for the ```UserKeysClient``` and ```ObservableUserKeysClient``` methods.  Standard unit tests are also added

### CODE REVIEW

**Comment by ryangribble** [comment]
I pushed up the final change, to implement the extension methods as mentioned in #1106 

Mac and Linux builds failed but I can't see anything obvious...  @shiftkey can you spot anything or do they just need a kick?

**Comment by Haacked** [comment]
> I pushed up the final change

The description still says "Do not merge!" :laughing:

**Comment by ryangribble** [comment]
> The description still says "Do not merge!" :laughing:
Ah yeah, I meant I'd finished coding but still needed to rebase once the other PR's were merged

Which is now done!  Finger's crossed the builds behave this time :grinning:

**Comment by ryangribble** [comment]
Damn, we were on a roll there for several days where the linux/osx builds didnt have any issues, but it seems we are back to the ```System.TypeInitializationException: The type initializer for 'System.Collections.Generic.List`1' threw an exception. ---> System.Threading.ThreadAbortException: ``` issue :angry:

I don't think it's related to any changes, just another case of #1076

**Comment by ryangribble** [comment]
Cool thanks for giving the builds a kick!

@shiftkey are you happy with the obsolete approach?
@Haacked i kept the extension methods pretty much as they were so I assume that works for you guys in the desktop client.  Would be interested to know if you use all 3 extension methods or whether any of them could be dropped...

File: Octokit/Helpers/ModelExtensions.cs [hunk 1]
```diff
@@ -43,6 +46,7 @@ public static bool HasSameDataAs(this SshKey key, SshKey otherKey)
             return keyData != null && keyData == otherKey.GetKeyData();
         }
 
+        [Obsolete("This method is obsolete. Please use PublicKey.GetKeyData() instead.")]
```
Haacked: This is a private method. No need  to obsolete it.

Haacked:
@ryangribble after looking into it, we only use `HasSameDataAs`. Let's drop the extension methods and I'll just move them into our own code.

ryangribble:
> Let's drop the extension methods and I'll just move them into our own code.
Done

ryangribble:
Any other feedback before I hit the button?

Haacked:
:shipit:

ryangribble:
Woohoo, my first self merged PR.  That deserves a celebration :beers: 

LGTM

Issue #1102, opened 2016-02-10
Get commit SHA from branch is wrong


michael-kokorin:
If I use the following code:
```
var commitSha = client.Repository.GetBranch(
				GetSetting(GitHubItSettingKeys.RepositoryOwner),
				GetSetting(GitHubItSettingKeys.RepositoryName),
				branchId)
				.Result.Commit.Sha;

			var sources = client.Repository.Content.GetArchive(
				GetSetting(GitHubItSettingKeys.RepositoryOwner),
				GetSetting(GitHubItSettingKeys.RepositoryName),
				ArchiveFormat.Zipball,
				commitSha)
				.Result;
```

and branchId is different from master then I always get sources from master branch.
This behavior started from version 0.16 and till now. Version 0.15 gets correct result

ryangribble:
Looks like that function inadvertently passes ```String.Empty``` rather than the ```reference``` parameter into the API call 

https://github.com/octokit/octokit.net/blob/master/Octokit/Clients/RepositoryContentsClient.cs#L241

In the meantime you should be able to use the overload that takes an additional ```TimeSpan``` argument (pass in 60 minutes if you want the same behaviour as the overload you are wanting to use), to work around the issue.

shiftkey:
@ryangribble nice spot - yep, that's a bug

shiftkey:
Added the `easy-fix` label as I think this is good for someone new to contribute. I'd also like to see a unit test for this to verify that we're passing through the right value.

shiftkey:
Fixed by #1113

PR #1113, opened 2016-02-19
Fix passing wrong reference in GetArchive method of RepositoryContentsClient

michael-kokorin:
Fix is finished and tested. Also unit test is added.


michael-kokorin:
Fixes bug #1102

shiftkey:
@michael-kokorin flawless victory!

Issue #1076, opened 2016-01-31
tracking TravisCI build issues


shiftkey:
Currently I'm just turning the TravisCI builds on and off again through the portal when something goes awry. That's not great.

Opening this to track these failures better so I don't need to do this as often.

shiftkey:
Timeout? https://gist.github.com/shiftkey/777ce24a06cb1f4c60ca

shiftkey:
"Got a SIGSEGV while executing native code." https://gist.github.com/shiftkey/ab637423fd610e1dc0ce

ryangribble:
Here's a new one we hadn't seen before.  Perhaps just temporary issue communicating with nuget
**[Linux]** The command "nuget restore Octokit-Mono.sln" failed and exited with 139 during .
https://travis-ci.org/octokit/octokit.net/jobs/115926780

ryangribble:
**[Linux]** Timeout

Same as [this one previously](https://github.com/octokit/octokit.net/issues/1076#issuecomment-178228432)

https://travis-ci.org/octokit/octokit.net/jobs/116884393

M-Zuber:
[Linux] "Got a SIGSEGV while executing native code." https://gist.github.com/M-Zuber/7efea02cbf47fb310131982bbfcbd63b
https://travis-ci.org/octokit/octokit.net/builds/129625518

dampir:
[MAC OS] 

> Assertion at metadata.c:3643, condition `ptr' not met

https://gist.github.com/dampir/b96ee54c5153d8b6f5aed65340ba6e9e

https://travis-ci.org/octokit/octokit.net/jobs/136667692

shiftkey:
Nothing really changed in here since 2016 so I'm going to close this out.

PR #1118, opened 2016-03-06
Fixed Issue 1104

### LINKED ISSUE #1104: configure-integration-tests tries to remove a value right after I set it
This flow around optional values seems counter-intuitive:

>
**[value] found as the configured application Secret
Want to change this? Press Y, otherwise we'll move on:** y
**Change the application Secret to use for the integration tests:** [new-value]
**Want to remove this optional value, press Y:**
```

If the user has pressed `Y` here, `Want to remove...` shouldn't be an option...

### PR DESCRIPTION
**Pull request by Anubhav10**
Made changes in .\script\configure-integration-tests.ps1.
For an optional value it gives the option to change or remove the value, for a necessary value it doesn't give an option.

Fixes #1104

### DIFF
--- a/script/configure-integration-tests.ps1
+++ b/script/configure-integration-tests.ps1
@@ -1,4 +1,4 @@
-﻿# TODO: this should indicate whether a variable is required or optional
+# TODO: this should indicate whether a variable is required or optional
 
 function SetVariable([string]$key, [string]$value)
 {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by shiftkey** [comment]
While we're in here, could you look at making the whitespace consistent? It's probably my fault that there's some inconsistencies in here...

File: script/configure-integration-tests.ps1 [hunk 1]
```diff
@@ -43,21 +43,34 @@ function VerifyEnvironmentVariable([string]$friendlyName, [string]$key, [bool]$o
...
+	        	elseif ($reset -eq "R")
+		        { 
+        			SetVariable $key $null
+	        	}
+             }
+            
+        }
+	   else
+	   {
+	    $reset = Read-Host -Prompt "Want to change this value , press Y, otherwise we'll move on"
```
shiftkey: Let's tweak this message: `Press Y to change this value, otherwise we'll move on"

shiftkey:
@Anubhav10 this looks good - just that one improvement to the message and some cleanup of the whitespace and I think this is good to go!

script/configure-integration-tests.ps1
```diff
@@ -1,4 +1,4 @@
-﻿# TODO: this should indicate whether a variable is required or optional
+# TODO: this should indicate whether a variable is required or optional
```
shiftkey: This is an old comment from before I actually implemented the optional parameters. Feel free to remove this :fire:

Anubhav10:
I'll do the suggested changes. Yesterday, the Travis CI build was failing, how come it passed today? I tried to understand why it was failing but I couldn't.

ryangribble:
Unfortunately we are sometimes seeing unrelated failures on the linux and/or OSX travis CI builds.
#1076

Anubhav10:
@shiftkey The whitespace inconsistency was in my code and was my fault. I have removed the TODO comment and changed the message.

shiftkey:
@Anubhav10 excellent, thanks for this!

Issue #1130, opened 2016-03-07
update issue sample to not use octokit repository


shiftkey:
https://github.com/octokit/octokit.net/blob/master/docs/issues.md#create

The placeholders in the code snippet should be something like `"owner"` and `"name"` to indicate that they should be changed (or fail to complete successfully).

cc @rahutchinson who brought this to my attention

shiftkey:
Fixed by #1131

rahutchinson:
:+1:

PR #1131, opened 2016-03-07
#1130 Change issue sample placeholder to indicate that they should be changed

AlexP11223:
I changed placeholders in the issue sample, issue #1130.


shiftkey:
@AlexP11223 thanks!

PR #1140, opened 2016-03-10
Added enum ItemStateFilter to differentiate between search and list APIs

### LINKED ISSUE #1082: SearchIssuesRequest with State = ItemState.All
https://github.com/octokit/octokit.net/blob/master/Octokit/Clients/SearchClient.cs#L52

Searching issues via the search api does not look like it is working when setting the state parameter to `ItemState.All`, this only seems to work for `ItemState.Open` or `ItemState.Closed`.

e.g. `https://github.com/search?q=state%3Aall+repo%3Arails%2Frails&type=Issues`

Does not yield any results. I have tried out the code below too:

````
            var repo = new RepositoryCollection();
            repo.Add("rails", "rails");
            var issueSearch = new SearchIssuesRequest
            {
                 Repos = repo,
                 State = ItemState.All,
                 Created = DateRange.GreaterThan(new DateTime(DateTime.Now.Year, 01, 01).AddYears(-1)),
            };
            var railsIssues = await _githubClient.Search.SearchIssues(issueSearch);
````

However if I set the `State` to either `ItemState.Open` ItemState.Closed` it will work, or if I omit it, it will also work. I think when it is set to ItemState.All we need to *not* add it to the request parameter.

The api seems to correlate to what I am saying to? https://help.github.com/articles/searching-issues/#search-based-on-whether-an-issue-or-pull-request-is-open

I hope that makes sense. Let me know if I need to explain further.

### PR DESCRIPTION
**Pull request by prayankmathur**
Fixes #1082 
I have added an enum "ItemStateFilter"
and tried to explain it in the xml docs.
You can review it.

### CODE REVIEW

**Comment by ryangribble** [comment]
Fixes #1082

**Comment by ryangribble** [comment]
so after having a look at the changes on the PR I have to say that I am backflipping a little on what i said in the issue #1082 !!

It looks like almost all usages of ```ItemState``` only allow values of ```Open``` or ```Closed```.  Eg issues, milestones and pull request objects, as well as the ```NewXXX``` and ```UpdateXXX``` Request classes for these.  It doesn't really make sense to me, to use an enum called ```ItemStateFilter``` (that s described as being used for search filtering) on all of these objects and new/update message classes.  It looks like pretty much the ONLY place that seems to allow ```All``` as an option to ```State``` is the List Issues/Milestones/Pull Requests type calls.

So Im wondering if perhaps we should:

- Keep ```ItemState``` enum but flag the ```All``` value as ```[Deprecated]``` (to be removed in future)
- Keep using this ```ItemState``` enum on all the existing places such as ```Issue``` ```Milestone``` and ```PullRequest``` objects and the ```NewXXX``` ```UpdateXXX``` request classes for them
- Introduce the new ```ItemStateFilter``` enum, allowing ```open,closed,all``` and use it only on the ``XXXRequest``` classes that are used for the "list" methods

That means less "overall" change, and the more sensible approach that ```ItemState``` is used for actual items and creating/updating them... whilst the ```ItemStateFilter``` is used when listing them

**Comment by prayankmathur** [comment]
@ryangribble 
So basically, just reverse the names of the enum !! :P

**Comment by ryangribble** [comment]
Yep pretty much, and ```[Obsolete]``` the ```ItemState.All``` item

**Comment by prayankmathur** [comment]
@ryangribble 
But then we come back to the same issue, that the user who is using octokit doesn't get all the issues on using ItemState.All

**Comment by ryangribble** [comment]
> @ryangribble 
But then we come back to the same issue, that the user who is using octokit doesn't get all the issues on using ItemState.All

I'm not sure I follow?   The search API would no longer take ItemState as an option (it would be ItemStateFilter)  so they wouldn't be able to specify it anymore?

**Comment by prayankmathur** [comment]
@ryangribble 
Sorry, My bad !!
-------
I have made the changes. Take a look.

**Comment by ryangribble** [comment]
hi @prayankmathur, as with the other PR it would be great if you could give this one a nice title and description when you can!  :grinning: 

In terms of your latest changes, Im still seeing several uses of `ItemStateFilter` where I dont think they are valid.

To recap:
`ItemStateFilter` allows values Open, Closed, All and is used on operations that can accept all 3 values - namely the "List" calls under Issues, Milestones, PullRequest API clients
(eg `client.Issues.GetAll(new IssueRequest(State: ItemStateFilter.All))`

`ItemState` allows only Open and Closed (with "All" being marked `[Deprecated]`, to be removed next release) and is used on operations where only Open or Closed are valid states.
The Search API client is the example that triggered this issue to be raised, 
eg `client.Search.SearchIssues(new SearchIssuesRequest(State: ItemState.Open))` 
however there are also several other use cases, such as updating an Issue/Milestone/PullRequest and creating a Milestone where the state can only be open or closed.  

Your latest changes show ```IssueUpdate``` ```MilestoneUpdate``` ```NewMilestone``` and ```PullRequestUpdate``` classes as all using ```ItemStateFilter``` but shouldn't they be using ```ItemState``` since only ```open``` and ```closed``` are valid values for all of these?

**Comment by prayankmathur** [comment]
@ryangribble 
Thanks for clearing that out. I got a little confused.
I have made the necessary changes.
Take a look.

And i hope you like the title. :grinning:

File: Octokit/Models/Response/Issue.cs [hunk 1]
```diff
@@ -64,11 +65,16 @@ public Issue(Uri url, Uri htmlUrl, Uri commentsUrl, Uri eventsUrl, int number, I
...
         /// <summary>
-        /// Whether the issue is open or closed.
+        /// Whether the issue is open , closed or all.
         /// </summary>
         public ItemState State { get; protected set; }
 
         /// <summary>
+        /// Whether the issue is open or close
+        /// </summary>
+        public ItemStateFilter FilteredState {get; protected set;}
```
ryangribble: Don't need this anymore?

Octokit/Models/Request/IssueRequest.cs
```diff
@@ -140,6 +140,27 @@ public enum ItemState
         /// </summary>
```
ryangribble: Should probably add a XmlDoc comment on this guy stating "Items that are open OR closed"

Octokit/Models/Request/IssueRequest.cs
```diff
@@ -121,17 +121,17 @@ public enum IssueFilter
...
     /// <summary>
-    /// The range of states that an issue can be in.
+    /// Search API Filter
     /// </summary>
-    public enum ItemState
+    public enum ItemStateFilter
     {
         /// <summary>
-        /// Isuses that are open (default).
+        /// Issues that are open (default).
```
ryangribble: Perhaps say "Items" rather than "Issues"

Also Im not sure if we should say "(default)" here, since the default behaviour would technically be based on the place of usage (eg on the ```State``` field of ```IssueRequest``` ```PullRequestRequest``` etc is where you could mention what the default value is for that operation)

Octokit/Models/Request/PullRequestRequest.cs
```diff
@@ -13,15 +13,15 @@ public class PullRequestRequest : RequestParameters
...
         {
-            State = ItemState.Open;
+            State = ItemStateFilter.Open;
             SortProperty = PullRequestSort.Created;
             SortDirection = SortDirection.Descending;
         }
 
         /// <summary>
-        /// "open" or "closed" to filter by state. Default is "open".
+        /// "open" , "closed" or "All" to filter by state. Default is "open".
```
ryangribble: use lowercase for consistency, also there is an extraneious space space after "open"

eg:
> /// "open", "closed" or "all" to filter by state.  Default is "open".

Octokit/Models/Request/MilestoneRequest.cs
```diff
@@ -13,12 +13,12 @@ public class MilestoneRequest : RequestParameters
...
         public MilestoneRequest()
         {
-            State = ItemState.Open;
+            State = ItemStateFilter.Open;
             SortProperty = MilestoneSort.DueDate;
             SortDirection = SortDirection.Ascending;
         }
 
-        public ItemState State { get; set; }
+        public ItemStateFilter State { get; set; }
```
ryangribble: This is where I would mention the default

Octokit/Models/Response/Issue.cs
```diff
@@ -10,7 +10,7 @@ public class Issue
     {
         public Issue() { }
 
-        public Issue(Uri url, Uri htmlUrl, Uri commentsUrl, Uri eventsUrl, int number, ItemState state, string title, string body, User user, IReadOnlyList<Label> labels, User assignee, Milestone milestone, int comments, PullRequest pullRequest, DateTimeOffset? closedAt, DateTimeOffset createdAt, DateTimeOffset? updatedAt, int id, bool locked)
+        public Issue(Uri url, Uri htmlUrl, Uri commentsUrl, Uri eventsUrl, int number, ItemState state , ItemStateFilter filteredState, string title, string body, User user, IReadOnlyList<Label> labels, User assignee, Milestone milestone, int comments, PullRequest pullRequest, DateTimeOffset? closedAt, DateTimeOffset createdAt, DateTimeOffset? updatedAt, int id, bool locked)
```
ryangribble: Extraneous space?  `ItemState state[remove-this-space], ItemStateFilter`

Octokit/Models/Response/Issue.cs
```diff
@@ -64,11 +65,16 @@ public Issue(Uri url, Uri htmlUrl, Uri commentsUrl, Uri eventsUrl, int number, I
         public int Number { get; protected set; }
 
         /// <summary>
-        /// Whether the issue is open or closed.
+        /// Whether the issue is open , closed or all.
```
ryangribble: Revert comment back to previous value
> /// Whether the issue is open or closed.

ryangribble:
Looking really good!  

Down to very nitpick type comments now... Ive flagged a few whitespace and XmlDoc tidy ups


@shiftkey - im thinking since this is a breaking change anyhow (replacing ```ItemState``` with ```ItemStateFilter``` on the `xxxSearch` Request objects... perhaps there is no need to `Obselete` the `ItemState.All` value and we can simply remove it.  Do you agree?

Octokit/Models/Request/IssueRequest.cs
```diff
@@ -140,6 +140,27 @@ public enum ItemState
...
+
+        /// <summary>
+        /// Issues that are closed
+        /// </summary>
+        Closed,
+
+        /// <summary>
+        /// All the issues. The option is Obsolete
+        /// </summary>
+        [Obsolete("The method is Obsolete. Dont specify the state in case of all queries")]
```
ryangribble: As per my other comment Im thinking we actually can just remove this guy and not need to Obsolete it.  Waiting for @shiftkey to confirm.  If we DO need to obsolete it though, the warning message should be reworded, since this is not a "method".  Something like
> This value is obsolete and will be removed in a future version.  For Request operations please use ItemStateFilter.All.  For Search, Create and Update operations, "all" is not a valid option.

prayankmathur:
@ryangribble 
I made the changes specified by you.
Thanks for helping me out !! :smile:

shiftkey:
@ryangribble thanks for working on this. I'll have a look over it in the next day or so and see if I can answer your questions as well...

ryangribble:
Just a reminder @shiftkey I was waiting for your guidance on whether the `ItemState.All` needs to be obsoleted, since it's already a "breaking change" replacing `ItemState` with `ItemStateFilter` on any place that All was ever a valid option anyway...

> im thinking since this is a breaking change anyhow (replacing ItemState with ItemStateFilter on the xxxSearch Request objects)... perhaps there is no need to Obselete the ItemState.All value and we can simply remove it as part of the breaking change. There shouldn't really be anyone using ItemState.All on Issue/PR/Milestone update/create anyway since it isn't a valid operation (not sure how the API handles it if requested actually!). Do you agree with not needing to Obsolete ItemState.All ?

shiftkey:
@ryangribble haven't forgotten about this, aiming to get to it today

Octokit/Models/Request/IssueRequest.cs
```diff
@@ -121,23 +121,45 @@ public enum IssueFilter
     }
 
     /// <summary>
-    /// The range of states that an issue can be in.
+    /// Search API Filter
```
shiftkey: This comment seems to be incorrect - it's used for issues/milestones/pull requests.

prayankmathur: @shiftkey
Updated the comment.
Is it correct now ?

shiftkey:
@ryangribble looking over this change, I'm :thumbsup: to just removing this as it'll be covered by the breaking change

Octokit/Models/Request/IssueRequest.cs
```diff
@@ -121,23 +121,45 @@ public enum IssueFilter
     }
 
     /// <summary>
-    /// The range of states that an issue can be in.
+    /// Range of states for "Issues", "Milestones" and "Pull Request" API.
```
shiftkey: Don't worry about the quotation marks here

prayankmathur: @shiftkey Removed the quotation marks. :smile:

Octokit/Models/Response/Issue.cs
```diff
@@ -64,7 +64,7 @@ public Issue(Uri url, Uri htmlUrl, Uri commentsUrl, Uri eventsUrl, int number, I
         public int Number { get; protected set; }
 
         /// <summary>
-        /// Whether the issue is open or closed.
+        /// Whether the issue is "open", "closed" or "all".
```
shiftkey: If we're going to `[Obsolete]` `All` here, or just remove it, we should update the docs accordingly...

shiftkey:
Just one more code comment to tidy up and I think this is good.

@ryangribble how does this read for the breaking changes documentation:

> Decoupled request and response models usage of `ItemState` enum due to API rules. When querying for issues, milestones or pull requests these now use `ItemStateFilter`, and response models continue to use `ItemState`.

prayankmathur:
@shiftkey 
Which one ?
The one you mentioned in the diff ?

shiftkey:
@prayankmathur correct, this one https://github.com/octokit/octokit.net/pull/1140#discussion_r57465473

prayankmathur:
@shiftkey 
Will remove the quotation marks right away.

PS:- Could you have a look at the final proposal, i have uploaded the draft to be on the safe side

prayankmathur:
@shiftkey
Made the changes.Take a look.

shiftkey:
@ryangribble over to you ✊

Octokit/Models/Request/IssueRequest.cs
```diff
@@ -121,23 +121,45 @@ public enum IssueFilter
...
-        /// Isuses that are closed.
+        /// Issues that are closed
         /// </summary>
         Closed,
 
         /// <summary>
-        /// All the issues.
+        /// All the issues. The option is Obsolete
         /// </summary>
+        [Obsolete("The value is Obsolete and will be removed in a future release as it is not a valid option for Search, Create and Update operations")]
```
ryangribble: As agreed by @shiftkey we can just DROP the `All` option here, no need to mark it `[Obsolete]` anymore

Octokit/Models/Request/IssueRequest.cs
```diff
@@ -121,23 +121,45 @@ public enum IssueFilter
...
+    }
+
+    /// <summary>
+    /// Items that are open OR closed
     /// </summary>
     public enum ItemState
     {
         /// <summary>
-        /// Isuses that are open (default).
+        /// Issues that are open
```
ryangribble: Let's word this "Items that are open"

Octokit/Models/Request/IssueRequest.cs
```diff
@@ -121,23 +121,45 @@ public enum IssueFilter
...
         /// </summary>
         Open,
 
         /// <summary>
-        /// Isuses that are closed.
+        /// Issues that are closed
```
ryangribble: Let's word this "Items that are closed"

Octokit/Models/Response/Issue.cs
```diff
@@ -64,7 +64,7 @@ public Issue(Uri url, Uri htmlUrl, Uri commentsUrl, Uri eventsUrl, int number, I
         public int Number { get; protected set; }
 
         /// <summary>
-        /// Whether the issue is open or closed.
+        /// Whether the issue is open, closed or all.
```
ryangribble: all is not a valid option here so this comment can be reverted back to what it was

Octokit/Models/Request/PullRequestRequest.cs
```diff
@@ -13,15 +13,15 @@ public class PullRequestRequest : RequestParameters
...
         {
-            State = ItemState.Open;
+            State = ItemStateFilter.Open;
             SortProperty = PullRequestSort.Created;
             SortDirection = SortDirection.Descending;
         }
 
         /// <summary>
-        /// "open" or "closed" to filter by state. Default is "open".
+        /// "open", "closed" or "all" to filter by state. Default is "open".
```
ryangribble: Perhaps word this consistently with the other `Request` objects, eg:

> Which PullRequests to get.  The default is <see cref="ItemStateFilter.Open"/>.

ryangribble:
Hi guys, sorry for the delay I had a little break over the long weekend :)

It's looking great now @prayankmathur... 
Ive marked up a couple of (final!) comment tweaks, and also noted that as per above, no need to `[Obsolete]` the `ItemState.All` option anymore - just :fire: it

prayankmathur:
@ryangribble 
I have made the necessary changes. Take a look.

ryangribble:
And with that, I think we are done with this one!  Thanks for persisting through all the nitty gritty review :laughing: 

LGTM

prayankmathur:
@ryangribble 
The more I am able to improve octokit the better. :grinning:

shiftkey:
release_notes: Add `importers` field to `IMiscellaneousClient.Meta` response

alfhenrik:
@shiftkey it seems this fix didn't make it in to the **Breaking Changes** section of the release notes (I realised as it broke a couple of my scripts after I upgraded to latest), mind having a look and adding it in there?

shiftkey:
@alfhenrik ugh, yep, missed this one.

Thanks for letting me know, I'll edit the GitHub release to include that change...

shiftkey:
For reference, I added this comment to the GitHub release:

> `ItemStateFilter` has been introduced to separate it from the search APIs, which still use `ItemState`. This affects `IssueRequest.State`, `PullRequestRequest.State` and `MilestoneRequest.State`. You'll need to rename any enum usages from `ItemState` to `ItemStateFilter`.

@alfhenrik can I get a :thumbsup: that this matches with what you were seeing?

alfhenrik:
I was using `RepositoryIssueRequest` but it inherits the `State` property from `IssueRequest` so all good.

Issue #1029, opened 2015-12-23
Implement Migration - Migrations client


shiftkey:
https://developer.github.com/v3/orgs/migrations/

I'm not 100% sure what this is all about, but it's in preview mode so it's not that urgent

[Update]
This API was moved out to a new location in a documentation update on.around the 19 February 2016 
https://developer.github.com/v3/migration/migrations/

ryangribble:
> I'm not 100% sure what this is all about, but it's in preview mode so it's not that urgent

FYI :

https://help.github.com/enterprise/2.4/admin/guides/migrations/about-migrations/
> A migration is the process of transferring data from a source location (either a GitHub.com organization or a GitHub Enterprise instance) to a target GitHub Enterprise instance. Migrations can be used to transfer your data when changing platforms or upgrading hardware on your instance.

shiftkey:
@ryangribble :metal: thanks!

ryangribble:
Similar to #1024 you need to be on GitHub Enterprise to do this one :wink:

devkhan:
@ryangribble Can I still help writing models, requests, tests, etc.?

I know I won't be able to to do integration tests, but still. I'll submit them to your fork.

ryangribble:
Sure, feel free to create all the models classes interfaces and so on and let me know your branch etc.

I wasnt planning on doing this one for a while anyway, as im not really across what these migrations are about (migrating code from one github to another involving zips and files etc i think!) and dont really have a use for them at the moment...  Once I've done the #1024 this issue will be the last major Enterprise API missing, so for completeness sake I do want to do it... but definitely not as a priority

ryangribble:
Implemented by #1141

PR #1141, opened 2016-03-11
Migrations API support

devkhan:
Refer #1029 .

Migrations API: https://developer.github.com/v3/migration/migrations/

About migrations: https://help.github.com/enterprise/2.4/admin/guides/migrations/about-migrations/

To-do:

- [x] Add models for Migrations request and response.
- [x] Add `IEnterpriseMigrationsClient`.
- [x] Implement `IEnterpriseMigrationsClient`.
- [x] Implement Reactive client.
- [x] Re-order methods.
- [x] Refactor (Naming methods and file organization).
- [x] Documentation.
- [x] Unit tests (~~in progress, **help needed**~~).
- [x] Integration tests (~~I don't have access to an Enterprise instance~~, some pointers needed).
- [x] Run `.\build.cmd FixProjects` (for Mono and Xamarin projects).
- [x] Final wiring of parts.
- [ ] Find someone who will actually use it. (otherwise its just pointless).

I also need to discuss some of things which pop-up in my head every now and then. :pray: 

/cc @ryangribble @shiftkey


devkhan:
@ryangribble The build is failing for other projects, do I need add all the files there too?

ryangribble:
Yeah, you have to run the "fake" build script target FixProjects to add any files added to octokit solution, to the mono/core/etc projects 

Instructions in the [contributors doc] (https://github.com/octokit/octokit.net/blob/master/CONTRIBUTING.md)

ryangribble:
So I was reading the documentation about this feature and it turns out I was incorrect in saying you need github enterprise for this API 

These API calls cover the process of extracting repositories from github.com or github enterprise,  that can later be imported into github enterprise. The API is in preview mode only and on covers the export side, there doesn't appear to be anything on import side yet. 

Also this preview API isn't yet available on github enterprise as of latest 2.5 version (according to docs at least)  so at the moment it seems only possible to do against github.com (usually new features make it into the next ghe feature release so I'd expect to see them in 2.6, potentially with the import side implemented as well or possibly not) 

Good news is this means you should be able to do all the testing and integration tests etc yourself @devkhan

devkhan:
Thanks @ryangribble, I will try my best. And I'd would still need your and others' help.

devkhan:
@ryangribble @shiftkey Can anybody tell me why `CanBeDeserialized()` test is failing?

Stack trace:

canbedeserialized

It fails here:

constructorcache

ryangribble:
Hey @devkhan 

Great job getting a good head start on this :)  Implementing a whole new "client" is a pretty maojr job, so there are quite a few things we'll need to cover to get you over the line!

A couple of points to start with:

#### API Client Structure
We try to structure the API clients in octokit to be inline with the GitHub API.  If you have a look at the sidebar in the API docs, the ```Migrations``` client lives under a menu item called `Migration` and does not belong under the ```Enterprise``` API.
This means, you actually need to create a "IMigrationClient" that lives in `IGitHubClient`, then a `IMigrationClient` that lives in `MigrationsClient`.  Yes this is confusing since there is Migration->Migrations and also confusing since the docs refer to it as the `Enterprise Migration API`, but we live and die by the API structure in the sidebar, therefore when using Octokit we should match that structure.  eg ```client.Migrations.Migration.Start()```

image

As an aside, it looks like this Migration/Migrations API actually got moved to this top level location late February, previously it was under the `Organizations` API.  I see there is also a new `Source Import` API under `Migration` that I will raise an issue for shortly!

#### Naming things (TM)
If you have a look at some other ```Client``` APIs for inspiration, you'll see there are some naming conventions in Octokit which should be consistent.  For example, since it's the ```Migrations``` API, you wouldnt call the `List Migrations` method ```GetMigrations()``` (since that would be ```client.Migration.Migrations.GetMigrations()``` !!! ).  We normally name the "list" methods ```Get()``` (for a single instance) ```GetAll()``` (for all instances) and ```GetAllForCurrent()``` (for all instances for the current user, when it's something a user can own, like repositories, publickeys etc.  So i'd suggest the names for the methods in this API should be
Start a migration: ```Start()```
List of migrations: ```GetAll()```
Status of migration: ```Get()```
Download migration archive: ```GetArchive()```
Delete migration archive: ```DeleteArchive()```
Unlock repository: ```UnlockRepository()```

#### Order of Methods
In a similar line of thought, I personally try to have the "order" of the method calls in my client, match those shown in the API.  So in your case, you've listed them as 
Status, List, Start, Download, Delete, Unlock
whereas the API docs list them in a slightly different order 
Start, List, Status, Download, Delete, Unlock
Personally I'd prefer to match the order to the API docs as it makes it easier to review and check for completeness etc :grinning:

#### Other Notes
- Dont forget to set the custom accepts header to enable this preview functionality, otherwise it wont actually work when you do it "for real".  Have a look at the ```AcceptHeaders.ProtectedBranchesApiPreview``` as an example on how to do this.  Basically you will add a new entry into that ```AcceptHeaders``` helper class, then on your API methods you will need to pass that accept header to the ```ApiConnection.Post/Get/Delete``` methods
image

- With your serialization problem, the tiniest little things can sometimes cause so much trouble!!!  
If you have a look at all other ```Response``` model objects, you will notice they have a default/parameterless constructor!  That's all your problem boils down to... 1 line!  haha :sob: 

  Internally the ```SimpleJsonSerializer``` builds up that ```ConstructorCache``` and in your ```Migration``` class, it can't find a constructor to use, since you only provided one that takes all the parameters!  Adding a default constructor should get you past this hurdle!
```public Migration() { }```.  Same issue will be with your ```StartMigrationRequest``` request object.  If you want to have a unit test that tests the deserialization, it will need a default ctor

**Comment by devkhan** [comment]
@ryangribble Thanks for the detailed feedback. I was also eagerly waiting for someone to get me into the "mindset". Now that I know much better about the philosophy of Octokit, I'll adhere to it.

> ... Migrations client lives under a menu item called Migration and does not belong under the Enterprise API.
> This means, you actually need to create a "IMigrationClient" that lives in IGitHubClient, then a IMigrationClient that lives in MigrationsClient...

Thank god, it came up before running `.\build.cmd FixProjects`
I know it my fault that I didn't first checked the docs, but since you said an Enterprise instance is needed for this, I just assumed it is a part of Enterprise API. I'll keep this in mind in the future.

Also, a question, does it mean we have to create an IMigrationsClient and an ISourceImportClient (may be in the future) inside an IMigrationClient and that will go under the IGitHubClient?

> Naming things (TM) and Order of methods

I was having a hard time Naming Things (TM), its now cleared now, and I'll follow the conventions from now on. :+1: 

--------------
And, I was about to ask where to put the AcceptHeader, thanks for clearing that out.

> If you have a look at all other Response model objects, you will notice they have a default/parameterless constructor! That's all your problem boils down to... 1 line! haha :sob:

**That was such a nightmare. Thank you so much.** :100:

**Comment by ryangribble** [comment]
> Thank god, it came up before running .\build.cmd FixProjects

Indeed!  We have an [open issue #1092](https://github.com/octokit/octokit.net/issues/1092) suggesting to make "FixProjects" build task remove items from the slave projects, if the items no longer exist on disk... we've all "been there" when we rename something and have to go hacking around in those other projects to set the world right again.  Unfortunately I am too chicken to tackle the F# build script to see what's involved in doing that!

> Also, a question, does it mean we have to create an IMigrationsClient and an ISourceImportClient (may be in the future) inside an IMigrationClient and that will go under the IGitHubClient?

Yep!  I think this is the first time they've given us 2 such similarly named things as parent/child, but the structure is a "Migration" entry, with "Migrations" and "Source Import" under that.  So although it's a bit odd to have `client.Migrations.Migration.GetAll()` etc, that's what they've come up with, so that's what we'll implement :grimacing:

**Comment by devkhan** [comment]
@ryangribble The GitHub API docs list the Migration->Migrations API as the "Enterprise Migrations API", but does not list it under the Enterprise section, so should I add into a new section for Migration or let it stay in the Enterprise section?

**Comment by ryangribble** [comment]
Yeah, I mentioned earlier it is confusing since they doi call it the "Enterprise Migrations API" but with octokit.net, we base the Clients class structure on the API structure (as shown in the sidebar on the doc site) and not the name... so `GitHubClient.Migration.Migrations` is what you should go with (and similarly move all the models, tests etc out of the `Enterprise` folders)

**Comment by ryangribble** [comment]
Also from a consistency point of view, in the Client interfaces and implementations you've got method parameters on separate lines

``` csharp
Task<Migration> Start(
   string org,
   StartMigrationRequest migration);
```

Whereas to be consistent with the existing Clients in the code base, they should be
```csharp 
Task<Migration> Start(string org, StartMigrationRequest migration);
```

devkhan:
Thanks. As the GitHub APiI docs says.

And, I just put them on the next line to improve readability because of long parameters list. But project's conventions are more important, I'll change them. Should I change this in the tests also?

Octokit.Tests/Models/MigrationTests.cs
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    public class StartMigrationTests
+    {
+        const string migrationRequestJson = @"{
+  ""repositories"": [
+    ""octocat/Hello-World""
+  ],
+  ""lock_repositories"": true
+}";
```
ryangribble: indentation

ryangribble:
> Should I change this in the tests also?
I was only referring to the method declarations, and the test methods dont take any parameters so it;s not applicable to those...

Octokit.Tests/Models/MigrationTests.cs
devkhan: :+1:

ryangribble: Sorry, to clarify I'm saying the indentation here doesn't look correct...

devkhan: I know, I just copy-pasted from the docs, I'll correct them.

devkhan:
@ryangribble The GitHub API docs refer the Migrations API as the "Enterprise Migrations API", so while I will keep it out of the enterprise directory, should I prefix it with Enterprise or not as ...
> we live and die by the API structure in the sidebar, therefore when using Octokit we should match that structure. 

So, what should I do?

ryangribble:
> So, what should I do?

The mentions of "Enterprise Migrations" is only at the landing page for the parent "Migration" page.  Everywhere else shows "Migrations" - eg the URL, the sidebar, the actual page title etc.  This seems to be the first example I can find where the "title" of an API as shown on it's parent's landing page is not consistent with it's name in the side bar and on it's own page!

This API was already only recently moved/relocated to this spot, so I assume this is it's new home now, and if anything it may be losing the "Enterprise" nomenclature... Perhaps @shiftkey or @Haacked can get the inside word on whether this is being renamed "to" or "away" from "Enterprise" Migrations API on it's parent page?

This is the only place it's called "Enterprise Migrations"

image



Everywhere else it's just "Migrations"

image

image

image


I think your latest revision is great though - `IMigrationClient` being the top level parent, and `IMigrationsClient` being the lower level migrations API.  The amusing thing is that due to our policy in octokit of naming the API client property in the singular, we still end up with `client.Migration.Migration.GetAll()` :sob: 

But I do think the way you've got it now is the most consistent and inline with the standards set throughout octokit (rather than breaking the singular property naming rule or calling it EnterpriseMigration eventhough its not called that in most places in the doc)...  so Im happy with how it's ended up!

ryangribble:
This PR is looking great now!!!  Great effort on the testing coverage and everything is looking nicely consistent and thorough!  :+1: 

I think there's just the one indentation issue with the json in the model test... and the fact that other PR merges have caused you to be out of date with master which you'll need to rebase or merge (sorry!)

Ill pull this down and run through the tests etc myself in the next couple of days

devkhan:
@ryangribble I still have to write the integration tests, though. I'll try to get them done as soon as possible.

devkhan:
@ryangribble I've added the integration tests but I have a couple of questions:

- The `GetArchive()` test is failing. Maybe it is returning a string and I'm doing something wrong. Can you please look into it? [Here](https://github.com/octokit/octokit.net/pull/1141/files?diff=unified#diff-2182b0ee397ae547553b88f76159c299R64) is the API method, see if needs modification.
- The DELETE based end-points throws "Accept header error", but how do we specify the preview header in the DELETE methods, all the other preview API's also doesn't specify it in their DELETE method.

ryangribble:
@devkhan ill pull your stuff down and have a look through it in the next couple of days

thanks!

Octokit.Tests.Integration/Clients/MigrationsClientTests.cs
```diff
@@ -0,0 +1,147 @@
...
+using Octokit.Tests.Integration;
+using Xunit;
+
+public class MigrationsClientTests
+{
+    public class TheStartNewMethod
+    {
+        readonly IGitHubClient _gitHub;
+
+        public TheStartNewMethod()
```
ryangribble: This should reflect the method name which was renamed, so `TheStartMethod()`

Octokit.Tests.Integration/Clients/MigrationsClientTests.cs
```diff
@@ -0,0 +1,147 @@
...
+    {
+        readonly IGitHubClient _gitHub;
+
+        public TheDeleteArchiveMethod()
+        {
+            _gitHub = Helper.GetAuthenticatedClient();
+        }
+
+        [IntegrationTest]
+        public async Task DeletesArchive()
```
ryangribble: Probably include "Can" in the test name, since all the others have it?  `CanDeleteArchive`

Octokit.Tests.Integration/Clients/MigrationsClientTests.cs
```diff
@@ -0,0 +1,147 @@
...
+    {
+        readonly IGitHubClient _gitHub;
+
+        public TheUnlockRepositoryMethod()
+        {
+            _gitHub = Helper.GetAuthenticatedClient();
+        }
+
+        [IntegrationTest]
+        public async Task UnlocksRepository()
```
ryangribble: Probably include "Can" in the test name, since all the others have it?  `CanUnlockRepository`

Octokit/Helpers/AcceptHeaders.cs
```diff
@@ -15,5 +15,7 @@ public static class AcceptHeaders
         public const string StarCreationTimestamps = "application/vnd.github.v3.star+json";
 
         public const string CommitReferenceSha1Preview = "application/vnd.github.chitauri-preview+sha";
+
+        public const string MigrationsApiPreview = "  application/vnd.github.wyandotte-preview+json";
```
ryangribble: oops!  There's a few spaces at the beginning of this string!

Octokit/Models/Request/Enterprise/StartMigrationRequest.cs
```diff
@@ -0,0 +1,70 @@
...
+        public StartMigrationRequest()
+        { }
+
+        public StartMigrationRequest(
+            IReadOnlyList<string> repositories
+            ) :
+            this(repositories, false, false)
+        { }
+
+        public StartMigrationRequest(
```
ryangribble: Id say this ctor can be ditched.  If we just provide one where you specify repositories (and the 2 bools are default to false) and one where you specify all 3 parameters, I think that is enough...

Octokit/Models/Response/Enterprise/Migration.cs
```diff
@@ -0,0 +1,98 @@
+﻿using System.Collections.Generic;
```
ryangribble: This file is still living in /Enterprise/ folder in the source tree

Octokit/Models/Request/Enterprise/StartMigrationRequest.cs
devkhan: That was added because the `CanBeDeserialized` test was failing. You told me about this fix, remember?

Octokit.Tests/Clients/MigrationsClientTests.cs
```diff
@@ -0,0 +1,187 @@
...
+        {
+            [Fact]
+            public void RequestsCorrectUrl()
+            {
+                var connection = Substitute.For<IApiConnection>();
+                var client = new MigrationsClient(connection);
+
+                client.Get("fake", 69);
+
+                connection.Received().Get<Migration>(
```
ryangribble: Now you're using the Accepts Header, all the mock `Connection.Received()` checks will need to include that parameter as well (and just a reminder that in the unit test, you should hardcode the expected accept header string ie dont just use the `AcceptHeaders` helper class... otherwise if there was ever a typo in it, the unit tests would pass!  So we specify it "again" on the unit test side)

Octokit.Tests/Clients/MigrationsClientTests.cs
```diff
@@ -0,0 +1,187 @@
...
+            {
+                var connection = Substitute.For<IApiConnection>();
+                var client = new MigrationsClient(connection);
+                var migrationRequest = new StartMigrationRequest(new List<string> { "fake/repo" });
+
+                client.Start("fake", migrationRequest);
+
+                connection.Received().Post<Migration>(
+                    Arg.Is<Uri>(u => u.ToString() == "orgs/fake/migrations"),
+                    migrationRequest);
```
ryangribble: Minor nitpick but since this test's purpose is to assert the correct Url is hit, you shouldnt "also" verify the request body was passed (there is another test for that later).  So you can use Arg.IsAny<StartMigrationRequest>()` here

Octokit.Reactive/Clients/IObservableMigrationClient.cs
```diff
@@ -0,0 +1,13 @@
...
+{
+    public interface IObservableMigrationClient
+    {
+        /// <summary>
+        /// A client for GitHub's Migrations API
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="https://developer.github.com/v3/migration/migrations/">Enterprise License API documentation</a> for more information.
+        /// </remarks>
+        IObservableMigrationsClient Migration { get; }
```
ryangribble: This property name should be `Migrations` (this will fix the failing convention test which ensures that all properties/methods on the normal interfaces are also present on the observable interfaces.  Currently it thinks `Migrations` is missing, because it's called `Migration` here but `Migrations` on the normal client interface...

Octokit.Tests.Integration/Clients/MigrationsClientTests.cs
```diff
@@ -0,0 +1,147 @@
...
+        public TheStartNewMethod()
+        {
+            _gitHub = Helper.GetAuthenticatedClient();
+
+        }
+
+        [IntegrationTest]
+        public async Task CanStartNewMigration()
+        {
+            var organization = Environment.GetEnvironmentVariable("OCTOKIT_GITHUBORGANIZATION");
```
ryangribble: You dont need to hit the environment vars directly here... you can use `Helper.Organization` (which behind the scenes has already retrieved the env vars etc)

Octokit/Clients/MigrationsClient.cs
```diff
@@ -0,0 +1,133 @@
...
+        /// <param name="repo">The repo to unlock.</param>
+        /// <returns></returns>
+        public Task UnlockRepository(string org, int id, string repo)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(org, "org");
+            Ensure.ArgumentNotNullOrEmptyString(repo, "repo");
+
+            var endpoint = ApiUrls.EnterpriseMigrationUnlockRepository(org, id, repo);
+
+            return ApiConnection.Delete(endpoint);
```
ryangribble: Need to specify the preview accepts header here.  Currently the `ApiConnection.Delete()` doesn't have an override that allows this.  One is actually being added as part of another PR that is pending currently: #1185 so you could copy those changes to `IApiConnection` `ApiConnection` `IConnection` and `Connection` classes to get your side going (or wait until that PR is merged, then when you merge or rebase master you will pick it up)

ryangribble:
Hey @devkhan , sorry for the delay in getting to this one.  Ive made a number of comments against file diffs for you to go through which should get most stuff working.

Regarding the integration tests, there are a few things here:
- they currently are creating a lot of migration requests - eg almost every test is requesting a migration of all repo's in the configured organization.
- they also assume the organization HAS repos which potentially may not be the case...
- the download archive method - I think you cant download the archive until the migration status is "exported".  I had to add a while loop to that test, sleep for a few seconds and check the status again (using `GetMigration`) until it reported "exported" and then I was able to call the DownloadArchive function.  
- the DownloadArchive API call issues a redirect to the download link for the archive, andoctokit automatically follows redirects.  In my case at least, this was an 80+MB file (as i had a clone of octokit.net repo in my test organisation) and took a few minutes to download

For the tests to be succesful and efficient we cant rely on or pre-suppose any existence of data, beyond the information provided through integration test setup (ie user, password, apikey, organization).  So ideally we want to create whatever data we need for the tests, and wherever possible we want to clean up after ourselves as well (eg who wants to have 100s or 1000s of repos in their test org, because they ran the test a number of times).  Luckily we have some examples in octokit of doing this, and some helper classes and methods to help you out :grinning: 

We want to use the "CreateTheWorld()" type approach, that we are using in some other ingration tests (a good one to look at is `MergingClientTests.cs` .  We'll essentially have a single test class for this `MigrationsClientTests` with a constructor that will setup all the data it needs (eg repositories, content in those repositories and kick off the migration).  Then you will have the normal test methods you've already written for the other methods, and finally implement `IDisposable` and clean up after ourselves where we can (eg we can delete the repos but we cant really stop the pending migrations, however it could be that deleting the repos deletes those migrations anyway, we'll have to see).  It shouldnt be a big deal anyway since the API docs say that any created migration archives are removed after 7 days anyway, so the main thing for me is that our test repos get removed once we're done.

We have helper classes for `RepositoryContext` and an extension method on `IGitHubClient` that can create these temporary repositories, and they implement `IDisposable` which will then delete the repositories on github.com when the tests wind up (well more precisely, when your test class Dispose() method calls Dispose() on the contexts).  You can steal/borrow some code from the `CreateTheWorld()` implementation in MergingClientTests.cs to populate those repos with a couple of commits/content so they arent empty... and the last thing to do in the constructor phase, is actually initiate the migration request (since all the other test methods need a migration to be requested).

Then have your normal test methods, to run through the `Get`, `GetAll`, `DownloadArchive` (remembering to check/wait for appropriate exported status) and `UnlockRepository` test methods, against that list of created/known repositories.  (although the usual approach is a separate test class per method, in situation like this I reckon it's OK to make all the test methods in the one class, so they can use the contexts setup in the constructor.  Another option would be to have the test method classes, inherit the base class, so they can access the contexts... either would work.

Implement `IDisposable` on the test class, so you can override `Dispose()` and explicitly dispose the contexts (which will delete the repositories from the user's test organization on github.com) and will hopefully leave things mostly tidy over many test ezxecutions.

I've knocked up a [quick gist](https://gist.github.com/ryangribble/fc14239ecad3f65d96f5ebd2fecf722e) to demonstrate what I'm talking about.

Octokit/Models/Request/Enterprise/StartMigrationRequest.cs
ryangribble: Soryy I wasnt clear... I didnt mean the parameterless constructor (which is required for Json deserialiser) Im only suggesting to drop the one that takes the `List<string>` and 1 `bool`

devkhan:
@ryangribble Thanks a lot for the thorough review, I'll get back to you once I fix all these issues.

Octokit.Reactive/Clients/IObservableMigrationClient.cs
devkhan: Will do it, but I've also mentioned it in #1207 as this is a special case caused by the API docs.

ryangribble: agree that this is (the only one ive seen so far) a weird naming thing in github API that makes it hard for us to follow our rules.  But the convention test will fail if you dont have the same property name in `IObservableMigrationClient` that you used in `IMigrationClient`

ryangribble:
Also question for @shiftkey ... should the `status` field be implemented as an enum in octokit?

Im thinking yes, since we have things like `ItemState` as an enum, eventhough the api returns that as a text field.  

In the case of `Migratoin` `status`, the API docs do specifically list all the possible values, so using an `enum` would be the way to go?

https://developer.github.com/v3/migration/migrations/#get-the-status-of-a-migration

shiftkey:
> should the status field be implemented as an enum in octokit?

I like this, as judging by the documentation it should have a finite set of values.

devkhan:
@ryangribble I've re-implemented the integration tests using the **world-building** approach. All the tests are passing except `GetArchive()`, that must be because that end-point actually redirects to the `tar.gz` archive, and since Octokit is configured to auto-follow redirect, it never actually gets a `string` in the result. What should I do here? Download the file and give it to the user or restrict the auto-redirect?

Whichever may be the case, I'll need you help implementing it.

ryangribble:
i believe other download URLs like the files on "Releases" API will follow the redirect and download the file so Id assume similar can be done here.  I've not looked into it deeply myself but if you have a look at the Releases API i think there would be some similarities there

devkhan:
Other API such as "Releases" actually return an object which contains a `download_url`, but this end-point directly downloads a file. Now, there are two options:

* Either download the file ourselves by setting the content-type as `application/x-gzip` and setting the return type as `Task<File>`.
* Or do not follow the redirect and return the URL returned in the `302` response.

devkhan:
@ryangribble I noticed that there is a `GetRedirect()` method on `ApiConnection`, which serves our exact purpose but doesn't have an overload which takes the `accepts` parameter. So should I add that overload and use that method?

shiftkey:
@devkhan 

> I noticed that there is a GetRedirect() method on ApiConnection, which serves our exact purpose but doesn't have an overload which takes the accepts parameter. But it is marked obsolete as Octokit now follows redirects automatically, so what should be done?

This was a decision made a while ago when repository redirects were added. We decided to make the client behave the same and never surface `301`/`302` responses to the consumer - simply follow these and return the content at the end.

> So, while Octokit has the ability to download raw files, it can't be utilised as the Get() methods only return the deserialized objects.

What sort of raw object are we talking about here? I ask because for our repository archive downloads, we return a `Task<byte[]>` which the user then converts into the expected archive format.

@shiftkey Can you clarify on what should be done?

devkhan:
> our repository archive downloads, we return a Task<byte[]> which the user then converts into the expected archive format.

@shiftkey Thanks for clearing that out, so it means I should return the archive as a `byte[]` and let the user decide how it wants to save it?

shiftkey:
@devkhan I think that's the best way to do it here, but I haven't delved deeply into this PR (apologies!). With the repository archive endpoint, the user can specify the format to download too (ZIP or TAR) - is that an option here as well?

devkhan:
I don't think so as the API only mention the response as `application/x-gzip`. I'll let the user decide then.

shiftkey:
@devkhan if there's no option then the xml-docs should just explain the details to the user. Keep it simple!

devkhan:
@ryangribble @shiftkey I've changed the `GetArchvie()` to return a byte array and I've also tested it manually by saving it to a file and it contain the repos defined in the migration, so everything works fine now.

This is complete, AFAIK, you can now review and merge. Let me know if I've missed something.

devkhan:
@shiftkey Should I add `ApiOptions` overloads (for pagination support) to this client as well?

ryangribble:
Sounds good, but how about raising a separate issue and doing it on a new branch/PR so we can get this one over the line without any bigger changes

im just reviewing this one now...

devkhan:
Sure. I was trying to implement that and I am confused on many points so it's better to open a new PR.

ryangribble:
hey @devkhan it's looking pretty good and the integration tests run ok for me.  Only have 2 items of feedback:

- The skipped integration test - we dont really want to have tests flagged as Skip as generally it indicates "something wrong" or needs to be addressed.  I think in this case, just remove the whole `[IntegrationTest]` attribute from that guy, and treat it like a helper method (eg like `CreateTheWorld()`)

- When creating the `RepositoryContext` items you should use the helper function to give them a timestamped name.  This way each time the test runs it has unique names for the repos, and has no chance of clashing with repos that might already exist (eg from a previously failed test run or anyting else).  So basically just change those 3 repo creations to use the helper function like this:
``` csharp
_repos.Add(await _gitHub.CreateRepositoryContext(_orgName, new NewRepository(Helper.MakeNameWithTimestamp("migrationtest-repo1"))
```

Octokit.Tests.Integration/Clients/MigrationsClientTests.cs
```diff
@@ -0,0 +1,132 @@
...
+    private async Task CreateTheWorld()
+    {
+        _repos.Add(await _gitHub.CreateRepositoryContext(_orgName, new NewRepository("migrationtest-repo1")
+        {
+            GitignoreTemplate = "VisualStudio", LicenseTemplate = "mit"
+        }));
+        _repos.Add(await _gitHub.CreateRepositoryContext(_orgName, new NewRepository("migrationtest-repo2")
+        {
+            GitignoreTemplate = "VisualStudio",
+            LicenseTemplate = "mit"
```
ryangribble: the first repo creation has the GitIgnoreTemplate, LicenseTemplate on the same line, but the next 2 have these on separate lines.  would be nice to :lipstick: these to be consistent with eachother

shiftkey: :thumbsup: one property per line

devkhan:
@ryangribble Can you please restart the Travis build?

ryangribble:
@devkhan sorry for the stuff around but can you get this up to date with master again please and we can merge it in?

devkhan:
@ryangribble Sure, I'll get to it as soon as I get some time off my exams.

devkhan:
@ryangribble The Linux build is failing, maybe a mono issue. Please, look into it. :pray:

ryangribble:
This looks good now...  @shiftkey did you want to confirm you were happy with the "raw" byte array return or shall I just hit the button?

devkhan:
Yeah, another PR has been merged since my rebase. Fixed.

ryangribble:
I'm going to merge this in as it's been a long hard slog for you @devkhan!

Good work on implementing a brand new API from scratch!

LGTM

devkhan:
@ryangribble @shiftkey 

It was a great experience, I hope to contribute more in future.
Thanks a :100: to both of you. It wouldn't have been possible without you.

PR #1185, opened 2016-03-14
Implemented Lock/Unlock Functionality for Issues

prayankmathur:
@shiftkey 
I have added the unit and integration tests for issue #896 . however the build is not running successfully.

### DIFF
--- a/Octokit.Tests.Integration/Clients/IssuesClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/IssuesClientTests.cs
@@ -62,6 +62,26 @@ public class IssuesClientTests : IDisposable
     }
 
     [IntegrationTest]
+    public async Task CanLockAndUnlockIssue()
+    {
+         var newIssue = new NewIssue("a test issue") { Body = "A new unassigned issue" };
+         var issue = await _issuesClient.Create(_context.RepositoryOwner, _context.RepositoryName, newIssue);
+         var retrieved = await _issuesClient.Get(_context.RepositoryOwner, _context.RepositoryName, issue.Number);
+ 
+         Assert.Equal(false, issue.Locked);
+ 
+         await _issuesClient.LockIssue(_context.RepositoryOwner, _context.RepositoryName, issue.Number);
+         retrieved = await _issuesClient.Get(_context.RepositoryOwner, _context.RepositoryName, issue.Number);
+         Assert.NotNull(retrieved);
+         Assert.Equal(true, issue.Locked);
+ 
+         await _issuesClient.UnlockIssue(_context.RepositoryOwner, _context.RepositoryName, issue.Number);
+         retrieved = await _issuesClient.Get(_context.RepositoryOwner, _context.RepositoryName, issue.Number);
+         Assert.NotNull(retrieved);
+         Assert.Equal(false, issue.Locked);
+     }
+ 
+     [IntegrationTest]
     public async Task CanListOpenIssuesWithDefaultSort()
     {
         var newIssue1 = new NewIssue("A test issue1") { Body = "A new unassigned issue" };
--- a/Octokit.Tests.Integration/Clients/IssuesClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/IssuesClientTests.cs
@@ -65,18 +65,17 @@ public class IssuesClientTests : IDisposable
     {
         var newIssue = new NewIssue("a test issue") { Body = "A new unassigned issue" };
         var issue = await _issuesClient.Create(_context.RepositoryOwner, _context.RepositoryName, newIssue);
-        var retrieved = await _issuesClient.Get(_context.RepositoryOwner, _context.RepositoryName, issue.Number);
         Assert.False(issue.Locked);
  
         await _issuesClient.Lock(_context.RepositoryOwner, _context.RepositoryName, issue.Number);
-        retrieved = await _issuesClient.Get(_context.RepositoryOwner, _context.RepositoryName, issue.Number);
+        var retrieved = await _issuesClient.Get(_context.RepositoryOwner, _context.RepositoryName, issue.Number);
         Assert.NotNull(retrieved);
-        Assert.True(issue.Locked);
+        Assert.True(retrieved.Locked);
  
         await _issuesClient.Unlock(_context.RepositoryOwner, _context.RepositoryName, issue.Number);
         retrieved = await _issuesClient.Get(_context.RepositoryOwner, _context.RepositoryName, issue.Number);
         Assert.NotNull(retrieved);
-        Assert.False(issue.Locked);
+        Assert.False(retrieved.Locked);
      }
  
      [IntegrationTest]
--- a/Octokit.Tests.Integration/Clients/IssuesClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/IssuesClientTests.cs
@@ -64,21 +64,20 @@ public class IssuesClientTests : IDisposable
     [IntegrationTest]
     public async Task CanLockAndUnlockIssue()
     {
-         var newIssue = new NewIssue("a test issue") { Body = "A new unassigned issue" };
-         var issue = await _issuesClient.Create(_context.RepositoryOwner, _context.RepositoryName, newIssue);
-         var retrieved = await _issuesClient.Get(_context.RepositoryOwner, _context.RepositoryName, issue.Number);
+        var newIssue = new NewIssue("a test issue") { Body = "A new unassigned issue" };
+        var issue = await _issuesClient.Create(_context.RepositoryOwner, _context.RepositoryName, newIssue);
+        var retrieved = await _issuesClient.Get(_context.RepositoryOwner, _context.RepositoryName, issue.Number);
+        Assert.False(issue.Locked);
  
-         Assert.Equal(false, issue.Locked);
+        await _issuesClient.Lock(_context.RepositoryOwner, _context.RepositoryName, issue.Number);
+        retrieved = await _issuesClient.Get(_context.RepositoryOwner, _context.RepositoryName, issue.Number);
+        Assert.NotNull(retrieved);
+        Assert.True(issue.Locked);
  
-         await _issuesClient.LockIssue(_context.RepositoryOwner, _context.RepositoryName, issue.Number);
-         retrieved = await _issuesClient.Get(_context.RepositoryOwner, _context.RepositoryName, issue.Number);
-         Assert.NotNull(retrieved);
-         Assert.Equal(true, issue.Locked);
- 
-         await _issuesClient.UnlockIssue(_context.RepositoryOwner, _context.RepositoryName, issue.Number);
-         retrieved = await _issuesClient.Get(_context.RepositoryOwner, _context.RepositoryName, issue.Number);
-         Assert.NotNull(retrieved);
-         Assert.Equal(false, issue.Locked);
+        await _issuesClient.Unlock(_context.RepositoryOwner, _context.RepositoryName, issue.Number);
+        retrieved = await _issuesClient.Get(_context.RepositoryOwner, _context.RepositoryName, issue.Number);
+        Assert.NotNull(retrieved);
+        Assert.False(issue.Locked);
      }
  
      [IntegrationTest]
--- a/Octokit.Tests.Integration/Reactive/ObservableIssuesClientTests.cs
+++ b/Octokit.Tests.Integration/Reactive/ObservableIssuesClientTests.cs
@@ -72,6 +72,23 @@ public class ObservableIssuesClientTests : IDisposable
         Assert.Equal("Modified integration test issue", updateResult.Title);
     }
 
+     [IntegrationTest]
+     public async Task CanLockAndUnlockIssues()
+     {
+         var newIssue = new NewIssue("Integration Test Issue");
+ 
+         var createResult = await _client.Create(_context.RepositoryOwner, _context.RepositoryName, newIssue);
+         Assert.Equal(false, createResult.Locked);
+ 
+         await _client.LockIssue(_context.RepositoryOwner, _context.RepositoryName, createResult.Number);
+         var lockResult = await _client.Get(_context.RepositoryOwner, _context.RepositoryName, createResult.Number);
+         Assert.Equal(true, lockResult.Locked);
+ 
+         await _client.UnlockIssue(_context.RepositoryOwner, _context.RepositoryName, createResult.Number);
+         var unlockIssueResult = await _client.Get(_context.RepositoryOwner, _context.RepositoryName, createResult.Number);
+         Assert.Equal(false, unlockIssueResult.Locked);
+    }
+ 
     public void Dispose()
     {
         _context.Dispose();
--- a/Octokit.Tests.Integration/Reactive/ObservableIssuesClientTests.cs
+++ b/Octokit.Tests.Integration/Reactive/ObservableIssuesClientTests.cs
@@ -78,15 +78,15 @@ public class ObservableIssuesClientTests : IDisposable
          var newIssue = new NewIssue("Integration Test Issue");
  
          var createResult = await _client.Create(_context.RepositoryOwner, _context.RepositoryName, newIssue);
-         Assert.Equal(false, createResult.Locked);
+         Assert.False(createResult.Locked);
  
-         await _client.LockIssue(_context.RepositoryOwner, _context.RepositoryName, createResult.Number);
+         await _client.Lock(_context.RepositoryOwner, _context.RepositoryName, createResult.Number);
          var lockResult = await _client.Get(_context.RepositoryOwner, _context.RepositoryName, createResult.Number);
-         Assert.Equal(true, lockResult.Locked);
+         Assert.True(lockResult.Locked);
  
-         await _client.UnlockIssue(_context.RepositoryOwner, _context.RepositoryName, createResult.Number);
+         await _client.Unlock(_context.RepositoryOwner, _context.RepositoryName, createResult.Number);
          var unlockIssueResult = await _client.Get(_context.RepositoryOwner, _context.RepositoryName, createResult.Number);
-         Assert.Equal(false, unlockIssueResult.Locked);
+         Assert.False(unlockIssueResult.Locked);
     }
  
     public void Dispose()
--- a/Octokit.Tests/Reactive/ObservableIssuesClientTests.cs
+++ b/Octokit.Tests/Reactive/ObservableIssuesClientTests.cs
@@ -329,11 +329,61 @@ public class ObservableIssuesClientTests
             var gitHubClient = Substitute.For<IGitHubClient>();
             var client = new ObservableIssuesClient(gitHubClient);
 
-            Assert.Throws<ArgumentNullException>(() => client.Create(null, "name", new NewIssue("title")));
-            Assert.Throws<ArgumentException>(() => client.Create("", "name", new NewIssue("x")));
-            Assert.Throws<ArgumentNullException>(() => client.Create("owner", null, new NewIssue("x")));
-            Assert.Throws<ArgumentException>(() => client.Create("owner", "", new NewIssue("x")));
-            Assert.Throws<ArgumentNullException>(() => client.Create("owner", "name", null));
+            Assert.Throws<ArgumentNullException>(() => client.Update(null, "name", 42, new IssueUpdate()));
+            Assert.Throws<ArgumentException>(() => client.Update("", "name", 42, new IssueUpdate()));
+            Assert.Throws<ArgumentNullException>(() => client.Update("owner", null, 42, new IssueUpdate()));
+            Assert.Throws<ArgumentException>(() => client.Update("owner", "", 42, new IssueUpdate()));
+            Assert.Throws<ArgumentNullException>(() => client.Update("owner", "name", 42, null));
+        }
+    }
+
+    public class TheLockIssueMethod
+    {
+        [Fact]
+        public void LockIssue()
+        {
+            var gitHubClient = Substitute.For<IGitHubClient>();
+            var client = new ObservableIssuesClient(gitHubClient);
+
+            client.LockIssue("fake", "repo", 42);
+            gitHubClient.Issue.Received().LockIssue("fake", "repo", 42);
+        }
+
+        [Fact]
+        public void EnsuresArgumentsNotNull()
+        {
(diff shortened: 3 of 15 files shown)


prayankmathur:
@shiftkey 

The build passed the checks here i guess.
Take a look at the unit and integration tests.

ryangribble:
Hi @prayankmathur , could I suggest that you reword the pull request title and description to contain a little bit more details about what it's about?  
You certainly want to link to the issue #896 but also would be good to say a few words here about it so everyone doesnt have to click through to the issue to see what this PR is about.

eg something like this:

Title:
> Implement Lock/Unlock functionality for Issues

Description:
> This PR implements the locking and unlocking funcitonality for Issues
Fixes #896
Also includes Unit and Integration Tests


Also from a quick look, this pull request includes your changes for ther other PR about ItemState/ItemStateFilter etc, which makes it hard to review the diffs in isolation.  Ideally you would either do the work on separate branches, or once one PR gets merged, you can merge master into your branch and we will then see the "true" differences in this PR

prayankmathur:
@ryangribble

Changed the title. Sorry for the trouble. Actually i did branch out this issue, but it turned out that i branched it from my master branch which consisted of the previous bug(which i hadn't branched).
If you could take the trouble of going through the other PR first, then it could solve the issue.

BTW, is there any problem with this [function](https://github.com/octokit/octokit.net/blob/master/Octokit.Tests/Reactive/ObservableIssuesClientTests.cs#L327)

prayankmathur:
@ryangribble 
@shiftkey 
I found that this [function](http://github.com/octokit/octokit.net/blob/master/Octokit.Tests/Reactive/ObservableIssuesClientTests.cs#L327) was not correct.
So corrected that too in the same PR.
Hope that it is fine.

shiftkey:
@prayankmathur given it's not related to this feature, a separate PR would be nice for a couple of reasons:

 - we can review it separate to this
 - given it's a simpler change, it's likely to get merged in quicker

I'll leave it up to you whether you'd like to extract that to a new branch.

prayankmathur:
@shiftkey 
I ll open a separate PR.

ryangribble:
Just noting for reference that the status of this PR is
- Waiting for #1140 to be merged and then have this PR's commits tidied up (due to this PR being built ontop of #1140)
- Waiting on @prayankmathur to split out the test fix for `UpdateIssue` to another PR and remove the commit from this one

prayankmathur:
@ryangribble 
I was waiting for #1140 to be merged.
But now i have opened a separate PR for the UpdateMethod function correction.

shiftkey:
@prayankmathur this looks like it contains work from #1140 (open) and #1196 (merged). Are you able to rebase the branch so that this only contains the work related to the locking and unlocking of issues?

Octokit.Tests/Reactive/ObservableIssuesClientTests.cs
```diff
@@ -329,11 +329,61 @@ public void EnsuresArgumentsNotNull()
...
-            Assert.Throws<ArgumentException>(() => client.Create("", "name", new NewIssue("x")));
-            Assert.Throws<ArgumentNullException>(() => client.Create("owner", null, new NewIssue("x")));
-            Assert.Throws<ArgumentException>(() => client.Create("owner", "", new NewIssue("x")));
-            Assert.Throws<ArgumentNullException>(() => client.Create("owner", "name", null));
+            Assert.Throws<ArgumentNullException>(() => client.Update(null, "name", 42, new IssueUpdate()));
+            Assert.Throws<ArgumentException>(() => client.Update("", "name", 42, new IssueUpdate()));
+            Assert.Throws<ArgumentNullException>(() => client.Update("owner", null, 42, new IssueUpdate()));
+            Assert.Throws<ArgumentException>(() => client.Update("owner", "", 42, new IssueUpdate()));
+            Assert.Throws<ArgumentNullException>(() => client.Update("owner", "name", 42, null));
+        }
```
prayankmathur: @shiftkey 
Why is this function showing changes.
I had opened a seperate PR to change it which has been merged.
The function has changed also [here](https://github.com/octokit/octokit.net/blob/master/Octokit.Tests/Reactive/ObservableIssuesClientTests.cs#L327)

prayankmathur:
@shiftkey 
I have made the necessary changes. Please take a look.

Octokit.Tests.Integration/Clients/IssuesClientTests.cs
```diff
@@ -62,6 +62,26 @@ public IssuesClientTests()
     }
 
     [IntegrationTest]
+    public async Task CanLockAndUnlockIssue()
+    {
+         var newIssue = new NewIssue("a test issue") { Body = "A new unassigned issue" };
+         var issue = await _issuesClient.Create(_context.RepositoryOwner, _context.RepositoryName, newIssue);
+         var retrieved = await _issuesClient.Get(_context.RepositoryOwner, _context.RepositoryName, issue.Number);
+ 
+         Assert.Equal(false, issue.Locked);
```
shiftkey: You can use `Assert.False` or `Assert.True` here to make these a bit more succinct.

Octokit.Tests.Integration/Reactive/ObservableIssuesClientTests.cs
```diff
@@ -72,6 +72,23 @@ public ObservableIssuesClientTests()
         Assert.Equal("Modified integration test issue", updateResult.Title);
     }
 
+     [IntegrationTest]
+     public async Task CanLockAndUnlockIssues()
+     {
+         var newIssue = new NewIssue("Integration Test Issue");
+ 
+         var createResult = await _client.Create(_context.RepositoryOwner, _context.RepositoryName, newIssue);
+         Assert.Equal(false, createResult.Locked);
```
shiftkey: `Assert.True` or `Assert.False` in this test too

Octokit/Clients/IIssuesClient.cs
```diff
@@ -163,5 +163,25 @@ public interface IIssuesClient
...
+
+        /// <summary>
+        /// Locks an issue for the specified repository. Issue owners and users with push access can lock an issue.
+        /// </summary>
+        /// <remarks>https://developer.github.com/v3/issues/#lock-an-issue</remarks>
+        /// <param name="owner">The owner of the repository</param>
+        /// <param name="name">The name of the repository</param>
+        /// <param name="number">The issue number</param>
+        /// <returns></returns>
+        Task LockIssue(string owner, string name, int number);
```
shiftkey: Given this usage will look like this:

```
await github.Issue.LockIssue("owner", "name", 1);
```

Perhaps we can drop the `Issue` suffix here, to give:

```
github.Issue.Lock("owner", "name", 1);
```

Thoughts @haacked @ryangribble?

Octokit/Clients/IssuesClient.cs
```diff
@@ -225,5 +225,37 @@ public Task<Issue> Update(string owner, string name, int number, IssueUpdate iss
...
+        /// <param name="owner">The owner of the repository</param>
+        /// <param name="name">The name of the repository</param>
+        /// <param name="number">The issue number</param>
+        /// <returns></returns>
+        public Task LockIssue(string owner, string name, int number)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
+            Ensure.ArgumentNotNullOrEmptyString(name, "name");
+
+            return ApiConnection.Put(ApiUrls.IssueLock(owner, name, number));
```
shiftkey: As per the [blog post](https://developer.github.com/changes/2016-02-11-issue-locking-api/) we need to pass in a custom `Accept` header here (as well as for the `Unlock` action below)

Octokit/Clients/IIssuesClient.cs
ryangribble: Yep I like the sound of `.Lock()` and `.Unlock()` :+1:

Octokit/Clients/IssuesClient.cs
ryangribble: Yep, see [AcceptHeaders.cs`(https://github.com/octokit/octokit.net/blob/master/Octokit/Helpers/AcceptHeaders.cs) helper class and other usages, for how we can do this

prayankmathur:
@shiftkey @ryangribble 
Should i rename the reactive function also ?
Because then it's usage would be like
>client.Lock(owner, name, 1);

Currently its like 
>client.LockIssue(owner, name, 1)

ryangribble:
Yes you should rename it too.  In integration or unit tests, you are directly referencing the `IObserableIssuesClient` but just remember that "regular" users of octokit will be accessing both the standard or observable client, through the "proper" heierarchy
eg `ObservableGitHubClient.Issue.Lock()`

prayankmathur:
@ryangribble 
Yeah, forgot that :grin: 
will make the changes.

ryangribble:
Also dont forget you've got merge conflicts so either merge master into your branch, or rebase your branch on master and force push

prayankmathur:
@ryangribble 
I resolved the issues too and have made the changes. Please take a look.

Octokit/Helpers/AcceptHeaders.cs
```diff
@@ -14,6 +14,9 @@ public static class AcceptHeaders
 
         public const string StarCreationTimestamps = "application/vnd.github.v3.star+json";
 
+        public const string IssueLockingUnlockingApiPreview = "application/vnd.github.the-key-preview+json";
+
         public const string CommitReferenceSha1Preview = "application/vnd.github.chitauri-preview+sha";
+
```
ryangribble: Extra whitespace line introduced here :)

Octokit.Tests/Reactive/ObservableIssuesClientTests.cs
```diff
@@ -337,6 +337,56 @@ public void EnsuresArgumentsNotNull()
         }
     }
 
+    public class TheLockIssueMethod
```
ryangribble: I guess this now should be "TheLockMethod"

Octokit.Tests/Reactive/ObservableIssuesClientTests.cs
```diff
@@ -337,6 +337,56 @@ public void EnsuresArgumentsNotNull()
...
+            var client = new ObservableIssuesClient(gitHubClient);
+
+            Assert.Throws<ArgumentNullException>(() => client.LockIssue(null, "name", 42));
+            Assert.Throws<ArgumentException>(() => client.LockIssue("", "name", 42));
+            Assert.Throws<ArgumentNullException>(() => client.LockIssue("owner", null, 42));
+            Assert.Throws<ArgumentException>(() => client.LockIssue("owner", "", 42));
+        }
+    }
+
+    public class TheUnlockIssueMethod
```
ryangribble: "TheUnlockMethod"

Octokit.Tests/Reactive/ObservableIssuesClientTests.cs
```diff
@@ -337,6 +337,56 @@ public void EnsuresArgumentsNotNull()
...
+    {
+        [Fact]
+        public void UnlockIssue()
```
ryangribble: Wording for the unit tests should read TheBlahMethod....DoesSomething

So...
TheUnlockMethod->Unlock**s**Issue

(note Unlocks rather than Unlock)

Octokit.Tests/Reactive/ObservableIssuesClientTests.cs
```diff
@@ -337,6 +337,56 @@ public void EnsuresArgumentsNotNull()
         }
     }
 
+    public class TheLockIssueMethod
+    {
+        [Fact]
+        public void LockIssue()
```
ryangribble: "Lock**s**Issue"

ryangribble:
Looks like you're ontop of the test failures which is :+1: 

Ive highlighted a couple of REALLY minor consistency issues around naming unit tests (now the methods have been renamed)

prayankmathur:
@ryangribble 
The changes are made successfully. You can now finally take a look. :grinning:

Octokit.Tests.Integration/Clients/IssuesClientTests.cs
```diff
@@ -61,6 +61,25 @@ public IssuesClientTests()
...
+    {
+        var newIssue = new NewIssue("a test issue") { Body = "A new unassigned issue" };
+        var issue = await _issuesClient.Create(_context.RepositoryOwner, _context.RepositoryName, newIssue);
+        var retrieved = await _issuesClient.Get(_context.RepositoryOwner, _context.RepositoryName, issue.Number);
+        Assert.False(issue.Locked);
+ 
+        await _issuesClient.Lock(_context.RepositoryOwner, _context.RepositoryName, issue.Number);
+        retrieved = await _issuesClient.Get(_context.RepositoryOwner, _context.RepositoryName, issue.Number);
+        Assert.NotNull(retrieved);
+        Assert.True(issue.Locked);
```
shiftkey: We should be checking `retrieved.Locked` here, right?

Octokit.Tests.Integration/Clients/IssuesClientTests.cs
```diff
@@ -61,6 +61,25 @@ public IssuesClientTests()
...
+ 
+        await _issuesClient.Unlock(_context.RepositoryOwner, _context.RepositoryName, issue.Number);
+        retrieved = await _issuesClient.Get(_context.RepositoryOwner, _context.RepositoryName, issue.Number);
+        Assert.NotNull(retrieved);
+        Assert.False(issue.Locked);
```
shiftkey: `retrieved.Locked` here too

Octokit.Tests.Integration/Clients/IssuesClientTests.cs
```diff
@@ -61,6 +61,25 @@ public IssuesClientTests()
...
     [IntegrationTest]
+    public async Task CanLockAndUnlockIssue()
+    {
+        var newIssue = new NewIssue("a test issue") { Body = "A new unassigned issue" };
+        var issue = await _issuesClient.Create(_context.RepositoryOwner, _context.RepositoryName, newIssue);
+        var retrieved = await _issuesClient.Get(_context.RepositoryOwner, _context.RepositoryName, issue.Number);
+        Assert.False(issue.Locked);
+ 
+        await _issuesClient.Lock(_context.RepositoryOwner, _context.RepositoryName, issue.Number);
+        retrieved = await _issuesClient.Get(_context.RepositoryOwner, _context.RepositoryName, issue.Number);
```
shiftkey: Actually, you should be returned the updated `Issue` from `Lock` and `Unlock` here - to simplify this test (and do less network calls) you could assign this returned value to `retrieved`...

Octokit/Clients/IssuesClient.cs
```diff
@@ -225,5 +225,37 @@ public Task<Issue> Update(string owner, string name, int number, IssueUpdate iss
...
+        /// <param name="owner">The owner of the repository</param>
+        /// <param name="name">The name of the repository</param>
+        /// <param name="number">The issue number</param>
+        /// <returns></returns>
+        public Task<Issue> Lock(string owner, string name, int number)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
+            Ensure.ArgumentNotNullOrEmptyString(name, "name");
+
+            return ApiConnection.Put<Issue>(ApiUrls.IssueLock(owner, name, number), AcceptHeaders.IssueLockingUnlockingApiPreview);
```
shiftkey: This is not quite the right overload, and sadly `ApiConnection` doesn't quite have an easy way to do this.

This works for me:

```
return ApiConnection.Put<Issue>(ApiUrls.IssueLock(owner, name, number), new object(), null, AcceptHeaders.IssueLockingUnlockingApiPreview);
```

Octokit/Clients/IIssuesClient.cs
```diff
@@ -163,5 +163,25 @@ public interface IIssuesClient
...
+
+        /// <summary>
+        /// Locks an issue for the specified repository. Issue owners and users with push access can lock an issue.
+        /// </summary>
+        /// <remarks>https://developer.github.com/v3/issues/#lock-an-issue</remarks>
+        /// <param name="owner">The owner of the repository</param>
+        /// <param name="name">The name of the repository</param>
+        /// <param name="number">The issue number</param>
+        /// <returns></returns>
+        Task<Issue> Lock(string owner, string name, int number);
```
shiftkey: This endpoint doesn't return any data, so it needs to return a `Task`.

Octokit/Clients/IssuesClient.cs
```diff
@@ -225,5 +225,37 @@ public Task<Issue> Update(string owner, string name, int number, IssueUpdate iss
...
+        /// <param name="owner">The owner of the repository</param>
+        /// <param name="name">The name of the repository</param>
+        /// <param name="number">The issue number</param>
+        /// <returns></returns>
+        public Task Unlock(string owner, string name, int number)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
+            Ensure.ArgumentNotNullOrEmptyString(name, "name");
+
+            return ApiConnection.Delete(ApiUrls.IssueLock(owner, name, number), AcceptHeaders.IssueLockingUnlockingApiPreview);
```
ryangribble: > he name is the least-worst option that came to mind given it'll clash with the other

The inconsistency of overloads for Post Put Delete and Get makes me :cry: too!

Rather than introduce the first/only occurence of this "{verb}With" terminology to get around the overload collisions, perhaps in this case we could just go with adding an overload for
`Delete(Uri uri, object data, string accepts)`
and call it as `ApiConnection.Delete(ApiUrls.IssueLock(owner, name, number), null, AcceptHeaders.IssueLockingUnlockingApiPreview)`
To me this seems slightly less ugly than adding in a new terminology

Or alternatively, perhaps we could introduce a whole new set of "With" overloads, and start to obsolete the old ones

eg

```
GetWithCustomAccept()
PutWithCustomAccept()
DeleteWithCustomAccept()
PostWithCustomAccept()
```

Octokit/Clients/IssuesClient.cs
ryangribble: Does the body have to be `new object()` or can it just be `null` ?

Octokit/Clients/IIssuesClient.cs
ryangribble: Given that the API does return a HTTP 204 NoContent response when "succesful" the way ive seen this implemented in octokit in other places (eg [OrganizationMembers.Publicize()](https://github.com/octokit/octokit.net/blob/829677c9292ca503806aac385f49e37ad3730fdd/Octokit/Clients/OrganizationMembersClient.cs#L289)) is that the Octokit call returns a Task<bool> then internally it calls the `Connection.Put` and checks the HTTP response.  If a 204 was received then "true" is returned, otherwise false (or an exception of course, for 404/500/other stuff handled by the plumbing inside Connection class)

ryangribble:
Regarding the actual implementation of Lock and Unlock methods (which are API PUT methods that return a 204 response on success) I suggest we should implement them similarly to  [OrganizationMembers.Publicize()](https://github.com/octokit/octokit.net/blob/829677c9292ca503806aac385f49e37ad3730fdd/Octokit/Clients/OrganizationMembersClient.cs#L289)

In other words, check for that 204 response and return a `Task<bool>` to indicate success

prayankmathur:
@ryangribble 
@shiftkey 
Right now, I am implementing the >Delete(Uri uri, object data, string accepts)
And then later on we can open up a new PR.
Is that Fine ?

Octokit/Clients/IIssuesClient.cs
shiftkey: :thumbsup:

Octokit/Clients/IssuesClient.cs
shiftkey: > we could just go with adding an overload for `Delete(Uri uri, object data, string accepts)` and then pass null as the body

I don't think we can pass `null` here, but everything else is :thumbsup: from me

prayankmathur:
@shiftkey 
@ryangribble 
Just a gentle reminder to go see the changes that i made in this PR.

ryangribble:
> Also should I implement the Lock/Unlock methods as suggested by @ryangribble which would return 
Task<bool>

Yes please :grinning: @shiftkey also gave this a :+1: so we are in agreement that when the API docs mention a 204 response as success... you can check for that and return `Task<bool>` accordingly.  Link to existing example: [OrganizationMembers.Publicize()](https://github.com/octokit/octokit.net/blob/829677c9292ca503806aac385f49e37ad3730fdd/Octokit/Clients/OrganizationMembersClient.cs#L289)

Octokit/Http/Connection.cs
```diff
@@ -485,6 +485,22 @@ Task<IApiResponse<T>> SendDataInternal<T>(object body, string accepts, string co
         }
 
         /// <summary>
+        /// Performs an asynchronous HTTP DELETE request that expects an empty response.
+        /// </summary>
+        /// <param name="uri">URI endpoint to send request to</param>
+        /// <param name="data">The object to serialize as the body of the request</param>
+        /// <param name="accepts">Specifies accept response media type</param>
+        /// <returns>The returned <seealso cref="HttpStatusCode"/></returns>
+        public async Task<HttpStatusCode> Delete(Uri uri,object data, string accepts)
```
ryangribble: missing space in between uri and object

SolutionInfo.cs
```diff
@@ -6,10 +6,8 @@
 [assembly: AssemblyVersionAttribute("0.19.0")]
 [assembly: AssemblyFileVersionAttribute("0.19.0")]
 [assembly: ComVisibleAttribute(false)]
-namespace System
-{
-    internal static class AssemblyVersionInformation
-    {
+namespace System {
+    internal static class AssemblyVersionInformation {
```
ryangribble: Is this change intended?  Probably best to revert this one

Octokit/Http/Connection.cs
shiftkey: This will be handled by the next `FormatCode` call, so let's :boot: addressing this...

SolutionInfo.cs
shiftkey: This is the eternal struggle between the FAKE script and `FormatCode` - see https://github.com/shiftkey/Octokit.CodeFormatter/issues/2 for how I'd like to tackle this...

shiftkey:
I'm happy with the integration test now passing, and the minor stuff remaining can be sorted out later.

Thanks @prayankmathur!

prayankmathur:
@shiftkey
:+1:

PR #1190, opened 2016-03-15
Add ApiOption overloads to methods on IUserEmailsClient

dampir:
Refer #1182 .

To-do:

- [x]  Add an overload for Task<IReadOnlyList<EmailAddress>> GetAll() on IUserEmailsClient that use ApiOptions parameter.
- [x]  Add unit and integration tests for this new method (need some clarification)
- [x]  Add an overload for IObservable<EmailAddress> on IObservableUserEmailsClient that use ApiOptions parameter.
- [x]  Add unit and integration tests for this new method (need some clarification)

/cc @shiftkey

### DIFF
--- a/Octokit.Tests.Integration/Clients/UserEmailsClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/UserEmailsClientTests.cs
@@ -39,24 +39,9 @@ namespace Octokit.Tests.Integration.Clients
 
             var emails = await _emailClient.GetAll(options);
 
-            Assert.Equal(1, emails.Count);
+            Assert.NotEmpty(emails);
         }
-
-        [IntegrationTest]
-        public async Task ReturnsCorrectCountOfEmailsWithStart()
-        {
-            var options = new ApiOptions
-            {
-                PageSize = 5,
-                PageCount = 1,
-                StartPage = 2
-            };
-
-            var emails = await _emailClient.GetAll(options);
-
-            Assert.Equal(0, emails.Count);
-        }
-
+        
         const string testEmailAddress = "hahaha-not-a-real-email@foo.com";
 
         [IntegrationTest(Skip = "this isn't passing in CI - i hate past me right now")]
--- a/Octokit.Tests.Integration/Clients/UserEmailsClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/UserEmailsClientTests.cs
@@ -15,6 +15,14 @@ namespace Octokit.Tests.Integration.Clients
             Assert.NotEmpty(emails);
         }
 
+        [IntegrationTest]
+        public async Task CanGetEmailWithApiOptions()
+        {
+            var github = Helper.GetAuthenticatedClient();
+            var emails = await github.User.Email.GetAll(ApiOptions.None);
+            Assert.NotEmpty(emails);
+        }
+
         const string testEmailAddress = "hahaha-not-a-real-email@foo.com";
 
         [IntegrationTest(Skip = "this isn't passing in CI - i hate past me right now")]
--- a/Octokit.Tests.Integration/Clients/UserEmailsClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/UserEmailsClientTests.cs
@@ -6,6 +6,14 @@ namespace Octokit.Tests.Integration.Clients
 {
     public class UserEmailsClientTests
     {
+        private readonly IUserEmailsClient _emailClient;
+
+        public UserEmailsClientTests()
+        {
+            var github = Helper.GetAuthenticatedClient();
+            _emailClient = github.User.Email;
+        }
+
         [IntegrationTest]
         public async Task CanGetEmail()
         {
@@ -22,6 +30,58 @@ namespace Octokit.Tests.Integration.Clients
             var emails = await github.User.Email.GetAll(ApiOptions.None);
             Assert.NotEmpty(emails);
         }
+
+        [IntegrationTest]
+        public async Task ReturnsCorrectCountOfEmailsWithoutStart()
+        {
+            var options = new ApiOptions
+            {
+                PageSize = 5,
+                PageCount = 1
+            };
+
+            var emails = await _emailClient.GetAll(options);
+
+            Assert.Equal(1, emails.Count);
+        }
+
+        [IntegrationTest]
+        public async Task ReturnsCorrectCountOfEmailsWithStart()
+        {
+            var options = new ApiOptions
+            {
+                PageSize = 5,
+                PageCount = 1,
+                StartPage = 2
+            };
+
+            var emails = await _emailClient.GetAll(options);
+
+            Assert.Equal(0, emails.Count);
+        }
+
+        //[IntegrationTest]
+        //public async Task ReturnsDistinctResultsBasedOnStartPage()
+        //{
+        //    var startOptions = new ApiOptions
+        //    {
+        //        PageSize = 5,
+        //        PageCount = 1
+        //    };
+
+        //    var firstPage = await _emailClient.GetAll(startOptions);
+
+        //    var skipStartOptions = new ApiOptions
+        //    {
+        //        PageSize = 5,
+        //        PageCount = 1,
+        //        StartPage = 2
+        //    };
+
+        //    var secondPage = await _emailClient.GetAll(skipStartOptions);
+
+        //    Assert.Equal(firstPage[0].Email, secondPage[0].Email);
+        //}
 
         const string testEmailAddress = "hahaha-not-a-real-email@foo.com";
 
--- a/Octokit.Tests.Integration/Clients/UserEmailsClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/UserEmailsClientTests.cs
@@ -17,17 +17,14 @@ namespace Octokit.Tests.Integration.Clients
         [IntegrationTest]
         public async Task CanGetEmail()
         {
-            var github = Helper.GetAuthenticatedClient();
-
-            var emails = await github.User.Email.GetAll();
+            var emails = await _emailClient.GetAll();
             Assert.NotEmpty(emails);
         }
 
         [IntegrationTest]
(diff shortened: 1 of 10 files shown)


Octokit.Tests/Clients/UserEmailsClientTests.cs
```diff
@@ -19,7 +19,26 @@ public void GetsCorrectUrl()
...
+                    PageCount = 1
+                };
+
+                var connection = Substitute.For<IApiConnection>();
+                var client = new UserEmailsClient(connection);
+
+                client.GetAll(options);
+
+                connection.Received(1)
+                    .GetAll<EmailAddress>(Arg.Is<Uri>(u => u.ToString() == "user/emails"), Arg.Is<ApiOptions>(apiOptions => apiOptions == options));
```
devkhan: I think that `Arg.Any<ApiOptions>()` is sufficient here, instead of the elaborate lambda. But its fine for total correctness, I guess.

dampir: Tnx, @devkhan, you're absolutely right, usage of Any is more convenient in this situation. Fixed in ad3b7823.

dampir:
@shiftkey, I have one question about integration tests for IObservableUserEmailsClient. 

In aa3ce0d you're added test ReturnsDistinctResultsBasedOnStartPage for IObservableReleaseClient.

How can I implement such tests for IObservableUserEmailsClient and IUserEmailsClient? 
I use my fake github account for integration test and there is exists only one email.

shiftkey:
@dampir yeah, email addresses are tricky. Let's not worry about integration testing pagination on those.

ryangribble:
Just a query regarding adding the additional overload to ```Connection.GetResponse()``` and changing the ```ApiOptions.None``` to a Singleton implementation.  

Since these weren't in @shiftkey's example implementation I'm wondering whether they should be included here or not.  

There are already lots of overloads on GetResponse etc so we should only add them if absolutely necessary.  Im not sure if in this case it was necessary or just a convenience?

Similarly with the singleton for ApiOptions.None - performance improvement is surely negligible and not really a major concern.  The other point about unit testing... we have plenty of examples of ```Model``` classes that need to be compared in Unit Tests and some discussions have occured around implementing ```IEquality``` for Model classes or using extension/comparison methods etc.  Im not sure that the singleton approach for this one specific edge case (so that you can use object equality of `options == ApiOptions.None`) is worth it?

I defer to @shiftkey for opinion on both matters, I am mainly coming at it from the point of "if this is the first time we are introducing something" it should be thought about in the context of the whole project and not just this PR

shiftkey:
> Since these weren't in @shiftkey's example implementation I'm wondering whether they should be included here or not.

Unless there's a compelling reason to add them I'd like to remove these changes.

The big issue with introducing the `ApiOptions.None` singleton is that `ApiOptions`'s properties are still mutable - potentially introducing strange behaviour if you happen to mutate the global instance - so I'd rather it just return a new instance each time you `get` a value.

The alternative here is to remove the setters so you can enforce read-only values, but the necessary `ctors` to support this felt more verbose so that's why I ended up here.

dampir:
> @dampir yeah, email addresses are tricky. Let's not worry about integration testing pagination on those. 

Tnx, I removed unnecessary tests.

@ryangribble
@shiftkey

I am fully agree with you about "ApiOptions" Singleton (especially, mutability breaks Singleton approach) and "GetResponse" overload. Both of these changes were made in order to implement needed unit/integration tests.

I removed Singleton and overload of "GetResponse" extension. Also I rewrote necessary tests a little bit.

Octokit.Tests.Integration/Clients/UserEmailsClientTests.cs
```diff
@@ -6,15 +6,57 @@ namespace Octokit.Tests.Integration.Clients
...
+        {
+            var options = new ApiOptions
+            {
+                PageSize = 5,
+                PageCount = 1
+            };
+
+            var emails = await _emailClient.GetAll(options);
+
+            Assert.Equal(1, emails.Count);
```
shiftkey: This test will likely break if the user's account has more than one account - what about if we just did a `Assert.NotEmpty(emails)` here too?

Octokit.Tests.Integration/Clients/UserEmailsClientTests.cs
```diff
@@ -6,15 +6,57 @@ namespace Octokit.Tests.Integration.Clients
...
+            var options = new ApiOptions
+            {
+                PageSize = 5,
+                PageCount = 1,
+                StartPage = 2
+            };
+
+            var emails = await _emailClient.GetAll(options);
+
+            Assert.Equal(0, emails.Count);
```
shiftkey: This one is less likely to break than the previous one, but let's drop it anyway...

Octokit.Tests/Reactive/ObservableUserEmailsClientTests.cs
```diff
@@ -18,25 +18,35 @@ private static ObservableUserEmailsClient CreateFixtureWithNonReactiveClient()
...
+                client.GetAll(ApiOptions.None);
+
+                github.Connection.Received(1).Get<List<EmailAddress>>(_expectedUri,
+                    Arg.Is<Dictionary<string, string>>(dictionary => dictionary.Count == 0), null);
             }
         }
 
         public class TheAddMethod
         {
-            public IGitHubClient GitHubClient;
```
shiftkey: Nice spot!

shiftkey:
Those changes look good. Added a couple of comments around the integration tests.

I'll get @ryangribble to have another look once those are addressed.

Octokit.Tests.Integration/Clients/UserEmailsClientTests.cs
dampir: Yes, sure, I going to fix it right now!

Octokit.Tests.Integration/Clients/UserEmailsClientTests.cs
dampir: Ok, I had some doubts about this test case too. I going to remove it, thx!

shiftkey:
@dampir just an FYI - there's no need to @mention me in these PRs and comments unless there's a specific question you need answered from me - I get the notifications either way :grin:

dampir:
@shiftkey Ohh, sorry for that! I don't know that GitHub so smart. I not going to do that in future!

Octokit.Tests/Clients/UserEmailsClientTests.cs
```diff
@@ -1,14 +1,14 @@
...
 using System.Threading.Tasks;
+using NSubstitute;
 using Xunit;
 
 namespace Octokit.Tests.Clients
 {
     public class UserEmailsClientTests
     {
-        public class TheGetAllMethod
+        public class TheGetAllMethods
```
ryangribble: Can we revert this back to `TheGetAllMethod` rather than pluralising it.  Not because there aren't more than one method now, but more just for consistency with the rest of the code base!

Octokit/Models/Request/ApiOptions.cs
```diff
@@ -5,7 +5,7 @@
 namespace Octokit
 {
     [DebuggerDisplay("{DebuggerDisplay,nq}")]
-    public class ApiOptions
+    public class ApiOptions
```
ryangribble: whitespace?

Octokit.Tests.Integration/Reactive/ObservableUserEmailsClientTests.cs
```diff
@@ -7,15 +7,37 @@ namespace Octokit.Tests.Integration
 {
     public class ObservableUserEmailsClientTests
     {
+        private readonly ObservableUserEmailsClient _emailClient
+            = new ObservableUserEmailsClient(Helper.GetAuthenticatedClient());
```
ryangribble: This shared `_emailClient` is used by 2 of the 3 tests in this class, with the other 1 test declaring it's own client...

My personal preference is to share as little between tests as possible, as it makes them more straight forward and resiliient to unintended changes down the track...  however im not saying you have to do it to my preference, but whatever you do should at least be consistent :grinning:

ryangribble:
Ive reviewed the changes and it's looking pretty good!  

My "consistency is king!" OCD shone through a little though sorry :grinning:

To explain where that's coming from: on a code base such as this one, consistency and clarity are very important - they enable new contributors and users/consumers to more easily gain an understanding as to how octokit works.  There will often be a better or more efficient way to write test code - however if you're going to be the "first instance" of deviating from the general approach adhered to within a class/file (or in an ideal world even an APIclient or the whole codebase!), it should be intentional and communicated.  

So I guess what Im saying is, lets definitely try to keep consistency within the same file, if not the whole API client (and ideally code base!)

Octokit.Tests/Clients/UserEmailsClientTests.cs
dampir: Yeah, I changed it :+1: 
Find one more TheGetAllMethods in [GistsClientTests.cs](https://github.com/octokit/octokit.net/blob/615b6900b9cbcdb6fe6e647c245b77c3ef7f8b38/Octokit.Tests/Clients/GistsClientTests.cs#L27)

Octokit/Models/Request/ApiOptions.cs
dampir: For some reasons VS do not show **ending** whitespaces in own diff tool. Is my mistake, fixed.

Octokit.Tests.Integration/Reactive/ObservableUserEmailsClientTests.cs
dampir: Yeah, I agree, I like the consistency too :+1: Fixed,

ryangribble:
Thanks for going the extra mile to address my consistency nitpicks!

LGTM

dampir:
> Thanks for going the extra mile to address my consistency nitpicks!

@shiftkey,
@ryangribble 
Thx for your patient and detailed explanations! :+1:

PR #1195, opened 2016-03-16
New method on RepositoryCommits client - Get the sha1 of a commit reference

### LINKED ISSUE #1189: New method on RepositoryCommits client
https://developer.github.com/v3/repos/commits/#get-the-sha-1-of-a-commit-reference

Added to the [API docs](https://github.com/github/developer.github.com/commit/524e1d748156a8cabfdb01a4ca112c7408bb24b8) on 27 February 2016

```
GET /repos/:owner/:repo/commits/:ref
```

Note this is a preview API and requires a custom accept header to activate the functionality
` application/vnd.github.chitauri-preview+sha`

rogertinsley:
**Title**
New method on RepositoryCommits client - Get the sha1 of a commit reference. Fixes #1189 

**Description**

- [x] Add a sha1 method to IRepositoryCommitsClient
- [x] Add unit and integration tests for this new method
- [x] Add a sha1 method to IObservableRepositoryCommitsClients
- [x] Update AcceptHeaders to include the preview http accept header

This is my first foray into contributing to OSS. Any feedback is greatly appreciated.


Octokit/Clients/IRepositoryCommitsClient.cs
```diff
@@ -50,5 +50,14 @@ public interface IRepositoryCommitsClient
...
         Task<IReadOnlyList<GitHubCommit>> GetAll(string owner, string name, CommitRequest request);
+
+        /// <summary>
+        /// Get the SHA-1 of a commit reference
+        /// </summary>
+        /// <param name="owner">The owner of the repository</param>
+        /// <param name="name">The name of the repository</param>
+        /// <param name="reference">The repository reference</param>
+        /// <returns></returns>
+        Task<string> Sha1(string owner, string name, string reference);
```
shiftkey: Could we add a verb to this method? Say something like `GetSha`?

shiftkey:
@rogertinsley welcome!

To be perfectly honest, the only bits of feedback I have are minor - this is really nice work!

One suggestion - if you wanted to add another integration test where you:

 - create a repository
 - make a commit through the API, get the new SHA
 - verify the SHA from this new endpoint matches what you got earlier

I think that'll give us enough coverage for this endpoint to close this out.

rogertinsley:
Thanks @shiftkey, I'll make the changes that you've suggested :+1:

Octokit.Tests/Clients/RepositoriesClientTests.cs
```diff
@@ -790,5 +790,35 @@ public void GetsCorrectUrl()
                 await Assert.ThrowsAsync<ArgumentException>(() => client.EditBranch("owner", "repo", "", update));
             }
         }
+
+        public class TheSha1Method
+        {
+            [Fact]
+            public async Task EnsureNonNullArguments()
```
ryangribble: Minor note that the wording should be "Ensures" rather than "Ensure".  Since the test shoudl read TheGetSha1Method.EnsuresNonNullArguments() :grinning:

Also this test is checking Null and Empty arguments.  In other places ive seen us implement this as 2 test methods eg `EnsuresNonNullArguments()` and `EnsuresNonEmptyArguments()`

ryangribble:
Great first contribution... I've agree with @shiftkey regarding `GetSha1()` and the integration test... and I've also raised one minor nitpick on consistency/wording

M-Zuber:
:clap: Nice stuff!
Welcome to OSS :smile:

rogertinsley:
I've made the changes you both have suggested.

Are you happy with the integration test approach? I've used a similar approach to other tests in the same class, that is to use `CreateTheWorld` to create repo & commit. Get the Sha1 and compare that it hasn't changed.

Octokit.Tests/Clients/RepositoriesClientTests.cs
```diff
@@ -790,5 +790,41 @@ public void GetsCorrectUrl()
                 await Assert.ThrowsAsync<ArgumentException>(() => client.EditBranch("owner", "repo", "", update));
             }
         }
+
+        public class TheSha1Method
+        {
+            [Fact]
+            public async void EnsuresNonNullArguments()
```
shiftkey: Make this `async Task` as this is more test-runner friendly...

Octokit.Tests.Integration/Clients/RepositoryCommitsClientTests.cs
```diff
@@ -158,6 +166,17 @@ public TestsWithNewRepository()
             Assert.Equal(0, result.BehindBy);
         }
 
+        [IntegrationTest]
+        public async Task GetSha1FromRepository()
+        {
+            await CreateTheWorld();
```
shiftkey: I'm not sure how much this will break, but I'd love to get the `Reference` that `CreateTheWorld` returns - it should contain a `Sha`.

So rather than invoking `GetSha1` twice below, you invoke it once and compare that SHA to the one you have here.

Does that make sense?

Octokit.Tests/Clients/RepositoriesClientTests.cs
shiftkey: And the CI test won't fail like it currently is...

rogertinsley:
I've updated `CreateTheWorld` to return a `Task<Reference>` and compare it against the `GetSha1` and the class `TheSha1Method` methods should be async

Octokit.Tests/Clients/RepositoriesClientTests.cs
```diff
@@ -790,5 +790,40 @@ public void GetsCorrectUrl()
                 await Assert.ThrowsAsync<ArgumentException>(() => client.EditBranch("owner", "repo", "", update));
             }
         }
+
+        public class TheSha1Method
```
ryangribble: Should be `TheGetSha1Method`

rogertinsley: Thanks - updated it, I'm getting there :smile:

ryangribble:
This looks good!  

But I noticed there are no Unit or Integration tests for the `ObservableRepositoryCommitsClient`.

I realise this was like that before this PR... Some parts of the code base (particularly observable implementations) don't have tests around them, but as we touch these areas ideally we should start to correct that problem.  Without tests there really could be something lurking in there making the observable implementation not work and we dont want to find out from end users if possible!

So at a minimum, I'd like to see tests added for the Observable `GetSha1()` method that has been added... And ideally - tests for ALL the methods in this observable client could be added!  :grinning:

If you have a look at some existing examples of Observable client unit and integration tests you will find that the unit tests of observable would have a `CallsIntoClient` type test (that checks the underlying regular client method is called with appropriate args), as well as `EnsuresNonNullArguments` and `EnsuresNonEmptyArguments` tests as appropriate (depending on what parameters the call takes and whether they have checks for nulls/empty string).  Also a test for the `Ctor` guarding against a null client argument.

Im not sure where @shiftkey stands on integration tests for observable clients - given that alot are missing, and it's pretty much only ever calling the regular client anyway... perhaps all we need are unit tests for observable clients... 

But in my contributions I do typically add integration tests, I pretty much copy/paste the regular client test, modify it to use the observable client and also like to write it so it's obvious the item is an observable (eventhough you could simply `await` it inline).  This is just personal preference.

``` csharp
var observable = _fixture.GetSha1(x,x,x);
var sha1 = await observable;
```

rogertinsley:
I've added unit and integrations tests for the `ObservableRepositoryCommitsClient`, specifically the `GetSha1` method.

shiftkey:
> Im not sure where @shiftkey stands on integration tests for observable clients - given that alot are missing, and it's pretty much only ever calling the regular client anyway... perhaps all we need are unit tests for observable clients...

This is essentially my views - if the implementation is just calling the `Task`-based API, then additional integration tests for the `Observable`s are just doubling up. But yes, :thumbsup: to additional unit tests.

ryangribble:
Oops!  Looks like just a minor issue there with the "NotNull" test checking empty strings, and the "NonEmpty" test checking nulls!  If you can just fix that up we are good to go

Also given @shiftkey's comment let's ditch the integration test for the observable implementation, since it's just calling the real client anyway.  SORRY!

rogertinsley:
No problem - I've removed the integration test and updated the unit test. 

I've learnt a lot about the codebase and how to contribute with this PR and you guys haven't scared me off either :grin: so I'm going to pick up some more `up-for-grabs` :smile:

shiftkey:
@rogertinsley this all looks good! I can't wait for the next one.

I'll let @ryangribble have the final word.

ryangribble:
> I've learnt a lot about the codebase and how to contribute with this PR and you guys haven't scared me off either

Great to hear!  Bring on the next one, this one LGTM
LGTM

PR #1196, opened 2016-03-17
TheUpdateMethod.EnsureArgumentsNotNull function changed

prayankmathur:
I found the function [here](https://github.com/octokit/octokit.net/blob/master/Octokit.Tests/Reactive/ObservableIssuesClientTests.cs#L327) to be incorrect. Therefore corrected it.

--EDIT--

I also ran **./build FormatCode** just in case i missed some formatting things.

### DIFF
--- a/Octokit.Reactive/Clients/Enterprise/IObservableEnterpriseLdapClient.cs
+++ b/Octokit.Reactive/Clients/Enterprise/IObservableEnterpriseLdapClient.cs
@@ -33,7 +33,7 @@ namespace Octokit.Reactive
         /// <param name="userName">The username to sync LDAP mapping</param>
         /// <returns>The <see cref="LdapSyncResponse"/> of the queue request.</returns>
         IObservable<LdapSyncResponse> QueueSyncUserMapping(string userName);
-        
+
         /// <summary>
         /// Update the LDAP mapping for a team on a GitHub Enterprise appliance (must be Site Admin user).
         /// </summary>
--- a/Octokit.Reactive/Clients/Enterprise/ObservableEnterpriseLdapClient.cs
+++ b/Octokit.Reactive/Clients/Enterprise/ObservableEnterpriseLdapClient.cs
@@ -49,7 +49,7 @@ namespace Octokit.Reactive
         {
             return _client.QueueSyncUserMapping(userName).ToObservable();
         }
-        
+
         /// <summary>
         /// Update the LDAP mapping for a team on a GitHub Enterprise appliance (must be Site Admin user).
         /// </summary>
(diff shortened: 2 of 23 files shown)


devkhan:
@prayankmathur I think you applied the patch for it in the same branch of another PR. You should have created another branch for this PR, as it will unnecessary show the old commits.

prayankmathur:
@devkhan 
Actually i removed the changes from the previous pull request and opened a new PR so that changes can be merged quickly, as suggested by @shiftkey .

shiftkey:
@prayankmathur yep, but I'd rather just deal with the single commit - instead of all the preceding commits.

Here's a way to reset this branch to just contain that commit:

```
// if you have your remotes already configured, skip this step
git remote add upstream https://github.com/octokit/octokit.net.git -f 
// ensure we're on the right branch
git checkout UpdateIssue
// reset this branch to the latest version in development
git reset upstream/master --hard
// add just one commit to the new branch
git cherry-pick 5118d8d
// push these changes to reset the branch
git push origin UpdateIssue --force
```

prayankmathur:
@shiftkey
Thanks !! .
Made the changes accordingly.

ryangribble:
Apart from fixing the mentioned issue, this commit also seems to have applied some whitespace/bracket reformatting across a number of files.  

This isn't necessarily a problem but I do like to see it explicitly mentioned in your PR comments, if you've done something like that, firstly to give an indication that it's a deliberate action on your part, and not some inadvertent/unintended action that you weren't even aware of... and secondly so that you can give your reasoning/justification as to how you made the change and why etc... (eg did you run something over the entire code base and these were the only files that weren't "in spec" or otherwise how come "only" ~20 files are affected?)

prayankmathur:
@ryangribble 
I ran this command too **./build FormatCode**
Actually i was not hoping for so many changes because i was hoping other contributors would be knowing this and i read it from the gitter chat group.
I ll mention it from next time what all changes i am making.

shiftkey:
@prayankmathur I generally don't obsess over everyone running `FormatCode` as part of every pull request. For me, it was there to run periodically to ensure everything was up to (a) standard.

As you can see, all it caught here was a bunch of excess whitespace - so this is minor thing for me.

@ryangribble I'm not going to stress about this too much. Thanks for being on top of it!

ryangribble:
> @ryangribble I'm not going to stress about this too much. Thanks for being on top of it!

All good!  My approach with such "off topic" changes is I'll normally question them if the PR author didnt mention them themselves...  If they do mention it, then they aren't "off topic" anymore and thus no need to question!

And yeah, I hadnt actually seen the FormatCode build command until I was poling around the other day too... perhaps it should be added to the contributors.md.  Not as a "you must run this" but more so as a "if you want to, here it is"

shiftkey:
@ryangribble captured in #1201

PR #1197, opened 2016-03-17
introduce Code of Conduct

shiftkey:
If you're not familiar with it, the [Contributor Covenant](http://contributor-covenant.org/) is a good starting point for communities to identify the values they uphold, as well as what happens when conflicts arise.

Opening this up to start a discussion on what should be in it.
- ~~[ ] discussion about things to add/change/remove~~
- ~~[ ] introduce preamble which is Octokit-specific~~
- [x] replace `[INSERT EMAIL ADDRESS]` with actual email address
- [x] mention this in `README.md`

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,74 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
+contributors and maintainers pledge to making participation in our project and
+our community a harassment-free experience for everyone, regardless of age, body
+size, disability, ethnicity, gender identity and expression, level of experience,
+nationality, personal appearance, race, religion, or sexual identity and
+orientation.
+
+## Our Standards
+
+Examples of behavior that contributes to creating a positive environment
+include:
+
+* Using welcoming and inclusive language
+* Being respectful of differing viewpoints and experiences
+* Gracefully accepting constructive criticism
+* Focusing on what is best for the community
+* Showing empathy towards other community members
+
+Examples of unacceptable behavior by participants include:
+
+* The use of sexualized language or imagery and unwelcome sexual attention or
+advances
+* Trolling, insulting/derogatory comments, and personal or political attacks
+* Public or private harassment
+* Publishing others' private information, such as a physical or electronic
+  address, without explicit permission
+* Other conduct which could reasonably be considered inappropriate in a
+  professional setting
+
+## Our Responsibilities
+
+Project maintainers are responsible for clarifying the standards of acceptable
+behavior and are expected to take appropriate and fair corrective action in
+response to any instances of unacceptable behavior.
+
+Project maintainers have the right and responsibility to remove, edit, or
+reject comments, commits, code, wiki edits, issues, and other contributions
+that are not aligned to this Code of Conduct, or to ban temporarily or
+permanently any contributor for other behaviors that they deem inappropriate,
+threatening, offensive, or harmful.
+
+## Scope
+
+This Code of Conduct applies both within project spaces and in public spaces
+when an individual is representing the project or its community. Examples of
+representing a project or community include using an official project e-mail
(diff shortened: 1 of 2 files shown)


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,74 @@
...
+
+Examples of unacceptable behavior by participants include:
+
+* The use of sexualized language or imagery and unwelcome sexual attention or
+advances
+* Trolling, insulting/derogatory comments, and personal or political attacks
+* Public or private harassment
+* Publishing others' private information, such as a physical or electronic
+  address, without explicit permission
+* Other conduct which could reasonably be considered inappropriate in a
```
M-Zuber: Serious question: Some of use come from a more relaxed definition of `professional setting` in general, or at least when it comes to projects done on private time.
Is it worth it/is there a way at all to reword this bit so as not to cause confusion due to the dry (American based) meaning?

shiftkey: Given this is a "virtual" group of contributors and spanning just about all the continents, I'm open to elaborating on what this might mean to our situation. This also feels like a bit of a catch-all phrase, so perhaps we can be more specific here. 

Will ponder on it, but open to suggestions...

caseycraig:
NEO I BELIEVE

CODE_OF_CONDUCT.md
Haacked: I'd like to leave it as-is so we stay true to the original CoC. We take a pretty relaxed approach to our "professional setting" so telling jokes and being silly is fine as long as the jokes aren't demeaning or exclusive of others along the lines described. So probably fine to make the joke about VB the programming language, but not jokes that demean women.

Also, we understand that people make mistakes and have different ideas about what is considered appropriate. We're not going to hell-ban you for the first infraction unless it's way over the line.

> Project maintainers are responsible for clarifying the standards of acceptable
> behavior and are expected to take appropriate and fair corrective action in
> response to any instances of unacceptable behavior.

Consider this document the start of the conversation, not the end. For the most part, I don't see anything in here so far that goes against the spirit of this. Everyone that's participated has been pretty great. We've had a few bad eggs, but most were still somewhat respectful.

Haacked:
:sparkles: I'm excited to see this added. We want to signal to people that they are welcome to contribute and that we care about having an inclusive welcoming environment for all contributors.

M-Zuber:
:shipit:

Haacked:
@shiftkey should the email address be support@github.com?

shiftkey:
@Haacked I'd love it to be more specific than that, so that it goes directly to the right people

Haacked:
> @Haacked I'd love it to be more specific than that, so that it goes directly to the right people

Please don't make it go just to me. :stuck_out_tongue: What did you have in mind?

shiftkey:
@Haacked something like `conduct+octokit.net@github.com` but I need to bug some people internally to ensure we're all on the same page before confirming it...

M-Zuber:
@shiftkey bump

shiftkey:
Yeah, this fell off our radar due to Other Priorities™.

I'm taking next week off (computer-free) so I'll pick this up when I'm back.

SeanKilleen:
Bumping this; would be nice to actually get it in.

For the sake of merging this, could it make sense to use `support@github.com` for now and add a follow-up issue to add a more specific e-mail address when one is available? Seems like a quick win to get this in; plus, you know that someone will be manning the e-mail there.

If you want to give them the ability to filter, maybe using an e-mail trick like `support+octokitnet+conduct@github.com` could do what you're looking for? Would still go to support but they'd be able to route based on the `to` address.

ryangribble:
I agree it'd be nice to merge this in.  Obviously GitHub staffers will need to make the call about what email address to use (and whether we can use support@github.com for now)...  ping @shiftkey @Haacked

shiftkey:
Updated the guidelines to use a real email address. Aside from some logistics work on my end this is ready to merge. Over to you @haacked.

Haacked:
🎉

PR #1198, opened 2016-03-17
Add ApiOption overloads to methods on IAuthorizationsClient

dampir:
Refer #1149  .

To-do:

- [x]  Add overload Task<IReadOnlyList<Authorization>> GetAll(ApiOptions options) on IAuthorizationsClient.
- [x]  Add overload IObservable<Authorization> GetAll(ApiOptions options) on IObservableAuthorizationsClient.
- [x] Add integration tests for these new methods.

I have one question about integration tests for (Observable)AuthorizationsClient. There are no any integration tests for these clients, so should I add it?

### DIFF
--- a/Octokit.Reactive/Clients/IObservableAuthorizationsClient.cs
+++ b/Octokit.Reactive/Clients/IObservableAuthorizationsClient.cs
@@ -19,6 +19,19 @@ namespace Octokit.Reactive
         IObservable<Authorization> GetAll();
 
         /// <summary>
+        /// Get all <see cref="Authorization"/>s for the authenticated user. This method requires basic auth.
+        /// </summary>
+        /// <remarks>
+        /// See <a href="http://developer.github.com/v3/oauth/#list-your-authorizations">API documentation</a> for more
+        /// details.
+        /// </remarks>
+        /// <param name="options">Options for changing the API response</param>
+        /// <returns>An <see cref="Authorization"/></returns>
+        [SuppressMessage("Microsoft.Design", "CA1024:UsePropertiesWhereAppropriate",
+            Justification = "It's an API call, so it's not a property.")]
+        IObservable<Authorization> GetAll(ApiOptions options);
+
+        /// <summary>
         /// Get a specific <see cref="Authorization"/> for the authenticated user. This method requires basic auth.
         /// </summary>
         /// <remarks>
--- a/Octokit.Reactive/Clients/IObservableAuthorizationsClient.cs
+++ b/Octokit.Reactive/Clients/IObservableAuthorizationsClient.cs
@@ -13,7 +13,7 @@ namespace Octokit.Reactive
         /// See <a href="http://developer.github.com/v3/oauth/#list-your-authorizations">API documentation</a> for more
         /// details.
         /// </remarks>
-        /// <returns>An <see cref="Authorization"/></returns>
+        /// <returns>A list of <see cref="Authorization"/>s for the authenticated user.</returns>
         [SuppressMessage("Microsoft.Design", "CA1024:UsePropertiesWhereAppropriate",
             Justification = "It's an API call, so it's not a property.")]
         IObservable<Authorization> GetAll();
@@ -26,7 +26,7 @@ namespace Octokit.Reactive
         /// details.
         /// </remarks>
         /// <param name="options">Options for changing the API response</param>
-        /// <returns>An <see cref="Authorization"/></returns>
+        /// <returns>A list of <see cref="Authorization"/>s for the authenticated user.</returns>
         [SuppressMessage("Microsoft.Design", "CA1024:UsePropertiesWhereAppropriate",
             Justification = "It's an API call, so it's not a property.")]
         IObservable<Authorization> GetAll(ApiOptions options);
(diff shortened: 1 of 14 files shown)


Octokit.Reactive/Clients/IObservableAuthorizationsClient.cs
```diff
@@ -19,6 +19,19 @@ public interface IObservableAuthorizationsClient
...
 
         /// <summary>
+        /// Get all <see cref="Authorization"/>s for the authenticated user. This method requires basic auth.
+        /// </summary>
+        /// <remarks>
+        /// See <a href="http://developer.github.com/v3/oauth/#list-your-authorizations">API documentation</a> for more
+        /// details.
+        /// </remarks>
+        /// <param name="options">Options for changing the API response</param>
+        /// <returns>An <see cref="Authorization"/></returns>
```
M-Zuber: I'm not sure why but in the rest of the lib the `<returns>` are left empty.
We should either stick with that - or start here as change (which would necessitate going back and changing the entire code base)

M-Zuber:
Other than my comment (which is more a question really) it LGTM!

> I have one question about integration tests for (Observable)AuthorizationsClient. There are no any integration tests for these clients, so should I add it?

My (non-maintainer) opinion is - if the tests follow existing conventions and coding styles - test away.

Octokit.Reactive/Clients/IObservableAuthorizationsClient.cs
dampir: I updated method's documentation.

ryangribble:
> I have one question about integration tests for (Observable)AuthorizationsClient. There are no any integration tests for these clients, so should I add it?

My vote is yes!  Ideally we want to have unit and integration tests for everything.  Some older stuff doesnt have that coverage, but the next time we touch something in that area, ideally we would also add the tests for that client, at a minimum implementing the tests for the methods you've touched (eg in this case the `GetAll()` method, but if time allows writing "full" tests for all of the methods in that Client would be great as it will increase our coverage

Sometime soon I'd like to write something that will highlight the clients/methods that dont have unit or integration test coverage, and raise issues in the repo for them to be implemented

shiftkey:
Seconding everything @ryangribble just said above...

dampir:
I added new integration tests in last commits.

Octokit/Clients/AuthorizationsClient.cs
```diff
@@ -33,10 +33,30 @@ public AuthorizationsClient(IApiConnection apiConnection) : base(apiConnection)
...
+        /// </remarks>
+        /// <param name="options">Options for changing the API response</param>
+        /// <exception cref="AuthorizationException">
+        /// Thrown when the current user does not have permission to make the request.
+        /// </exception>
+        /// <exception cref="ApiException">Thrown when a general API error occurs.</exception>
+        /// <returns>A list of <see cref="Authorization"/>s for the authenticated user.</returns>
+        public Task<IReadOnlyList<Authorization>> GetAll(ApiOptions options)
+        {
+            Ensure.ArgumentNotNull(options, "options");
```
ryangribble: I wonder with these `ApiOptions` implementations, if there should actually be a unit test for this exception throw.  (And same in [`ObservableAuthorizationsClient`](https://github.com/octokit/octokit.net/pull/1198/files#diff-32f89c741486cc29188bdd64ec7b2ad9R45))

`TheGetAllMethod ... EnsuresArgumentsNotNull()`

Octokit.Tests/Clients/AuthorizationsClientTests.cs
```diff
@@ -37,6 +37,19 @@ public void GetsAListOfAuthorizations()
...
+            public void RequestsCorrectUrlWithApiOptions()
+            {
+                var client = Substitute.For<IApiConnection>();
+                var authEndpoint = new AuthorizationsClient(client);
+                
+                authEndpoint.GetAll(ApiOptions.None);
+
+                client.Received().GetAll<Authorization>(
+                    Arg.Is<Uri>(u => u.ToString() == "authorizations"),
+                    Args.ApiOptions);
```
ryangribble: This test and the one before it (`RequestsCorrectUrl()`) are both checking the same thing.  Perhaps this one should check that the specific ApiOptions object specified was passed through to the client.  Eg instead of using ApiOptions.None, declare one containing some page size/skip settings, then verify it is actually seen on the `Received()` call?

I realise you were just following the reference implementation but Im just wondering before we go through and implement ALL the pagination PR's we should settle on a good set of test coverage.  There are only 1 or 2 merged so far, so we can easily go back and add the extra unit test to those if we decide it's needed.  The main thing is, it would be done on all the remaining issues as they are merged.

My 2c is that as much as possible gets covered by the unit tests! :grinning:

dampir: I'm fully agree, will be fixed in the next commits. Now I unmarked items in PR todo list, so WIP :)

Octokit/Clients/AuthorizationsClient.cs
dampir: I'm fully agree, fixed in the next commits. Also, I reviewed other places where ApiOptions is used and update the tests accordingly.

ryangribble:
Looking awesome now :grinning: And thanks for taking care of those additional tests on the previously implemented ApiOptions clients as well.

Great job @dampir!

LGTM

PR #1205, opened 2016-03-22
Redundant "this." qualifier was removed.

dampir:
Resharper has some useful code analysis that used to eliminate code redundancies. I used Resharper to find redundant "this." qualifiers and delete them.

Also, I want to deliver a few PR that are connected with some of redundancies in Octokit.Net, Have I open issue for each of them? Or I could create just PR without appropriate issue?

/cc @shiftkey


shiftkey:
@dampir 

> Have I open issue for each of them? Or I could create just PR without appropriate issue?

Save time, just go straight to the PR - especially for things like cleanup and/or refactoring.

ryangribble:
:+1:

PR #1206, opened 2016-03-22
Redundant string.Format() call was removed.

dampir:
Resharper has some useful code analysis that used to eliminate code redundancies. I used Resharper to find "Redundant string.Format() call" and delete them.


ryangribble:
Pretty clear cut, may as well merge these ones quickly :grinning:

PR #1209, opened 2016-03-22
Inconsistent modifiers declaration order was fixed.

dampir:
Resharper has some useful code analysis that used to eliminate code redundancies. I used Resharper to find "Inconsistent modifiers declaration order" issues and fix them.


ryangribble:
:+1:

PR #1210, opened 2016-03-23
Fix common redundancies by resharper

dampir:
I've removed some common redundancies of code from Octokit.Net.

List of redundancies:

1. Redundant explicit array creation in argument of 'params' parameter.
2. Redundant type qualifiers.
3. Empty constructors.
4.  Redundant explicit array type specification.
5. Redundant member initializers.
6.  Redundant collection list initializers.
7.  Unused "using" directives were removed (in modified files).
8. Some misspellings were fixed.

This PR was made in order to support code quality in Octokit.Net.

### DIFF
--- a/Octokit.Tests.Integration/Clients/OrganizationMembersClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/OrganizationMembersClientTests.cs
@@ -42,7 +42,7 @@ namespace Octokit.Tests.Integration.Clients
 
                 // There shouldnt be any members that are in both groups if the role filter works correctly
                 var membersInBoth = adminMembers.Select(a => a.Id).Intersect(normalMembers.Select(n => n.Id));
-                Assert.True(!membersInBoth.Any());
+                Assert.NotEmpty(membersInBoth);
             }
 
             [IntegrationTest(Skip = "TwoFactor filter can't be used unless the requester is an organization owner")]
--- a/Octokit.Tests.Integration/Clients/OrganizationMembersClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/OrganizationMembersClientTests.cs
@@ -1,10 +1,5 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
+﻿using System.Linq;
 using System.Threading.Tasks;
-using Octokit;
-using Octokit.Tests.Integration;
-using Octokit.Tests.Integration.Helpers;
 using Xunit;
 
 namespace Octokit.Tests.Integration.Clients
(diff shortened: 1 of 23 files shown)


Haacked:
In general, each of these should be individual pull requests. However, in this case, the PR is small enough I'll allow it. :smile: I'll let @shiftkey take a look and merge it.

dampir:
Yep, I thought about separate PR for each type of changes, so I'm going to create small PR in further, because all changes are small and quite simple. There are some common code cleanups that I want to provide.

Octokit.Tests.Integration/Clients/OrganizationMembersClientTests.cs
```diff
@@ -47,7 +42,7 @@ public TheGetAllMethod()
 
                 // There shouldnt be any members that are in both groups if the role filter works correctly
                 var membersInBoth = adminMembers.Select(a => a.Id).Intersect(normalMembers.Select(n => n.Id));
-                Assert.True(membersInBoth.Count() == 0);
+                Assert.True(!membersInBoth.Any());
```
shiftkey: :lipstick: let's make this `Assert.NotEmpty(membersInBoth);`

dampir: Done!

devkhan:
@dampir Can you make the list in the first comment a to-do list? Or if everything is done, ignore this.

dampir:
@devkhan, everything is done.

shiftkey:
I have some reservations about the `SimpleJson` changes, mostly due to that file being an external dependency we pull in. Are you able to revert that change?

Haacked:
> I have some reservations about the SimpleJson changes, mostly due to that file being an external dependency we pull in. Are you able to revert that change?

Good catch! Yes, we don't want to change that file. That breaks NuGet's ability to detect if there are any changes or not when upgrading the package.

dampir:
Changes has been reverted. I'll exclude that file from Resharer's analysis.

Octokit.Tests.Integration/Clients/OrganizationMembersClientTests.cs
```diff
@@ -47,7 +42,7 @@ public TheGetAllMethod()
 
                 // There shouldnt be any members that are in both groups if the role filter works correctly
                 var membersInBoth = adminMembers.Select(a => a.Id).Intersect(normalMembers.Select(n => n.Id));
-                Assert.True(membersInBoth.Count() == 0);
+                Assert.NotEmpty(membersInBoth);
```
shiftkey: Apologies: I inverted this assert earlier - this should be `Assert.Empty`

dampir: Done!

shiftkey:
@dampir excellent, thanks!

PR #1211, opened 2016-03-23
Add ApiOptions overloads to ITeamsClient

### LINKED ISSUE #1181: Add ApiOption overloads to methods on ITeamsClient

- Add an overload for this method on `ITeamsClient` with a new parameter `ApiOptions`
- Implement the new method on the class which implements `ITeamsClient`
- Add an overload for this method on `IObservableTeamsClient` with a new parameter `ApiOptions`
- Implement the new method on the class which implements `IObservableTeamsClient`

- [ ] `Task<IReadOnlyList<Team>> GetAll(String org)`
- [ ] `Task<IReadOnlyList<Team>> GetAllForCurrent()`
- [ ] `Task<IReadOnlyList<User>> GetAllMembers(Int32 id)`
- [ ] `Task<IReadOnlyList<Repository>> GetAllRepositories(Int32 id)`

devkhan:
Fixes #1181 .

Checklist:

- [x] Add an overload for this method on `ITeamsClient` with a new parameter `ApiOptions`
- [x] Implement the new method on the class which implements `ITeamsClient`
- [x] Add an overload for this method on `IObservableTeamsClient` with a new parameter `ApiOptions`
- [x] Implement the new method on the class which implements `IObservableTeamsClient`
- [x] Reactive client tests.
- [x] Integration tests.

As I was doing this, a couple of questions sprung up in my mind:

- ~~After adding `ApiOptions` overload to the methods, we change the implementation of the old method to just call the new one with `ApiOptions.None`. Why don't we just write one method and use default values of `ApiOptions.None`?~~ see discussion
- ~~Secondly, modifying the test for the overloads requires you to check for `Args.ApiOptions`, neither `ApiOptions.None` nor `Arg.Any<ApiOptions>(ApiOptions.None)` work. Why is that so?~~ see discussion

/cc @shiftkey @ryangribble

### DIFF
--- a/Octokit.Reactive/Clients/IObservableTeamsClient.cs
+++ b/Octokit.Reactive/Clients/IObservableTeamsClient.cs
@@ -27,9 +27,19 @@ namespace Octokit.Reactive
         /// <summary>
         /// Returns all <see cref="Team" />s for the current org.
         /// </summary>
+        /// <param name="org">Organization to list all teams of.</param>
         /// <exception cref="ApiException">Thrown when a general API error occurs.</exception>
         /// <returns>A list of the orgs's teams <see cref="Team"/>s.</returns>
         IObservable<Team> GetAll(string org);
+
+        /// <summary>
+        /// Returns all <see cref="Team" />s for the current org.
+        /// </summary>
+        /// <param name="org">Organization to list all teams of.</param>
+        /// <param name="options">Options to change API behaviour.</param>
+        /// <exception cref="ApiException">Thrown when a general API error occurs.</exception>
+        /// <returns>A list of the orgs's teams <see cref="Team"/>s.</returns>
+        IObservable<Team> GetAll(string org, ApiOptions options);
 
         /// <summary>
         /// Returns all <see cref="Team" />s for the current user.
@@ -40,15 +50,36 @@ namespace Octokit.Reactive
         IObservable<Team> GetAllForCurrent();
 
         /// <summary>
+        /// Returns all <see cref="Team" />s for the current user.
+        /// </summary>
+        /// <param name="options">Options to change API behaviour.</param>
+        /// <exception cref="ApiException">Thrown when a general API error occurs.</exception>
+        /// <returns>A list of the user's <see cref="Team"/>s.</returns>
+        [SuppressMessage("Microsoft.Design", "CA1024:UsePropertiesWhereAppropriate")]
+        IObservable<Team> GetAllForCurrent(ApiOptions options);
+
+        /// <summary>
         /// Returns all members of the given team. 
         /// </summary>
-        /// <param name="id">The team identifier</param>
         /// <remarks>
         /// https://developer.github.com/v3/orgs/teams/#list-team-members
         /// </remarks>
+        /// <param name="id">The team identifier</param>
         /// <exception cref="ApiException">Thrown when a general API error occurs.</exception>
         /// <returns>A list of the team's member <see cref="User"/>s.</returns>
         IObservable<User> GetAllMembers(int id);
+
+        /// <summary>
+        /// Returns all members of the given team. 
+        /// </summary>
+        /// <remarks>
+        /// https://developer.github.com/v3/orgs/teams/#list-team-members
+        /// </remarks>
+        /// <param name="id">The team identifier</param>
+        /// <param name="options">Options to change API behaviour.</param>
+        /// <exception cref="ApiException">Thrown when a general API error occurs.</exception>
+        /// <returns>A list of the team's member <see cref="User"/>s.</returns>
+        IObservable<User> GetAllMembers(int id, ApiOptions options);
 
         /// <summary>
         /// Returns newly created <see cref="Team" /> for the current org.
@@ -116,9 +147,19 @@ namespace Octokit.Reactive
         /// <summary>
         /// Returns all team's repositories.
         /// </summary>
+        /// <param name="id">Team Id.</param>
         /// <exception cref="ApiException">Thrown when a general API error occurs.</exception>
         /// <returns>The team's repositories</returns>
         IObservable<Repository> GetAllRepositories(int id);
+
+        /// <summary>
+        /// Returns all team's repositories.
+        /// </summary>
+        /// <param name="id">Team Id.</param>
+        /// <param name="options">Options to change API behaviour.</param>
+        /// <exception cref="ApiException">Thrown when a general API error occurs.</exception>
+        /// <returns>The team's repositories</returns>
+        IObservable<Repository> GetAllRepositories(int id, ApiOptions options);
 
         /// <summary>
         /// Remove a repository from the team
--- a/Octokit.Reactive/Clients/ObservableTeamsClient.cs
+++ b/Octokit.Reactive/Clients/ObservableTeamsClient.cs
@@ -47,6 +47,8 @@ namespace Octokit.Reactive
         /// <returns>A list of the orgs's teams <see cref="Team"/>s.</returns>
         public IObservable<Team> GetAll(string org)
         {
+            Ensure.ArgumentNotNullOrEmptyString(org, "org");
+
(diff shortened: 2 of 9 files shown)


dampir:
> After adding ApiOptions overload to the methods, we change the implementation of the old method to just call the new one with ApiOptions.None. Why don't we just write one method and use default values of ApiOptions.None?

@devkhan, I am agree with such approach, it could be very convenient, because instead of 2 method in interface we will have only one with default parameter. Since there are not so many overloads were added, we can do refactor easily at this stage.

shiftkey:
I'd rather not introduce optional parameters here for fear of [versioning issues](http://haacked.com/archive/2010/08/10/versioning-issues-with-optional-arguments.aspx/) down the track, and also for having to write code like this inside the implementation:

```csharp
public Task<IReadOnlyList<Release>> GetAll(string owner, string name, ApiOptions options = null)
{
   options = options ?? ApiOptions.None;
   // other code here
}
```

devkhan:
I agree on this for null values, but according to current implementation, if we keep the default value `ApiOptions.None`, we wouldn't need to write the `??` syntax and the non-null check. Are there any other problems which will arise due to this?

Haacked:
@devkhan the link @shiftkey provides points out some of the potential compatibility issues with optional arguments. They were designed for COM interop, not for public APIs.

```csharp
public Task<IReadOnlyList<Release>> GetAll(string owner, string name, ApiOptions options);
public Task<IReadOnlyList<Release>> GetAll(string owner, string name);
```

Another benefit of having overloads is with the optional parameter approach you can't distinguish between the case where the user passed in a null value explicitly by mistake, or if they left it out.

For example, the following is a mistake and should throw an `ArgumentNullException`. Clearly the programmer made a mistake.

```csharp
ApiOptions options = null;
client.GetAll("foo", "bar", options);
```

They either should set `options` to something, or just omit that argument. Having two overloads makes that clear.

```csharp
client.GetAll("foo", "bar", null); // Throws ArgumentNullException
client.GetAll("foo", "bar"); // Just fine. Programmer's intentions are clear.
```

devkhan:
@Haacked I read the article, it made clear the intentions and problems which arise due to optional parameters. I guess I was just being nostalgic due to my C++ experience.

Thanks to both of you.

Octokit.Reactive/Clients/IObservableTeamsClient.cs
```diff
@@ -27,11 +27,21 @@ public interface IObservableTeamsClient
         /// <summary>
         /// Returns all <see cref="Team" />s for the current org.
         /// </summary>
+        /// <param name="org">Organization to list all teams of.</param>
```
shiftkey: Nice catch!

Haacked:
> nostalgic due to my C++ experience

Heh. I've never had the experience of being [nostalgic for C++](http://robey.lag.net/2012/10/09/cpp-quirks-remembered.html). :wink:

I was glad to move on. Out of curiosity, what about C++ makes the optional parameters different?

devkhan:
> what about C++ makes the optional parameters different?

Nothing. It was at the time when I only knew C++ and less code seemed better. :disappointed:

shiftkey:
@devkhan in response to your second question:

> Secondly, modifying the test for the overloads requires you to check for `Args.ApiOptions`, neither `ApiOptions.None` nor `Arg.Any<ApiOptions>(ApiOptions.None)` work. Why is that so?

Because `ApiOptions.None` is just a plain old instance of `ApiOptions`, testing for a specific instance is tricky. You *could* check it has the default values set, but as the whole purpose of the class is to just carry around values, I didn't really focus on it here.

Our helper  `Args.ApiOptions` is just a wrapper around `Args.Any<ApiOption>()` - if you wanted to test for specific values on `ApiOptions` you could use `Arg.Any<ApiOptions>(o => o.Page == 2)` to make this clearer. I'll leave it up to you how far you want to take this...

devkhan:
@shiftkey Thanks for clearing that out.

Sidenote: I've sent you an email. Can you please check? :pray:

shiftkey:
@devkhan I can't see anything interesting in my inbox

devkhan:
@shiftkey Is my proposal really that boring?

shiftkey:
@devkhan apologies, that's not what I meant. I saw the stuff about your proposal but couldn't find anything else...

devkhan:
@shiftkey Well, now I've uploaded the final proposal on GSoC website. Hope you like it. :pray:

Octokit.Reactive/Clients/ObservableTeamsClient.cs
```diff
@@ -47,8 +47,22 @@ public IObservable<Team> Get(int id)
         /// <returns>A list of the orgs's teams <see cref="Team"/>s.</returns>
         public IObservable<Team> GetAll(string org)
         {
+            return GetAll(org, ApiOptions.None);
```
devkhan: I also thought of adding it, but since it will be checked in the internal `GetAll`, I skipped it. Should I add here too?

shiftkey: Yes please

Octokit.Tests.Integration/Reactive/ObservableTeamsClientTests.cs
```diff
@@ -10,13 +10,13 @@ public class ObservableTeamsClientTests
 {
     public class TheGetMembersMethod
     {
-        readonly Team team;
+        readonly Team _team;
```
shiftkey: This is a coding style thing, which is cool and all but there's no other changes around here.

Can we revert this to simplify the diff?

ryangribble:
Although, there are plenty of existing cases of naming class variables with the underscore, and its also as per the [CoreFx coding style](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md) which @Haacked mentoned we are generally following...  

My take is, as long as the PR has made the class consistent (ie all cases of private members are using underscore notation) that it is OK and actually means we are standardising towards the coreFx standards.

In this case there is only 1 private variable so the change does look isolated but at the same time, it IS consistent (as the entire file is now using underscore notation).  Although the entire PR is not consistent because the regular client integration tests have 2 cases of local class variables which werent converted to underscore notation (only the reactive client tests were).  So personally I'd say I am happy with moving any instances to underscore notation where they exist, as long as all instances are done in the PR... but I dont presume to override @shiftkey's request :grinning: just proividing my 2 cents

My personal preference is to use the underscore, which is also something that resharper suggests, and as mentioned the coreFx coding style also defines.  Similar to that other discussion and blog post by @Haacked between `string.Format` and `String.Format` I think the conclusion was, no point fighting everyone over your personal preference - if we can just point to some other standard then it's generally a lot less effort :grinning: 

Perhaps if this change is reverted then we can have a meta discussion to talk about how seriously we adopt the CoreFx style or not etc?

shiftkey:
@ryangribble okay, I'll bite :grin:

You're right, this isn't a blocker. And it's only one situation. But I'd like these PRs to focus on adding the `ApiOptions` overloads wherever possible so I can get the release out as soon as possible. That's why I'm calling out coding style things in these PRs.

> My personal preference is to use the underscore, which is also something that resharper suggests, and as mentioned the coreFx coding style also defines.

This is one place I don't like the R# defaults, let alone the CoreFX coding style, as I feel it adds unnecessary noise to the code - and not everyone runs R#.

Happy to have a discussion on this and where my thoughts are at. I'd also remove all usages of `private` for example, as that's considered the default behaviour for visibility...

ryangribble:
> But I'd like these PRs to focus on adding the ApiOptions overloads wherever possible so I can get the release out as soon as possible. That's why I'm calling out coding style things in these PRs.

Makes sense... I'll raise a meta discussion about coding styles for further deliberations :grinning:

shiftkey:
@devkhan @ryangribble I won't hold up merging this for the single change. Let's hash it out in that related issue.

PR #1212, opened 2016-03-23
Add ApiOption overloads to methods on ICommitStatusClient

SamTheDev:
This implementation targets the issue [#1150](https://github.com/octokit/octokit.net/issues/1150)

- [x] Add an overload for the GetAll method on ICommitStatusClient with a new parameter ApiOptions
- [x] Implement the the new method on the class which implements ICommitStatusClient
- [x] Add an overload for the GetAll method on IObservableCommitStatusClient with a new parameter ApiOptions
- [x] Implement the new method on the class which implements IObservableCommitStatusClient
- [x] Implement the RequestsCorrectUrlWithApiOptions unit test
- [x] Implement the integration tests

Ps: this is my first contribution to this repo, your remarks are welcome
@shiftkey

### DIFF
--- a/Octokit.Reactive/Clients/ObservableCommitStatusClient.cs
+++ b/Octokit.Reactive/Clients/ObservableCommitStatusClient.cs
@@ -31,8 +31,8 @@ namespace Octokit.Reactive
             Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
             Ensure.ArgumentNotNullOrEmptyString(name, "name");
             Ensure.ArgumentNotNullOrEmptyString(reference, "reference");
-                        
-            return _connection.GetAndFlattenAllPages<CommitStatus>(ApiUrls.CommitStatuses(owner, name, reference));
+
+            return GetAll(owner, name ,reference, ApiOptions.None);                  
         }
 
         /// <summary>
--- a/Octokit.Reactive/Clients/ObservableCommitStatusClient.cs
+++ b/Octokit.Reactive/Clients/ObservableCommitStatusClient.cs
@@ -28,7 +28,31 @@ namespace Octokit.Reactive
         /// <returns></returns>
         public IObservable<CommitStatus> GetAll(string owner, string name, string reference)
         {
+            Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
+            Ensure.ArgumentNotNullOrEmptyString(name, "name");
+            Ensure.ArgumentNotNullOrEmptyString(reference, "reference");
+                        
             return _connection.GetAndFlattenAllPages<CommitStatus>(ApiUrls.CommitStatuses(owner, name, reference));
+        }
+
+        /// <summary>
+        /// Retrieves commit statuses for the specified reference. A reference can be a commit SHA, a branch name, or
+        /// a tag name.
+        /// </summary>
+        /// <remarks>Only users with pull access can see this.</remarks>
+        /// <param name="owner">The owner of the repository</param>
+        /// <param name="name">The name of the repository</param>        
+        /// <param name="reference">The reference (SHA, branch name, or tag name) to list commits for</param>
+        /// <param name="options">Options for changing the API response</param>
+        /// <returns></returns>
+        public IObservable<CommitStatus> GetAll(string owner, string name, string reference, ApiOptions options)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
+            Ensure.ArgumentNotNullOrEmptyString(name, "name");
+            Ensure.ArgumentNotNullOrEmptyString(reference, "reference");
+            Ensure.ArgumentNotNull(options, "options"); 
+
+            return _connection.GetAndFlattenAllPages<CommitStatus>(ApiUrls.CommitStatuses(owner, name, reference),options);
         }
 
         /// <summary>
--- a/Octokit.Tests.Integration/Reactive/ObservableCommitStatusClientTests.cs
+++ b/Octokit.Tests.Integration/Reactive/ObservableCommitStatusClientTests.cs
@@ -1,0 +1,88 @@
+﻿using System.Reactive.Linq;
+using System.Threading.Tasks;
+using Octokit.Reactive;
+using Xunit;
+
+namespace Octokit.Tests.Integration.Reactive
+{
+    public class ObservableCommitStatusClientTests
+    {
+        public class TheGetAllMethod
+        {
+            readonly ObservableCommitStatusClient _commitStatusClient;
+            const string owner = "octokit";
+            const string name = "octokit.net";
+            const string reference = "master";
+
+            public TheGetAllMethod()
+            {
+                var github = Helper.GetAuthenticatedClient();
+                _commitStatusClient = new ObservableCommitStatusClient(github);
+            }
+
+            [IntegrationTest]
+            public async Task ReturnCommitStatus()
+            {
+                var commitStatus = await _commitStatusClient.GetAll(owner, name, reference).ToList();
+
+                Assert.NotEmpty(commitStatus);
+            }
+
+            [IntegrationTest]
+            public async Task ReturnsCorrectCountOfCommitStatusWithoutStart()
+            {
+                var options = new ApiOptions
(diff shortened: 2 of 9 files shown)


Octokit/Clients/CommitStatusClient.cs
```diff
@@ -36,7 +36,29 @@ public Task<IReadOnlyList<CommitStatus>> GetAll(string owner, string name, strin
...
+        /// </summary>
+        /// <remarks>
+        /// https://developer.github.com/v3/repos/statuses/#list-statuses-for-a-specific-ref
+        /// </remarks>
+        /// <param name="owner">The owner of the repository</param>
+        /// <param name="name">The name of the repository</param>
+        /// <param name="options">Options for changing the API response</param>
+        /// <param name="reference">The reference (SHA, branch name, or tag name) to list commits for</param>
+        /// <returns></returns>
+        public Task<IReadOnlyList<CommitStatus>> GetAll(string owner, string name, ApiOptions options, string reference)
```
shiftkey: Can you change this method signature to be `(owner, name, reference, options)`?

SamTheDev: Ok

Octokit.Reactive/Clients/ObservableCommitStatusClient.cs
```diff
@@ -28,10 +28,34 @@ public ObservableCommitStatusClient(IGitHubClient client)
         /// <returns></returns>
         public IObservable<CommitStatus> GetAll(string owner, string name, string reference)
         {
+            Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
+            Ensure.ArgumentNotNullOrEmptyString(name, "name");
+            Ensure.ArgumentNotNullOrEmptyString(reference, "reference");
```
shiftkey: Rather than duplicating all these rules, `GetAll(owner, name, reference)` should just call `GetAll(owner, name, reference, ApiOptions.None)`

shiftkey:
@SamTheDev this looks like it's progressing nicely :metal:

Octokit.Reactive/Clients/ObservableCommitStatusClient.cs
SamTheDev: > Rather than duplicating all these rules, GetAll(owner, name, reference) should just call GetAll(owner, name, reference, ApiOptions.None)

Yea agreed i will fix that in the next commits

shiftkey: Sorry, I should clarify this before you go and fix it.

 - The line below is very similar to the actual implementation inside the new `GetAll` method you've defined. So rather than call `_connection.GetAndFlattenPages<T>(...)` here, why not just call `GetAll(owner, name, reference, ApiOptions.None);` here?

 - We've been discussing what to do with the `Ensure` usages in the case where we call a different overload. We've settled on keeping these around, even if they are duplicated checks. So don't clean them up here.

Make sense?

SamTheDev: > We've been discussing what to do with the Ensur e usages in the case where we call a different overload. We've settled on keeping these around, even if they are duplicated checks. So don't clean them up here.

I am not sure why to !, a link to the discussion would be helpful, thanks (y)

SamTheDev:
Hi @shiftkey I need some help/codeReview with the IntegrationTests part pls

Octokit.Reactive/Clients/ObservableCommitStatusClient.cs
shiftkey: @SamTheDev here's the related discussion https://github.com/octokit/octokit.net/pull/1218#discussion_r57460248

SamTheDev: @shiftkey And could you take a look at the Integration Tests :pray: ?

Octokit.Tests.Integration/Reactive/ObservableCommitStatusClientTests.cs
```diff
@@ -0,0 +1,88 @@
...
+namespace Octokit.Tests.Integration.Reactive
+{
+    public class ObservableCommitStatusClientTests
+    {
+        public class TheGetAllMethod
+        {
+            readonly ObservableCommitStatusClient _commitStatusClient;
+            const string owner = "octokit";
+            const string name = "octokit.net";
+            const string reference = "master";
```
shiftkey: Given `master` will actually change over time, we'd need to be careful here about what we're testing.

For example, you could lock this to 1335f37c6cde55de09c086cda283e70e0dfc339e which has 5 statuses and won't change. This means the first three tests will pass currently.

Octokit.Tests.Integration/Reactive/ObservableCommitStatusClientTests.cs
```diff
@@ -0,0 +1,88 @@
...
+
+                Assert.NotEmpty(commitStatus);
+            }
+
+            [IntegrationTest]
+            public async Task ReturnsCorrectCountOfCommitStatusWithoutStart()
+            {
+                var options = new ApiOptions
+                {
+                    PageSize = 5,
```
shiftkey: Given we're working with five commit statuses, you could change this `PageSize` to 2 so that you can get these other tests working as expected...

Octokit.Tests.Integration/Reactive/ObservableCommitStatusClientTests.cs
```diff
@@ -0,0 +1,88 @@
...
+                var skipStartOptions = new ApiOptions
+                {
+                    PageSize = 5,
+                    PageCount = 1,
+                    StartPage = 2
+                };
+
+                var secondPage = await _commitStatusClient.GetAll(owner, name, reference,skipStartOptions).ToList();             
+
+                Assert.NotEqual(firstPage[0].Id, secondPage[0].Id);
```
shiftkey: With this one here, we just want to ensure we're getting different results. Working with a page size of two should get this passing as well, just drop the index checks that are outside this range.

Octokit.Tests/Clients/CommitStatusClientTests.cs
```diff
@@ -20,8 +20,27 @@ public void RequestsCorrectUrl()
...
+                {
+                    PageSize = 1,
+                    PageCount = 1,
+                    StartPage = 1
+                };
+
+                client.GetAll("fake", "repo", "sha", options);
+
+                connection.Received()
+                    .GetAll<CommitStatus>(Arg.Is<Uri>(u => u.ToString() == "repos/fake/repo/commits/sha/statuses"),options);
```
shiftkey: Swap `options` out for `Args.ApiOptions` here

shiftkey:
@SamTheDev I added a bunch of suggestions to your initial tests, what you have there is rather close

SamTheDev:
@shiftkey  I've implemented your suggestions, thx man (y)

Octokit.Tests.Integration/Reactive/ObservableCommitStatusClientTests.cs
```diff
@@ -0,0 +1,85 @@
...
+            {
+                var options = new ApiOptions
+                {
+                    PageSize = 2,
+                    PageCount = 1
+                };
+
+                var commitStatus = await _commitStatusClient.GetAll(owner, name ,reference , options).ToList();
+
+                Assert.Equal(5, commitStatus.Count);
```
shiftkey: This test will fail as we changed the page size above...

shiftkey:
@SamTheDev almost there, just that one last test to update!

Octokit.Tests.Integration/Reactive/ObservableCommitStatusClientTests.cs
SamTheDev: I forget to commit that one...My bad

shiftkey:
@SamTheDev excellent, I'm going to hit the big green button now. Thanks for the contribution!

SamTheDev:
WooooHoo... Look mom i made it to master, ...Thx to you @shiftkey 

panda

PR #1213, opened 2016-03-24
Added ApiOption overloads to methods on IRepositoryPagesClient and IObservableRepositoryPagesClient

prayankmathur:
@shiftkey 
@ryangribble 

Refer #1178 

Added the methods as well as tests (reactive methods too).


Octokit.Reactive/Clients/ObservableRepositoryPagesClient.cs
```diff
@@ -54,6 +54,25 @@ public IObservable<PagesBuild> GetAll(string owner, string repositoryName)
...
+        /// See the <a href="https://developer.github.com/v3/repos/pages/#list-pages-builds">API documentation</a> for more information.
+        /// </remarks>
+        /// <returns></returns>
+        public IObservable<PagesBuild> GetAll(string owner, string repositoryName, ApiOptions options)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
+            Ensure.ArgumentNotNullOrEmptyString(repositoryName, "repositoryName");
+            Ensure.ArgumentNotNull(options, "options");
+
+            return _connection.GetAndFlattenAllPages<PagesBuild>(ApiUrls.RepositoryPageBuilds(owner, repositoryName), options);
```
shiftkey: While we're in here, can you change the method body of `GetAll(string owner, string repositoryName)` to point to this - to reduce the amount duplicated code?

Octokit/Clients/RepositoryPagesClient.cs
```diff
@@ -50,10 +50,29 @@ public Task<IReadOnlyList<PagesBuild>> GetAll(string owner, string repositoryNam
...
+        /// </remarks>
+        /// <returns></returns>
+        public Task<IReadOnlyList<PagesBuild>> GetAll(string owner, string repositoryName, ApiOptions options)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
+            Ensure.ArgumentNotNullOrEmptyString(repositoryName, "repositoryName");
+            Ensure.ArgumentNotNull(options, "options");
+
+            var endpoint = ApiUrls.RepositoryPageBuilds(owner, repositoryName);
+            return ApiConnection.GetAll<PagesBuild>(endpoint, null, AcceptHeaders.StableVersion, options);
```
shiftkey: We should be able to get away with just using the `GetAll<T>` overload that takes `endpoint` and `options`, right?

prayankmathur: Yeah, my bad :grin: 
Have made the changes

prayankmathur:
@shiftkey 
Made the changes. Take a look.

shiftkey:
@prayankmathur looks like some tests need to be updated

prayankmathur:
@shiftkey 
Yeah, I was doing that.
Also, should there be a change [here](https://github.com/octokit/octokit.net/blob/master/Octokit.Reactive/Clients/ObservableReleasesClient.cs#L36) too, like the one that you suggested.

shiftkey:
@prayankmathur you're right - I'll correct that and open a separate PR for it...

Octokit.Tests/Clients/RepositoryPagesClientTests.cs
```diff
@@ -31,7 +31,7 @@ public void RequestsCorrectUrl()
             }
         }
 
-        public class TheGetBuildsMethod
+        public class TheGetAllBuildsMethod
```
shiftkey: :lipstick: this should be `TheGetAllMethod`

shiftkey:
@prayankmathur this is looking good.

We don't seem to have any integration tests for the `RepositoryPagesClient`, but that seems to be covered by the `ObservableRepositoryPagesClientTests` tests. Would you mind adding that in now, at least just for the new `GetAll` methods we're adding here?

prayankmathur:
@shiftkey 
I have added the integration tests as you said. Please take a look.

Octokit/Clients/IRepositoryPagesClient.cs
```diff
@@ -33,6 +34,19 @@ public interface IRepositoryPagesClient
...
         /// <returns></returns>
         Task<IReadOnlyList<PagesBuild>> GetAll(string owner, string repositoryName);
+
+        /// <summary>
+        /// Gets all build metadata for a given repository
+        /// </summary>
+        /// <param name="owner">The owner of the repository</param>
+        /// <param name="repositoryName">The name of the repository</param>
+        /// <param name="options">Options for changing the behaviour of the API</param>
+        ///  <remarks>
```
ryangribble: extraneous whitespace here? between /// and <remarks> there are 2 spaces instead of 1.  Seems to be the case on a few of the xmldoc comments in multiple files

Also should the description of options be the same between observable and regular client concrete and interfaces?  Currently across the 4 places, there are 3 different wordings

> Options for changing the behaviour of the API

> Options to change the behaviour of the API

> Options to change the API response

prayankmathur:
@ryangribble 
I changed the description to

>Options to change the API response

And i guess it should be same.

Octokit.Reactive/Clients/IObservableRepositoryPagesClient.cs
```diff
@@ -26,6 +27,18 @@ public interface IObservableRepositoryPagesClient
...
         /// <returns></returns>
         IObservable<PagesBuild> GetAll(string owner, string repositoryName);
+
+        /// <summary>
+        /// Gets all build metadata for a given repository
+        /// </summary>
+        /// <param name="owner">The owner of the repository</param>
+        /// <param name="repositoryName">The name of the repository</param>
+        /// <param name="options">Options to change the API response</param>
+        ///  <remarks>
```
ryangribble: whitespace

Octokit.Reactive/Clients/ObservableRepositoryPagesClient.cs
```diff
@@ -50,7 +50,26 @@ public IObservable<PagesBuild> GetAll(string owner, string repositoryName)
...
+            return GetAll(owner, repositoryName, ApiOptions.None);
+        }
+
+        /// <summary>
+        /// Gets all build metadata for a given repository
+        /// </summary>
+        /// <param name="owner">The owner of the repository</param>
+        /// <param name="repositoryName">The name of the repository</param>
+        /// <param name="options">Options to change the API response</param>
+        ///  <remarks>
```
ryangribble: whitespace

prayankmathur:
@ryangribble 
I have removed the whitespaces. I copy pasted the description from the function above and therefore the whitespace crept in everywhere.

Please take a look.

ryangribble:
Looks good now.  Yeah I realise often these little inconsistencies are already in the files, but whenever we are touching files I like to clean whatever we can up as we go :+1:

prayankmathur:
@ryangribble 
@shiftkey 
Why are there no integration tests for the methods in the `RespositoryPagesClient` ?
I have added the tests for the `GetAll`

devkhan:
@prayankmathur There is some code at places which are not covered by tests. I have encountered this situation in the TeamsClient also, so I think we should add tests wherever they are missing.

shiftkey:
@prayankmathur 

> Why are there no integration tests for the methods in the `RespositoryPagesClient`?

Probably something that was missed at the time. Given this is a read-only interface it should be easy to use Octokit (or something similar) and craft some simple tests to get our coverage up...

prayankmathur:
@shiftkey 
So should i do the honors and include them too ? :grinning:

ryangribble:
You know the answer's gonna be yes :grinning:  We always like moar tests

shiftkey:
> So should i do the honors and include them too ?

prayankmathur:
@shiftkey 
@ryangribble 
I have added some tests for the functions.
Please take a look.
Also is there a way to check `GetLatest` returns the latest build, because i couldn't find a way.

Octokit.Tests/Clients/RepositoryPagesClientTests.cs
```diff
@@ -50,8 +50,9 @@ public void RequestsCorrectUrl()
                 var connection = Substitute.For<IApiConnection>();
                 var client = new RepositoryPagesClient(connection);
 
-                await Assert.ThrowsAsync<ArgumentNullException>(() => client.Get(null, "name"));
-                await Assert.ThrowsAsync<ArgumentNullException>(() => client.Get("owner", null));
+                await Assert.ThrowsAsync<ArgumentNullException>(() => client.GetAll(null, "name", new ApiOptions()));
+                await Assert.ThrowsAsync<ArgumentNullException>(() => client.GetAll("owner", null, new ApiOptions()));
+                await Assert.ThrowsAsync<ArgumentNullException>(() => client.GetAll("owner", "name", null));
             }
```
ryangribble: Can you also add a `EnsuresNonEmptyArguments()` test function that passes `""` into the variuos parameters and ensures exceptions are thrown?

Octokit.Tests.Integration/Clients/RepositoryPagesClientTests.cs
```diff
@@ -0,0 +1,119 @@
...
+            _repositoryPagesClient = github.Repository.Page;
+        }
+
+        [IntegrationTest]
+        public async Task ReturnsMetadata()
+        {
+            var data = await _repositoryPagesClient.Get(owner, name);
+            Assert.Equal("https://api.github.com/repos/octokit/octokit.net/pages", data.Url);
+        }
+    }
```
ryangribble: Missing whitespace line between these 2 classes

ryangribble:
Ive made a couple of minor comments.  Also it seems like there are no unit tests for `Observable` client.

> Also is there a way to check GetLatest returns the latest build, because i couldn't find a way.

Im not too familiar with the `Pages` side of things, I'll have a look but it could be as  you say, no way to know if it's truly the latest or not.  As long as you are calling the right endpoint etc (as verified by unit tests) you have to trust the github api will return latest when we ask for it :grinning:

shiftkey:
@ryangribble @prayankmathur looks like there's some limitations with accessing the Pages API based on the [docs](https://developer.github.com/v3/repos/pages/):

>  Information about the site and the builds can only be accessed by **authenticated owners**, even though the websites are public.

This is going to cause headaches, and perhaps we need to do this setup:

 - create the repository 
 - publish something to `gh-pages`
 - check build staus
 - teardown the repository

This will likely mean most of the tests here are excessive...

prayankmathur:
@shiftkey 
So i should remove those `excessive tests` and then maybe open up a separate issue to create this setup ?

shiftkey:
@prayankmathur that sounds great! Leave one test with an `"These tests require repository admin rights - see https://github.com/octokit/octokit.net/issues/XXX for discussion"` so we can link to the related discussion...

prayankmathur:
@shiftkey 
I think that appveyor is not running integration tests.
could you please take a look.

prayankmathur:
@shiftkey 
How do I know that appveyor is running/skipping the test that i mentioned to skip ?

shiftkey:
@prayankmathur it's defined in AppVeyor's config file - https://github.com/octokit/octokit.net/blob/master/appveyor.yml#L3-L7

shiftkey:
I'm not sure what the current state of this PR is, so I'm moving this to #1304 so I can merge this new code in (with the skipped test).

PR #1218, opened 2016-03-25
ObservableReleasesClient.GetAll should use overloads properly

shiftkey:
See comment: https://github.com/octokit/octokit.net/pull/1213#issuecomment-200981143

 - [x] fix impacted tests

### DIFF
--- a/Octokit.Reactive/Clients/ObservableReleasesClient.cs
+++ b/Octokit.Reactive/Clients/ObservableReleasesClient.cs
@@ -33,7 +33,7 @@ namespace Octokit.Reactive
             Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
             Ensure.ArgumentNotNullOrEmptyString(name, "name");
 
-            return GetAll(owner, name, ApiOptions.None);            
+            return GetAll(owner, name, ApiOptions.None);
         }
 
         /// <summary>
--- a/Octokit.Tests/Reactive/ObservableReleasesClientTests.cs
+++ b/Octokit.Tests/Reactive/ObservableReleasesClientTests.cs
@@ -29,7 +29,7 @@ namespace Octokit.Tests.Reactive
                 client.GetAll("fake", "repo");
 
                 gitHubClient.Connection.Received(1).Get<List<Release>>(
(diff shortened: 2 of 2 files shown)


Octokit.Reactive/Clients/ObservableReleasesClient.cs
```diff
@@ -33,7 +33,7 @@ public IObservable<Release> GetAll(string owner, string name)
             Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
             Ensure.ArgumentNotNullOrEmptyString(name, "name");
```
dampir:
```cs
`  Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
              Ensure.ArgumentNotNullOrEmptyString(name, "name");`
Should we save these checks if we them in overloaded method too?

shiftkey: Personally I'd lean towards "no", but I'll let @Haacked chime in here as well and see if we agree or not on this...

Haacked: I think of these as being like contracts. I want them as close to the call-site as possible. Semantically, I think of these arguments being invalid for this method.

Eventually, we'll replace many of these with Fody NullGuard which will throw the exception at the call site.

shiftkey: > I think of these as being like contracts. I want them as close to the call-site as possible. 

:thumbsup:

shiftkey:
Just rebased this due to a merge conflict. @ryangribble can I get a :thumbsup: here?

ryangribble:
> Just rebased this due to a merge conflict. @ryangribble can I get a :thumbsup: here?

Given the tests are passing, I wasn't going to look too deeply, but I did notice that the title of commit and PR "ensure observable method calls overload with ApiOptions" isnt actually evident in the code changes - which are only tidying up whitespace formatting and changing the mock asertion to ANY dictionary rather than empty dictionary.

Im assuming the actual change to call the correct overload was already in master, and the rebase is hiding that here?  Should the PR be reworded to "tidy up/tweak test" ?

shiftkey:
@ryangribble yeah, looks like #1212 also touched these files. Given the difference is now minor, I'll skip it.

PR #1221, opened 2016-03-25
Unused "using" directives were removed in whole solution.

dampir:
In order to get code more clean I deleted unused "using" directives in whole solution. I hope there will not any merge issues. If so, pls reject this PR, I will do that more carefully.


Haacked:
Thanks! In general I'd rather we just clean these up as we go along. Unused `using` statements don't hurt anything. Having said that, this definitely appeals to my OCD tendencies so I'll allow it just this once! :smile:.

PR #1222, opened 2016-03-25
Unused ArgumentNotNullOrEmptyString checks were removed

dampir:
In order to get code more clean I deleted unused ArgumentNotNullOrEmptyString checks wherever ApiOptions is ised as parameter for overloads GetAll(...).


Haacked:
These are definitely being used.

Issue #1151, opened 2016-03-14
Add ApiOption overloads to methods on IDeploymentsClient


shiftkey:
As part of completing #760 a number of overloads need to be added to the client interfaces.

As an example, this has been added to [`IReleasesClient`](https://github.com/octokit/octokit.net/commit/03ea83148995711fc47148489e2123e65f33bd82)and [`IObservableReleasesClient`](https://github.com/octokit/octokit.net/commit/aa3ce0db9210b778414e290af7f17c4a646bdeec).

The checklist for porting each method is:

- Add an overload for this method on `IDeploymentsClient` with a new parameter `ApiOptions`
- Implement the new method on the class which implements `IDeploymentsClient`
- Add an overload for this method on `IObservableDeploymentsClient` with a new parameter `ApiOptions`
- Implement the new method on the class which implements `IObservableDeploymentsClient`

This needs to be done for each of these methods:

- [ ] `Task<IReadOnlyList<Deployment>> GetAll(String owner, String name)`

devkhan:
Can be closed now.

ryangribble:
Fyi for anyone not aware of this github feature, if when you mention an issue in PR, you use words like "fixes #1151" or "resolves #1151" "fixes #issue" then the issue is automatically closed when PR merged. So including those words in your PR will avoid needing to manually close out issues 😁

devkhan:
@ryangribble I know about that, but that didn't work if the PR opener doesn't have appropriate access rights to close the issue. Maybe I'm wrong, I don't know for sure.

dampir:
@ryangribble tnx, I will do that in future. For some reasons I've used "refers" instead of "fixes" :smile: 

@devkhan As far as in [github blog post](https://github.com/blog/1506-closing-issues-via-pull-requests) nothing mentioned about rights, I guess auto close should work.

Maybe you confused with this: [Closing issues via commit messages](https://github.com/blog/1506-closing-issues-via-pull-requests)?

shiftkey:
> I know about that, but that didn't work if the PR opener doesn't have appropriate access rights to close the issue.

I don't believe this is the case. Please mention `Fixes #XXX` in the PR title.

PR #1223, opened 2016-03-25
Add ApiOption overloads to methods on IDeploymentsClient

dampir:
Refer #1151 .

To-do:

- [x] Add overload Task> GetAll(ApiOptions options) on IDeploymentsClient.
- [x]  Add overload IObservable GetAll(ApiOptions options) on IObservableDeploymentsClient.

Some additional checks for count of received parameters were added in ObservableDeploymentsClientTests. If these are ambiguous checks, pls, let me know :+1:

### DIFF
--- a/Octokit.Reactive/Clients/ObservableDeploymentsClient.cs
+++ b/Octokit.Reactive/Clients/ObservableDeploymentsClient.cs
@@ -31,11 +31,7 @@ namespace Octokit.Reactive.Clients
         /// <returns>All the <see cref="Deployment"/>s for the specified repository.</returns>
         public IObservable<Deployment> GetAll(string owner, string name)
         {
-            Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
-            Ensure.ArgumentNotNullOrEmptyString(name, "name");
-
-            return _connection.GetAndFlattenAllPages<Deployment>(
-                ApiUrls.Deployments(owner, name));
+            return GetAll(owner, name, ApiOptions.None);
         }
 
         /// <summary>
--- a/Octokit.Reactive/Clients/ObservableDeploymentsClient.cs
+++ b/Octokit.Reactive/Clients/ObservableDeploymentsClient.cs
@@ -31,6 +31,9 @@ namespace Octokit.Reactive.Clients
         /// <returns>All the <see cref="Deployment"/>s for the specified repository.</returns>
         public IObservable<Deployment> GetAll(string owner, string name)
         {
+            Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
+            Ensure.ArgumentNotNullOrEmptyString(name, "name");
+
             return GetAll(owner, name, ApiOptions.None);
         }
 
--- a/Octokit.Reactive/Clients/ObservableDeploymentsClient.cs
+++ b/Octokit.Reactive/Clients/ObservableDeploymentsClient.cs
@@ -39,6 +39,27 @@ namespace Octokit.Reactive.Clients
         }
 
         /// <summary>
+        /// Gets all the deployments for the specified repository. Any user with pull access
+        /// to a repository can view deployments.
+        /// </summary>
+        /// <remarks>
+        /// http://developer.github.com/v3/repos/deployments/#list-deployments
+        /// </remarks>
+        /// <param name="owner">The owner of the repository</param>
+        /// <param name="name">The name of the repository</param>
+        /// <param name="options">Options for changing the API response</param>
+        /// <returns>All the <see cref="Deployment"/>s for the specified repository.</returns>
+        public IObservable<Deployment> GetAll(string owner, string name, ApiOptions options)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
+            Ensure.ArgumentNotNullOrEmptyString(name, "name");
+            Ensure.ArgumentNotNull(options, "options");
+
+            return _connection.GetAndFlattenAllPages<Deployment>(
+                ApiUrls.Deployments(owner, name), options);
+        }
+
+        /// <summary>
         /// Creates a new deployment for the specified repository.
         /// Users with push access can create a deployment for a given ref.
         /// </summary>
--- a/Octokit.Tests.Integration/Clients/DeploymentsClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/DeploymentsClientTests.cs
@@ -1,4 +1,6 @@
 ﻿using System;
+using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using Octokit;
 using Octokit.Tests.Integration;
@@ -9,7 +11,7 @@ public class DeploymentsClientTests : IDisposable
 {
     private readonly IDeploymentsClient _deploymentsClient;
     private readonly RepositoryContext _context;
-    private readonly Commit _commit;
+    private readonly ICollection<Commit> _commits;
 
     public DeploymentsClientTests()
     {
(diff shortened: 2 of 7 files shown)


Octokit.Tests.Integration/Clients/DeploymentsClientTests.cs
```diff
@@ -17,33 +19,38 @@ public DeploymentsClientTests()
 
         _deploymentsClient = github.Repository.Deployment;
         _context = github.CreateRepositoryContext("public-repo").Result;
+        _commits = new List<Commit>();
 
-        var blob = new NewBlob
+        for (int i = 0; i < 6; i++)
```
shiftkey: Given not all of the tests need this range of commits, do you think it'd be useful to introduce some new code here for the more complex deployment scenarios?

Octokit.Tests/Clients/DeploymentsClientTests.cs
```diff
@@ -94,11 +119,11 @@ public void PostsToDeploymentsUrl()
         {
             var connection = Substitute.For<IApiConnection>();
             var client = new DeploymentsClient(connection);
-            var expectedUrl = "repos/owner/name/deployments";
+            var expectedUrl = ApiUrls.Deployments("owner", "name");
```
shiftkey: Don't change this to call the same code that the app does - let's keep this as-is to verify we're also formatting the URL correctly...

dampir: Fixed.

Octokit.Tests.Integration/Clients/DeploymentsClientTests.cs
dampir: Do you mean, I should move creation of these additional commits from test constructor to each ApiOption test method?

shiftkey: @dampir while I like making this initial code more generic (and extending it to do multiple commits) I think we can extract this from the constructor and just have a standalone method or methods.

A couple of reasons:

 - we're creating 6 commits for every test run, even though many of them only need one
 - by moving the creation of the commits closer to the test itself, we can make the intent clearer
 - we're doing a lot of `.Result` here, which blocks the current thread as you can't await in a constructor. By moving this async code closer into the test itself, we can leverage async/await better - which the test runner supports

dampir: All reasons sounds reasonable :) I'll refactor it.

dampir:
I've done some refactorings, pls, check this out whenever possible.

Octokit.Tests/Clients/DeploymentsClientTests.cs
```diff
@@ -45,24 +50,44 @@ public void RequestsCorrectUrl()
         {
             var connection = Substitute.For<IApiConnection>();
             var client = new DeploymentsClient(connection);
-            var expectedUrl = "repos/owner/name/deployments";
+            var expectedUrl = ApiUrls.Deployments(owner, name);
```
shiftkey: Let's put this back to their expected format, instead of using `ApiUrls` here.

Octokit.Tests/Clients/DeploymentsClientTests.cs
```diff
@@ -45,24 +50,44 @@ public void RequestsCorrectUrl()
...
+        }
 
-            client.GetAll("owner", "name");
-            connection.Received(1).GetAll<Deployment>(Arg.Is<Uri>(u => u.ToString() == expectedUrl));
+        [Fact]
+        public void RequestsCorrectUrlWithApiOptions()
+        {
+            var connection = Substitute.For<IApiConnection>();
+            var client = new DeploymentsClient(connection);
+            var expectedUrl = ApiUrls.Deployments(owner, name);
```
shiftkey: Let's put this back to their expected format, instead of using `ApiUrls` here.

Octokit/Clients/DeploymentsClient.cs
```diff
@@ -34,10 +34,27 @@ public DeploymentsClient(IApiConnection apiConnection)
         /// <returns>All the <see cref="Deployment"/>s for the specified repository.</returns>
         public Task<IReadOnlyList<Deployment>> GetAll(string owner, string name)
         {
+            return GetAll(owner, name, ApiOptions.None);
```
shiftkey: Like we mentioned earlier, can you bring back the `Ensure` arguments for `owner` and `name` here?

Octokit.Tests/Reactive/ObservableDeploymentsClientTests.cs
```diff
@@ -25,43 +27,88 @@ public TheGetAllMethod()
...
+                    StartPage = 1,
+                    PageCount = 1,
+                    PageSize = 1
+                };
+
+                _client.GetAll(owner, name, options);
+                _githubClient.Connection
+                             .Received(1)
+                             .Get<List<Deployment>>(Arg.Is(expectedUri),
+                                                         Arg.Is<IDictionary<string, string>>(dictionary => dictionary.Count == 2), null);
```
shiftkey: :lipstick: indenting (and bring the `null` onto it's own line)

Octokit.Tests/Reactive/ObservableDeploymentsClientTests.cs
```diff
@@ -25,43 +27,88 @@ public TheGetAllMethod()
...
+                options = new ApiOptions
+                {
+                    StartPage = 1
+                };
+
+                _client.GetAll(owner, name, options);
+                _githubClient.Connection
+                             .Received(1)
+                             .Get<List<Deployment>>(Arg.Is(expectedUri),
+                                                         Arg.Is<IDictionary<string, string>>(dictionary => dictionary.Count == 1), null);
```
shiftkey: :lipstick: indenting (and bring the `null` onto it's own line)

Octokit.Tests/Reactive/ObservableDeploymentsClientTests.cs
```diff
@@ -25,43 +27,88 @@ public TheGetAllMethod()
...
+                options = new ApiOptions
+                {
+                    PageCount = 1
+                };
+
+                _client.GetAll(owner, name, options);
+                _githubClient.Connection
+                             .Received(1)
+                             .Get<List<Deployment>>(Arg.Is(expectedUri),
+                                                         Arg.Is<IDictionary<string, string>>(dictionary => dictionary.Count == 0), null);
```
shiftkey: :lipstick: indenting (and bring the `null` onto it's own line)

Octokit.Tests/Clients/DeploymentsClientTests.cs
```diff
@@ -45,24 +50,44 @@ public void RequestsCorrectUrl()
...
+            client.GetAll(owner, name, options);
+            connection.Received(1)
+                .GetAll<Deployment>(Arg.Is<Uri>(u => u == expectedUrl), options);
         }
     }
 
     public class TheCreateMethod
     {
-        readonly NewDeployment newDeployment = new NewDeployment("aRef");
+        private readonly NewDeployment _newDeployment = new NewDeployment("aRef");
```
shiftkey: I'm not a fan of using `_` or `m_` prefixes for private fields like this. Could you change this back?

Octokit.Reactive/Clients/ObservableDeploymentsClient.cs
```diff
@@ -31,11 +31,28 @@ public ObservableDeploymentsClient(IGitHubClient client)
         /// <returns>All the <see cref="Deployment"/>s for the specified repository.</returns>
         public IObservable<Deployment> GetAll(string owner, string name)
         {
+            return GetAll(owner, name, ApiOptions.None);
```
shiftkey: Like we mentioned earlier, can you bring back the `Ensure` arguments for `owner` and `name` here?

Octokit.Tests/Clients/DeploymentsClientTests.cs
dampir: No problem. So I will not change that in future.

Octokit/Clients/DeploymentsClient.cs
dampir: Yep)

Octokit.Tests/Clients/DeploymentsClientTests.cs
dampir: Oh, I missed some places, sorry! Fixed.

Octokit.Tests/Reactive/ObservableDeploymentsClientTests.cs
dampir: Fixed (I guess :) )

dampir:
@shiftkey I've fixed all remarks, pls take a look.

ryangribble:
Looks OK to me and ive retried the travis build.  Will leave to @shiftkey to bring this one home since he had raised specific points

shiftkey:
@dampir LGTM

Issue #1194, opened 2016-03-16
[META] Cleaning up obsoleted items


ryangribble:
Just opening a discussion to review the items marked `[Obsolete]`, since most of them will have been more than 3 months old by the time of the next release (and many of them are 1 year+ old too)... it must be time to :rocket: :fire: some of them :grinning: 

We've currently got 80 occurences across 39 files, and by my (manual count) about 14 actual functional areas.

Here's a list of the areas and the number of days since they were obsoleted

```
IAuthorizationsClient.RevokeAllApplicationAuthentications() 43 days
IOrganizationMembersClient.GetAll(string org, string filter) 454 days 
IRepositoriesClient.CommitStatus 83 days
IRepositoryContentsClient.GetArchiveLink() 289 days
ISshKeysClient and associated SshKey class, helpers etc 29 days
ITeamsClient.IsMember() 299 days

GitHubClient.Release 82 days
GitHubClient.GitDatabase 82 days

ApiConnection.GetRedirect() 98 days
ApiUrls.Members() 454 days
IConnection.Get<T>(... allowAutoRedirect) 290 days

SearchIssuesRequest ctor(term, owner, name) 241 days

Repository.SubscribersCount 378 days
Repository.Organization 384 days
```

And here's the raw "Find in Files" output

```
  \Octokit\Clients\AuthorizationsClient.cs(384):        [Obsolete("This feature is no longer supported in the GitHub API and will be removed in a future release")]
  \Octokit\Clients\IAuthorizationsClient.cs(233):        [Obsolete("This feature is no longer supported in the GitHub API and will be removed in a future release")]
  \Octokit\Clients\IOrganizationMembersClient.cs(66):        [Obsolete("No longer supported, use GetAll(string, OrganizationMembersFilter) instead")]
  \Octokit\Clients\IRepositoriesClient.cs(198):        [Obsolete("Use Status instead")]
  \Octokit\Clients\IRepositoryContentsClient.cs(104):        [Obsolete("Use GetArchive to download the archive instead")]
  \Octokit\Clients\IRepositoryContentsClient.cs(127):        [Obsolete("Use GetArchive to download the archive instead")]
  \Octokit\Clients\IRepositoryContentsClient.cs(152):        [Obsolete("Use GetArchive to download the archive instead")]
  \Octokit\Clients\ISshKeysClient.cs(22):        [Obsolete("This method is obsolete. Please use User.Keys.Get(int) instead.")]
[...]

PR #1224, opened 2016-03-27
Cleanup obsolete items

M-Zuber:
per #1194 (specifically [this comment](https://github.com/octokit/octokit.net/issues/1194#issuecomment-199810285))

- one small item to draw attention to is [this question](https://github.com/octokit/octokit.net/issues/1194#issuecomment-202014873)

### DIFF
--- a/Octokit.Reactive/Clients/IObservableOrganizationMembersClient.cs
+++ b/Octokit.Reactive/Clients/IObservableOrganizationMembersClient.cs
@@ -51,15 +51,6 @@ namespace Octokit.Reactive
         IObservable<User> GetAll(string org, OrganizationMembersFilter filter);
 
         /// <summary>
-        /// Obsolete, <see cref="GetAll(string,OrganizationMembersFilter)"/>
-        /// </summary>
-        /// <param name="org">The login for the organization</param>
-        /// <param name="filter">The user filter</param>
-        /// <returns>The users</returns>
-        [Obsolete("No longer supported, use GetAll(string, OrganizationMembersFilter) instead")]
-        IObservable<User> GetAll(string org, string filter);
-
-        /// <summary>
         /// <para>
         /// List all users who are members of an organization. A member is a user that
         /// belongs to at least 1 team in the organization.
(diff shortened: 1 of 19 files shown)


ryangribble:
> one small item to draw attention to is this question

As per my reply on #1194 I suggest we should just :fire: the `ApiExtensions.GetRedirect()` method

M-Zuber:
@ryangribble done

ryangribble:
Oops, it looks like there is one more reference in `ApiConnection` and `IApiConnection` with `GetRedirect()`

The problem is, the concrete implementation was made obsolete 111 days ago, but the interface member hasnt been marked obsolete at all! :sob: 

But if we have to keep this, then it means we have to keep the `ApiExtension` one around as well, but that seems kinda pointless since the `GetRedirect()` will never actually get a HTTP 302 response anyway and thus will always throw an exception

@shiftkey what do you reckon to :fire:ing [`IApiConnection.GetRedirect()`](https://github.com/octokit/octokit.net/blob/master/Octokit/Http/IApiConnection.cs#L304), eventhough it wasnt marked obsolete?  Note that the concrete implementation, as well as multiple other places regarding Redirects and ArchiveLinks HAVE been obsolete for 100+ days

M-Zuber:
:sob: 
That's what I get for pushing my machines RAM too hard. The error in VS hadn't show up by the time the commit was pushed....

I'll wait for @shiftkey's response (may have to wait until after #build :smile: ) and follow his recommendation

shiftkey:
> what do you reckon to :fire:ing IApiConnection.GetRedirect(), eventhough it wasnt marked obsolete? Note that the concrete implementation, as well as multiple other places regarding Redirects and ArchiveLinks HAVE been obsolete for 100+ days

I'd be breaking my own rule here, and covering over an oversight of mine. Let's add the `[Obsolete]` to the interface and leave it alone for this release.

ryangribble:
so @M-Zuber just chasing up on this one, I take it you are good to rebase to latest master and also tackle the redirect methods according to @shiftkey's comments

M-Zuber:
I'm good to do the work, but on the road ATM. 
Will attempt to get to it in the next day or two

M-Zuber:
@shiftkey @ryangribble rebased and updated code with regards to `.GetRedirect`.
Would love if I could get some :eyes: on the changes to make sure that the changes are correct.

shiftkey:
This all seems good to me. @ryangribble anything else to add here?

ryangribble:
All good!  I love that feeling of cleaning things up :grinning:

ryangribble:
LGTM

PR #1225, opened 2016-03-28
Some misspells in ApiUrls and ApiConnection were fixed.

dampir:
I have fixed some misspells in ApiUrls.cs and ApiConnection.cs to make code more clean.


shiftkey:
@dampir excellent pick up! 👏

PR #1226, opened 2016-03-29
Some ArgumentNotNullOrEmptyString cheсks were added in ObservableAssigneesClient

dampir:
After the discussion https://github.com/octokit/octokit.net/pull/1218/commits/42c3a2b728a30def1e8e4385804eeb77b8162c32,  I checked usage of ApiOptions and add necessary null checking.

/cc @shiftkey


ryangribble:
Looks good, we also normally have 2 unit test methods, to assert these (non null and non empty)  checks, have a look at other examples (sorry I'd link you one but on my phone 😀)

dampir:
@ryangribble I've added new ObservableAssigneesClientTests. Pls, take a look.

ryangribble:
:+1: :+1: :+1: :+1: :+1:

PR #1229, opened 2016-03-30
Fix Ensure methods parameters

dampir:
I found and fixed some issues when Ensure.ArgumentNotNull and Ensure.ArgumentNotNullOrEmptyString were used.

/cc @shiftkey, @ryangribble


ryangribble:
Nice work! :+1:

PR #1230, opened 2016-03-30
Fix some misspellings in code, comments and xml docs.

dampir:
I've found and fixed some misspellings inside Octokit solution. There are a lot of changes, but they all very simple, one or two letter have been changed only. Hope this PR would be useful.

/cc @ryangribble, @shiftkey

### DIFF
--- a/Octokit.Tests.Integration/Clients/RepositoryForksClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/RepositoryForksClientTests.cs
@@ -40,7 +40,7 @@ namespace Octokit.Tests.Integration.Clients
             [IntegrationTest]
             public async Task ForkCreatedForUserLoggedIn()
             {
-                // The fork is created asynchronically by github and therefore it cannot 
+                // The fork is created asynchronously by github and therefore it cannot 
                 // be certain that the repo exists when the test ends. It is therefore deleted
                 // before the test starts instead of after.
                 Helper.DeleteRepo(Helper.Credentials.Login, "octokit.net");
@@ -57,7 +57,7 @@ namespace Octokit.Tests.Integration.Clients
             [OrganizationTest]
             public async Task ForkCreatedForOrganization()
             {
-                // The fork is created asynchronically by github and therefore it cannot 
+                // The fork is created asynchronously by github and therefore it cannot 
                 // be certain that the repo exists when the test ends. It is therefore deleted
                 // before the test starts.
                 Helper.DeleteRepo(Helper.Organization, "octokit.net");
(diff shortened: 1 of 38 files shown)


Octokit.Tests.Integration/Clients/RepositoryForksClientTests.cs
```diff
@@ -41,7 +40,7 @@ public class TheCreateMethod
             [IntegrationTest]
             public async Task ForkCreatedForUserLoggedIn()
             {
-                // The fork is created asynchronially by github and therefore it cannot 
+                // The fork is created asynchronically by github and therefore it cannot
```
shiftkey: `asynchronously`

Octokit.Tests.Integration/Clients/RepositoryForksClientTests.cs
```diff
@@ -58,7 +57,7 @@ public class TheCreateMethod
             [OrganizationTest]
             public async Task ForkCreatedForOrganization()
             {
-                // The fork is created asynchronially by github and therefore it cannot 
+                // The fork is created asynchronically by github and therefore it cannot
```
shiftkey: `asynchronously`

Octokit/Models/Request/SearchRepositoriesRequest.cs
```diff
@@ -479,9 +479,9 @@ public enum Language
         [SuppressMessage("Microsoft.Naming", "CA1704:IdentifiersShouldBeSpelledCorrectly", MessageId = "Edn")]
         Edn,
         Eiffel,
-        [SuppressMessage("Microsoft.Naming", "CA1704:IdentifiersShouldBeSpelledCorrectly", MessageId = "Elixer")]
-        [SuppressMessage("Microsoft.Naming", "CA1704:IdentifiersShouldBeSpelledCorrectly", MessageId = "Elixer")]
-        Elixer,
+        [SuppressMessage("Microsoft.Naming", "CA1704:IdentifiersShouldBeSpelledCorrectly", MessageId = "Elixir")]
+        [SuppressMessage("Microsoft.Naming", "CA1704:IdentifiersShouldBeSpelledCorrectly", MessageId = "Elixir")]
+        Elixir,
```
shiftkey: These are going to be breaking changes for the unfortunate few which might be trying to use these - we need to leave the old ones around and mark them as obsolete

dampir: I have one remarks about it. Yes, it is breaking changes, but this fully incorrect items (**Elixer**, Unified **Paralel ** C), these items never works correctly and if somebody use these then he doesn't  get any search results. Anyway, I reverted older items back and added new correct items. Hope they will be removed in further release.

shiftkey:
Just a heads up: #1238 has a fix for the Travis issue - feel free to cherry-pick 8ec4bf388c674c1ce5f09f8c8ad88a470ce7f4d9 so you can verify this PR is still correct

dampir:
@shiftkey, all checks, pls, check this out whenever you want :)

shiftkey:
@dampir thanks!

PR #1235, opened 2016-04-03
Add "importers" field to MiscellaneousClient.Meta response

### LINKED ISSUE #1234: New property on Meta endpoint response
While poking around the [API docs](https://developer.github.com/v3/meta/) I found that the `Meta` endpoint under `MiscellaneousClient` has an additional property (`importer`) that is not yet implemented in octokit.net

It also looks like the `GetMetadata` method is not covered by unit tests, and the integration test only asserts 1 of the 6 returned properties

ryangribble:
Adds new field "Importers" to the Metadata request of MiscellaneousClient

API docs: https://developer.github.com/v3/meta/

Fixes #1234 

- [x] Added `Importers` field to `Meta` response model and updated `ctor`

- [x] Added unit test for `GetMetadata` method

- [x] Enhanced integration tests for `GetMetadata` method
 - Revealed a bug with an existing field `GitHubServicesSha` not being deserialised properly, added a fix to this PR

- [x] Added unit tests for `ObservableMiscellaneousClient` for all methods
No tests for this observable client existed, so all methods were implemented

- [x] Fixed `ObservableMiscellaneousClient` constructor to be inline with all other obersvable API clients
(Marked old constructor as `[Obsolete]` so it can be removed in a later release)

### DIFF
--- a/Octokit.Reactive/Clients/ObservableMiscellaneousClient.cs
+++ b/Octokit.Reactive/Clients/ObservableMiscellaneousClient.cs
@@ -9,7 +9,7 @@ namespace Octokit.Reactive
     {
         readonly IMiscellaneousClient _client;
 
-        [Obsolete("Please use another constructor")]
+        [Obsolete("Please use the IGitHubClient overload constructor")]
         public ObservableMiscellaneousClient(IMiscellaneousClient client)
         {
             Ensure.ArgumentNotNull(client, "client");
--- a/Octokit.Reactive/Clients/ObservableMiscellaneousClient.cs
+++ b/Octokit.Reactive/Clients/ObservableMiscellaneousClient.cs
@@ -9,11 +9,19 @@ namespace Octokit.Reactive
     {
         readonly IMiscellaneousClient _client;
 
+        [Obsolete("Please use another constructor")]
         public ObservableMiscellaneousClient(IMiscellaneousClient client)
         {
             Ensure.ArgumentNotNull(client, "client");
 
             _client = client;
+        }
+
+        public ObservableMiscellaneousClient(IGitHubClient client)
+        {
+            Ensure.ArgumentNotNull(client, "client");
+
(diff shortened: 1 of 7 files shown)


shiftkey:
Just a heads up: #1238 has a fix for the Travis issue - feel free to cherry-pick 8ec4bf388c674c1ce5f09f8c8ad88a470ce7f4d9 so you can verify this PR is still correct

Octokit.Reactive/Clients/ObservableMiscellaneousClient.cs
```diff
@@ -9,6 +9,7 @@ public class ObservableMiscellaneousClient : IObservableMiscellaneousClient
     {
         readonly IMiscellaneousClient _client;
 
+        [Obsolete("Please use another constructor")]
```
shiftkey: Let's make this really specific - what about something like: `[Obsolete("Please use the IGitHubClient overload ctor")]`

shiftkey:
> Revealed a bug with an existing field `GitHubServicesSha` not being deserialised properly, added a fix to this PR

It's weird that this is failing - the UpperCamelCase to upper_camel_case looks like it should work here but perhaps I'm missing something. Anyway, happy to take this in here...

shiftkey:
@ryangribble this all seems great :sparkles: just the little message which is really, really minor

ryangribble:
> It's weird that this is failing - the UpperCamelCase to upper_camel_case looks like it should work here but perhaps I'm missing something. Anyway, happy to take this in here...

I think the capital H in "GitHub" was causing it to be looking for `git_hub_services_sha` rather than `github_services_sha`

But yeah in any case, it was `null` before this change, but is now populated correctly.  And it's a greeat example of why integration tests should check all fields and not just a couple!

shiftkey:
> I think the capital H in "GitHub" was causing it

Ah yes, that makes sense :skull:

Issue #1231, opened 2016-03-31
Provide uniform name for constructors test classes


dampir:
I found some inconsistency in naming of classes that tests client constructors. So,
30 tests uses "TheConstructor" class name for test client constructors.
37 tests uses "TheCtor" class name for test client constructors.
1 test use TheCtorMethod" class name for test client constructor in [ObservableCommitsClientTests.cs](https://github.com/octokit/octokit.net/blob/aa133286459612c32351cde75dca399b84e093ef/Octokit.Tests/Reactive/ObservableCommitsClientTests.cs).

In order to eliminate inconsistency in these cases, I propose put names in order and rename all such classes to "TheCtor" (as most popular class name at this moment).

devkhan:
Can we add inconsistencies like this to be resolved in FormatCode command? Automating this will remove any inconsistencies introduced by humans. Is something like this possible?

dampir:
@devkhan Guess after we choose name convention for these classes, we can add additional checks in Conventions test rather in FormatCode. Anyway, it would be very cool have such additional check.

M-Zuber:
Checks like this are in theory possible, _but_ would require a lot of hard coding (not that that is a bad thing, just something to keep in mind).

And I second @dampir's suggestion that they go into `Conventions` and not `FormatCode`

ryangribble:
Yep considering we aren't "making" people run the format code step, nor is there an easy way to enforce that, adding things to convention tests would be the preferred option

shiftkey:
> Can we add inconsistencies like this to be resolved in FormatCode command?

There's no real project-specific rules defined in [`Octokit.CodeFormatter`](https://github.com/shiftkey/octokit.codeformatter) and writing Roslyn analyzers is a bit tricky if you're not familiar with the concepts.

> Guess after we choose name convention for these classes, we can add additional checks in Conventions test rather in FormatCode.

This feels a bit easier to add and will be caught earlier.

shiftkey:
Closing this after the work done in #1246. Please open a new issue if there are further ideas about this.

PR #1246, opened 2016-04-06
Add check of client constructor test class name

dampir:
Refer #1231

In order to provide uniform name for constructors test classes, I've created new convention test checks  
that each client test class must contain a nested TheCtor class.

- [x] Add new convention test.
- [x] Fix all red tests (Rename/Add needed tests classes)
- [x] Remove unused usings in modified classes.
- [x] Find way to detect tests assembly in proper way.
- [x] Each test class TheCtor should have method EnsuresNonNullArguments that test a client constructor against null arguments.
- [x] All unused "using" directives in modified classes were removed in order to get code more robust.

Ready to any discussion about implementation of convention test ClientConstructorTests.cs, especially about GetTestConstructorsClasses method. 

/cc @shiftkey, @ryangribble

### DIFF
--- a/Octokit.Tests/GitHubClientTests.cs
+++ b/Octokit.Tests/GitHubClientTests.cs
@@ -1,17 +1,15 @@
 ﻿using System;
-using System.Net.Http.Headers;
 using System.Threading.Tasks;
 using NSubstitute;
 using Octokit.Internal;
 using Xunit;
-using Xunit.Extensions;
 using System.Collections.Generic;
 
 namespace Octokit.Tests
 {
     public class GitHubClientTests
     {
-        public class TheConstructor
+        public class TheCtor
         {
             [Fact]
             public void CreatesAnonymousClientByDefault()
--- a/Octokit.Tests/GitHubClientTests.cs
+++ b/Octokit.Tests/GitHubClientTests.cs
@@ -17,26 +17,26 @@ namespace Octokit.Tests
                 Assert.Throws<ArgumentNullException>(() => new GitHubClient((IConnection)null));
                 Assert.Throws<ArgumentNullException>(() => new GitHubClient((ProductHeaderValue)null));
 
-                var productHeaderValue = new ProductHeaderValue("UnitTest");
+                var productInformation = new ProductHeaderValue("UnitTest");
                 var baseAddress = new Uri("http://github.com");
                 var credentialStore = Substitute.For<ICredentialStore>();
 
-                Assert.Throws<ArgumentNullException>(() => new GitHubClient(productHeaderValue, (ICredentialStore)null));
+                Assert.Throws<ArgumentNullException>(() => new GitHubClient(productInformation, (ICredentialStore)null));
                 Assert.Throws<ArgumentNullException>(() => new GitHubClient(null, credentialStore));
                 
-                Assert.Throws<ArgumentNullException>(() => new GitHubClient(productHeaderValue, (Uri)null));
+                Assert.Throws<ArgumentNullException>(() => new GitHubClient(productInformation, (Uri)null));
                 Assert.Throws<ArgumentNullException>(() => new GitHubClient(null, baseAddress));
 
                 Assert.Throws<ArgumentNullException>(() => new GitHubClient(null, (ICredentialStore)null));
                 Assert.Throws<ArgumentNullException>(() => new GitHubClient(null, (Uri)null));
 
-                Assert.Throws<ArgumentNullException>(() => new GitHubClient(productHeaderValue, null, null));
+                Assert.Throws<ArgumentNullException>(() => new GitHubClient(productInformation, null, null));
                 Assert.Throws<ArgumentNullException>(() => new GitHubClient(null, credentialStore, null));
                 Assert.Throws<ArgumentNullException>(() => new GitHubClient(null, null, baseAddress));
                 
                 Assert.Throws<ArgumentNullException>(() => new GitHubClient(null, credentialStore, baseAddress));
-                Assert.Throws<ArgumentNullException>(() => new GitHubClient(productHeaderValue, null, baseAddress));
-                Assert.Throws<ArgumentNullException>(() => new GitHubClient(productHeaderValue, credentialStore, null));
+                Assert.Throws<ArgumentNullException>(() => new GitHubClient(productInformation, null, baseAddress));
+                Assert.Throws<ArgumentNullException>(() => new GitHubClient(productInformation, credentialStore, null));
             }
 
             [Fact]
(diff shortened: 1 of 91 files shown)


ryangribble:
From my perspective this is :gem:! 

I've checked all the added
`Ctor` tests are `new`ing up the correct client :+1: very thorough and consistent! 

My only comment is the way you locate the assembly by using `EventsClientTests` type. There's nothing special/significant about that type so I'd rather some alternative way of finding the assembly either by name or a more common/top level class or something like that

dampir:
@ryangribble, thanks, I am very glad :+1: 

I am agree, the way I've located the assembly is smell a little bit :smirk:. I found that way here: https://github.com/octokit/octokit.net/blob/f354d1bf00ff7606b46489e3a915e1428414bc47/Octokit.Tests.Conventions/SyncObservableClients.cs#L121

I'll think a little bit and will try to deliver more common way.

ryangribble:
> I'll think a little bit and will try to deliver more common way. 

I wasn't aware this is how some other convention tests are doing it, they were before my involvement.  To me it would "seem" nicer to locate the assembly in a less specific way but maybe I'm over thinking it :grinning:

if you think of a nicer way let us know, otherwise from the perspective that this is already being done elsewhere, I won't bikeshed over this!

dampir:
Yeah, I'm agree about GitHubClientTests, it is looks very good for this purpose :+1: . Fixed!

ryangribble:
Cool! I wonder whether we should change all the other Convention tests over to using those top level types too (perhaps in another PR, or perhaps on this one if it is reworded to "Add convention test for Ctor constructor and tidy up other convention tests"...

dampir:
> I wonder whether we should change all the other Convention tests over to using those top level types too instead of EventsClient/ObservableEventsClient etc (perhaps in another PR, or perhaps on this one if it is reworded to "Add convention test for Ctor constructor and tidy up other convention tests"...

I've marked it in my notes "Tidy up conv. tests", so it would be my next PR :+1: 
By the way, I've found some incosistency in method's name that test constructor, I going to add additional check for that, if nobody minds :smile:

shiftkey:
> (what's so special about events client since it seems to be used EVERYWHERE in the convention tests @shiftkey ?? :grinning:)

I cannot recall why - but it goes way, way back to 2375a16aa05e0bb7211d089b43b9ced1a97af45d. :thumbsup: to using `IGitHubClient`

shiftkey:
> I've marked it in my notes "Tidy up conv. tests", so it would be my next PR 

:thumbsup: to addressing our other usages after this PR lands

dampir:
In order to deliver more consistency I've added new convention check. Now each test class "TheCtor" should have method EnsuresNonNullArguments that test a client constructor against null arguments. Please take a look.

Octokit.Tests/GitHubClientTests.cs
```diff
@@ -1,19 +1,41 @@
...
-        public class TheConstructor
+        public class TheCtor
         {
             [Fact]
+            public void EnsuresNonNullArguments()
+            {
+                Assert.Throws<ArgumentNullException>(() => new GitHubClient((IConnection)null));
+                Assert.Throws<ArgumentNullException>(() => new GitHubClient((ProductHeaderValue)null));
+
+                Assert.Throws<ArgumentNullException>(() => new GitHubClient(new ProductHeaderValue("UnitTest"), (ICredentialStore)null));
```
shiftkey: Do you think we could simplify these rather than creating new values everywhere they're used?

e.g.

```csharp
var header = new ProductHeaderValue("UnitTest");

Assert.Throws<ArgumentNullException>(() => new GitHubClient(header, (ICredentialStore)null));
```

This might help with the readability of these tests...

shiftkey:
Aside from the merge conflict and the suggestion above, this is ready to merge! :metal:

Octokit.Tests/GitHubClientTests.cs
dampir: I agree :+1:  Fixed.

dampir:
@shiftkey what I can do in order to fix merge conflict?

devkhan:
@dampir Maybe rebasing can help, try this:

```powershell
git checkout ctor-conventions 
git rebase upstream/master 
## Resolve any merge conflicts that arise, mark them resolved with:
git add . 
git rebase --continue
## Repeat above two steps until done.
```

To be on the safe side, try this on a temp branch forked from the main first.

dampir:
@devkhan , great thanks, rebase passed.

devkhan:
@dampir This looks odd as it contains commits from other PRs also. Did you update master from the main repo before rebasing?

dampir:
@devkhan , yes, I did. It is error?

devkhan:
I don't completely understand what happened here, as this PR contains commits from other PRs also. But as there are no conflicts, maybe it will auto-resolve once merged. I think someone with experience can help out, @ryangribble @shiftkey ?

P. S.: You should look into the failing builds so it can be merged quickly.

shiftkey:
@devkhan @dampir 

> I don't completely understand what happened here, as this PR contains commits from other PRs also. 

I think this is due to me squashing the PRs yesterday when doing the merge. This wasn't intentional, however I do like how it's simplified the history...

shiftkey:
@devkhan I think @dampir needs to update `upstream/master` or ensure they're using the right remote for the rebase - my psychic debugger thinks that an earlier `master` was used here, which would explain the commits we're seeing here...

dampir:
Unfitunately, I can't say what I did exactly, because it was my first "git rebase" at all. Now I am not so sure that I did right "git pull" from upstream before rebase. Can I do something that can help fix this situation?

devkhan:
> There were some errors after rebase (not about merge, but about push), so I searched errors over internet and tried to fix that.

That maybe because you need to `--force` a push after a rebase (as you have effectively changed the history).

dampir:
@devkhan , I going to use --force next time, thanks!

dampir:
@shiftkey I've revert rebase and repeat it again today. Please, check this out whenever possible :smile:

Octokit.Tests/GitHubClientTests.cs
shiftkey: Did this change get lost?

Octokit/Clients/RepositoryCommitsClient.cs
```diff
@@ -9,11 +9,12 @@ namespace Octokit
     /// <remarks>
     /// See the <a href="http://developer.github.com/v3/repos/commits/">Repository Commits API documentation</a> for more information.
     /// </remarks>
-    public class RepositoryCommitsClient : IRepositoryCommitsClient
+    public class RepositoryCommitsClient : ApiClient, IRepositoryCommitsClient
```
shiftkey: Nice spot! :eyes:

shiftkey:
@dampir just that one comment about the cleanup and this is good to go. I'll hold off on other merges so that we can get this in!

Octokit.Tests/GitHubClientTests.cs
dampir: Unfortunately, yes! Fixed :+1:

dampir:
@shiftkey thank you, all remarks were fixed. Ready to merge, I guess :+1:

shiftkey:
@dampir yes, just waiting on CI to pass

Issue #1111, opened 2016-02-15
Trying to get content fail 


mathieukempe:
Using 
https://developer.github.com/v3/repos/contents/

This used to work with 0.10.0

<pre>
var api = new ApiConnection(new Connection(new ProductHeaderValue("something"), new InMemoryCredentialStore(new Credentials(_token))));

 string url = string.Format("https://api.github.com/repos/{0}/{1}/contents/{2}?ref={3}", owner, repo, path, branch);
            
 var blob = api.Get<Octokit.Blob>(new Uri(url)).Result;
</pre>

Using the same call with curl I can retrieve the content of file but using octokit 0.18.0 api.Get<Octokit.Blob>(new Uri(url)).Result just hang forever.

shiftkey:
@mathieukempe not sure whether this is an outstanding issue, but here's how I'd do this without using `ApiConnection` directly (this is using 0.18.0):

```csharp
var client = new GitHubClient("something");
client.Credentials = new Credentials(_token);
var contents = await client.Repository.Content.GetAllContents(owner, repo, path, branch)
var file = contents[0].Content;
```

mathieukempe:
@shiftkey thanks for the example of code
The issue seems to be when I call the code in the controller asp.net mvc 5

```c#
    public class HomeController : Controller
    {
        public ActionResult Index()
        {
            var client = new GitHubClient(new ProductHeaderValue("something"), new InMemoryCredentialStore(new Credentials("my api key")));

            var content = client.Repository.Content.GetAllContents("Selz", "drachma",
                "src/Drachma.Web/Resources/Buttons.resx").Result;



            return View();
        }

```
The call to GetAllContents hang
The same code in a console app work.

shiftkey:
@mathieukempe to be honest, I'm not sure how this might be deadlocking.

We have the `ConfigureAwait` setup [inside GetAll](https://github.com/octokit/octokit.net/blob/master/Octokit/Http/ApiConnection.cs#L147-L148) so perhaps making this method `async/await` might change the behaviour (rather than blocking using `.Result`).

zzzprojects:
There is a lot of ConfigureAwait(false) missing in all the projects.

By example, the method "client.Repository.Content.GetAllContents(...)" didn't work in my WinForm application because of UI Blocking.

I just fixed RedirectHandler.SendAsync to add ConfigureAwait to all methods and it worked.

shiftkey:
[...]

PR #1248, opened 2016-04-07
Always Be Auditing Your Awaits

shiftkey:
As discussed in #1111, there are some places where we use `await` but do not specify the context correctly.

This is me scanning the codebase and doing three things with each scenario I find:

 - if possible, drop the `async/await` usage completely
 - if not possible to remove, ensure `ConfigureAwait` is set
 - ensure that the `await` and `ConfigureAwait` are on the same line - to make finding and verifying these usages easier in the future

 - [x] integration tests passing

cc @ryangribble as I found myself refactoring in some of the Enterprise API clients

### DIFF
--- a/Octokit/Clients/AssigneesClient.cs
+++ b/Octokit/Clients/AssigneesClient.cs
@@ -66,8 +66,7 @@ namespace Octokit
 
             try
             {
-                var response = await Connection.Get<object>(ApiUrls.CheckAssignee(owner, name, assignee), null, null)
-                                               .ConfigureAwait(false);
+                var response = await Connection.Get<object>(ApiUrls.CheckAssignee(owner, name, assignee), null, null).ConfigureAwait(false);
                 return response.HttpResponse.IsTrue();
             }
             catch (NotFoundException)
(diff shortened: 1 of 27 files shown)


shiftkey:
For reference, `ConfigureAwait(false)` is what libraries such as Octokit need to do when doing async/await internally, and we know we don't need the "context". You can read more about this stuff from [this MSDN link](https://msdn.microsoft.com/en-us/magazine/jj991977.aspx)

ryangribble:
All looks good :+1: 

> as I found myself refactoring in some of the Enterprise API clients

Not sure what you were referring to, I only saw the mentinoed `async`/`await`/`ConfigureAwait(false)` changes

shiftkey:
@ryangribble I just wanted to call out that I was able to drop some usages of `async` / `await` and there might have been a reason to keep them around...

ryangribble:
ah cool, there was no reason to them I think at the time I didnt really realise that if you are just `await`ing the call and returning it anyway, yuo can just remove the `async` and `await` and simply return the `awaitable` call.  There's only the odd few cases where we do something with the response before returning.

All changes on this PR look :+1: to me

shiftkey:
release_notes: added `ConfigureAwait(false)` to numerous locations to prevent deadlock scenarios for ASP.NET and WinForms users

PR #1250, opened 2016-04-07
Fix a broken link

SamTheDev:
Fix the Testing Documentation's guide link after merging/deleting the rewrite-contributing branch.


ryangribble:
Nice pick up on the broken link, although your change is still linking to shiftkey/octokit.net fork rather than octokit/octokit.net ??

PR #1247, opened 2016-04-06
Added ApiOption overloads to methods on IRepositoryCommitsClient

prayankmathur:
Refer to issue #1173

- [x] Add overloads for `Task<IReadOnlyList<GitHubCommit>> GetAll(String owner, String name)` (reactive too)
- [x] Add overloads for `Task<IReadOnlyList<GitHubCommit>> GetAll(String owner, String name, CommitRequest request)`
- [x] Add unit test
- [x] Add integration tests.

//cc @shiftkey @ryangribble

Issue #1208, opened 2016-03-22
Lowercase namespace


M-Zuber:
At some point a folder called `fixtures` was introduced.
This lead to a number of classes having the namespace`Octokit.Tests.Integration.fixtures`

Will make a PR to fix once #1144 is merged (as it introduces some more files to that namespace)

([Affected Files - potentially more](https://github.com/octokit/octokit.net/search?utf8=%E2%9C%93&q=namespace+Octokit.Tests.Integration.fixtures))

Haacked:
I would love to remove all namespaces from unit and integration tests. They really don't serve a purpose for our tests. Namespaces are useful for libraries that you reference in order to disambiguate class names. But you never reference a set of unit tests for your own code. At least, we don't need to support that if you do for some wild reason. :stuck_out_tongue:

M-Zuber:
Would you be open to one fat PR that rips out all namespace parts lower than what matches the project? 
Or shall I split it up?

Haacked:
Split it up. I worry that one big fat PR will cause problems for everybody else. :)

shiftkey:
Just a heads up - removing the namespaces is likely to cause a whole bunch of indenting changes to the test suite. It's a minor nuisance but things like `git blame` will then be full of lies...

M-Zuber:
What do you mean by `indenting changes`?

M-Zuber:
An example of what things look like now is
![image](https://cloud.githubusercontent.com/assets/4746537/14064333/f1fd22b0-f407-11e5-849b-4b315e86fe3b.png)
@Haacked do you want to just drop the `.Clients` or drop the namespace altogether (like :arrow_down: )?
![image](https://cloud.githubusercontent.com/assets/4746537/14064338/16c428c8-f408-11e5-80ad-9bd2b3ac5549.png)

shiftkey:
@M-Zuber that's exactly the issue, and why I kinda put it aside when I started finding these 😢

M-Zuber:
IMO I would do the first option. having a namespace that matches the project name does not create noise (at least to me) - and that is what is included in the default.

Is there a setting that prevents VS from tacking on namespace parts to match the file system?

shiftkey:
> Is there a setting that prevents VS from tacking on namespace parts to match the file system?

[Unfortunately it seems to be R# only](http://stackoverflow.com/a/6246679/1363815)

M-Zuber:
[...]

PR #1261, opened 2016-04-10
Add ApiOption overloads to methods on IGistsClient

### LINKED ISSUE #1156: Add ApiOption overloads to methods on IGistsClient

- Add an overload for this method on `IGistsClient` with a new parameter `ApiOptions`
- Implement the new method on the class which implements `IGistsClient`
- Add an overload for this method on `IObservableGistsClient` with a new parameter `ApiOptions`
- Implement the new method on the class which implements `IObservableGistsClient`

- [ ] `Task<IReadOnlyList<Gist>> GetAll()`
- [ ] `Task<IReadOnlyList<Gist>> GetAll(DateTimeOffset since)`
- [ ] `Task<IReadOnlyList<GistHistory>> GetAllCommits(String id)`
- [ ] `Task<IReadOnlyList<GistFork>> GetAllForks(String id)`
- [ ] `Task<IReadOnlyList<Gist>> GetAllForUser(String user)`
- [ ] `Task<IReadOnlyList<Gist>> GetAllForUser(String user, DateTimeOffset since)`
- [ ] `Task<IReadOnlyList<Gist>> GetAllPublic()`
- [ ] `Task<IReadOnlyList<Gist>> GetAllPublic(DateTimeOffset since)`
- [ ] `Task<IReadOnlyList<Gist>> GetAllStarred()`
- [ ] `Task<IReadOnlyList<Gist>> GetAllStarred(DateTimeOffset since)`

SamTheDev:
Fixes #1156 

- [x]  Add overloads for the GetAll* methods in the IGistsClient with the the ApiOptions parameter.
- [x] Implement the the GetAll* methods in the GistsClient class.
- [x] Add an overload for the GetAll* methods to the IObservableGistsClient  with the ApiOptions parameter.
- [x] Implement the GetAll* methods in the ObservableGistsClient  class.
- [x] Implement the needed tests in the GistsClientTests class.
- [x] Implement the required integration tests in the ObservableGistsClientTests class.

### DIFF
--- a/Octokit.Tests.Integration/Reactive/ObservableGistClientTests.cs
+++ b/Octokit.Tests.Integration/Reactive/ObservableGistClientTests.cs
@@ -285,5 +285,146 @@ namespace Octokit.Tests.Integration.Reactive
                 Assert.NotEqual(firstPublicGistsPage[3].Id, secondPublicGistsPage[3].Id);
             }
         }
+
+        public class TheGetAllStarredMethod
+        {
+            readonly ObservableGistsClient _gistsClient;
+
+            public TheGetAllStarredMethod()
+            {
+                var github = Helper.GetAuthenticatedClient();
+
+                _gistsClient = new ObservableGistsClient(github);
+            }
+
+            [IntegrationTest]
+            public async Task ReturnsStartedGists()
+            {
+                var gists = await _gistsClient.GetAllStarred().ToList();
+
+                Assert.NotEmpty(gists);
+            }
+
+            [IntegrationTest]
+            public async Task ReturnsCorrectCountOfStartedGistsWithoutStart()
+            {
+                var options = new ApiOptions
+                {
+                    PageSize = 5,
+                    PageCount = 1
+                };
+
+                var gists = await _gistsClient.GetAllStarred(options).ToList();
+
+                Assert.Equal(5, gists.Count);
+            }
+
+            [IntegrationTest]
+            public async Task ReturnsCorrectCountOfStartedGistsWithStart()
+            {
+                var options = new ApiOptions
+                {
+                    PageSize = 5,
+                    PageCount = 1,
+                    StartPage = 2
+                };
+
+                var gists = await _gistsClient.GetAllStarred(options).ToList();
+
+                Assert.Equal(5, gists.Count);
+            }
+
+            [IntegrationTest]
+            public async Task ReturnsDistinctResultsBasedOnStartPage()
+            {
+                var startOptions = new ApiOptions
+                {
+                    PageSize = 5,
+                    PageCount = 1
+                };
+
+                var firstStartedGistsPage = await _gistsClient.GetAllStarred(startOptions).ToList();
+
+                var skipStartOptions = new ApiOptions
+                {
+                    PageSize = 5,
+                    PageCount = 1,
+                    StartPage = 2
+                };
+
+                var secondStartedGistsPage = await _gistsClient.GetAllStarred(skipStartOptions).ToList();
+
+                Assert.NotEqual(firstStartedGistsPage[0].Id, secondStartedGistsPage[0].Id);
+                Assert.NotEqual(firstStartedGistsPage[1].Id, secondStartedGistsPage[1].Id);
+                Assert.NotEqual(firstStartedGistsPage[2].Id, secondStartedGistsPage[2].Id);
+                Assert.NotEqual(firstStartedGistsPage[3].Id, secondStartedGistsPage[3].Id);
+                Assert.NotEqual(firstStartedGistsPage[4].Id, secondStartedGistsPage[4].Id);
+            }
+
+            [IntegrationTest]
+            public async Task ReturnsStartedGistsSince()
+            {
+                var since = new DateTimeOffset(new DateTime(2016, 1, 1));
+                var gists = await _gistsClient.GetAllStarred(since).ToList();
+
+                Assert.NotEmpty(gists);
+            }
+
+            [IntegrationTest]
+            public async Task ReturnsCorrectCountOfStartedGistsSinceWithoutStart()
+            {
+                var options = new ApiOptions
+                {
+                    PageSize = 5,
+                    PageCount = 1
+                };
+                var since = new DateTimeOffset(new DateTime(2016, 1, 1));
+                var gists = await _gistsClient.GetAllStarred(since, options).ToList();
+
+                Assert.Equal(5, gists.Count);
+            }
+
+            [IntegrationTest]
+            public async Task ReturnsCorrectCountOfStartedGistsSinceWithStart()
+            {
+                var options = new ApiOptions
+                {
+                    PageSize = 5,
+                    PageCount = 1,
+                    StartPage = 2
+                };
+                var since = new DateTimeOffset(new DateTime(2016, 1, 1));
+                var gists = await _gistsClient.GetAllStarred(since, options).ToList();
+
+                Assert.Equal(5, gists.Count);
+            }
+
+            [IntegrationTest]
+            public async Task ReturnsDistinctStartedGistsSinceBasedOnStartPage()
+            {
+                var startOptions = new ApiOptions
+                {
+                    PageSize = 5,
+                    PageCount = 1
+                };
+                var since = new DateTimeOffset(new DateTime(2016, 1, 1));
+                var firstStartedGistsPage = await _gistsClient.GetAllStarred(since, startOptions).ToList();
+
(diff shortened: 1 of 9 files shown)


SamTheDev:
In order to pass the `since ` `DateTimeOffset ` parameter as long with the `ApiOptions` parameter to some of the `GetAll*` methods, An overload to the `GetAndFlattenAllPages `method that accept both a `ParametersDictionary `and An `ApiOptions`seems to be missing !
should i add it as an extension method to the `ConnectionExtensions `class ?
or do something like this ? :
```
 public IObservable<Gist> GetAll(DateTimeOffset since, ApiOptions options)
        {

            var request = new GistRequest(since);
            return _connection.GetAndFlattenAllPages<Gist>(ApiUrls.Gist(), Pagination.Setup(request.ToParametersDictionary(), options));
        }
```

shiftkey:
@SamTheDev your needed `GetAndFlattenAllPages` changes seems like it'll overlap with what's been done in #1247 (in particular https://github.com/octokit/octokit.net/pull/1247#discussion_r58698529). 

We're close to getting that sorted, so if you can hold off on that until we merge you should be able to pull in those changes...

SamTheDev:
@shiftkey Ok, got it (y)

SamTheDev:
@shiftkey conflicts resolved, ready for review/merge :}

Octokit.Tests/Clients/GistsClientTests.cs
```diff
@@ -1,321 +1,573 @@
...
-using Octokit;
 using Octokit.Internal;
 using System;
 using System.Collections.Generic;
 using System.Net;
 using System.Threading.Tasks;
 using Xunit;
 
-public class GistsClientTests
+namespace Octokit.Tests.Clients
```
shiftkey: We're having an ongoing discussion about dropping namespaces from all our test classes - https://github.com/octokit/octokit.net/issues/1208 - it kind of goes against the standards for .NET but we think it has some benefits in places like the test runner output.

Are you able to drop this namespace and re-indent this file? It should simplify this diff a bit as well...

Octokit.Tests.Integration/Reactive/ObservableGistClientTests.cs
```diff
@@ -0,0 +1,720 @@
+﻿using System;
+using System.Reactive.Linq;
+using System.Threading.Tasks;
+using Octokit.Reactive;
+using Xunit;
+
+namespace Octokit.Tests.Integration.Reactive
```
shiftkey: As mentioned below, can we drop this namespace from our new test class and re-indent this file?

Octokit.Tests/Helpers/Arg.cs
```diff
@@ -53,6 +53,21 @@ public static NewRepository NewRepository
             get { return Arg.Is<Dictionary<string, string>>(d => d.Count == 0); }
         }
 
+        public static Dictionary<string, string> DictionaryWithSince
```
shiftkey: I like `DictionaryWithApiOptions` for the general case, but these feel very specific to gists.

Should these live here, or do you think they should live closer to the tests themselves?

Octokit/Clients/GistsClient.cs
```diff
@@ -87,7 +87,7 @@ public Task<Gist> Fork(string id)
         /// <param name="id">The id of the gist</param>
         public Task Delete(string id)
         {
-            Ensure.ArgumentNotNull(id, "id");
+            Ensure.ArgumentNotNullOrEmptyString(id, "id");
```
shiftkey: Nice spot :sparkles:

Octokit/Clients/GistsClient.cs
```diff
@@ -234,7 +392,7 @@ public Task<IReadOnlyList<GistFork>> GetAllForks(string id)
         /// <param name="gistUpdate">The update to the gist</param>
         public Task<Gist> Edit(string id, GistUpdate gistUpdate)
         {
-            Ensure.ArgumentNotNull(id, "id");
+            Ensure.ArgumentNotNullOrEmptyString(id, "id");
```
shiftkey: And nice work picking up these ones too!

shiftkey:
@SamTheDev this is getting close - the namespace changes are the only real blocker, but those new test helpers are an interesting question...

Octokit.Tests/Clients/GistsClientTests.cs
SamTheDev: I see, I need to get use to this ugly underline then :/
image

Octokit.Tests/Helpers/Arg.cs
SamTheDev: yes indeed both `DictionaryWithApiOptions `and `DictionaryWithApiOptionsAndSince `are specific to  `GitsClient`, it doesn't feel right to put them in `Arg.cs`, should i define them in both `GistsClientTests `and `ObservableGistsTests `or remove them entirely and use the expression each time ?

shiftkey: @SamTheDev if you feel like they are being used many times and the readability is necessary, define them in the classes. Otherwise just inline them.

Octokit.Tests/Clients/GistsClientTests.cs
shiftkey: We'll circle around to that later and get R# to play along (I'm pretty sure that's R#, it is on my machine at least)

Octokit.Tests/Helpers/Arg.cs
SamTheDev: @shiftkey done.

SamTheDev:
@shiftkey one last thing and this probably not the right place to ask, but haven't you guys thought of better implementation to the `ArgumentNotNullOrEmptyString `method so you won't pass the parameter name each time ? using reflexion may be ?

SamTheDev:
Or using something like this:
 ```
public static void ArgumentNotNull([ValidatedNotNull]object value)
        {
            if (value != null) return;

            throw new ArgumentNullException(NameOf(() => value));
        }
        public static string NameOf<T>(Expression<Func<T>> memberExpression)
        {            
            return ((MemberExpression)memberExpression.Body).Member.Name;
        }
        public static void ArgumentNotNullOrEmptyString([ValidatedNotNull]string value)
        {            
            
            ArgumentNotNull(value);
            if (!string.IsNullOrWhiteSpace(value)) return;

            throw new ArgumentException("String cannot be empty", NameOf(()=>value));
        }
```

or this ain't worth the effort ?

**Comment by shiftkey** [comment]
@SamTheDev not really, but rather than using reflection I'd recommend switching our usages over to `nameof(arg)` from C# 6. We migrated to VS2015 a while ago, and now that things have settled down I think we can hit the switch on using features like that...

**Comment by SamTheDev** [comment]
@shiftkey that's what i 've tried at first but using `nameof `inside a method will always return the name of the method's parameter's name  instead of the passing parameter name :

```
public static void ArgumentNotNullOrEmptyString([ValidatedNotNull]string value)
{                        
      ArgumentNotNull(value);
      if (!string.IsNullOrWhiteSpace(value)) return;
         throw new ArgumentException("String cannot be empty", nameof(value));
}
```

`nameof(value)` here will return always "value", and i couldn't found a hack around that :/

**Comment by shiftkey** [comment]
@SamTheDev you need to do it up a level, e.g.

```csharp
Ensure.ArgumentNotNullOrEmptyString(id, nameof(id));
```

SamTheDev:
@shiftkey i guess, there are no escape from passing 2 params then

shiftkey:
@SamTheDev correct

M-Zuber:
What about extension methods? (other than being ugly :wink:)

shiftkey:
@M-Zuber 

> What about extension methods? 

Extension methods are just syntactic sugar on top of regular methods, so I'm not sure how they'd get around the same problems as mentioned above. However I'd love to be proven wrong if it helps make things more readable...

shiftkey:
@SamTheDev aside from the merge conflict, I think this is good to go!

shiftkey:
@SamTheDev @M-Zuber feel free to move this discussion about `Ensure` out to a fresh issue, if you want to explore it further...

SamTheDev:
@shiftkey conflicts resolved.

shiftkey:
@SamTheDev looks like a new convention test is in your way:

```
Octokit.Tests.Conventions.MissingClientConstructorTestMethodException : Constructor test method is missing GistsClientTests.
```

See #1246 for more information

SamTheDev:
@shiftkey fixed

PR #1265, opened 2016-04-11
Ensure that Commit.Compare(..) sets MergeBaseCommit correctly.

### LINKED ISSUE #1258: Octokit.CompareResult does not include merge_base_commit field of the GitHub Api
Hi, 

GitHub compare Api (https://developer.github.com/v3/repos/commits/#compare-two-commits) specifies the "merge_base_commit" field. Unfortunately, OctoKit.CompareResult (return value of await IGitHubClient.Repository.Commit.Compare(..)), does not contain a property for the "merge_base_commit" field. The closest it has is "MergedBaseCommit", however it always returns "null".

Are there any plans to properly fill this value and fix the naming convention?

Thanks,

kivancmuslu:
MergeBaseCommit was spelled incorrectly in CompareResult causing Octokit to pass
"null" no matter what the comparison is. This commit makes MergedBaseCommit (the
previous spelling) obsolete, introduces MergeBaseCommit property and adds a test
to ensure that MergeBaseCommit is set properly in CompareResult.

In the test, I specifically compared a commit with itself since the test only
cares whether the property is filled correctly or not, it is not testing whether
the internal computation is correct or not.

Closes #1258.

### DIFF
--- a/Octokit.Tests.Integration/Clients/RepositoryCommitsClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/RepositoryCommitsClientTests.cs
@@ -21,6 +21,13 @@ public class RepositoryCommitsClientTests
         }
 
         [IntegrationTest]
+        public async Task CanGetMergeBaseCommit()
+        {
+            var compareResult = await _fixture.Compare("octokit", "octokit.net", "65a22f4d2cff94a286ac3e96440c810c5509196f", "65a22f4d2cff94a286ac3e96440c810c5509196f");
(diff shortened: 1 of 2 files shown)


shiftkey:
> In the test, I specifically compared a commit with itself since the test only
cares whether the property is filled correctly or not, it is not testing whether
the internal computation is correct or not.

:thumbsup: nice work!

shiftkey:
@kivancmuslu thanks for submitting this! I literally have no feedback to give aside from this:

thumbs-up-jackie

Issue #1024, opened 2015-12-23
Implement Enterprise - Management Console client


shiftkey:
New in GitHub Enterprise 2.4

https://developer.github.com/v3/enterprise/management_console/

```
/enterprise/management_console/
 - POST /setup/api/start
 - POST /setup/api/upgrade
 - GET /setup/api/configcheck
 - POST /setup/api/configure
 - GET /setup/api/settings
 - GET /setup/api/maintenance
 - POST /setup/api/maintenance
 - GET /setup/api/settings/authorized-keys
 - POST /setup/api/settings/authorized-keys
 - DELETE /setup/api/settings/authorized-keys
```

This has some additional security notes, for whoever ends up implementing this:

> You must explicitly set the port number when making API calls to the Management Console. If SSL is enabled on your Enterprise instance, the port number is `8443`; otherwise, the port number is `8080`.
> 
> If you don't want to provide a port number, you'll need to configure your tool to automatically follow redirects.

devkhan:
@ryangribble I would also like to work on this, can I help?

ryangribble:
Do you have access to a GitHub Enterprise environment (this is on premise version of github) ?

devkhan:
I guess not. :cry:

ryangribble:
Yeah if you have a look at the API docs linked above, there are some specific APIs that only the on premises "GitHub Enterprise" supports.  We use GHE at my work so I've been working on fleshing out the support for these.  Happy for anyone to help of course, but obviously there are a few barriers to entry - namely that you have a GitHub Enterprise license!

PR #1269, opened 2016-04-13
Begin implementation of Enterprise ManagementConsole API

ryangribble:
This is the beginnings of adding support for #1024 the GitHub Enterprise [Management Console API](https://developer.github.com/v3/enterprise/management_console/)

To begin with Ive only added the Get/Set Maintenance Mode functions and the actual plumbing around how we can actually integrate with the management console API.

Im aware there are merge conflicts as I havent rebased to latest master yet.  We are far from done with this PR so Ill do that in good time.  For now I just want to get the [WIP] up to find out from @Haacked and @shiftkey how far off track i've taken things :trollface: : 
### Notes

This API is significantly different to all the "regular" github APIs.  There are special requirements around connectivity and authentication and it also doesnt have fully fleshed out calls that take multiple parameters, but in more than one place simply have a single parameter that takes a json string specifying multiple parameters.  To top it off it had some really strange handling around date formats and some other things.

Im definitely after guidance on some of the design decisions ive made...
### Summary

Read below for large detail on the inner workings and design decisions I made, but in a nut-shell someone using octokit.net against a GitHub Enterprise server an now simply call the following methods on the same `GitHubClient` client they have already created for regular API access.

``` csharp
// Some examples...

var consolePassword = "MyPassword";

// Get maintenace status
var maintenance = await 
    _github.Enterprise.ManagementConsole.GetMaintenanceMode(consolePassword);

// Set maintenance mode ON now
var maintenance = await
    _github.Enterprise.ManagementConsole.EditMaintenanceMode(
        new UpdateMaintenanceRequest(true, MaintenanceDate.Now()),
        EnterpriseHelper.ManagementConsolePassword);

// Set maintenance mode OFF
var maintenance = await
    _github.Enterprise.ManagementConsole.EditMaintenanceMode(
        new UpdateMaintenanceRequest(),
        consolePassword);

// Schedule maintenance for next wednesday 5pm
var maintenance = await
    _github.Enterprise.ManagementConsole.EditMaintenanceMode(
        new UpdateMaintenanceRequest(true, MaintenanceDate.FromChronicValue("Next wednesday at 5pm")),
        consolePassword);

// Schedule maintenance mode in 10 minutes, with a DateTimeOffset
var start = DateTimeOffset.Now.AddMinutes(10);
var maintenance = await 
    _github.Enterprise.ManagementConsole.EditMaintenanceMode(
        new UpdateMaintenanceRequest(true, MaintenanceDate.FromDateTimeOffset(start)),
        consolePassword);
```
#### Network access
- TCP Port
  Since octokit handles following redirects, we don't have any issues around handling the special ports of `8080` and `8443` so we are good on this front and can simply target this API at the regular server address the user is passing into the `GitHubClient` as octokit will follow redirects as needed
- API Uri's
  Not as lucky on the Uri front though... The management console api calls must be targetted at `http://my.enterprise.server/setup/api` ie they can't have the `/api/v3/` in the URL as all other calls do.
  Unfortunatley in the guts of octokit.net's `GitHubClient` and `Connection` classes, this `/api/v3/` is added to any custom URL specified by the user and saved in the `baseUri` of the `Connection` object.
  Originally my implementation made it so that you had to declare a new/special `GitHubClient` using a new constructor essentially stating that you wanted to use it for management console access, and this essentially passed a `bool` through which prevented the `/api/v3/` being put in the uri.  But of course this then meant that all the other API clients on this `GitHubClient` wouldnt actually work... and similarly it meant that the `client.Enterprise.ManagementConsole` client wouldn't work on a "regular" `GitHubClient` you were using to call the normal APIs.
  My latest iteration was actually exploiting behaviour of the dotnet `Uri` appending behaviour, where if an endpoint relative Uri starts with a leading forward slash, it actually causes any Uri on the `baseUri` to be stripped off.  We've encountered the odd bug over time where inadvertently including these leading slashes on Urls in the `ApiUri` helper class has caused endpoints to not work on github enterprise, but now it seemed like this behaviour might actually help us here.  Rather than just magially include the leading slashes on the management console Uris, i implemented a function `CorrectEndpointForManagementConsole(endpoint)` that is called in each management console client method, to add the leading slash if not already there.  This does seem a bit hacky but really it is a fairly reliable way to allow a user to specify one `GitHubClient` object and credentials that work against the regular APIs on their enterprise instance, then when they use the ManagementConsole API on the same client, we manipulate the Uri via the endpoint (ie not manipulating the baseUri) and provide the management console password each time.  Plus there is zero behavioral change for anyone using octokit against github.com (ie not an enterprise user).  As an octokit contributor and enterprise user I felt like this has achieved good usability and is an acceptable code base compromise.  At least it's a starting point to see how @shiftkey and others feel :)
#### Authenticating

The management console has a special password to access it.  For API use, this password is specified as the api key, however it appears it cant be specified in the "normal" way the user/password or OAUTH token can be (ie using our `Credentials` object which inserts it in a header in base64/UTF8 form).  I was only able to get things successfully working, by having `?api_key=MyPassword` on the end of the Uri (or potentially it would also work being embedded in the server name `https://api_key:MyPassword@hostname` but that causes similar problems as above if we want to try and have the one `GitHubClient` object able to work with all APIs
So with the way i've implemented it on the end of the Url, it actually doesnt work too badly as it means the existing credentials provided on creating the `GitHubClient` can stay, and are used for regular API access, and every management console method takes a `managementConsolePassword` parameter.  I actually like this as it's more inline with the seriousness of some of these management console activities - eg along the lines of "provide your password everytime you do these management related things"
The irksome thing is that the password is essentiall in clear text on the end of the Url but im not sure if that's really any different to it being in the HTTP headers for regular calls anyway...
#### What the heck is up with parameters containing json strings?!!

The Set Maintenance Mode call takes a single `maintenance` parameter which is defined as follows:

> _maintenance_ `string` _Required_ A JSON string with the attributes enabled and when.

I tried and tried and I just could NOT get things working in the normal way where we have a request object, which gets serialized to json and put in the body of the HTTP Post.  I had many incarnations providing a class with a public `Maintenance` parameter that then contained the c# object class with the `Enabled` and `When` parameters on it.  When that didnt work I tried simply having a public `string` Getter parameter that deserialized the class to json (didnt work), so then moved to constructing the json string in a more basic way - both normal format and fully escaped, I tried doing the parameter dictionary type approach like used in search requests, I tried the default content type and `application/json` content types but in all cases github would throw a HTTP 400 Bad Request.  
Eventually I found that if I just specified the string in the url-form-encoded format of `maintenace="{"enabled":true, "when":"now"}"` it worked, so that's where I'm at at the moment.  I plan to tidy up the json string to not be created manually and instead call a serializer but for now I have something working so didnt want to wait in pushing it up.
If anyone is able to show me how to get this working in a more "normal" way for octokit where the request object is serialized using the builtin `SimpleJsonSerializer` in the normal pipeline, I would love to see what I missed bercause I feel like I tried so many attempts and still couldnt crack it without going the form encoded route.

This is definitely important to sort out a strategy for, because the [Settings](https://developer.github.com/v3/enterprise/management_console/#retrieve-settings) endpoint on the management console is a real doozy containing literally 50+ items in a hierarchical json structure that all have to get packed into a single json string on a `settings` property in the API call.  I am NOT looking forward to that endpoint!
#### What the heck is up with the time handling?!!

The response to a scheduled maintenance mode event includes this field 
`"scheduled_time": "Tuesday, January 22 at 15 => 34 -0800"`
In any other github API call ive encountered, the dates are in a lovely and easily parseable ISO 8601 format 
In this case we not only have the the full day and month names (this will pose issues if github can run in other cutures/languages), the silly formatting like comma and `=>` the non standard timezone offset information (in dotnet land this should be `-08:00` rather than `-0800` but most importantly THE YEAR is missing!  This plays havoc with attempting to actually parse this into a date.  Now if we hardcode a culture like `en-US` we can get around the full day/month names, and we can "almost" use `ParseExact()` to match most of the formatting issues, except the lack of year means we cant use `DateTimeOffset.ParseExact()` but if we use `DateTime.ParseExact()` the time zone offset is not handled well.  
Given that the day of week and month are mentioned, we "could" potentially take an educated guess at the year, since eg if it says Tuesday the 12th April, we can figure out what YEAR in the future the 12th April will fall on a Tuesday and presume it to be this... then we could shove that year into the string.  This is clearly getting ridiculous though (espcially because this field is coming back from the API and we want it to be deserialised properly in the Json deserializer) so I took the easy way out and simply made it a `string`!!!
In runtime testing this behaviour it's even more odd because through the API I can scheduled maintenance for a date in 1/1/2017 and the API response `Sunday, January 1 at 20:00` indicates it really is that 2017 date - the fact it says Sunday is the confirmation.  Yet the management console UI will show it as 1/1/2016 but in a scheduled status.  If it really was 20-16 we should either now BE in maintenance mode (not scheduled) or it should reject the update since it's in the past.  So im thinking even the management console UI itself is using this API call with the missing date and just assuming it's the current year.
Now does it make sense to schedule maintenance for so far in the future?  Not really!  But given the API lets you do it, it's not really our place to limit this in any way.  My hope is that this API call gets modified to include a properly parseable ISO 8601 formatted date like all other dates in the normal APIs, and the management console GUI fixed up to know the proper year as well :grinning: In the meantime Im thinking we have to leave this field as a `string` because parsing it is too difficult and not 100% reliable.
#### Other design decisions
- Specifying "when" parameter in EditMaintenanceMode()
  The SetMaintenance method `when` parameter can handle "any" string that the [mojombo/chronic](https://github.com/mojombo/chronic) ruby library can parse.  This includes lots of humanised things like "next tuesday at 5pm" and "3pm on the monday before friday in 3 weeks time".  Luckily for our api case we simply have to take whatever string the user says and pass it through.  But ive also added specific suport to passing in a `DateTimeOffset` object since that is more aligned to our c# nature, and also the special `now` keyword as indicated by the API docs.  To do this i created a `MaintenanceDate` object used as the `When` parameter on the `UpdateMaintenanceRequest` object, and helper functions `MaintenanceDate.FromDateTimeOffset()` `MaintenanceDate.FromChronicString()` and `MaintenanceDate.Now()`
- MaintenanceModeStatus as an enum
  There appear to be finite values of `off` `on` and `scheduled` so I made this an enum.  The API docs dont definitively state the finite values though...  This isnt too different to other enums we've implemented though
- Added environment var for management console password for integration tests
  added to the powershell helper script and the `EnterpriseHelper` class, so the integration tests can run

That's about it for now.  Please have a read and provide feedback...  I'm looking at you @shiftkey :grinning:


ryangribble:
@shiftkey this is another one where im pretty much waiting for you or @Haacked or someone "official" to provide some feedback/help, if you can add it to your list :grinning:

ryangribble:
@shiftkey just a gentle reminder about this PR 😀  I know you've been travelling etc, I've been a bit preoccupied with some other things recently myself, but should be ready to get back into finishing off the enterprise api's now, pending advice/feedback on what I'd found with this PR

ryangribble:
@shiftkey I'd like to knock over more of the enterprise maintenance stuff for the next release but I really need your guidance on everything I mentioned in the PR comments, as the other endpoints have similar issues to this maintenance mode one. Don't want to be a nag but incase it's fallen off the radar I wanted to bump it 😉

Octokit.Tests.Integration/Clients/Enterprise/EnterpriseManagementConsoleClientTests.cs
```diff
@@ -0,0 +1,91 @@
...
+    [GitHubEnterpriseTest]
+    public async Task CanSetMaintenanceModeOff()
+    {
+        // Set maintenance mode OFF now
+        var maintenance = await
+            _github.Enterprise.ManagementConsole.EditMaintenanceMode(
+                new UpdateMaintenanceRequest(),
+                EnterpriseHelper.ManagementConsolePassword);
+
+        Assert.NotNull(maintenance);
```
shiftkey: :fire: - the next assert will cover this

Octokit.Tests.Integration/Clients/Enterprise/EnterpriseManagementConsoleClientTests.cs
```diff
@@ -0,0 +1,91 @@
...
+
+    public EnterpriseManagementConsoleClientTests()
+    {
+        _github = EnterpriseHelper.GetAuthenticatedClient();
+    }
+
+    [GitHubEnterpriseTest]
+    public async Task CanGetMaintenanceMode()
+    {
+        var maintenance = await _github.Enterprise.ManagementConsole.GetMaintenanceMode("Password01");
```
shiftkey: Should this string be `EnterpriseHelper.ManagementConsolePassword` ?

Octokit.Tests.Integration/Clients/Enterprise/EnterpriseManagementConsoleClientTests.cs
```diff
@@ -0,0 +1,91 @@
...
+        var maintenance = await
+            _github.Enterprise.ManagementConsole.EditMaintenanceMode(
+                new UpdateMaintenanceRequest(true, MaintenanceDate.Now()),
+                EnterpriseHelper.ManagementConsolePassword);
+
+        Assert.NotNull(maintenance);
+        Assert.Equal(maintenance.Status, MaintenanceModeStatus.On);
+
+        // Ensure maintenance mode is OFF
+        await _github.Enterprise.ManagementConsole.EditMaintenanceMode(
```
shiftkey: We don't seem to be doing any asserting here? Will the request will fail here if the GHE server is not in the right state?

Octokit/Clients/Enterprise/EnterpriseManagementConsoleClient.cs
```diff
@@ -0,0 +1,68 @@
...
+        public Uri CorrectEndpointForManagementConsole(Uri endpoint)
+        {
+            Ensure.ArgumentNotNull(endpoint, "endpoint");
+
+            if (ApiConnection.Connection.BaseAddress != null &&
+                ApiConnection.Connection.BaseAddress.ToString().EndsWith("/api/v3/", StringComparison.OrdinalIgnoreCase))
+            {
+                // We need to get rid of the /api/v3/ for ManagementConsole requests
+                // if we specify the endpoint starting with a leading slash, that will achieve this
+                return string.Concat("/", endpoint.ToString()).FormatUri();
```
shiftkey: :lipstick: maybe just a `new Uri("/" + endpoint.ToString());` here to simplify it

Octokit/Clients/Enterprise/EnterpriseManagementConsoleClient.cs
```diff
@@ -0,0 +1,68 @@
...
+    /// <remarks>
+    /// See the <a href="https://developer.github.com/v3/enterprise/management_console/">Enterprise Management Console API documentation</a> for more information.
+    ///</remarks>
+    public class EnterpriseManagementConsoleClient : ApiClient, IEnterpriseManagementConsoleClient
+    {
+        public EnterpriseManagementConsoleClient(IApiConnection apiConnection)
+            : base(apiConnection)
+        { }
+
+        public Uri CorrectEndpointForManagementConsole(Uri endpoint)
```
shiftkey: Can this be internal?

Octokit/Clients/Enterprise/EnterpriseManagementConsoleClient.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// <remarks>
+        /// https://developer.github.com/v3/enterprise/management_console/#check-maintenance-status
+        /// </remarks>
+        /// <returns>The <see cref="MaintenanceModeResponse"/>.</returns>
+        public Task<MaintenanceModeResponse> EditMaintenanceMode(UpdateMaintenanceRequest maintenance, string managementConsolePassword)
+        {
+            Ensure.ArgumentNotNull(maintenance, "maintenance");
+            Ensure.ArgumentNotNullOrEmptyString(managementConsolePassword, "managementConsolePassword");
+
+            var endpoint = ApiUrls.EnterpriseManagementConsoleMaintenance(managementConsolePassword);
```
shiftkey: :lipstick: rather than doing this as two steps, you could pass in the base address here and move that logic out so it's just one call:

```csharp
public static Uri EnterpriseManagementConsoleMaintenance(string managementConsolePassword, Uri baseAddress)
{
    if (baseAddress != null
        && baseAddress.ToString().EndsWith("/api/v3/", StringComparison.OrdinalIgnoreCase))
    {
        // note: leading slash here means the /api/v3/ prefix inherited from baseAddress is ignored
        return "/setup/api/maintenance?api_key={0}".FormatUri(managementConsolePassword);
    }
     
    return "setup/api/maintenance?api_key={0}".FormatUri(managementConsolePassword);
}
```

And this usage becomes: 

```csharp
var endpoint = ApiUrls.EnterpriseManagementConsoleMaintenance(managementConsolePassword, ApiConnection.Connection.BaseAddress);
```

It's a bit more duplication inside `ApiUrls`, but perhaps there's other GHE-based APIs that we need to do this for...

Octokit.Tests.Integration/Clients/Enterprise/EnterpriseManagementConsoleClientTests.cs
```diff
@@ -0,0 +1,91 @@
...
+        Assert.NotNull(maintenance);
+        Assert.Equal(maintenance.Status, MaintenanceModeStatus.Scheduled);
+
+        // Ensure maintenance mode is OFF
+        await _github.Enterprise.ManagementConsole.EditMaintenanceMode(
+            new UpdateMaintenanceRequest(),
+            EnterpriseHelper.ManagementConsolePassword);
+    }
+
+    [GitHubEnterpriseTest]
```
shiftkey: I think these tests need a specific attribute that verifies the existence of `OCTOKIT_GHE_CONSOLEPASSWORD`

Octokit/Models/Request/Enterprise/UpdateMaintenanceRequest.cs
```diff
@@ -0,0 +1,153 @@
...
+        }
+
+        public bool Enabled { get; protected set; }
+        
+        public MaintenanceDate When { get; protected set; }
+
+        [SuppressMessage("Microsoft.Globalization", "CA1304:SpecifyCultureInfo", MessageId = "System.String.ToLower")]
+        public string AsJsonString()
+        {
+            var json = string.Format(CultureInfo.InvariantCulture,
```
shiftkey: It might sound excessive, but perhaps a `StringBuilder` in here would make some of this `Concat` code easier to read...

Octokit/Models/Request/Enterprise/UpdateMaintenanceRequest.cs
```diff
@@ -0,0 +1,153 @@
...
+        /// <summary>
+        /// Static helper to create a <see cref="MaintenanceDate"/> using the humanised forms supported by the <a href="https://github.com/mojombo/chronic">mojombo/chronic library</a> used by the GitHub API
+        /// such as "this friday at 5pm" or "5 minutes before midday tomorrow"
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="https://developer.github.com/v3/enterprise/management_console/#enable-or-disable-maintenance-mode">Enable or disable maintenance mode</a> API documentation for more information.
+        /// </remarks>
+        public static MaintenanceDate FromChronicValue(string chronicValue)
+        {
+            return new MaintenanceDate(chronicValue);
```
shiftkey: `Ensure.ArgumentNotNullOrEmptyString`

shiftkey:
@ryangribble apologies again for the delay on this - the approach seems fine, added a few comments

Digging into the original PR comment:

#### API Uri's

I'm fine with isolating that rewriting behaviour as near to the Management Console functionality as possible. We might have a better idea for handling it down the track.

#### Authenticating

Also fine. [Just something we need to live with, it seems.](https://developer.github.com/v3/enterprise/management_console/#authentication)

> The irksome thing is that the password is essentiall in clear text on the end of the Url but im not sure if that's really any different to it being in the HTTP headers for regular calls anyway...

If you're using HTTPS and everything is setup correctly on the server, after the [initial handshake](https://en.wikipedia.org/wiki/Transport_Layer_Security#TLS_handshake) between server and client the contents of requests (whether it's the request line or the `Authorization` header) are considered encrypted.

#### What the heck is up with parameters containing json strings?!!

Gosh that's some beautiful API design :trollface:

> If anyone is able to show me how to get this working in a more "normal" way for octokit where the request object is serialized using the builtin `SimpleJsonSerializer` in the normal pipeline, I would love to see what I missed bercause I feel like I tried so many attempts and still couldnt crack it without going the form encoded route.

Pretty much all of Octokit's serialization has been built off the assumption you're serializing an object to the request body. This is outside that sandbox, so I think we're stuck for now. If SimpleJson is all you need, perhaps we could make the serializer a first class feature on the client, so the code might look like this:

```csharp
public Task<MaintenanceModeResponse> EditMaintenanceMode(UpdateMaintenanceRequest maintenance, string managementConsolePassword)
{
    // guard clauses and other setup
 
    var payload = "maintenance=" + this.Serializer.Serialize(maintenance);

    return ApiConnection.Post<MaintenanceModeResponse>(endpoint,  payload);
}
```

I'd need to think about what that might break in terms of testability by baking it into `ApiClient`, and what other scenarios this would help us cover. Also, are there more scenarios where the JSON sent needs to be wildly flexible beyond our simple serializer attributes? Perhaps this won't be worth investing in...

#### What the heck is up with the time handling?!!

I'll just leave this here:



Okay, now to actually be constructive.

 - When sending a value as part of the request, it sounds like we should be able to send an ISO 8601 string, right?
 - When receiving a value as part of the response, why bother being clever with the parsing? Just serve it up as a string, if the API is going to try and be clever with what it presents.

I've ask the team about this API and whether others have felt the same pain.

#### Specifying "when" parameter in EditMaintenanceMode()

These all seem fair. I'd perhaps drop `FromChronicString` in case it does get obsoleted, but that's a minor feel.

#### MaintenanceModeStatus as an enum

#### Added environment var for management console password for integration tests

:thumbsup: to both

Octokit.Tests.Integration/Clients/Enterprise/EnterpriseManagementConsoleClientTests.cs
ryangribble: From memory I wrote these tests to always return maintenance mode status back to a consistent state afterwards, so this was just ensuring matinenance mode was OFF after the test which turned it ON.  Or are you saying we should actually assert this here anyway?  Or potentially do it using a `Context` helper class, with the destructor implementing the "reset" logic?

Octokit/Clients/Enterprise/EnterpriseManagementConsoleClient.cs
ryangribble: There was some issue with one of the code analysis rules not letting you form Uri's from strings but Ill take another look

Octokit/Models/Request/Enterprise/UpdateMaintenanceRequest.cs
ryangribble: I think this was left in this state through sheer frustration, but im sure it should be possible to at least have a class that serializes to a json string using our existing SimpleJsonSerializer

ryangribble:
thanks for the comments.  Ill revisit this soon with those comments in mind and see what a fresh perspective brings.

I think I've already got a branch somewhere where I at least used the serializer to build the JSON string but from memory I still couldnt get it to be accepted in the request body and had to go with the URLEncoded approach.

Octokit/Clients/Enterprise/EnterpriseManagementConsoleClient.cs
ryangribble: sounds good

Octokit/Models/Request/Enterprise/UpdateMaintenanceRequest.cs
```diff
@@ -0,0 +1,153 @@
...
+            }
+
+            json = string.Concat("{", json, "}");
+
+            return json;
+        }
+
+        public string AsNamedFormEncodingString()
+        {
+            var value = string.Format(CultureInfo.InvariantCulture, "maintenance={0}", this.AsJsonString());
```
ryangribble: i was thinking perhaps like how we have the `RequestParameters` class that has the `ToParametersDictionary()` that we could maybe extend this (or add a new similar class) that would provide a `ToFormUrlEncoded()` method on it, which would provide a way to have any request class be able to have it's key/value pairs formatted up to suit the required style here (eg  key=value&key2=value2)

Octokit.Tests.Integration/Clients/Enterprise/EnterpriseManagementConsoleClientTests.cs
shiftkey: > Or are you saying we should actually assert this here anyway? Or potentially do it using a Context helper class, with the destructor implementing the "reset" logic?

I'm neutral on this, but I guess if "ensure maintenance mode is off" needs to be done with each test it should be done in teardown. Just me misreading the test a bit.

Octokit/Clients/Enterprise/EnterpriseManagementConsoleClient.cs
shiftkey: Huh. Weird.

Octokit.Tests.Integration/Clients/Enterprise/EnterpriseManagementConsoleClientTests.cs
BigAlInTheHouse: What should you need me do?

ryangribble:
OK I think all comments have been addressed 

I've also changed the request object side of things so the structure matches what the API expects, and I can use `SimpleJsonSerializer` to build the json string.  

The usage is now `new MaintenanceUpdate(new MaintenanceUpdateDetails(true, "next tuesday at 5pm"))` - the nested class isn't ideal but that's how the API object needs to be (ie top level item name `maintenance` that then contains the `enabled`/`when` attributes).

I still need to UrlFormEncode the request rather than have it in the request body, but at least it's now using the normal serializer

Octokit.Tests.Integration/Helpers/MaintenanceModeContext.cs
```diff
@@ -0,0 +1,31 @@
...
+        }
+
+        private IConnection _connection;
+
+        private bool _initialState;
+        
+        public void Dispose()
+        {
+            // Ensure maintenance mode is restored to desired state
+            EnterpriseHelper.SetMaintenanceMode(_connection, _initialState);
```
khellang: Why is this setting it to the exact same thing as the ctor?

Octokit/Models/Request/Enterprise/UpdateMaintenanceRequest.cs
```diff
@@ -0,0 +1,107 @@
...
+        /// <summary>
+        /// Maintenance request details to enable/disable maintenance at a specified time (only applicable when enabling maintenance)
+        /// </summary>
+        /// <param name="enabled">true to enable, false to disable</param>
+        /// <param name="when">A <see cref="DateTimeOffset"/> specifying when maintenance mode will be enabled.</param>
+        /// <remarks>If enabled is false, the when parameter is ignored and maintenance is turned off immediately</remarks>
+        public UpdateMaintenanceRequestDetails(bool enabled, DateTimeOffset when)
+        {
+            Enabled = enabled;
+            When = when.ToUniversalTime().ToString("yyyy-MM-ddTHH:mm:ssZ",
```
khellang: Isn't this (more or less) the ["round-trip (`ToString("o")`) format"](https://msdn.microsoft.com/en-us/library/az4se3k1(v=vs.110).aspx#Roundtrip) spelled out?

ryangribble: indeed it is :blush: 

I'll see if the GHE API is happy to just receive the out of the box `.ToString("o")` format

Octokit.Tests.Integration/Helpers/MaintenanceModeContext.cs
ryangribble: this is a helper context class to be used via a using statement wrapping a test.  

So a test that sets maintenance ON, needs to ensure that maintenance was OFF to begin with, hence the intial state and setting it in the `ctor`.

But also we dont want to leave GHE in maintenance mode after a test (since that is pretty useless for any subsequent tests) so the destructor should be ensuring maintenance mode is OFF again.  Which make me realise Im currently setting the intial state in destructor but actually should always be setting OFF :grinning:

khellang: > Which make me realise Im currently setting the intial state in destructor but actually should always be setting OFF :smile: 

:+1:

Octokit/Models/Request/Enterprise/UpdateMaintenanceRequest.cs
ryangribble: > I'll see if the GHE API is happy to just receive the out of the box .ToString("o") format

@khellang : No dice! :stuck_out_tongue: 

An internal server error 500 is thrown when using this format.

All of these values cause an error
```
2016-08-16T21:15:06.9332018+00:00 // ToUniversalTime().ToString("o")
2016-08-17T07:05:06.9332018+10:00 // .ToString("o")
2016-08-16T21:15:06+00:00 // UTC with custom format removing the fffffff component
2016-08-17T07:05:06+10:00 // Local time with custom format removing the fffffff component
2016-08-16T21:15:06+0000 // UTC with custom format removing the fffffff component and removing colon in time zone
2016-08-17T07:05:06+1000 // Local time with custom format removing the fffffff component and removing colon in time zone
2016-08-16T21:15:06+00 // UTC with custom format removing the fffffff component and only 2 digits for time zone
2016-08-17T07:05:06+10 // Local time with custom format removing the fffffff component and only 2 digits for time zone
```
These formats work:
```
`2016-08-16T21:15:06Z` // .ToUniversalTime() with custom format using a static "Z" for zulu timezone
`2016-08-16T21:15:06.9332018Z` // .ToUniversalTime() with custom format including fffffff and using a static "Z" for zulu timezone
```

So yeah basically it's the timezone designator that the GHE API seems to only accept when it is "Z" and not any of the other (valid ISO8601) values.

ryangribble: @khellang can you confirm you're good with my attempts above basically proving that it does seem to require being done this way and not the built in "O" format?

khellang: Hey, if it doesn't work, it doesn't work. How am I to say what should and shouldn't work :wink: I think I'd ping the API team at GitHub about this, though, as it seems to me that it should be more forgiving with the values it accepts.

ryangribble:
Rebased and assigned to @shiftkey for review :trollface:

shiftkey:
Moving this to #2010 to see if we can actually land this

PR #1273, opened 2016-04-19
Method of assembly determination was changed in order to use types of top level

dampir:
Refers #1246 

As was discussed in #1246 top level types should be used to determine assembly name in conventions tests. I've changed type from IEventsClient to IGitHubClient and GitHubClientTests types.

/cc @shiftkey @ryangribble

### DIFF
--- a/Octokit.Tests.Conventions/ModelTests.cs
+++ b/Octokit.Tests.Conventions/ModelTests.cs
@@ -97,7 +97,7 @@ namespace Octokit.Tests.Conventions
 
         public static IEnumerable<object[]> GetClientInterfaces()
         {
-            return typeof(GitHubClientTests)
+            return typeof(IGitHubClient)
                 .Assembly
                 .ExportedTypes
                 .Where(TypeExtensions.IsClientInterface)
--- a/Octokit.Tests.Conventions/ModelTests.cs
+++ b/Octokit.Tests.Conventions/ModelTests.cs
@@ -119,7 +119,7 @@ namespace Octokit.Tests.Conventions
         {
             var allModelTypes = new HashSet<Type>();
 
-            var clientInterfaces = typeof(GitHubClientTests).Assembly.ExportedTypes
+            var clientInterfaces = typeof(IGitHubClient).Assembly.ExportedTypes
                 .Where(type => type.IsClientInterface());
 
             foreach (var exportedType in clientInterfaces)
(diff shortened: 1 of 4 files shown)


Octokit.Tests.Conventions/ModelTests.cs
```diff
@@ -97,7 +97,7 @@ public void ResponseModelsHaveReadOnlyCollections(Type modelType)
 
         public static IEnumerable<object[]> GetClientInterfaces()
         {
-            return typeof(IEventsClient)
+            return typeof(GitHubClientTests)
```
shiftkey: This should be using something in the `Octokit` project, right? So something like `IGitHubClient`?

shiftkey: Yep, I'm missing some context here https://github.com/octokit/octokit.net/pull/1246#issuecomment-206358328

cc @ryangribble

dampir: @shiftkey nice spot, I think that usage of interface instead of implementation would be appropriate
here.

Octokit.Tests.Conventions/ModelTests.cs
```diff
@@ -119,7 +119,7 @@ private static IEnumerable<Type> GetModelTypes(bool includeRequestModels)
         {
             var allModelTypes = new HashSet<Type>();
 
-            var clientInterfaces = typeof(IEventsClient).Assembly.ExportedTypes
+            var clientInterfaces = typeof(GitHubClientTests).Assembly.ExportedTypes
```
ryangribble: This is from the wrong assembly, should be `IGitHubClient` ?

Octokit.Tests.Conventions/PaginationTests.cs
```diff
@@ -89,7 +89,7 @@ static bool MethodHasOverloadForApiOptions(MethodInfo methodInfo, ParameterInfo[
 
         public static IEnumerable<object[]> GetClientInterfaces()
         {
-            return typeof(IEventsClient).Assembly.ExportedTypes
+            return typeof(GitHubClientTests).Assembly.ExportedTypes
```
ryangribble: This is from the wrong assembly, should be `IGitHubClient` ?

ryangribble:
There's a couple more places where you've accidentally referred to a class in the wrong assembly and thus not found the right client interface classes :grinning:

Octokit.Tests.Conventions/PaginationTests.cs
dampir: Yeah, you're absolutely right!

dampir:
@ryangribble , you're absolutely right! Fixed!

PR #1280, opened 2016-04-22
Fix common redundancies

dampir:
Just continue my war with redundancies in code.

1) redundant @ prefix was removed
2) redundant explicit array type specifications were removed
3) redundant argument list was removed
4) redundant commas in object initializer were removed.

/cc @shiftkey


khellang:
:shipit::heart:

ryangribble:
LGTM!

Issue #1120, opened 2016-03-06
Add support for interacting with repositories by id.


Zoltu:
All of the commands that work with `https://api.github.com/repos/:owner/:repo/` also work with `https://api.github.com/repositories/:id/`.  If one desires to build an application that is resilient to change (e.g., repositories changing names, changing owners, owners changing name, etc.) then the application should be built against the `repositories` API rather than the `repos` API.  Unfortunately, it appears that this library only supports the `repos` API.

An application can be mostly resilient against change by saving the ID long term and then using the `repositories` API to convert it to an `owner/repo` right before making the API calls they really want.  This is still not great though as it can race and lead to very unexpected behavior if you are in the middle of a complex transaction (spanning multiple API calls) when the repository moves out from under you.

Please add support for making API calls via ID rather than owner/repo so developers can avoid unexpected behavior.

Example:
```cs
public interface IBlobsClient
{
	Task<BlobReference> Create(String owner, String name, NewBlob newBlob);
	Task<BlobReference> Create(String id, NewBlob newBlob);

	Task<Blob> Get(String owner, String name, String reference);
	Task<Blob> Get(String id, String reference);
}
```

Even better would be to abstract out the concept into a `Repo` object of some kind that can either be an ID or an owner/name.  That way the public surface area doesn't need to be twice as big.  However, deleting the old methods would be a breaking change and require a major version bump.

shiftkey:
Some initial thoughts:

 - the `/repositories/:id/` is not included in the response payload for a repository, e.g. https://api.github.com/repos/octokit/octokit.net or https://api.github.com/repositories/7528679
 - the only place I can find a mention of this is in the [search results API](https://developer.github.com/v3/search/#search-repositories)
 - I can see `octokit.rb` has support for passing an `Integer` [here](https://github.com/octokit/octokit.rb/blob/master/lib/octokit/repository.rb#L20-L21). I can't see an equivalent for `octokit.objc` but perhaps I'm not looking in the right spot.

> If one desires to build an application that is resilient to change (e.g., repositories changing names, changing owners, owners changing name, etc.) then the application should be built against the repositories API rather than the repos API.

This should be covered by the [API supporting redirects](https://developer.github.com/changes/2015-04-17-preview-repository-redirects/) which Octokit now follows correctly. 

>  This is still not great though as it can race and lead to very unexpected behavior if you are in the middle of a complex transaction (spanning multiple API calls) when the repository moves out from under you.

[...]

PR #1287, opened 2016-05-02
Fixed inconsistency in ApiUrls.cs

dampir:
As part of my work on #1120 I've done some improvements in ApiUrls.cs class. 

I've found that in order to support work with repo by their ID we should add 83 new methods in ApiUrls.cs that would duplicate methods that allow to work with repo by their owner and name identificators.

E.g.

We have the next method in ApiUrls.cs:

```cs
public static Uri RepoBranch(string owner, string name, string branchName)
{
      return "repos/{0}/{1}/branches/{2}".FormatUri(owner, name, branchName);
}
```

So we should add the new one:

```cs
public static Uri RepoBranch(int repositoryId, string branchName)
{
      return "repositories/{0}/branches/{1}".FormatUri(repositoryId, branchName);
}
```

Because of the fact that we should add 83 new methods, I've decided to generate them programmatically. And in order to get appropriate result of code generation, I've done some useful changes in ApiUrls.cs.

- [x]  Added XML documentation for needed methods.
- [x] Rename some method's parameters on order to eliminate conflicts of parameter's names.
- [x] Most of ApiUrls methods use "owner" and "name" as name of parameters that identify repository. But some of them use "repositoryName" or "repo" as parameter name. Now all of such methods use consistent parameter names: "owner" and "name".

BTW, I've implemented program that generate new 83 methods. Here is the [repo](https://github.com/dampir/OctokitUrlMethodsGenerator) and here is the [file](https://github.com/dampir/OctokitUrlMethodsGenerator/blob/master/new_methods.txt) with new generated methods. All methods contains appropriate XML documentation in many cases better than their original prototypes. I'm going to pull these methods in next PR after this one would be merged.

Now I'm thinking about how to validate them against their prototypes, how to check that each pair of methods returns equal results (e.g. RepoBranch(string owner, string name, string branchName) and RepoBranch(int repositoryId, string branchName) should return an equal JSONs if repositoryId is according to owner and name of repository).

/cc @shiftkey , @ryangribble

### DIFF
--- a/Octokit/Helpers/ApiUrls.cs
+++ b/Octokit/Helpers/ApiUrls.cs
@@ -557,11 +557,11 @@ namespace Octokit
         /// </summary>
         /// <param name="owner">The owner of the repository</param>
         /// <param name="name">The name of the repository</param>
-        /// <param name="milestoneNumber">The milestone number</param>
+        /// <param name="labelName">The name of label</param>
         /// <returns></returns>
-        public static Uri Label(string owner, string name, string milestoneNumber)
+        public static Uri Label(string owner, string name, string labelName)
         {
-            return "repos/{0}/{1}/labels/{2}".FormatUri(owner, name, milestoneNumber);
+            return "repos/{0}/{1}/labels/{2}".FormatUri(owner, name, labelName);
         }
 
         /// <summary>
--- a/Octokit/Helpers/ApiUrls.cs
+++ b/Octokit/Helpers/ApiUrls.cs
@@ -605,11 +605,11 @@ namespace Octokit
         /// </summary>
         /// <param name="owner">The owner of the repository</param>
         /// <param name="name">The name of the repository</param>
-        /// <param name="number">The milestone number</param>
+        /// <param name="labelName">The label name</param>
         /// <returns></returns>
-        public static Uri MilestoneLabels(string owner, string name, int number)
+        public static Uri MilestoneLabels(string owner, string name, int labelName)
         {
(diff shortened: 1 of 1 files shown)


Octokit/Helpers/ApiUrls.cs
```diff
@@ -556,60 +556,60 @@ public static Uri Milestones(string owner, string name)
         /// Returns the <see cref="Uri"/> that returns the specified label.
         /// </summary>
         /// <param name="owner">The owner of the repository</param>
-        /// <param name="repo">The name of the repository</param>
-        /// <param name="name">The milestone number</param>
+        /// <param name="name">The name of the repository</param>
+        /// <param name="milestoneNumber">The milestone number</param>
```
ryangribble: this should be labelName right?

dampir: Yes, sure. Fixed!

shiftkey: @dampir was the commit pushed?

dampir: @shiftkey, yes, now parameter has name "labelName" in Label method.Here is commit 		0da6f0e

shiftkey: @dampir apologies, I didn't see the comments being folded up so I assumed it was still to-be-done

dampir: @shiftkey no problem! :smile:

dampir:
Travis-ci falls again, please rerun it whenever possible.

PR #1288, opened 2016-05-04
Add actual "GetAllForRepository" Activity Feed

drasticactions:
Currently in EventsClient, "GetAllForRepository" actually gets the Issues activity feed, rather than the actual repository activity feed. This commit fixes that by adding "GetAllIssuesForRepository" and changing "GetAllForRepository" to correctly use the right endpoint.

### DIFF
--- a/Octokit/Clients/EventsClient.cs
+++ b/Octokit/Clients/EventsClient.cs
@@ -51,7 +51,7 @@ namespace Octokit
         /// Gets all the events for a given repository
         /// </summary>
         /// <remarks>
-        /// http://developer.github.com/v3/activity/events/#list-issue-events-for-a-repository
+        /// https://developer.github.com/v3/activity/events/#list-repository-events
         /// </remarks>
         /// <param name="owner">The owner of the repository</param>
         /// <param name="name">The name of the repository</param>
@@ -68,7 +68,7 @@ namespace Octokit
         /// Gets all the events for a given repository
         /// </summary>
         /// <remarks>
-        /// http://developer.github.com/v3/activity/events/#list-issue-events-for-a-repository
+        /// https://developer.github.com/v3/activity/events/#list-repository-events
         /// </remarks>
         /// <param name="owner">The owner of the repository</param>
         /// <param name="name">The name of the repository</param>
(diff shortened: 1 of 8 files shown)


M-Zuber:
Welcome and thank you!
I'd like to offer some pointers that will help get the ci build green and the PR merged. 

1. Some tests where impacted as the expected url has now changed. (when I am back in front of a proper machine I will try and link to them)
2. The new method should be added to the IEventsClient interface. 
3. IObservableEventsClient should also be updated, as well as the implementing class. 

If I can help in any way, just let me know.

drasticactions:
Yeah, realised that after I submitted it that I broke the tests, and then saw everywhere else it was impacted. I'll fix it after.

shiftkey:
@drasticactions no problem, it's what the tests are there for!

drasticactions:
Alright, test should be passing and code should be working now in the normal and observable classes. Let me know if it's all good.

Octokit/Clients/EventsClient.cs
```diff
@@ -80,7 +80,44 @@ public Task<IReadOnlyList<Activity>> GetAllForRepository(string owner, string na
             Ensure.ArgumentNotNullOrEmptyString(name, "name");
             Ensure.ArgumentNotNull(options, "options");
 
-            return ApiConnection.GetAll<Activity>(ApiUrls.IssuesEvents(owner, name),options);
+            return ApiConnection.GetAll<Activity>(ApiUrls.Events(owner, name), options);
```
shiftkey: I think the only issue I have with this PR is that we've changed the behaviour of the existing `GetAllForRepository` method. This is fine, we just need to call it out in the release notes.

shiftkey:
release_notes:

**Breaking Change:** `IEventsClient.GetAllForRepository` was incorrectly retrieving issue events before this release. Use the new `IEventsClient.GetAllIssuesForRepository` method if you still require issues, or continue to use `IEventsClient.GetAllForRepository` if you require all repository events...

shiftkey:
@drasticactions thanks for the contribution!

shiftkey:
And as it's your first contribution to Octokit, we should celebrate!

PR #1289, opened 2016-05-05
#1285 Updated the ReadTheDocs link to .io

### LINKED ISSUE #1285: readthedocs links now need to be .io, not .com
Just got an email with these details from the Read The Docs team (couldn't find a URL to point at):

> Starting today, Read the Docs will start hosting projects from subdomains on the domain readthedocs.io, instead of on readthedocs.org. This change addresses some security concerns around site cookies while hosting user generated data on the same domain as our dashboard.
>
> Changes to provide security against broader threats have been in place for a while, however there are still a few scenarios that can only be addressed by migrating to a separate domain.
>
> We implemented session hijacking detection and took precautions to limit cookie usage, but there are still a number of scenarios utilizing XSS and CSRF attacks that we aren't able to protect against while hosting documentation from subdomains on the readthedocs.org domain. Moving documentation hosting to a separate domain will provide more complete isolation between the two user interfaces.
>
> Projects will automatically be redirected, and this redirect will remain in place for the foreseeable future. Still, you should plan on updating links to your documentation after the new domain goes live.
>
> If you notice any problems with the changes, feel free to open an issue on our issue tracker: http://github.com/rtfd/readthedocs.org/issues. If you do notice any security issues, contact us at security@readthedocs.org with more information.

We're being redirected as expected, but we should update anywhere in the `README.md` or `CONTRIBUTING.md` docs that refer to this explicitly...

radu-matei:
Fixes #1285 
I looked to the rest of the documentation, but couldn't find another reference to Read The Docs, besides the one in the readme file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ In Xamarin Studio you can find this option under the project's context menu: **A
 
 ## Documentation
 
-Documentation is available at http://octokitnet.readthedocs.org/en/latest/.
+Documentation is available at http://octokitnet.readthedocs.io/en/latest/.
 
 ## Build
 


shiftkey:
@radu-matei thanks for checking!

PR #1291, opened 2016-05-11
Added new api urls to get access be repository id

dampir:
As part of my work on #1120 I've added new api urls to get access by repository id.

Here one example of new method

Protorype:
```cs
/// <summary>
/// Returns the <see cref="Uri"/> that returns the assets specified by the asset id.
/// </summary>
/// <param name="id">The id of the release asset</param>
/// <returns></returns>
public static Uri Asset(string owner, string name, int id)
{
        return "repos/{0}/{1}/releases/assets/{2}".FormatUri(owner, name, id);
}
```
New method:
```cs
/// <summary>
/// Returns the <see cref="Uri"/> that returns the assets specified by the asset id.
/// </summary>
/// <param name="repositoryId">The ID of the repository</param>
/// <param name="id">The id of the release asset</param>
/// <returns>The <see cref="Uri"/> that returns the assets specified by the asset id.</returns>
public static Uri Asset(int repositoryId, int id)
{
        return "repositories/{0}/releases/assets/{1}".FormatUri(repositoryId, id);
}
```
As you can see main difference in first part of url. Instead of url prefix "repos/{0}/{1}/" new method has
url prefix "repositories/{0}/" and use repositoryId parameter instead of owner and name parameters.

Also all new methods  have same XML documentation as their prototype.

PS: also there is one small update of XML documentation of Deployments method.

/cc @shiftkey , @ryangribble

### DIFF
--- a/Octokit/Helpers/ApiUrls.cs
+++ b/Octokit/Helpers/ApiUrls.cs
@@ -1918,5 +1918,933 @@ namespace Octokit
         {
             return "users/{0}/suspended".FormatUri(login);
         }
+
+        /// <summary>
+        /// Returns the <see cref="Uri"/> that returns the assets specified by the asset id.
+        /// </summary>
+        /// <param name="repositoryId">The ID of the repository</param>
+        /// <param name="id">The id of the release asset</param>
+        /// <returns>The <see cref="Uri"/> that returns the assets specified by the asset id.</returns>
+        public static Uri Asset(int repositoryId, int id)
+        {
+            return "repositories/{0}/releases/assets/{1}".FormatUri(repositoryId, id);
+        }
+
+        /// <summary>
+        /// Returns the <see cref="Uri"/> that returns all of the assignees to which issues may be assigned.
+        /// </summary>
+        /// <param name="repositoryId">The ID of the repository</param>
+        /// <returns>The <see cref="Uri"/> that returns all of the assignees to which issues may be assigned.</returns>
+        public static Uri Assignees(int repositoryId)
+        {
+            return "repositories/{0}/assignees".FormatUri(repositoryId);
+        }
+
+        /// <summary>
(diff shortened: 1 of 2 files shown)


Octokit/Helpers/ApiUrls.cs
```diff
@@ -1918,5 +1918,933 @@ public static Uri UserAdministrationSuspension(string login)
...
+        {
+            return "repositories/{0}/assignees".FormatUri(repositoryId);
+        }
+
+        /// <summary>
+        /// Returns the <see cref="Uri"/> for a specific blob.
+        /// </summary>
+        /// <param name="repositoryId">The ID of the repository</param>
+        /// <returns>The <see cref="Uri"/> for a specific blob.</returns>
+        public static Uri Blob(int repositoryId)
```
shiftkey: Perhaps the name on this should be `Blobs` to indicate it's not going to return a single value?

dampir: I agree, I'll fix that. Should I rename public static Uri Blob(string owner, string name) overload too?

M-Zuber: `public static Uri Blob(string owner, string name)` existed already, so you would need to add the `[Obsolete]` attribute - with a message pointing to the new method, and a new method of `public static Uri Blobs(string owner, string name)`.

dampir: @M-Zuber thank you! Done!

dampir:
@shiftkey all remarks are fixed.

dampir:
@shiftkey, @ryangribble could you merge it please? I want to start implement my GSOC proposal goal and this PR is the important part of it :pray:

shiftkey:
@dampir apologies for the delay!

dampir:
@shiftkey all fine, thanks! :smile:

PR #1292, opened 2016-05-12
Add repository property to Issue response model

### LINKED ISSUE #1243: Add Repository property to Issue response from API
So I want to query many repos within an org for issues with a given label. I could use `client.Issue.GetallForRepository` and call this for each repository, but this seems wasteful as I'm querying 51 repos.

Instead I would like to use `client.Issue.GetAllForOrganization` which worked fine, but now I have a list of issues and there's no easy way to tell which repo an issue came from.

Should org and repo data be included in the Issue object?

Tasks:

 - update `Issue` response to have `Repository` property
 - verify payload deserializes correctly
 - add/update integration test for new payload on `GetAllForOwnedAndMemberRepositories`
 - add/update integration test for new payload on `GetAllForCurrent`
 - add/update integration test for `GetAllForOrganization`

M-Zuber:
Fixes #1243

### DIFF
--- a/Octokit.Tests.Integration/Clients/IssuesClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/IssuesClientTests.cs
@@ -7,6 +7,7 @@ using Octokit;
 using Octokit.Tests.Integration;
 using Xunit;
 using Octokit.Tests.Integration.Helpers;
+using System.Collections.Generic;
 
 public class IssuesClientTests : IDisposable
 {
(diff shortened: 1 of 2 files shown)


Octokit.Tests.Integration/Clients/IssuesClientTests.cs
```diff
@@ -380,6 +381,40 @@ public IssuesClientTests()
         Assert.Equal(new Uri(string.Format(expectedUri, _context.RepositoryOwner, _context.RepositoryName, issue.Number, "events")), issue.EventsUrl);
     }
 
+    [IntegrationTest]
+    public async Task GetAlForCurrentContainsRepositoryData()
```
shiftkey: typo: `s/Al/All`

shiftkey:
Tests look good, just that one typo and this is good to go.

M-Zuber:
@shiftkey fixed

Issue #1263, opened 2016-04-11
improve documentation for interacting with Pages API


prayankmathur:
After having a discussion with @shiftkey and referring to the [docs](https://developer.github.com/v3/repos/pages/), we found that any normal integration tests won't be of any use as you need to be an authenticated owner of the repository to use the Pages API.

Refer to #1213 and especially conversation after this [comment](https://github.com/octokit/octokit.net/pull/1213#issuecomment-203437170) for more details.

shiftkey:
This is mentioned in the docs: https://developer.github.com/v3/repos/pages/

> Information about the site and the builds can only be accessed by authenticated owners, even though the websites are public.

PR #1304, opened 2016-05-18
Added ApiOption overloads to methods on I(Observable)RepositoryPagesClient

### LINKED ISSUE #1178: Add ApiOption overloads to methods on IRepositoryPagesClient

- Add an overload for this method on `IRepositoryPagesClient` with a new parameter `ApiOptions`
- Implement the new method on the class which implements `IRepositoryPagesClient`
- Add an overload for this method on `IObservableRepositoryPagesClient` with a new parameter `ApiOptions`
- Implement the new method on the class which implements `IObservableRepositoryPagesClient`

- [ ] `Task<IReadOnlyList<PagesBuild>> GetAll(String owner, String repositoryName)`

shiftkey:
Fixes #1178

Supersedes #1213 and addresses merge conflict


shiftkey:
@ryangribble can I get a :thumbsup: here before I merge this in?

ryangribble:
Looks pretty good, although I didn't see integration tests that actually test the pagination... Is that because there isn't a suitable repo that has enough responses?

shiftkey:
@ryangribble it's because anything related to Pages requires the account to be an administrator of the repository: https://github.com/octokit/octokit.net/issues/1263

ryangribble:
Yeah I saw the existing test was skipped, I guess I was just questioning whether there should still be the pagination test written, then also skipped? 

The code changes look fine though so 👍 if you aren't concerned about the missing test

shiftkey:
@ryangribble good point. I'll add the test so we don't forget it and then close this out...

PR #1308, opened 2016-05-20
Add ApiOption overloads to methods on I(Observable)ReleasesClient

### LINKED ISSUE #1169: Add ApiOption overloads to methods on IReleasesClient

- Add an overload for this method on `IReleasesClient` with a new parameter `ApiOptions`
- Implement the new method on the class which implements `IReleasesClient`
- Add an overload for this method on `IObservableReleasesClient` with a new parameter `ApiOptions`
- Implement the new method on the class which implements `IObservableReleasesClient`

- [x] `Task<IReadOnlyList<Release>> GetAll(String owner, String name)`
- [x] `Task<IReadOnlyList<Release>> GetAll(String owner, String name, ApiOptions options)`
- [ ] `Task<IReadOnlyList<ReleaseAsset>> GetAllAssets(String owner, String name, Int32 id)`

dampir:
Fixes #1169

To-do:

- [x]  Add overloads on IReleasesClient.
- [x]  Add overloads on IObservableReleasesClient.
        Also I've added await\async modifiers into unit tests class ReleasesClientTests.cs , because I suppose these modifiers should be there to get determined results of tests, shouldn't be?

/cc @shiftkey , @ryangribble all is done, but there are some strange failures in test cases.

### DIFF
--- a/Octokit.Tests/Clients/ReleasesClientTests.cs
+++ b/Octokit.Tests/Clients/ReleasesClientTests.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Collections.Generic;
 using System.IO;
 using System.Threading.Tasks;
 using NSubstitute;
@@ -249,7 +248,7 @@ namespace Octokit.Tests.Clients
                 await Assert.ThrowsAsync<ArgumentNullException>(() => client.GetAllAssets("owner", null, 1));
 
                 await Assert.ThrowsAsync<ArgumentNullException>(() => client.GetAllAssets(null, null, 1, null));
-                await Assert.ThrowsAsync<ArgumentNullException>(() => client.GetAllAssets(null, null, 1, Args.ApiOptions));
+                await Assert.ThrowsAsync<ArgumentNullException>(() => client.GetAllAssets(null, null, 1, ApiOptions.None));
                 await Assert.ThrowsAsync<ArgumentNullException>(() => client.GetAllAssets(null, "name", 1, null));
                 await Assert.ThrowsAsync<ArgumentNullException>(() => client.GetAllAssets("owner", null, 1, null));
 
--- a/Octokit.Tests/Clients/ReleasesClientTests.cs
+++ b/Octokit.Tests/Clients/ReleasesClientTests.cs
@@ -232,7 +232,8 @@ namespace Octokit.Tests.Clients
 
                 await client.GetAllAssets("fake", "repo", 1, options);
 
-                connection.Received().GetAll<ReleaseAsset>(Arg.Is<Uri>(u => u.ToString() == "repos/fake/repo/releases/1/assets"),
+                connection.Received().GetAll<ReleaseAsset>(
+                    Arg.Is<Uri>(u => u.ToString() == "repos/fake/repo/releases/1/assets"),
                     null,
                     "application/vnd.github.v3", options);
             }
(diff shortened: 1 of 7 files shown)


dampir:
Somehow these tests are working on my machine from time to time and do not work on CI. I've tried to understand what is wrong with RequestsTheCorrectUrlWithApiOptions test case, but I haven't saw any errors. It's looks like multithread error somewhere and at this moment I can't find out where it is.

shiftkey:
@dampir can you try defining the test as `public async Task` to see if that addresses it at all?

dampir:
@shiftkey yeah, I fixed that, it's my misspell. But anyway, the tests failed. As I see there is must be call received by ApiConnection Substitution. And these tests is working on my PC for now. But CI cannot pass them. I really confused such sutuation..

dampir:
@shiftkey, @ryangribble I've fixed errors in tests, accidentally I've used Args.ApiOptions instead of ApiOptions.None and it gives strange result: in parallel tests falls, but in single thread they executed successfully. I found this info [`Unit Test fails on Run All but not on Run Selected Tests with all selected`](http://stackoverflow.com/questions/14349512/unit-test-fails-on-run-all-but-not-on-run-selected-tests-with-all-selected) and fixed this error. So, now all tests are passed.

ryangribble:
Cool, nice find!  I'll review these tomorrow

shiftkey:
> I've fixed errors in tests, accidentally I've used Args.ApiOptions instead of ApiOptions.None and it gives strange result: in parallel tests falls, but in single thread they executed successfully. 

So I don't know the full answer here for the runtime quirks when running tests but I can at least recommend we don't use `Args.ApiOptions` (the wrapper around NSubstitute's assert helper `Arg.Any<ApiOption>()`) unless you're asserting, instead of `ApiOptions.None` (which is the Octokit model).

Not sure if we need to do anything else here, but something to be mindful of...

shiftkey:
Changes look good on my end. I'll let @ryangribble chime in, otherwise I'll merge it after my talk is done tomorrow.

Thanks @dampir!

dampir:
> Not sure if we need to do anything else here, but something to be mindful of...

Yeah, definitely no any changes have to be done in such situation, it is just interesting behaviour that blowed my mind for 2 days :smiley:. Maybe it would be useful information for someone :+1:

Octokit.Tests/Clients/ReleasesClientTests.cs
```diff
@@ -204,42 +204,71 @@ public void RequestsTheCorrectUrl()
...
+                    "application/vnd.github.v3", options);
             }
 
             [Fact]
             public async Task EnsuresNonNullArguments()
             {
                 var client = new ReleasesClient(Substitute.For<IApiConnection>());
-
+                
+                await Assert.ThrowsAsync<ArgumentNullException>(() => client.GetAllAssets(null, null, 1));
```
ryangribble: overkill on the null and empty string checks

Octokit.Tests/Reactive/ObservableReleasesClientTests.cs
```diff
@@ -181,17 +181,47 @@ public void RequestsTheCorrectUrl()
...
+                    Arg.Is<IDictionary<string, string>>(dictionary => dictionary.Count == 2), 
+                    null);
             }
 
             [Fact]
             public void EnsuresNonNullArguments()
             {
                 var client = new ObservableReleasesClient(Substitute.For<IGitHubClient>());
 
+                Assert.Throws<ArgumentNullException>(() => client.GetAllAssets(null, null, 1));
```
ryangribble: and here

dampir:
@ryangribble all remarks were fixed :+1:

ryangribble:
Pesky travis!  All :gem: now, thanks!

PR #1311, opened 2016-05-23
Redundant ApiConnection field was removed

dampir:
Just removed redundant ApiConnection field and use protected property ApiConection instead.

/cc @shiftkey , @ryangribble


ryangribble:
👍

PR #1313, opened 2016-05-23
Add reactions for commit comments

lrz-hal:
I have added feature to add reactions for commit comments. Please let me know if this is the right way you imagine so i can go on with issues, pull requests and so on.

Thx,

Martin


Octokit.Tests.Integration/Clients/RepositoryCommitsClientTests.cs
```diff
@@ -292,6 +292,8 @@ public TestsWithNewRepository()
             return await _github.Git.Commit.Create(Helper.UserName, _context.RepositoryName, newCommit);
         }
 
+       
+
```
dampir: Unintended newlines?

Octokit.Tests.Integration/Clients/RepositoryCommitsClientTests.cs
```diff
@@ -154,7 +154,7 @@ public TestsWithExistingRepositories()
             var sha1 = await _fixture.GetSha1("octokit", "octokit.net", "master");
 
             Assert.NotNull(sha1);
-        }
+        }
```
dampir: Unintended tab?

Octokit/Helpers/Enums.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+
+        Plus1 = 0,
+        Minus1 = 1,
+        Laugh = 2,
+        Confused = 3,
+        Heart = 4,
+        Hooray = 5
+    }
+
```
dampir: Unintended newline?

Octokit/Models/Response/CommitCommentReaction.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// </summary>
+        public int UserId { get; protected set; }
+
+        /// <summary>
+        /// The reaction type for this commit comment.
+        /// </summary>
+        public String Content { get; protected set; }
+
+
+
```
dampir: I think one empty line would be enough :smile:

dampir:
Hi @lrz-hal ! I see you do not implement Observable version of new methods, Unfortunately in this case CI cannot be pass successfully. Also you should add new unit tests for Observable client, as far as new integration tests for new methods.

Octokit/Models/Request/NewCommitCommentReaction.cs
```diff
@@ -0,0 +1,52 @@
...
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
+    public class NewCommitCommentReaction
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="NewCommitCommentReaction"/> class.
+        /// </summary>
+        /// <param name="content">The reaction type.</param>
+        public NewCommitCommentReaction(EnumReaction content)
+        {
+            Ensure.ArgumentNotNull(content, "content");
```
M-Zuber: Enums can not be null unless specified as <enumType>?
([see here](https://csharp.2000things.com/2011/01/11/208-you-can-make-any-value-type-nullable/) for more information)

M-Zuber: Actually you should be using [Paramater attributes](https://github.com/octokit/octokit.net/blob/f354d1bf00ff7606b46489e3a915e1428414bc47/Octokit/Helpers/ParameterAttribute.cs) on the enum value, which enables them to be automatically serialized.
[see here](https://github.com/octokit/octokit.net/blob/9231286fedce0c7a5f986d72cf1012a95941c62f/Octokit/Models/Request/MilestoneRequest.cs#L43) for an example

Octokit/Models/Response/CommitCommentReaction.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// <summary>
+        /// The UserId.
+        /// </summary>
+        public int UserId { get; protected set; }
+
+        /// <summary>
+        /// The reaction type for this commit comment.
+        /// </summary>
+        public String Content { get; protected set; }
```
M-Zuber: This should be `EnumReaction` type.

Would also suggest moving the enum definition into this file, and not it's own separate file.

lrz-hal:
Ok. Thanks for the notes. I will add this asap.

Octokit/Helpers/Enums.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Octokit
+{
+    public enum EnumReaction
```
M-Zuber: Naming things is hard (™ by @shiftkey :stuck_out_tongue_winking_eye:) but this one kinda grates on me.

Issue #1315, opened 2016-05-24
Do Contents support pagination?


dampir:
I guess that I've found a bug in Github API. I was implementing pagination support for IRepositoryContentsClient and found out that Contents do not support pagination.

This query

[](https://api.github.com/repos/octokit/octokit.net/contents?page=1&per_page=2)

return all contents of repo, but should return only 2 items.

/cc @shiftkey, @ryangribble

shiftkey:
Let's not worry about this one - because of the nuance of serving up a directory versus serving up a file, I don't expect pagination to be added here anytime soon.

PR #1262, opened 2016-04-10
[WIP] Add ApiOption overloads to methods on IRepositoryCommentsClient

prayankmathur:
Fixes #1172

To Do

- [x] Add overload to `IRepositoryCommentsClient` and `IObservableRespositoryCommentsClient`
- [x] Implement the above methods in the respective classes
- [x] Add unit tests for `IRepositoryCommentsClient`
- [x] Add unit tests for `IObservableRespositoryCommentsClient`
- [ ] Add integration tests for `IRepositoryCommentsClient`
- [ ] Add integration tests for `IObservableRepositoryCommentsClient`

Issue #1174, opened 2016-03-14
Add ApiOption overloads to methods on IRepositoryContentsClient


shiftkey:
As part of completing #760 a number of overloads need to be added to the client interfaces.

As an example, this has been added to [`IReleasesClient`](https://github.com/octokit/octokit.net/commit/03ea83148995711fc47148489e2123e65f33bd82)and [`IObservableReleasesClient`](https://github.com/octokit/octokit.net/commit/aa3ce0db9210b778414e290af7f17c4a646bdeec).

The checklist for porting each method is:

- Add an overload for this method on `IRepositoryContentsClient` with a new parameter `ApiOptions`
- Implement the new method on the class which implements `IRepositoryContentsClient`
- Add an overload for this method on `IObservableRepositoryContentsClient` with a new parameter `ApiOptions`
- Implement the new method on the class which implements `IObservableRepositoryContentsClient`

This needs to be done for each of these methods:

- [ ] `Task<IReadOnlyList<RepositoryContent>> GetAllContents(String owner, String name, String path)`
- [ ] `Task<IReadOnlyList<RepositoryContent>> GetAllContents(String owner, String name)`
- [ ] `Task<IReadOnlyList<RepositoryContent>> GetAllContentsByRef(String owner, String name, String path, String reference)`
- [ ] `Task<IReadOnlyList<RepositoryContent>> GetAllContentsByRef(String owner, String name, String reference)`

shiftkey:
Closing this out as the underlying API doesn't support pagination.

PR #1305, opened 2016-05-19
Add repositoryId overloads to methods on I(Observable)RepoCollaboratorsClient 

dampir:
As part of my work on #1120 I've added new overloads on I(Observable)RepoCollaboratorsClient to get access by repository id.

- [x] **Rename parameter "repo" to "name" (in order to deliver more consistency into the codebase).**

	 I've found out that there is the pair (owner, name) is used to identify repository. But some clients use pair (owner, repo) and it's some type of inconsistency.
	  So I've decided to fix these cases during my work on #1120.
- [x] **Update XML documentation of interface methods of clients (also synchronize XML docs of IRepoCollaboratorsClient and IObservableRepoCollaboratorsClient).**

	  There is some divergence between XML documentation of methods in IRepoCollaboratorsClient and IObservableRepoCollaboratorsClient. So I've decided 
	  to sync XML documentation of these classes during my work on #1120.
- [x] **Add overloads to IRepoCollaboratorsClient.**

	  Just add overloads of existing methods that use repositoryId to work with repo.
- [x] **Add overloads to IObservableRepoCollaboratorsClient.**

	  Just add overloads of existing methods that use repositoryId to work with repo.
- [x] **Add unit tests.**

	  I've added new unit tests that use repositoryId to work with repo that is just a full copy of existing tests that use (owner, name) key.
	  Also I've found out that not all methods are covered by tests and added them for new and for old methods.
- [x] **Add integration tests.**

	  I've added new integration tests that use repositoryId to work with repo that is just a full copy of existing tests that use (owner, name) key.
	  Also I've found out that not all methods are covered by tests and added them for new and for old methods.

I think that we should discuss and review this PR as thorough as we can because it's first PR in series of PRs that should add ability to work with repo by their Id. :smile: 

/cc @shiftkey, @ryangribble

PR #1316, opened 2016-05-24
Add ApiOption overloads to methods on I(Observable)RepositoryContentsClient

dampir:
Refer #1174

Despite on this #1262, I've do this by myself, because we should finish Pagination milestone in near future in order to start working on repositoryId functionality (#1120).

To-do:

- [x]  Add overloads on IRepositoryContentsClient.
- [x]  Add overloads on IObservableRepositoryContentsClient.
    
    I've add whole new integration test class for RepositoryCommentsClient, but I found out that [Contents](https://developer.github.com/v3/repos/contents/) **do not support pagination** (#1315), so I marked tests as skipped.

/cc @shiftkey , @ryangribble

### DIFF
--- a/Octokit.Reactive/Clients/ObservableRepositoryContentsClient.cs
+++ b/Octokit.Reactive/Clients/ObservableRepositoryContentsClient.cs
@@ -105,9 +105,32 @@ namespace Octokit.Reactive
             Ensure.ArgumentNotNullOrEmptyString(name, "name");
             Ensure.ArgumentNotNullOrEmptyString(path, "path");
 
+            return GetAllContents(owner, name, path, ApiOptions.None);
+        }
+
+        /// <summary>
+        /// Returns the contents of a file or directory in a repository.
+        /// </summary>
+        /// <remarks>
+        /// If given a path to a single file, this method returns a collection containing only that file.
+        /// </remarks>
+        /// <param name="owner">The owner of the repository</param>
+        /// <param name="name">The name of the repository</param>
+        /// <param name="path">The content path</param>
+        /// <param name="options">Options for changing the API response</param>
+        /// <returns>
+        /// A collection of <see cref="RepositoryContent"/> representing the content at the specified path
+        /// </returns>
+        public IObservable<RepositoryContent> GetAllContents(string owner, string name, string path, ApiOptions options)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
+            Ensure.ArgumentNotNullOrEmptyString(name, "name");
+            Ensure.ArgumentNotNullOrEmptyString(path, "path");
+            Ensure.ArgumentNotNull(options, "options");
+
             return _client
                 .Connection
-                .GetAndFlattenAllPages<RepositoryContent>(ApiUrls.RepositoryContent(owner, name, path));
+                .GetAndFlattenAllPages<RepositoryContent>(ApiUrls.RepositoryContent(owner, name, path), options);
         }
 
         /// <summary>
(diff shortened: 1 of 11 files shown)


dampir:
Tests are passed, but Mono SIGABRT fails again.

Octokit.Reactive/Clients/ObservableRepositoryContentsClient.cs
```diff
@@ -17,6 +17,8 @@ public class ObservableRepositoryContentsClient : IObservableRepositoryContentsC
         /// <param name="client"></param>
         public ObservableRepositoryContentsClient(IGitHubClient client)
         {
+            Ensure.ArgumentNotNull(client, "client");
```
ryangribble: nice pickup!

Octokit.Tests/Clients/RepositoryContentsClientTests.cs
```diff
@@ -68,35 +68,394 @@ public class TheGetReadmeHtmlMethod
...
             [Fact]
-            public async Task ReturnsContentsByRef()
+            public async Task ReturnsContents()
             {
-                List<RepositoryContent> result = new List<RepositoryContent> { new RepositoryContent() };
+                var result = new List<RepositoryContent> { new RepositoryContent() };
 
                 var connection = Substitute.For<IApiConnection>();
-                connection.GetAll<RepositoryContent>(Args.Uri).Returns(Task.FromResult(result.AsReadOnly() as IReadOnlyList<RepositoryContent>));
+                connection.GetAll<RepositoryContent>(Args.Uri, Args.ApiOptions).Returns(info => Task.FromResult(result.AsReadOnly() as IReadOnlyList<RepositoryContent>));
```
ryangribble: is there a need to declare `info =>` here?

dampir: No, at all! Fixed!

Octokit.Tests/Clients/RepositoryContentsClientTests.cs
```diff
@@ -68,35 +68,394 @@ public class TheGetReadmeHtmlMethod
...
+            }
+
+            [Fact]
+            public async Task EnsuresNonNullarguments()
+            {
+                var connection = Substitute.For<IApiConnection>();
+                var client = new RepositoryContentsClient(connection);
+
+                // public Task<IReadOnlyList<RepositoryContent>> GetAllContents(string owner, string name)
+                await Assert.ThrowsAsync<ArgumentNullException>(() => client.GetAllContents(null, null));
```
ryangribble: For the null checks dont we just need to check each argument is null, not also checking multiple arguments null at the same time?  So in this function all the ones with more than one argument being null at the same time, can be removed I think

Octokit.Tests/Clients/RepositoryContentsClientTests.cs
```diff
@@ -68,35 +68,394 @@ public class TheGetReadmeHtmlMethod
...
+                
+                // empty string checks
+
+                // public Task<IReadOnlyList<RepositoryContent>> GetAllContents(string owner, string name)
+                await Assert.ThrowsAsync<ArgumentException>(() => client.GetAllContents("", ""));
+                await Assert.ThrowsAsync<ArgumentException>(() => client.GetAllContents("", "name"));
+                await Assert.ThrowsAsync<ArgumentException>(() => client.GetAllContents("owner", ""));
+
+                // public Task<IReadOnlyList<RepositoryContent>> GetAllContents(string owner, string name, ApiOptions options)
+                await Assert.ThrowsAsync<ArgumentException>(() => client.GetAllContents("", "", (ApiOptions)null));
```
ryangribble: Same on all the empty string checks as above comment on null checks.  Each parameter needs to be checked, but not combinations of multiple parameters

Octokit/Models/Response/Readme.cs
```diff
@@ -30,6 +30,22 @@ internal Readme(ReadmeResponse response, IApiConnection client)
             htmlContent = new Lazy<Task<string>>(async () => await client.GetHtml(Url).ConfigureAwait(false));
         }
 
+        internal Readme(ReadmeResponse response, IConnection client)
```
ryangribble: What's the reasoning behind adding this additional constructor?  I notice it's only used in the Observable tests, but couldnt they just pass in the IApiConnection and use the existing constructor, rather than passing in an IConnection and having an entire new constructor (and mostly duplicated logic)?  Apologies if i've missed something, as im on my phone :grinning:

ryangribble:
Looking good, ive made a few comments.

BUT the big thing that's bothering me is... if the underlynig github API appears to not actually support pagination for these methods, then we can't implement the `ApiOptions` overloads as it would give users the misconception that they could request paginated results.  

@shiftkey might want to chime in on this, but my thoughts would be that we shouldnt implement this, until/unless it turns out to be a github API issue that is later resolved...

Octokit.Tests/Clients/RepositoryContentsClientTests.cs
dampir: yep, it is overkill really!

Octokit/Models/Response/Readme.cs
dampir: Yep, you're right! I've missed opportunity to use substitution of IApiConnection instead of new Readme constructor. Fixed :+1:

dampir:
@ryangribble all remarks were fixed, thanks!

@shiftkey I want to add that [Statistics](https://developer.github.com/v3/repos/statistics/) do not support pagination as I mentioned in #1305, also I opened an issue #1315 for further discussion, but maybe it is overkill and we can solve this question here.

shiftkey:
Looking at the docs for [Get Contents](https://developer.github.com/v3/repos/contents/#get-contents), you have these headers returned:





Apologies @dampir - I really don't think it's worth implementing this, and I should have caught it when opening the original issue...

PR #1317, opened 2016-05-24
Fixed incorrect names of tests in RepositoryDeployKeysClientTests

dampir:
I've fixed incorrect names of tests in RepositoryDeployKeysClientTests.

/cc @shiftkey, @ryangribble


ryangribble:
good pickup although I also see that variables inside the test are named `hooksCount` and so on, so probably should fix those up at the same time?

dampir:
@ryangribble good remark, fixed :smile: !

PR #1324, opened 2016-05-25
Add ApiOption overloads to methods on I(Observable)OrganizationsClient

### LINKED ISSUE #1165: Add ApiOption overloads to methods on IOrganizationsClient

- Add an overload for this method on `IOrganizationsClient` with a new parameter `ApiOptions`
- Implement the new method on the class which implements `IOrganizationsClient`
- Add an overload for this method on `IObservableOrganizationsClient` with a new parameter `ApiOptions`
- Implement the new method on the class which implements `IObservableOrganizationsClient`

- [ ] `Task<IReadOnlyList<Organization>> GetAll(String user)`
- [ ] `Task<IReadOnlyList<Organization>> GetAllForCurrent()`

dampir:
Fixes #1165 

To-do:

- [x]  Add overloads on IOrganizationsClient
- [x]  Add overloads on IObservableOrganizationsClient
- [x]  Add unit tests for these new methods
- [x]  Add integration tests for these new methods

    I had to use [GitHubEnterpriseTest] to get these tests done, because I should create organizations firstly.

/cc @shiftkey, @ryangribble

### DIFF
--- a/Octokit.Tests.Integration/Clients/OrganizationClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/OrganizationClientTests.cs
@@ -117,7 +117,7 @@ namespace Octokit.Tests.Integration.Clients
 
                 var secondPage = await _organizationsClient.GetAllForCurrent(skipStartOptions);
 
-                Assert.NotEqual(firstPage[0].Name, secondPage[0].Name);
+                Assert.NotEqual(firstPage[0].Login, secondPage[0].Login);
             }
         }
     }
--- a/Octokit.Tests.Integration/Clients/OrganizationClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/OrganizationClientTests.cs
@@ -1,0 +1,124 @@
+﻿using System.Threading.Tasks;
+using Xunit;
+
+namespace Octokit.Tests.Integration.Clients
+{
+    public class OrganizationClientTests
+    {
+        public class TheGetAllMethod
+        {
+            readonly IGitHubClient _github;
+            readonly IOrganizationsClient _organizationsClient;
+
+            public TheGetAllMethod()
+            {
+                _github = EnterpriseHelper.GetAuthenticatedClient();
+
+                _organizationsClient = _github.Organization;
+            }
+
+            [GitHubEnterpriseTest]
+            public async Task CanListOrganizations()
+            {
+                string orgLogin = Helper.MakeNameWithTimestamp("MyOrganization");
+                string orgName = string.Concat(orgLogin, " Display Name");
+
+                var newOrganization = new NewOrganization(orgLogin, EnterpriseHelper.UserName, orgName);
+                var organization = await
+                    _github.Enterprise.Organization.Create(newOrganization);
+
+                Assert.NotNull(organization);
+
+                var milestones = await _organizationsClient.GetAllForCurrent();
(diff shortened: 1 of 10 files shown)


ryangribble:
Have you got a github enterprise instance to run the integration tests against or would you like me to do that?

dampir:
No, I haven't. Unfortunately, I don't know a lot of about enterprise integration test procedure, so do it at your discretion :smile:

ryangribble:
> Anyway, if there is exists an possibility to provide me test enterprise credentials for period of GSoC, I think that I could run and debug enterprise integration tests by myself :+1:

[GitHub Enterprise](https://enterprise.github.com/home) is a version of GitHub that an organization hosts on their own servers (either on prem or cloud).  They do have a free 45 day trial though, so potenitally you could spin one up if you have AWS or Azure credit, or enough hardware at home for the minimum spec (16GB RAM).  But it really isn't necessary

I've pulled your stuff down and ran all the integration tests, you've just got one issue which ill highlight

Octokit.Tests.Integration/Clients/OrganizationClientTests.cs
```diff
@@ -0,0 +1,124 @@
...
+                var skipStartOptions = new ApiOptions
+                {
+                    PageSize = 1,
+                    PageCount = 1,
+                    StartPage = 2
+                };
+
+                var secondPage = await _organizationsClient.GetAllForCurrent(skipStartOptions);
+
+                Assert.NotEqual(firstPage[0].Name, secondPage[0].Name);
```
ryangribble: `.Name` is `null` on - you should use `.Login` field and then everything is green!  :tada:

dampir: It's cool, especially if take into account that I wrote these tests blindly :smile: 
Fixed!

dampir:
> GitHub Enterprise is a version of GitHub that an organization hosts on their own servers (either on prem or cloud). They do have a free 45 day trial though, so potenitally you could spin one up if you have AWS or Azure credit, or enough hardware at home for the minimum spec (16GB RAM).

Thank you for good explanation, now I really get it.

> But it really isn't necessary

If so, I really hope that there are not so many clients that should use Enterprise API (throughout my GSoC) :smile:

Octokit.Tests.Integration/Clients/OrganizationClientTests.cs
```diff
@@ -0,0 +1,124 @@
...
+                    PageCount = 1,
+                    StartPage = 2,
+                };
+
+                var organizations = await _organizationsClient.GetAllForCurrent(options);
+
+                Assert.Equal(1, organizations.Count);
+            }
+
+            [GitHubEnterpriseTest]
```
shiftkey: This seems reasonable. I'm not sure how we feel about creating multiple organizations for each test run - @ryangribble do you think this is worth doing simply to test pagination?

shiftkey:
Gonna merge this in and we can come back to that open question about running the tests cc @ryangribble

ryangribble:
I think it's ok creating the orgs, my test ghe instance already has a number of random organisations created from other tests I'd written previously... They don't really affect anything and anyone running integration tests on ghe should be using a dedicated  dev instance anyway

PR #1338, opened 2016-06-03
Add support for code searches without a search term

dsplaisted:
This change adds a parameter-less constructor to `SearchCodeRequest` and modifies `BaseSearchRequest` to handle the case where the 'Term' is null.  This supports the scenario of searching for a filename without specifying a search term, as in this search: https://github.com/dotnet/corefx/search?utf8=%E2%9C%93&q=filename%3Aproject.json


shiftkey:
@dsplaisted thanks for adding in the new test. This will go out with a release early next week! :metal:

PR #1335, opened 2016-05-30
stand alone reaction client

maddin2016:
First preview of an stand alone reaction client. Initially only for commit comments. I created this branch from #1325.

Fixes #1299 #1300 #1301 #1302 #1303

PR #1339, opened 2016-06-03
Add multiple assignees to issues

### LINKED ISSUE #1334: Support for multiple assignees
As per https://github.com/blog/2178-multiple-assignees-on-issues-and-pull-requests

This feature is amazeballs, and [we](https://github.com/Particular) want to start using it ASAP. Currently, we list all the assignees in a CSV string in the issue description (e.g.https://github.com/Particular/NServiceBus/issues/3342), and we have [some automation](https://github.com/Particular/PBot/blob/master/src/PBot/Issues/Pig.cs) to pull all of this out for each person to be able to see which issues they are involved in. We also include any issues to which someone is assigned. To allow us to gradually transition to using multiple assignees instead, we'd like this automation to understand multiple assignees too, but it looks like Octokit doesn't yet support this, would that be correct? I.e.

``` C#
public class RepositoryIssueRequest : IssueRequest
{
    public string Assignee { get; set; }
    public string Creator { get; set; }
    public string Mentioned { get; set; }
    public string Milestone { get; set; }
}
```

maddin2016:
Fixes #1334 
This is a first preview for add multiple assignees to an issue. I created a new class IssueAssignees because the key `assignees` is only added for this special api call. Maybe `AddAssignees` can live directly under `client.Issue`. Please tell me if this goes in the right direction.


shiftkey:
@maddin2016 I'm going to hold off on reviewing this until we get #1335 merged in.

Also cc @paladique

paladique:
Oh wow :sparkles: this is awesome work!

maddin2016:
Thank you : smile: What do you think where to place `AddAssignees`?

paladique:
@maddin2016 So I was :thought_balloon: about this last night too before I saw this! My vote is `client.Issue`, purely because of how the [API ](https://developer.github.com/v3/issues/assignees/)is set up and I'm assuming that's the first place users (or someone like me) would look

maddin2016:
Ok. Then I place the method under Issue. Is there something more to do?

paladique:
@maddin2016 I'll take a look over it, and make any comments I might have :question: on but I'll leave all final merges etc to @shiftkey and co.

ryangribble:
These 2 new calls actually belong under the `AssigneesClient`(https://github.com/octokit/octokit.net/blob/master/Octokit/Clients/AssigneesClient.cs).  

The API documentation [here](https://developer.github.com/v3/issues/assignees/) shows existing calls like "List Available Assignees" and "Check Assignees" which are implemented in this client already, so now these 2 new preview methods "Add assignees to issue" and "Remove assignees from issue" have been added, they should be added to this `AssignessClient` as well, not to `IssuesClient` directly.

I also am not sold on having a special `IssueAssignees` class that inherits `Issue` and adds the new `Assignees` property.  There are plenty of other examples where a particular field is on a response object and can sometimes be `null` depending on which API call was made or other factors... but we dont want to have too much split in the object model and have all these special objects that extend the base one and just add one field.  So I would suggest simpy adding the `Assignees` property to `Issue` model object.  

The new assignees field is also going to be returned on the ["Create an Issue"](https://developer.github.com/v3/issues/#create-an-issue) and ["Edit an Issue"](https://developer.github.com/v3/issues/#edit-an-issue) calls on IssuesClient, and I'd actually be quite surprised if it also wasn't going to be returned on the "List Issues" and "Get a single issue" methods.  The API docs don't indicate they have the preview functionality at the moment but it could even be a missed API doc update... or perhaps they really dont return the new field now, but surely in the future it must be planned (why wouldnt you want to know multiple assignees details when listing issues?!).  Therefore if we include `Assignees` on the `Issue` object, we wont need to change anything in the future if thoe calls do start returning the info


On a side note, I've got some local changes done for the Create and Edit issue calls, I might send a pull request to your fork @maddin2016 and then you can pull them into this PR?

ryangribble:
> I'd actually be quite surprised if it also wasn't going to be returned on the "List Issues" and "Get a single issue" methods

OK so I just did some poking... I added `Assignees` field to the `Issue` model class, and added the new preview header to the `IssuesClient.Get()` and `IssuesClient.GetAllForRepository()` calls and the new field IS populated in the response object :tada:   So definitely go with including this on the base issue and we can also set the accepts header on the Get/List calls.  

image


@shiftkey @paladique @shana could you possibly find out if an API docs update was missed for the Get and List issues calls supporting the new multiple assignees and taking the preview header?

maddin2016:
Ok. Then I roll back and add the `Assignees` property to Issue class. In the meantime i added method to remove assignees

Octokit.Tests/Reactive/ObservableAssigneesClientTests.cs
```diff
@@ -110,6 +110,68 @@ public void EnsuresNonEmptyArguments()
...
+            }
+
+            [Fact]
+            public void EnsuresNonNullArguments()
+            {
+                var githubClient = Substitute.For<IGitHubClient>();
+                var client = new ObservableAssigneesClient(githubClient);
+                var newAssignees = new AssigneesUpdate(new List<string>() { "assignee1", "assignee2" });
+
+                Assert.Throws<ArgumentNullException>(() => client.AddAssignees(null, null, 2, newAssignees));
```
ryangribble: these should be `RemoveAssignees` rather than `AddAssignees`

Octokit.Tests/Reactive/ObservableAssigneesClientTests.cs
```diff
@@ -110,6 +110,68 @@ public void EnsuresNonEmptyArguments()
...
```
ryangribble: you dont need to test combinations of more than one field being null or empty at a time, since the first assert to match, will throw an exception.

In other words if there are 3 input fields that have NotNull asserts, there should be 3 test cases (each one specifying null for 1 field and valid value for the other fields), eg:
null, "name", 2, newAssignee
"owner", null, 2, newAssignee
"owner", "name", 2, null

Octokit.Tests/Reactive/ObservableAssigneesClientTests.cs
```diff
@@ -110,6 +110,68 @@ public void EnsuresNonEmptyArguments()
...
+                Assert.Throws<ArgumentNullException>(() => client.AddAssignees(null, "name", 2, null));
+                Assert.Throws<ArgumentNullException>(() => client.AddAssignees("owner", null, 2, null));
+
+                Assert.Throws<ArgumentException>(() => client.AddAssignees("owner", "", 2, newAssignees));
+                Assert.Throws<ArgumentException>(() => client.AddAssignees("", "name", 2, newAssignees));
+                Assert.Throws<ArgumentException>(() => client.AddAssignees("", "", 2, newAssignees));
```
ryangribble: This one isnt needed since you already have checked both owner and name field being blank

Octokit/Models/Response/Issue.cs
```diff
@@ -10,7 +11,7 @@ public class Issue
     {
         public Issue() { }
 
-        public Issue(Uri url, Uri htmlUrl, Uri commentsUrl, Uri eventsUrl, int number, ItemState state, string title, string body, User closedBy, User user, IReadOnlyList<Label> labels, User assignee, Milestone milestone, int comments, PullRequest pullRequest, DateTimeOffset? closedAt, DateTimeOffset createdAt, DateTimeOffset? updatedAt, int id, bool locked, Repository repository)
+        public Issue(Uri url, Uri htmlUrl, Uri commentsUrl, Uri eventsUrl, int number, ItemState state, string title, string body, User closedBy, User user, IReadOnlyList<Label> labels, User assignee, Milestone milestone, int comments, PullRequest pullRequest, DateTimeOffset? closedAt, DateTimeOffset createdAt, DateTimeOffset? updatedAt, int id, bool locked, Repository repository, IReadOnlyList<User> assignees)
```
ryangribble: Ideally the parameter order should match the field order as per below, for consistency.  eg put the `assignees` parameter after `assignee` and before `milestone`

Octokit.Tests/Clients/AssigneesClientTests.cs
```diff
@@ -112,6 +112,60 @@ public class TheCheckAssigneeMethod
...
+                var connection = Substitute.For<IApiConnection>();
+                var client = new AssigneesClient(connection);
+
+                await client.AddAssignees("fake", "repo", 2, newAssignees);
+
+                connection.Received().Post<Issue>(Arg.Is<Uri>(u => u.ToString() == "repos/fake/repo/issues/2/assignees"), Arg.Any<object>(), "application/vnd.github.cerberus-preview");
+            }
+
+            [Fact]
+            public async Task EnsuresNonNullArguments()
```
ryangribble: these need to be reworked to test `null` and `""` values for any applicable field... at the moment `null` is not checked for `owner` or `repo`

Octokit.Tests/Clients/AssigneesClientTests.cs
```diff
@@ -112,6 +112,60 @@ public class TheCheckAssigneeMethod
...
+                var connection = Substitute.For<IApiConnection>();
+                var client = new AssigneesClient(connection);
+
+                await client.RemoveAssignees("fake", "repo", 2, removeAssignees);
+
+                connection.Received().Delete<Issue>(Arg.Is<Uri>(u => u.ToString() == "repos/fake/repo/issues/2/assignees"), Arg.Any<object>(), "application/vnd.github.cerberus-preview");
+            }
+
+            [Fact]
+            public async Task EnsuresNonNullArguments()
```
ryangribble: same here, rework to cover each parameter being `null` and `""`

Octokit.Tests.Integration/Clients/AssigneesClientTests.cs
```diff
@@ -38,4 +41,22 @@ public AssigneesClientTests()
         var assignees = await _github.Issue.Assignee.GetAllForRepository(_context.RepositoryOwner, _context.RepositoryName);
         Assert.True(assignees.Any(u => u.Login == Helper.UserName));
     }
+
+    [IntegrationTest]
+    public async Task CanAddAssignees()
```
ryangribble: Dont forget to add an integration test for `CanRemoveAssignees` method :grinning:

ryangribble:
Hey @maddin2016 looking good, I've marked up some comments.  

#1335 is the priority to get over the line first though...

Octokit.Tests/Reactive/ObservableAssigneesClientTests.cs
maddin2016: ok. i have copied this form another test. eg `ObservableRepositoryCommentsClientTests.cs`

Octokit.Tests.Integration/Clients/AssigneesClientTests.cs
maddin2016: Is it okto add the test into `AddAssignees`?

shiftkey: @maddin2016 :thumbsup:

Octokit.Tests.Integration/Clients/AssigneesClientTests.cs
```diff
@@ -4,16 +4,19 @@
 using Octokit.Tests.Integration;
 using Octokit.Tests.Integration.Helpers;
 using Xunit;
+using System.Collections.Generic;
```
shiftkey: :lipstick: sort these references so `System.*` appear first

Octokit.Tests.Integration/Clients/AssigneesClientTests.cs
```diff
@@ -4,16 +4,19 @@
...
 
 public class AssigneesClientTests
 {
     readonly IGitHubClient _github;
     readonly RepositoryContext _context;
+    private readonly IIssuesClient _issuesClient;
```
shiftkey: :lipstick: drop the `private` modifier here (it's a pet peeve of mine, but the other fields also lack this)

Octokit.Tests.Integration/Clients/AssigneesClientTests.cs
```diff
@@ -4,16 +4,19 @@
...
 
     public AssigneesClientTests()
     {
         _github = Helper.GetAuthenticatedClient();
         var repoName = Helper.MakeNameWithTimestamp("public-repo");
+        _issuesClient = _github.Issue;
```
shiftkey: If this is only referenced once, you might as well inline it into the method on line 51. Otherwise there's two usages in the test that can be replaced with this field. Up to you.

Octokit.Tests.Integration/Clients/AssigneesClientTests.cs
```diff
@@ -38,4 +41,28 @@ public AssigneesClientTests()
...
+        var issue = await _issuesClient.Create(_context.RepositoryOwner, _context.RepositoryName, newIssue);
+
+        Assert.NotNull(issue);
+
+        var addAssignees = await _github.Issue.Assignee.AddAssignees(_context.RepositoryOwner, _context.RepositoryName, issue.Id, newAssignees);
+
+        Assert.IsType<Issue>(addAssignees);
+
+        //Check if assignee was added to issue
+        Assert.True(addAssignees.Assignees.Where(x => x.Name == _context.RepositoryOwner).Any() == true);
```
shiftkey: :lipstick: `.Any()` has an overload which takes a predicate, so this can be simplified to:

```csharp
Assert.True(addAssignees.Assignees.Any(x => x.Name == _context.RepositoryOwner));
```

Octokit.Tests.Integration/Clients/AssigneesClientTests.cs
```diff
@@ -38,4 +41,28 @@ public AssigneesClientTests()
...
+
+        //Test to remove assignees
+        var removeAssignees = await _github.Issue.Assignee.RemoveAssignees(_context.RepositoryOwner, _context.RepositoryName, issue.Id, newAssignees);
+
+        //Check if assignee was removed
+        Assert.True(removeAssignees.Assignees.Where(x => x.Name == _context.RepositoryOwner).Any() == false);
```
shiftkey: Same thing here, but `Assert.False`

Octokit.Tests.Integration/Clients/AssigneesClientTests.cs
```diff
@@ -38,4 +41,28 @@ public AssigneesClientTests()
...
+    public async Task CanAddAndRemoveAssignees()
+    {
+        var newAssignees = new AssigneesUpdate(new List<string>() { _context.RepositoryOwner });
+        var newIssue = new NewIssue("a test issue") { Body = "A new unassigned issue" };
+
+        var issue = await _issuesClient.Create(_context.RepositoryOwner, _context.RepositoryName, newIssue);
+
+        Assert.NotNull(issue);
+
+        var addAssignees = await _github.Issue.Assignee.AddAssignees(_context.RepositoryOwner, _context.RepositoryName, issue.Id, newAssignees);
```
shiftkey: You'll need to use `issue.Number` here and on the other actions below (the user-friendly id) rather than `issue.Id` (essentially the database column).

Octokit.Tests.Integration/Clients/AssigneesClientTests.cs
shiftkey: Oh, and `x.Name` is not the user's account name - and not always populated from every endpoint.
Use `x.Login` here if you want to check the account name.

Octokit.Tests.Integration/Clients/IssuesClientTests.cs
```diff
@@ -86,23 +87,31 @@ public IssuesClientTests()
...
             Assert.NotNull(issue);
+            Assert.True(issue.Assignees.All(x => x.Login == _context.RepositoryOwner));
 
             var retrieved = await _issuesClient.Get(_context.RepositoryOwner, _context.RepositoryName, issue.Number);
             var all = await _issuesClient.GetAllForRepository(_context.RepositoryOwner, _context.RepositoryName);
             Assert.NotNull(retrieved);
-            Assert.True(all.Any(i => i.Number == retrieved.Number));
+            Assert.True(retrieved.Assignees.Count == 1);
+            Assert.True(retrieved.Assignees[0].Login == _context.RepositoryOwner);
+            Assert.True(all.Any(i => i.Number == retrieved.Number && i.Assignees.Count == 1 && i.Assignees[0].Login == _context.RepositoryOwner));
```
shiftkey: :lipstick: we're fetching both and doing a mix of assertions here:

- could we order this so we set and assert `retrieved`, and then set and assert `all`?
- could we use multiple `Assert.True` expressions here, rather than one big one sprinkled with `&&`s?

Octokit.Tests.Integration/Clients/IssuesClientTests.cs
```diff
@@ -86,23 +87,31 @@ public IssuesClientTests()
...
         }
         finally
         {
             var closed = _issuesClient.Update(_context.RepositoryOwner, _context.RepositoryName, issue.Number, new IssueUpdate { State = ItemState.Closed }).Result;
             Assert.NotNull(closed);
+            Assert.True(closed.Assignees.Count == 1);
```
shiftkey: :lipstick: `Assert.Equal(1, closed.Assignees.Count)` is nicer here because we get both values when it fails.

Octokit.Tests.Integration/Clients/IssuesClientTests.cs
```diff
@@ -86,23 +87,31 @@ public IssuesClientTests()
...
+            Assert.True(closed.Assignees[0].Login == _context.RepositoryOwner);
```
shiftkey: :lipstick: `Assert.Equal(_context.RepositoryOwner, closed.Assignees[0].Login)` will give us both values if this test fails

Octokit.Tests.Integration/Clients/IssuesClientTests.cs
```diff
@@ -233,6 +242,22 @@ public IssuesClientTests()
     }
 
     [IntegrationTest]
+    public async Task CanRetrieveIssueWithMultipleAssignees()
+    {
+        var issue = await _issuesClient.Get("octokit", "octokit.net", 1171);
+
+        Assert.True(issue.Assignees.Count == 2);
```
shiftkey: :lipstick: `Assert.Equal(2, issue.Assignees.Count)` to get the value if this test fails

Octokit.Tests.Integration/Clients/IssuesClientTests.cs
```diff
@@ -233,6 +242,22 @@ public IssuesClientTests()
...
+    }
+
+    [IntegrationTest]
+    public async Task CanRetrieveIssuesWithMultipleAssignees()
+    {
+        var issues = await _issuesClient.GetAllForRepository("octokit", "octokit.net");
+
+        Assert.True(issues.Any(x => x.Assignees.Count > 0));
```
shiftkey: Given `1171` is now a closed issue, is it worth keeping this around? Could we build a fresh repository to test `GetAll`, rather than querying for all the issues in the Octokit repository?

If not, I like this sort of test:

```csharp
var request = new RepositoryIssueRequest { State = ItemStateFilter.All };

var issues = await _issuesClient.GetAllForRepository("octokit", "octokit.net", request);

Assert.True(issues.Any(x => x.Assignees.Count > 1));
```

Octokit.Tests/Models/IssueTest.cs
```diff
@@ -257,7 +257,7 @@ public void CreatesAnIssueUpdateRequestObject()
 
             var update = issue.ToUpdate();
 
-            Assert.Null(update.Labels);
+            Assert.NotNull(update.Labels);
```
shiftkey: This is an interesting change - can you tell me what motivated this?

Octokit.Tests.Integration/Clients/AssigneesClientTests.cs
maddin2016: I'm also change `x.Name` to `x.Login`

Octokit.Tests.Integration/Clients/IssuesClientTests.cs
maddin2016: Done

shiftkey:
@maddin2016 this is looking really good. I focused on the integration tests just now as I plan to go on a bender cleaning them up and getting them working better/quicker, but I'll give the rest a pass tomorrow.

Octokit.Tests/Models/IssueTest.cs
maddin2016: Because in json String we add labels.
```
""labels"": [
{
    ""url"": ""https://api.github.com/repos/octocat/Hello-World/labels/bug"",
    ""name"": ""bug"",
    ""color"": ""f29513""
}
```
So this test will always fails if we test null.

shiftkey: Ok cool. Let me dig into this further as `ToUpdate` shouldn't create the `Labels` array by default unless it already exists on the issue. There's some [assumptions](https://developer.github.com/v3/issues/#edit-an-issue) about what we `POST` as part of editing an issue that may have changed (emphasis mine):

> Labels to associate with this issue. Pass one or more Labels to replace the set of Labels on this Issue. **Send an empty array (`[ ]`) to clear all Labels from the Issue.** NOTE: Only users with push access can set labels for issues. Labels are silently dropped otherwise.

ryangribble: That was actually my  commit I sent over to @maddin2016's branch. Sorry if it's not correct but it seemed to me .ToUpdate() shouldn't clear the labels from an issue?

shiftkey: @ryangribble it's all good, I just need to cast my mind back to the earlier discussions about this.

#718 is the original PR where I implementing this, and to TL;DR it:

 - initially, the result from `ToUpdate()` should have a value of `null` for `Labels` - this is how we say "don't change anything" to the server as part of the update, as `labels` won't be serialized
 - if you invoke `update.AddLabel("label")` we create a new list lazily, and this collection **replaces** whatever labels are previously set on the server
 - if you invoke `update.ClearLabels()` this creates an empty list, to indicate that all labels should be removed as part of the update 

It's might come across as a bit explicit, but I like the verbosity that it gives you in the code.

Anyway, coming back to this change. It looks like we're populating the `Labels` array from what exists in the issue, and while that's similar to the original intent I didn't want users to be manipulating the `Labels` collection directly to remove labels - I wanted to see them use the first-class `AddLabel` and `ClearLabel` methods on `IssueUpdate`.

ryangribble: Yeah that's a bit odd (IMO!).

I'd expect `.ToUpdate()` to give me a "perfect" replica of the issue, including Labels/Assignees/whatever that I can then use to update the `Issue`.  

I see what you're saying about prefering `AddLabel()` and `ClearLabels()` to direct list manipulation, but if you consider the use case of where I want to add a label to something, I cant just `ToUpdate()` and then `.AddLabel(newLabel)` because that would remove all the other labels.  So I would need more code to explicitly set the labels to whatever they were originally (which perhaps if im now in some other function I dont even have the original `Issue` object anymore, I only have the `IssueUpdate` I'd passed in?  It just seems like this behaviour is a bit odd and means `ToUpdate()` didnt actually create me a "usable" `IssueUpdate` request for my intended usage.  Conversely if we DID include the exiting labels, and possibly fleshed things out with a `RemoveLabel()` call as well as the `AddLabel()` and `ClearLabels()` you still have a way to do everything without direct manipulation but also cater for all intended usages?

It also makes me wonder now, if this is the way we currently handle multiple labels on an issue, does that mean the multiple Assignees functionality added in this PR should also take this approach?  Is the main use case REPLACING assignees or ADDING/REMOVING one of multiple assignees?  Id say the fact that the enhancement is here in the first place means multiple assignees are desirable so we wouldnt want to do things that remove other assignees...

Octokit/Models/Request/IssueUpdate.cs
```diff
@@ -27,10 +27,17 @@ public class IssueUpdate
         /// <remarks>
         /// Only users with push access can set the assignee for new issues. The assignee is silently dropped otherwise.
         /// </remarks>
-        [SerializeNull]
```
ryangribble: @shiftkey you might want to take a look at this too.  It's another "tweak" id made when pushing some changes over to @maddin2016 .  I think it's because now with the multiple assignees support, this field can actually be `null`, but we dont want to send it in as such or else it wiped out all the other assignees.  But I also may not have been aware of historic things similar to how I may have messed up the `Labels` handling currently being discussed!  So if you could check this out too and have a think about it etc to make sure I've done the right thing here?

shiftkey:
I still need to come back and ponder on the changes to serialization here. Apologies for the delay.

Oh, and there's a merge conflict here too @maddin2016 😢

Octokit.Tests/Models/IssueTest.cs
shana: Apologies for the drive-thru comment, but maybe what we should actually have is a standard group of `SetLabel`/`AppendLabel`/`ClearLabels` methods to clearly allow for all the possible ways of manipulating the list (replacing/appending/clearing) so that users won't be tempted to manipulate the list directly and can trust the methods to do what is required? And then apply the same pattern to all multiple categorization lists (assignees, etc).

adamralph:
I want to search for issues assigned to a particular user. Currently, using `RepositoryIssueRequest`, an issue will only be returned if the user is the first assignee on the issue. If the user is the 2nd, 3rd, 4th assignee, etc. the issue will not be returned. Will this PR fix that?

adamralph:
OK, strike my last comment. Issues are returned regardless of whether the user is the first assignee or not.

It was actually a filter in our code that excluded the issues, since it was examining `Issue.Assignee` which is indeed the first assignee.

I've worked around it for now, and I can see that this PR adds `Issue.Assignees` so all is good.

Octokit.Tests/Models/IssueTest.cs
ryangribble: this sounds good :+1:

maddin2016:
I would come back to this issue and have one question. In the `IssuesClient` it is possible to add `Assignees` in the `Create` and `Edit` method and also to get assignees for `Get` methods if we add the preview header for multiple assignees. But it is also possible to get a `reactions summary` if we add the accept header for `reactions`. So now the question how to add multiple `accept headers` for that calls?

maddin2016:
What if we change `accepts` from `string` to `array`. This is certainly not the last time where we have to add multiple `AcceptHeaders`, right?

ryangribble:
The preview period for multiple assignees is [over now](https://developer.github.com/changes/2016-07-12-ending-multiple-assignees-preview-period/), so I guess in this specific situation it's not an issue anymore 

It's still a valid question though, but the idea of adding all those overloads to take `string[]` in what is already a large number of overloads etc, I'd say it should only be done when/where necessary, assuming it actually works

maddin2016:
Thanks for that Info :thumbsup: So I will remove that accept headers. And I correct my comment from yesterday. Maybe it it's ok when we come in such a Situation where have to add multiple accept headers if we do something like this `AcceptHeaders.X + "," + AcceptHeaders.Y`

maddin2016:
@shiftkey and @ryangribble do you think that's ready for :ship: ? Because it sounds really important

jonnii:
@maddin2016 @shiftkey @ryangribble would be great to have this one in...

ryangribble:
Agreed.  After chatting with @shiftkey we agreed to let my label changes go in, although as mentioned above,we want to add a `RemoveLabel()` helper function on `IssueUpdate` to remove a single label from the list, so users aren't ever needing to directly manipulate the list contents.  Are you ok to do that @maddin2016  ?

maddin2016:
Hi @ryangribble, that's ok

maddin2016:
Hi @ryangribble, any news?

Octokit/Models/Request/IssueUpdate.cs
```diff
@@ -62,6 +69,37 @@ internal string DebuggerDisplay
...
+            {
+                Assignees = new List<string>();
+            }
+            else
+            {
+                Assignees.Clear();
+            }
+        }
+
+        /// <summary>
```
maddin2016: :+1:

Octokit/Models/Request/IssueUpdate.cs
```diff
@@ -100,6 +100,22 @@ public void ClearAssignees()
...
+        /// Removes the specified assignee from the issue
+        /// </summary>
+        /// <param name="name">The login of the assignee to remove</param>
+        public void RemoveAssignee(string name)
+        {
+            // lazily create the label array
+            if (Assignees == null)
+            {
+                Assignees = new List<string>();
+            }else
```
ryangribble: minor spacing issue (put `else` on it's own line)

Octokit/Models/Request/IssueUpdate.cs
maddin2016: :see_no_evil:

Eilon:
Can we add support for multiple-assignees in PullRequest.cs as well? It ought to be basically the same as the change for Issue.cs.

I desperately need this change for my app! 😄

ryangribble:
As far as the GitHub API is structured, PullRequests are the same as Issues regarding core concepts such as retrieving comments, assigning, labeling etc.  There is no separate API for PullRequests for these things, you actually just use the Issues API to do them for PullRequests too

So the good news is your change is already done.  The bad news is we STILL havent merged this, sorry!  I have xmas holidays after tomorrow and will definitely get ontop of the couple of PRs I need to review ;)

Eilon:
@ryangribble I think the `Octokit.PullRequest` type needs to be modified to have the new property, no? E.g. when I call `gitHubClient.PullRequest.GetAllForRepository(owner, name)` and get back an `IReadOnlyList<PullRequest>` I need that property, no?

ryangribble:
@Eilon yes I see what you're saying although the github API docs don't show that field on pull request objects. It might be an undocumented case I guess we can poke github.com and see if the field is in the response 

It's the same as labels and comments, eventhough they apply to PR's you get them from the issue call

maddin2016:
I have tested and the get response contains the assignees field and the comments count. It doesn't contains labels. What currently not works is to add assignees against the pull request api.

Eilon:
@maddin2016 are you saying the data for multiple PR assignees isn't even on the wire?

maddin2016:
@ryangribble, i think that adding multiple assignees for pull requests is yet not supported by the api. Maybe we should merge this one and open a new PR??

ryangribble:
As you say, the actual setting of multiple assignees can only be done via `Issues` API but at least the `Assignees` property is returned on PR's so we can add that field to `PullRequest` response class.  That is fine to be done on another PR to save making further changes on this one.  I'll do another/final review of this PR and look to merge it tonight

ryangribble:
@maddin2016 are you able to tick the box to enable me to push commits into this PR?  I've done some testing of this and written a couple of integration tests and also found that a couple of changes were needed to actually make this work!

ryangribble:
Thanks, Ive pushed up a few commits 
- added integration tests for `IssueUpdate.Assignees` usage
- fixed issue where `Assignees` and `Assignee` can't both be specified on the request.  API docs indicate `Assignee` is deprecated, so I have `[Obsoleted]` this field, and stopped including it in `Issue.ToUpdate()`
- I also ended up implementing the `Assignees` field on `PullRequest` response object since it was quick/easy

@maddin2016 are you able to review my changes and confirm the integration tests are running for you?

Octokit/Models/Response/PullRequest.cs
```diff
@@ -136,6 +137,11 @@ public PullRequest(Uri url, Uri htmlUrl, Uri diffUrl, Uri patchUrl, Uri issueUrl
         public User Assignee { get; protected set; }
 
         /// <summary>
+        ///The multiple users this pull request is assigned to.
+        /// </summary>
+        public IReadOnlyList<User> Assignees { get; protected set; }
```
Eilon: Yay!

Eilon: Does this need to also be in the ctor of the type to get populated?

maddin2016:
@ryangribble, integration tests work for me.

Octokit/Models/Response/PullRequest.cs
ryangribble: Nic pickup thanks! 

The ctor isn't actually use by the deserializer to create the response objects (it uses the field setters) but from a consistency point on view I should have include t new field in the ctor

ryangribble:
Wow this has gotta be one of the longest PR's we've had in ages but I think we are finally done!

Thanks @maddin2016 we got there in the end 😁 

LGTM

maddin2016:
Good things come to those who wait 😸

jonnii:
Would it be possible to cut a new release with this?

ryangribble:
Yep, hope to have a new release out soon!

PR #1341, opened 2016-06-04
Include reactions on issue comments

### LINKED ISSUE #1295: Add reactions to issue comments response payload
Github have added reactions to the issues, comments and pull requests. It would be a really cool feature to have, it's however opt in via header for now.

https://developer.github.com/changes/2016-05-12-reactions-api-preview/
https://developer.github.com/v3/issues/comments/#list-comments-on-an-issue

> An additional reactions object in the issue comment payload is currently available for developers to > preview. During the preview period, the APIs may change without advance notice. Please see the blog post for full details.

> To access the API you must provide a custom media type in the Accept header:
application/vnd.github.squirrel-girl-preview
The reactions key will have the following payload where url can be used to construct the API location for listing and creating reactions.

```
{
  "total_count": 5,
  "+1": 3,
  "-1": 1,
  "laugh": 0,
  "confused": 0,
  "heart": 1,
  "hooray": 0,
  "url": "https://api.github.com/repos/octocat/Hello-World/issues/comments/1/reactions"
}
```

alfhenrik:
- [x] Add Reactions object to IssueCommentsClient
- [x] Add Reactions object to ObservableIssueCommentsClient
- [x] Add tests
- [x] Add integration tests
- [x] Fix builds due to broken tests
- [x] Come to think of it, `IssueCommentReactions` should be renamed, so it can be used on the other relevant payloads.
- [x] Use `AcceptHeaders.ReactionsPreview` instead of local `ReactionsPreview` constant (once #1335 has been merged)

Fixes #1295

### DIFF
--- a/Octokit.Tests.Integration/Clients/IssueCommentsClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/IssueCommentsClientTests.cs
@@ -18,12 +18,30 @@ public class IssueCommentsClientTests
     }
 
     [IntegrationTest]
+    public async Task CanDeserializeIssueCommentWhenGettingAllForRepository()
+    {
+        var comments = await _issueCommentsClient.GetAllForRepository("alfhenrik-test", "repo-with-issue-comment-reactions");
+
+        Assert.True(comments.Count > 0);
+        var comment = comments[0];
+        Assert.NotNull(comment.Reactions);
+        Assert.Equal(3, comment.Reactions.TotalCount);
+        Assert.Equal(1, comment.Reactions.Plus1);
+        Assert.Equal(1, comment.Reactions.Hooray);
+        Assert.Equal(1, comment.Reactions.Heart);
+        Assert.Equal(0, comment.Reactions.Laugh);
+        Assert.Equal(0, comment.Reactions.Confused);
+        Assert.Equal(0, comment.Reactions.Minus1);
+    }
+
+    [IntegrationTest]
     public async Task CanDeserializeIssueComment()
     {
         var comments = await _issueCommentsClient.GetAllForIssue("alfhenrik-test", "repo-with-issue-comment-reactions", 1);
 
         Assert.True(comments.Count > 0);
         var comment = comments[0];
+        Assert.NotNull(comment.Reactions);
         Assert.Equal(3, comment.Reactions.TotalCount);
         Assert.Equal(1, comment.Reactions.Plus1);
         Assert.Equal(1, comment.Reactions.Hooray);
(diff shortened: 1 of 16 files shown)


alfhenrik:
RE: renaming `IssueCommentReactions`, perhaps just naming it `Reactions` is better? 

I know #1335 is adding a `Reaction` model, but that is different to this one...but is it going to get confusing?
 
@shiftkey @ryangribble thoughts?

ryangribble:
Yeah naming things is hard (tm) :trollface: 

Perhaps this one in the response payloads could be called `ReactionSummary` since it has counts of each type and the a TotalCount etc.

ryangribble:
Looking good... I guess we'll wait on #1335 for the `AcceptHeaders` value...

alfhenrik:
Yep, or I could add the `AcceptsHeader` in this PR (which would cause a slight merge conflict in #1335), but it looks like #1335 is almost ready for a merge so I have no problem with waiting 😄

ryangribble:
yep I was gonna say you could just add that here too, if you go by the same name as #1335 it wont be much of an issue

alfhenrik:
🎉 look at all those 💚 builds! Clearly @shiftkey's `mono` fix worked 😁

ryangribble:
@alfhenrik #1335 is now merged so you can merge/rebase and update your accepts header usage :+1:

alfhenrik:
👍 will do it tonight

alfhenrik:
PR updated to use new `AcceptHeaders.ReactionsPreview` instead of local variable. Just waiting for 💚 💚 from the running builds

Octokit.Tests.Integration/Clients/IssueCommentsClientTests.cs
```diff
@@ -0,0 +1,53 @@
...
+        var github = Helper.GetAuthenticatedClient();
+        _issueCommentsClient = github.Issue.Comment;
+    }
+
+    [IntegrationTest]
+    public async Task CanDeserializeIssueCommentWhenGettingAllForRepository()
+    {
+        var comments = await _issueCommentsClient.GetAllForRepository("alfhenrik-test", "repo-with-issue-comment-reactions");
+
+        Assert.True(comments.Count > 0);
```
shiftkey: :lipstick: `Assert.NotEmpty(comments)` ?

alfhenrik: Yes of course...not sure what I was thinking there 😕

shiftkey:
Just waiting on this to go green, everything else is :gem:

shiftkey:
Fantastic stuff @alfhenrik!

alfhenrik:
Oh, you are too kind sir! 😄

PR #1342, opened 2016-06-04
Add permission attribute to collaborators and team's repositories

### LINKED ISSUE #1319: Missing `permissions` attribute on repository collaborators and teams' repositories calls
I was writing a script to pull some auditing information from our github enterprise server when I stumbled into this one today.  We are missing the `permissions` section of the response on repository collaborators and team repositories, which indicates the admin/push/pull permissions.

Here are the items I found so far that have the `permissions` attribute.  It's also worth noting that this was in API preview mode and required a [custom accept header](https://developer.github.com/changes/2015-06-24-api-enhancements-for-working-with-organization-permissions/#preview-period) until [just recently](https://developer.github.com/changes/2016-01-05-api-enhancements-for-working-with-organization-permissions-are-now-official/) when the preview period ended.  In terms of GitHub Enterprise, on 2.5.x it is still preview, but is no longer preview on 2.6.x.  So depending on whether our intention is to only support the latest feature build of GHE or whether we are also trying to support 2.5.x, we might want to specify that accepts header for these calls...

This is also somewhat related to #1101 where a new `permission` parameter (a single string/permission, rather than the permissions structure mentioned i this issue) can be provided when adding a collaborator to a repo, to denote what level of access they have.

[List repository collaborators](https://developer.github.com/v3/repos/collaborators/#list-collaborators)

>   {
    "login": "octocat",
    "id": 1,
    "avatar_url": "https://github.com/images/error/octocat_happy.gif",
    ...
    "type": "User",
    "site_admin": false,
    **"permissions": {
      "pull": true,
      "push": true,
      "admin": false**
    }

I also found other places where this appears such as

[List team's repositories](https://developer.github.com/v3/orgs/teams/#list-team-repos)

>  {
    "id": 1296269,
    "owner": {
      "login": "octocat",
      "id": 1,
...
      "type": "User",
      "site_admin": false
    },
    "name": "Hello-World",
    "full_name": "octocat/Hello-World",
...
    "pushed_at": "2011-01-26T19:06:43Z",
    "created_at": "2011-01-26T19:01:12Z",
    "updated_at": "2011-01-26T19:14:43Z",
    **"permissions": {
      "admin": false,
      "push": false,
      "pull": true**
    }

Also there is a custom accept header version of the [Check teams access to repository](https://developer.github.com/v3/orgs/teams/#alternative-response-with-extra-repository-information) which can return the full repository information (including the permissions attributes) rather than just the 204/404 response indicating true/false.  Potentially we could implement this alongside the `Task<bool> IsRepositoryManagedByTeam()` call, to be a similar call but returning a `Task<Repository>` object

ryangribble:
Fixes #1319

Added/ensured `Permissions` are supported for:

## Repository Collaborators
Added `Permission` attribute to `User` object, updated tests for Repository Collaborators to ensure field is populated.  Although preview period has ended I decided to pass the preview accepts header anyway since GHE 2.5 requires it

## Team's Repositories
Found that `Repository` already had `Permissions` attribute and the Team's Repositories call was already working (once preview period ended).  Added accept header so it works on GHE 2.5 and added some integration tests to ensure the property is being deserialized

### DIFF
--- a/Octokit.Reactive/Clients/ObservableTeamsClient.cs
+++ b/Octokit.Reactive/Clients/ObservableTeamsClient.cs
@@ -224,7 +224,7 @@ namespace Octokit.Reactive
         {
             Ensure.ArgumentNotNull(options, "options");
 
-            return _connection.GetAndFlattenAllPages<Repository>(ApiUrls.TeamRepositories(id), options);
+            return _connection.GetAndFlattenAllPages<Repository>(ApiUrls.TeamRepositories(id), null, AcceptHeaders.OrganizationPermissionsPreview, options);
         }
 
         /// <summary>
--- a/Octokit.Reactive/Helpers/ConnectionExtensions.cs
+++ b/Octokit.Reactive/Helpers/ConnectionExtensions.cs
@@ -37,6 +37,15 @@ namespace Octokit.Reactive.Internal
             return GetPages(url, parameters, (pageUrl, pageParams) => connection.Get<List<T>>(pageUrl, pageParams, accepts).ToObservable());
         }
 
+        public static IObservable<T> GetAndFlattenAllPages<T>(this IConnection connection, Uri url, IDictionary<string, string> parameters, string accepts, ApiOptions options)
+        {
+            return GetPagesWithOptions(url, parameters, options, (pageUrl, pageParams, o) =>
+            {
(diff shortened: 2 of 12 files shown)


ryangribble:
This one is ready to merge in, can I get someone to :eyes: over this and :+1: if it looks alright?

PR #1344, opened 2016-06-05
Add repositoryId overloads to methods on I(Observable)RepositoryCommentsClient

dampir:
As part of my work on #1120 I've added new overloads on I(Observable)RepositoryCommentsClient to get access by repository id.

- [x] **Update XML documentation of interface methods of clients (also synchronize XML docs of IRepositoryCommentsClient and IObservableRepositoryCommentsClient).**

	  There is some divergence between XML documentation of methods in IRepositoryCommentsClient and IObservableRepositoryCommentsClient. So I've decided 
- [x] **Add overloads to IRepositoryCommentsClient.**

- [x] **Add overloads to IObservableRepositoryCommentsClient.**

- [x] **Add unit tests.**

/cc @shiftkey, @ryangribble


Octokit.Reactive/Clients/IObservableRepositoryCommentsClient.cs
```diff
@@ -13,81 +19,157 @@ public interface IObservableRepositoryCommentsClient
         /// <param name="owner">The owner of the repository</param>
         /// <param name="name">The name of the repository</param>
         /// <param name="number">The comment id</param>
-        /// <returns></returns>
+        /// <returns>A <see cref="IObservable{CommitComment}"/> of <see cref="CommitComment"/> for the specified repository.</returns>
```
shiftkey: I think this reads better with "**An** IObservable ..." rather than "**A** IObservable ..."

shiftkey: Rather than call this out everywhere for the file, I'll just come back and check after this comment is folded away

Octokit.Reactive/Clients/ObservableRepositoryCommentsClient.cs
```diff
@@ -25,7 +31,7 @@ public ObservableRepositoryCommentsClient(IGitHubClient client)
         /// <param name="owner">The owner of the repository</param>
         /// <param name="name">The name of the repository</param>
         /// <param name="number">The comment id</param>
-        /// <returns></returns>
+        /// <returns>A <see cref="IObservable{CommitComment}"/> of <see cref="CommitComment"/> for the specified repository.</returns>
```
shiftkey: Same thing here: "**An** IObservable ..." rather than "**A** IObservable ..."

Octokit.Reactive/Clients/ObservableRepositoryCommentsClient.cs
```diff
@@ -140,19 +229,46 @@ public IObservable<CommitComment> Update(string owner, string name, int number,
...
+
+        /// <summary>
         /// Deletes the specified Commit Comment
         /// </summary>
         /// <remarks>http://developer.github.com/v3/repos/comments/#delete-a-commit-comment</remarks>
         /// <param name="owner">The owner of the repository</param>
         /// <param name="name">The name of the repository</param>
         /// <param name="number">The comment id</param>
-        /// <returns></returns>
+        /// <returns>A <see cref="IObservable{Unit}"/></returns>
```
shiftkey: Add a comment on the end here "... indicating completion" because the `Unit` part really isn't interesting here...

Octokit.Reactive/Clients/ObservableRepositoryCommentsClient.cs
```diff
@@ -140,19 +229,46 @@ public IObservable<CommitComment> Update(string owner, string name, int number,
...
             return _client.Delete(owner, name, number).ToObservable();
         }
+
+        /// <summary>
+        /// Deletes the specified Commit Comment
+        /// </summary>
+        /// <remarks>http://developer.github.com/v3/repos/comments/#delete-a-commit-comment</remarks>
+        /// <param name="repositoryId">The ID of the repository</param>
+        /// <param name="number">The comment id</param>
+        /// <returns>A <see cref="IObservable{Unit}"/></returns>
```
shiftkey: Same thing as ⬆️

Octokit/Clients/IRepositoryCommentsClient.cs
```diff
@@ -19,74 +19,141 @@ public interface IRepositoryCommentsClient
...
+        Task<CommitComment> Get(int repositoryId, int number);
+
+        /// <summary>
         /// Gets Commit Comments for a repository.
         /// </summary>
         /// <remarks>http://developer.github.com/v3/repos/comments/#list-commit-comments-for-a-repository</remarks>
         /// <param name="owner">The owner of the repository</param>
         /// <param name="name">The name of the repository</param>
-        /// <returns></returns>
+        /// <returns>A <see cref="IReadOnlyPagedCollection{CommitComment}"/> of <see cref="CommitComment"/>s for the specified repository.</returns>
```
shiftkey: This method is technically returning `Task<IReadOnlyList<CommitComment>>` but I'm not really familiar enough with the markup in here to confirm what this should look like 😕 

Something like this?

> A <see cref="Task{IReadOnlyPagedCollection{CommitComment}}"/> for the specified repository. 

There's a few places where this is setup. Let me know if this doesn't work and I can dig deeper.

Octokit/Clients/IRepositoryCommentsClient.cs
```diff
@@ -19,74 +19,141 @@ public interface IRepositoryCommentsClient
...
+        /// <summary>
         /// Creates a new Commit Comment for a specified Commit.
         /// </summary>
         /// <remarks>http://developer.github.com/v3/repos/comments/#create-a-commit-comment</remarks>
         /// <param name="owner">The owner of the repository</param>
         /// <param name="name">The name of the repository</param>
         /// <param name="sha">The sha reference of commit</param>
         /// <param name="newCommitComment">The new comment to add to the commit</param>
-        /// <returns></returns>
+        /// <returns>A <see cref="CommitComment"/> for the specified commit.</returns>
```
shiftkey: Same thing here, gotta wrap this in a `Task<>`

Octokit/Clients/IRepositoryCommentsClient.cs
```diff
@@ -95,5 +162,14 @@ public interface IRepositoryCommentsClient
...
         /// <returns></returns>
         Task Delete(string owner, string name, int number);
+
+        /// <summary>
+        /// Deletes the specified Commit Comment
+        /// </summary>
+        /// <remarks>http://developer.github.com/v3/repos/comments/#delete-a-commit-comment</remarks>
+        /// <param name="repositoryId">The ID of the repository</param>
+        /// <param name="number">The comment id</param>
+        /// <returns></returns>
```
shiftkey: We could document the return value like this (same with the method above):

> A `<see cref="Task"/>` representing the asynchronous operation

shiftkey:
@dampir this is looking really good - just a few pointers on the documentation while we're in here updating things. I think once we settle on the reviews for the first couple of these PRs that it should really fly through!

dampir:
@shiftkey it is looks like I have to review all 26 existent repositoryId PRs to fix XML documentation things, because I've used similar pattern to write docs. So, I'm going through it as slowly and methodically as possible do it :+1:  But, I want add repositoryId PR for new Reactions clients (they are simple and small :+1: )

> I think once we settle on the reviews for the first couple of these PRs that it should really fly through!
I glad to hear that :+1:  Let's do this :smile:

Octokit/Clients/IRepositoryCommentsClient.cs
dampir: Unfortunately, VS do not understand such XML tags:

`<see cref="Task{IReadOnlyList{T}}"/>`

dampir: @shiftkey please see the pattern I've used here #1384, maybe it would be fine?

`A <see cref="Task{T}"/> of <see cref="IReadOnlyList{Reactions}"/> representing <see cref="Reaction"/>s for a specified issue.`

dampir:
@shiftkey I've updated XML docs according your remarks. Please review it whenever possible and if all ok, I'm going to pass through other PRs and update their docs.

Octokit/Clients/IRepositoryCommentsClient.cs
shiftkey: It's pretty close, but looking at the rendered markup doesn't really satisfy me (this isn't your fault):



Do you think it's worth keeping these return values around at all? It's all rather boring and repetitive...

dampir: I don't think so, it is really tedious and boring. I just tried to be self-consistent in work on GSoC. So, what we decide here? Clear <returns> tags at all or change it in the other manner?

shiftkey: Let's clear these `<return>` tags - I'm not sure it's worth all the back-and-forth for the value it gives us

dampir:
@shiftkey I've cleared XML docs here. I suppose I should clear <returns> in every PR that I've done for repository id, because they are similar these ones.

Octokit.Reactive/Clients/IObservableRepositoryCommentsClient.cs
```diff
@@ -15,10 +21,21 @@ public interface IObservableRepositoryCommentsClient
...
+             Justification = "Method makes a network request")]
         IObservable<CommitComment> Get(string owner, string name, int number);
 
         /// <summary>
+        /// Gets a single Repository Comment by number.
+        /// </summary>
+        /// <remarks>http://developer.github.com/v3/repos/comments/#get-a-single-commit-comment</remarks>
+        /// <param name="repositoryId">The ID of the repository</param>
+        /// <param name="number">The comment id</param>
+        /// <returns></returns>
```
shiftkey: Apologies, I should have been clearer - the whole `<returns></returns>` element can be removed from these docs.

shiftkey:
👍 to the additional integration tests here

Octokit.Reactive/Clients/IObservableRepositoryCommentsClient.cs
dampir: @shiftkey done :+1:

Octokit/Clients/IRepositoryCommentsClient.cs
Haacked: Weird, that syntax should work [according to this article](https://blogs.msdn.microsoft.com/ansonh/2006/09/11/c-xml-documentation-comments-faq/).

In any case, the return tags are currently worthless, but that's because we've simply been parroting the return type. We don't need to do that because the return type describes itself! Instead, we should use this tag to describe the intent of the returned value. But then again, that's usually already described in the summary. So :thumbsup: to getting rid of these and only use them when there's something particular interesting about the return value we need to describe that isn't described by the type.

PR #1348, opened 2016-06-07
Add repositoryId overloads to methods on I(Observable)RepositoryContentsClient

dampir:
As part of my work on #1120 I've added new overloads on I(Observable)RepositoryContentsClient to get access by repository id.

- [x] **Update XML documentation of interface methods of clients (also synchronize XML docs of IRepositoryContentsClient and IObservableRepositoryContentsClient).**

	  There is some divergence between XML documentation of methods in IRepositoryContentsClient and IObservableRepositoryContentsClient. So I've decided 
- [x] **Add overloads to IRepositoryContentsClient.**

- [x] **Add overloads to IObservableRepositoryContentsClient.**

- [x] **Add unit tests.**

/cc @shiftkey, @ryangribble

### DIFF
--- a/Octokit.Reactive/Clients/IObservableRepositoryContentsClient.cs
+++ b/Octokit.Reactive/Clients/IObservableRepositoryContentsClient.cs
@@ -16,14 +16,12 @@ namespace Octokit.Reactive
         /// </summary>
         /// <param name="owner">The owner of the repository</param>
         /// <param name="name">The name of the repository</param>
-        /// <returns>A <see cref="IObservable{Readme}"/> of <see cref="Readme"/> representing the README.md at the specified repository.</returns>
         IObservable<Readme> GetReadme(string owner, string name);
 
         /// <summary>
         /// Returns the HTML rendered README.
         /// </summary>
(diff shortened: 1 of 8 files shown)


dampir:
I haven't changed XML docs here, so lets it stay as is.

Octokit.Reactive/Clients/IObservableRepositoryContentsClient.cs
```diff
@@ -6,23 +6,40 @@ namespace Octokit.Reactive
...
+    /// </remarks>
     public interface IObservableRepositoryContentsClient
     {
         /// <summary>
         /// Returns the HTML rendered README.
         /// </summary>
         /// <param name="owner">The owner of the repository</param>
         /// <param name="name">The name of the repository</param>
-        /// <returns></returns>
+        /// <returns>A <see cref="IObservable{Readme}"/> of <see cref="Readme"/> representing the README.md at the specified repository.</returns>
```
shiftkey: Let's just remove the whole `<return></return>` elements wherever we're touching methods in these files.

shiftkey:
:thumbsup: to the added integration tests

PR #1351, opened 2016-06-08
Add repositoryId overloads to methods on I(Observable)RepositoryDeployKeysClient

dampir:
As part of my work on #1120 I've added new overloads on I(Observable)RepositoryDeployKeysClient to get access by repository id.

- [x] **Update XML documentation of interface methods of clients (also synchronize XML docs of IRepositoryDeployKeysClient and IObservableRepositoryDeployKeysClient).**

	  There is some divergence between XML documentation of methods in IRepositoryDeployKeysClient and IObservableRepositoryDeployKeysClient. So I've decided 
- [x] **Add overloads to IRepositoryDeployKeysClient.**

- [x] **Add overloads to IObservableRepositoryDeployKeysClient.**

- [x] **Add unit tests.**

/cc @shiftkey, @ryangribble

### DIFF
--- a/Octokit.Reactive/Clients/IObservableRepositoryDeployKeysClient.cs
+++ b/Octokit.Reactive/Clients/IObservableRepositoryDeployKeysClient.cs
@@ -26,6 +26,18 @@ namespace Octokit.Reactive
         IObservable<DeployKey> Get(string owner, string name, int number);
 
         /// <summary>
+        /// Get a single deploy key by number for a repository.
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="https://developer.github.com/v3/repos/keys/#get"> API documentation</a> for more information.
+        /// </remarks>
+        /// <param name="repositoryId">The ID of the repository.</param>
+        /// <param name="number">The id of the deploy key.</param>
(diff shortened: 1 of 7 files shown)


dampir:

Octokit.Reactive/Clients/IObservableRepositoryDeployKeysClient.cs
```diff
@@ -15,17 +21,31 @@ public interface IObservableRepositoryDeployKeysClient
         /// <param name="owner">The owner of the repository.</param>
         /// <param name="name">The name of the repository.</param>
         /// <param name="number">The id of the deploy key.</param>
+        /// <returns></returns>
```
shiftkey: Let's just remove these empty elements wherever we've created them here.

shiftkey:
:thumbsup: to the new integration tests, even they're all still skipped...

Octokit.Reactive/Clients/IObservableRepositoryDeployKeysClient.cs
dampir: @shiftkey done :+1:

PR #1353, opened 2016-06-08
Add closed_by key to Issue

### LINKED ISSUE #1350: How do I get an Issue "closed_by" data?
https://github.com/octokit/octokit.net/blob/master/docs/issues.md

I'm trying to retrieve an issue json with the "closed_by" data.  I can get a JSON value back, but it's missing the "closed_by" data.

maddin2016:
Fixes #1350 

Add `ClosedBy` key to Issue response if you request an single Issue.


ryangribble:
:+1: tested locally and confirmed the field is populated on Get of a single issue (not on list multiple issues though, but that is down to the API)

ryangribble:
release_notes: Added `ClosedBy` property to `Issue` response object

PR #1355, opened 2016-06-08
Add ApiOptions overloads to methods on I(Observable)RepositoryForksClient

### LINKED ISSUE #1176: Add ApiOption overloads to methods on IRepositoryForksClient

- Add an overload for this method on `IRepositoryForksClient` with a new parameter `ApiOptions`
- Implement the new method on the class which implements `IRepositoryForksClient`
- Add an overload for this method on `IObservableRepositoryForksClient` with a new parameter `ApiOptions`
- Implement the new method on the class which implements `IObservableRepositoryForksClient`

- [ ] `Task<IReadOnlyList<Repository>> GetAll(String owner, String repositoryName, RepositoryForksListRequest request)`

dampir:
Fixes #1176 

To-do:

- [x]  Add overloads on IRepositoryForksClient
- [x]  Add overloads on IObservableRepositoryForksClient
- [x]  Add unit tests for these new methods
- [x]  Add integration tests for these new methods
- [x]  Modified XML documentation of methods

/cc @shiftkey, @ryangribble

### DIFF
--- a/Octokit.Reactive/Clients/ObservableRepositoryForksClient.cs
+++ b/Octokit.Reactive/Clients/ObservableRepositoryForksClient.cs
@@ -57,7 +57,6 @@ namespace Octokit.Reactive
         {
             Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
             Ensure.ArgumentNotNullOrEmptyString(repositoryName, "repositoryName");
-            Ensure.ArgumentNotNull(request, "request");
 
             return GetAll(owner, repositoryName, request, ApiOptions.None);
         }
@@ -71,10 +70,10 @@ namespace Octokit.Reactive
         {
             Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
             Ensure.ArgumentNotNullOrEmptyString(repositoryName, "repositoryName");
-            Ensure.ArgumentNotNull(request, "request");
             Ensure.ArgumentNotNull(options, "options");
 
-            return _connection.GetAndFlattenAllPages<Repository>(ApiUrls.RepositoryForks(owner, repositoryName), request.ToParametersDictionary(), options);
+            return request == null ? _connection.GetAndFlattenAllPages<Repository>(ApiUrls.RepositoryForks(owner, repositoryName), options) :
+                _connection.GetAndFlattenAllPages<Repository>(ApiUrls.RepositoryForks(owner, repositoryName), request.ToParametersDictionary(), options);
         }
 
         /// <summary>
--- a/Octokit.Reactive/Clients/ObservableRepositoryForksClient.cs
+++ b/Octokit.Reactive/Clients/ObservableRepositoryForksClient.cs
@@ -4,6 +4,12 @@ using Octokit.Reactive.Internal;
 
 namespace Octokit.Reactive
 {
+    /// <summary>
+    /// A client for GitHub's Repository Forks API.
+    /// </summary>
+    /// <remarks>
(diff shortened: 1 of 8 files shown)


dampir:
@ryangribble could you review it and merge? Or we are have to waiting while @shiftkey review this too?

Octokit.Reactive/Clients/ObservableRepositoryForksClient.cs
```diff
@@ -25,14 +26,55 @@ public ObservableRepositoryForksClient(IGitHubClient client)
...
+        /// <summary>
+        /// Gets the list of forks defined for a repository
+        /// </summary>
+        /// <remarks>See <a href="http://developer.github.com/v3/repos/forks/#list-forks">API documentation</a> for more information.</remarks>
+        /// <returns></returns>
+        public IObservable<Repository> GetAll(string owner, string repositoryName, RepositoryForksListRequest request, ApiOptions options)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
+            Ensure.ArgumentNotNullOrEmptyString(repositoryName, "repositoryName");
+            Ensure.ArgumentNotNull(request, "request");
```
ryangribble: It looks like the previous implementation allowed the `request` object to be `null` however you are now enforcing that it can't be null which would be a breaking change for some existing users...

Octokit/Clients/RepositoryForksClient.cs
```diff
@@ -19,14 +19,55 @@ public RepositoryForksClient(IApiConnection apiConnection)
...
+        /// Gets the list of forks defined for a repository
+        /// </summary>
+        /// <remarks>See <a href="http://developer.github.com/v3/repos/forks/#list-forks">API documentation</a> for more information.</remarks>
+        /// <returns></returns>
+        public Task<IReadOnlyList<Repository>> GetAll(string owner, string repositoryName, RepositoryForksListRequest request, ApiOptions options)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
+            Ensure.ArgumentNotNullOrEmptyString(repositoryName, "repositoryName");
+            Ensure.ArgumentNotNull(request, "request");
+            Ensure.ArgumentNotNull(options, "options");
```
ryangribble: Same thing here, if the call previously allowed `options == null` then we need to preserve that behaviour

Octokit.Reactive/Clients/ObservableRepositoryForksClient.cs
```diff
@@ -25,14 +26,54 @@ public ObservableRepositoryForksClient(IGitHubClient client)
...
+        }
+
+        /// <summary>
+        /// Gets the list of forks defined for a repository
+        /// </summary>
+        /// <remarks>See <a href="http://developer.github.com/v3/repos/forks/#list-forks">API documentation</a> for more information.</remarks>
+        /// <returns></returns>
+        public IObservable<Repository> GetAll(string owner, string repositoryName, RepositoryForksListRequest request, ApiOptions options)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
```
dampir: @ryangribble yes, you're right :+1: , always forget about back compatibility, it is looks like I've been developing enterprise software enough to forgot that back compatibility is important part of software development. Fixed!

Octokit/Clients/RepositoryForksClient.cs
dampir: @ryangribble :+1: Fixed!

dampir:
@ryangribble I've fixed all remarks, but plesky travis fails again :smile:

ryangribble:
Looks good :+1:

PR #1356, opened 2016-06-09
Add repositoryId overloads to methods on I(Observable)RepositoryForksClient

dampir:
As part of my work on #1120 I've added new overloads on I(Observable)RepositoryForksClient to get access by repository id.

- [x] **Update XML documentation of interface methods of clients (also synchronize XML docs of IRepositoryForksClient and IObservableRepositoryForksClient).**

	  There is some divergence between XML documentation of methods in IRepositoryForksClient and IObservableRepositoryForksClient. So I've decided 
- [x] **Add overloads to IRepositoryForksClient.**

- [x] **Add overloads to IObservableRepositoryForksClient.**

- [x] **Add unit tests.**

/cc @shiftkey, @ryangribble

### DIFF
--- a/Octokit.Reactive/Clients/ObservableRepositoryForksClient.cs
+++ b/Octokit.Reactive/Clients/ObservableRepositoryForksClient.cs
@@ -35,7 +35,7 @@ namespace Octokit.Reactive
         /// </remarks>
         /// <param name="owner">The owner of the repository</param>
         /// <param name="name">The name of the repository</param>
-        /// <returns>A <see cref="IObservable{Repository}"/> of <see cref="Repository"/>s representing forks of specified repository.</returns>
+        /// <returns></returns>
         public IObservable<Repository> GetAll(string owner, string name)
         {
             Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
@@ -51,7 +51,7 @@ namespace Octokit.Reactive
         /// See <a href="http://developer.github.com/v3/repos/forks/#list-forks">API documentation</a> for more information.
         /// </remarks>
         /// <param name="repositoryId">The ID of the repository</param>
-        /// <returns>A <see cref="IObservable{Repository}"/> of <see cref="Repository"/>s representing forks of specified repository.</returns>
+        /// <returns></returns>
         public IObservable<Repository> GetAll(int repositoryId)
         {
             return GetAll(repositoryId, ApiOptions.None);
@@ -66,7 +66,7 @@ namespace Octokit.Reactive
         /// <param name="owner">The owner of the repository</param>
         /// <param name="name">The name of the repository</param>
         /// <param name="options">Options for changing the API response</param>
-        /// <returns>A <see cref="IObservable{Repository}"/> of <see cref="Repository"/>s representing forks of specified repository.</returns>
+        /// <returns></returns>
         public IObservable<Repository> GetAll(string owner, string name, ApiOptions options)
         {
             Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
@@ -84,7 +84,7 @@ namespace Octokit.Reactive
         /// </remarks>
         /// <param name="repositoryId">The ID of the repository</param>
         /// <param name="options">Options for changing the API response</param>
-        /// <returns>A <see cref="IObservable{Repository}"/> of <see cref="Repository"/>s representing forks of specified repository.</returns>
+        /// <returns></returns>
         public IObservable<Repository> GetAll(int repositoryId, ApiOptions options)
         {
             Ensure.ArgumentNotNull(options, "options");
@@ -101,7 +101,7 @@ namespace Octokit.Reactive
         /// <param name="owner">The owner of the repository</param>
         /// <param name="name">The name of the repository</param>
         /// <param name="request">Used to request and filter a list of repository forks</param>
-        /// <returns>A <see cref="IObservable{Repository}"/> of <see cref="Repository"/>s representing forks of specified repository.</returns>
+        /// <returns></returns>
         public IObservable<Repository> GetAll(string owner, string name, RepositoryForksListRequest request)
         {
             Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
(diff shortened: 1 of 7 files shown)


dampir:

Octokit.Reactive/Clients/ObservableRepositoryForksClient.cs
```diff
@@ -87,11 +116,25 @@ public IObservable<Repository> GetAll(string owner, string name, RepositoryForks
         /// <remarks>
         /// See <a href="http://developer.github.com/v3/repos/forks/#list-forks">API documentation</a> for more information.
         /// </remarks>
+        /// <param name="repositoryId">The ID of the repository</param>
+        /// <param name="request">Used to request and filter a list of repository forks</param>
+        /// <returns></returns>
+        public IObservable<Repository> GetAll(int repositoryId, RepositoryForksListRequest request)
+        {
```
shiftkey: `Ensure.ArgumentNotNull(request, "request")`

Octokit.Reactive/Clients/ObservableRepositoryForksClient.cs
```diff
@@ -87,11 +116,25 @@ public IObservable<Repository> GetAll(string owner, string name, RepositoryForks
...
+        /// <remarks>
+        /// See <a href="http://developer.github.com/v3/repos/forks/#list-forks">API documentation</a> for more information.
+        /// </remarks>
         /// <param name="owner">The owner of the repository</param>
         /// <param name="name">The name of the repository</param>
         /// <param name="request">Used to request and filter a list of repository forks</param>
         /// <param name="options">Options for changing the API response</param>
-        /// <returns>A <see cref="IObservable{Repository}"/> of <see cref="Repository"/>s representing forks of specified repository.</returns>
+        /// <returns></returns>
         public IObservable<Repository> GetAll(string owner, string name, RepositoryForksListRequest request, ApiOptions options)
```
shiftkey: `Ensure.ArgumentNotNull(request, "request")`

Octokit.Reactive/Clients/ObservableRepositoryForksClient.cs
```diff
@@ -103,6 +146,24 @@ public IObservable<Repository> GetAll(string owner, string name, RepositoryForks
...
+        /// </summary>
+        /// <remarks>
+        /// See <a href="http://developer.github.com/v3/repos/forks/#list-forks">API documentation</a> for more information.
+        /// </remarks>
+        /// <param name="repositoryId">The ID of the repository</param>
+        /// <param name="request">Used to request and filter a list of repository forks</param>
+        /// <param name="options">Options for changing the API response</param>
+        /// <returns></returns>
+        public IObservable<Repository> GetAll(int repositoryId, RepositoryForksListRequest request, ApiOptions options)
+        {
```
shiftkey: `Ensure.ArgumentNotNull(request, "request")`

Octokit.Reactive/Clients/ObservableRepositoryForksClient.cs
```diff
@@ -69,10 +82,26 @@ public IObservable<Repository> GetAll(string owner, string name, ApiOptions opti
...
+        /// <remarks>
+        /// See <a href="http://developer.github.com/v3/repos/forks/#list-forks">API documentation</a> for more information.
+        /// </remarks>
         /// <param name="owner">The owner of the repository</param>
         /// <param name="name">The name of the repository</param>
         /// <param name="request">Used to request and filter a list of repository forks</param>
-        /// <returns>A <see cref="IObservable{Repository}"/> of <see cref="Repository"/>s representing forks of specified repository.</returns>
+        /// <returns></returns>
         public IObservable<Repository> GetAll(string owner, string name, RepositoryForksListRequest request)
         {
```
shiftkey: `Ensure.ArgumentNotNull(request, "request")`

shiftkey:
Tests look good. Just some guard clauses that was previously overlooked and :fire: the `<returns>` tags and this is good to go!

Octokit.Reactive/Clients/ObservableRepositoryForksClient.cs
dampir: I think here it would be a breaking change, isn't it?

Octokit.Reactive/Clients/ObservableRepositoryForksClient.cs
dampir: I think here it would be a breaking change, isn't it?

Octokit.Reactive/Clients/ObservableRepositoryForksClient.cs
dampir: done!

Octokit.Reactive/Clients/ObservableRepositoryForksClient.cs
dampir: done!

dampir:
@shiftkey I've add only 2 new guards, but I think it would be a breaking change, because of initial method implementations:

`request == null
                ? ApiConnection.GetAll<Repository>(ApiUrls.RepositoryForks(owner, name), options) :
                ApiConnection.GetAll<Repository>(ApiUrls.RepositoryForks(owner, name), request.ToParametersDictionary(), options);`

What do you think? Should I add null checks?

Octokit.Reactive/Clients/ObservableRepositoryForksClient.cs
shiftkey: I'm not sure I understand - this is a new method we're adding here?

Octokit.Reactive/Clients/ObservableRepositoryForksClient.cs
shiftkey: Should we allow `null` here for `RepositoryForksListRequest`? We've been restricting this elsewhere in the codebase, and it feels weird to leave this one here. We can call this out in the release notes if you think that's important...

shiftkey:
> What do you think? Should I add null checks?

I think we should be consistent here with how we accept `null`s. We've been favouring throwing when someone passes a `null` for the request arguments, so I think we should do the same here.

dampir:
@shiftkey I agree with you, just I remember that there was other such case in codebase that allow pass null as correct parameter and we left it without additional null checks because of breaking changes (cannot find it right now). Anyway, I'm going to fix this in next 1-2 days :+1:

dampir:
@shiftkey done!

Issue #1249, opened 2016-04-07
Deployment and DeploymentStatus API enhancements preview period


ryangribble:
### Summary
Recently announced on this [blog post](https://developer.github.com/changes/#--deployment-and-deploymentstatus-api-enhancements-preview-period) the Deployment API has some enhancements that are now in preview mode.

API Docs:
[Create Deployment](https://developer.github.com/v3/repos/deployments/#create-a-deployment)
[Create Deployment Status](https://developer.github.com/v3/repos/deployments/#create-a-deployment)

#### Deployment API
`IDeploymentsClient.Create()`
`IObservableDeploymentsClient.Create()`

The request model object `NewDeployment` providing the input parameters for the deployment creation which now has 2 additional parameters:
> `transient_environment`	`boolean`
`production_environment`	`boolean`

#### DeploymentStatus API
`IDeploymentStatusClient.Create()`
`IObservableDeploymentStatusClient.Create()` 

The request model object `NewDeploymentStatus` providing the input parameters for the deployment status creation has a number of changes:
- 2 new parameters need to be added to the request model
> `environment_url`    `boolean`
`auto_inactive`    `boolean`

- renaming a field
> `target_url` -> `log_url`

Note that the old field will need to be marked `[Obsolete]` at this stage - it will be removed after having been obsolete for at least 3 octokit releases

- Adding a new value to `DeploymentState`
> inactive

Unit and Integration tests should also be updated to assert the new functionality works as expected!

Reminder that like all functionality in "preview mode" a custom accepts header needs to be specified (see `AcceptHeaders.cs`)
> To access the API during the preview period, you must provide a custom media type in the Accept header:
> `application/vnd.github.ant-man-preview+json`

### Working on this issue
This issue has been flagged as `up-for-grabs` as it is a reasonably easy issue for a new contributor to pick up, and be guided through the process of contributing to octokit
(that said, if an existing contributor wants to work on this, by all means go for it!)

Please comment here if you want to take this issue.

Please read and follow the [CONTRIBUTING guide](https://github.com/octokit/octokit.net/blob/master/CONTRIBUTING.md) guide and if you need any help, just ask!  Don't be afraid to open a `[WIP]` (work in progress) Pull Request once you've started implementing the changes

ErikSchierboom:
I'll try and have a go at this.

ryangribble:
Fixed by #1365

PR #1365, opened 2016-06-10
Deployment and DeploymentStatus API enhancements preview period. (#1249)

ErikSchierboom:
This PR implements the functionality described in issue #1249, which is to add new preview fields.

I was looking into adding tests for these methods, but how would I do that? If I look at the [deployments API](https://developer.github.com/v3/repos/deployments/#create-a-deployment), the response does not contain any of the new information. Is that correct?

### DIFF
--- a/Octokit.Tests.Integration/Clients/DeploymentStatusClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/DeploymentStatusClientTests.cs
@@ -58,13 +58,15 @@ public class DeploymentStatusClientTests : IDisposable
     [IntegrationTest]
     public async Task CanReadDeploymentStatuses()
     {
-        var newStatus = new NewDeploymentStatus(DeploymentState.Success);
+        var newStatus = new NewDeploymentStatus(DeploymentState.Success) { LogUrl = "http://test.com/log", EnvironmentUrl = "http:test.com/staging" };
         await _deploymentsClient.Status.Create(_context.RepositoryOwner, _context.RepositoryName, _deployment.Id, newStatus);
 
         var statuses = await _deploymentsClient.Status.GetAll(_context.RepositoryOwner, _context.RepositoryName, _deployment.Id);
 
         Assert.NotEmpty(statuses);
         Assert.Equal(DeploymentState.Success, statuses[0].State);
+        Assert.Equal(newStatus.LogUrl, statuses[0].LogUrl);
+        Assert.Equal(newStatus.EnvironmentUrl, statuses[0].EnvironmentUrl);
     }
 
     [IntegrationTest]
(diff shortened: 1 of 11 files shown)


ryangribble:
> If I look at the deployments API, the response does not contain any of the new information. Is that correct?

Try changing the `GetAll` methods to also include the new preview Accepts header, it could be the documentation wasn't updated fully because normally the extra information is also returned in payloads when specifying the preview header

ErikSchierboom:
@ryangribble Should I also be adding the fields added to the `NewDeployment` and `NewDeploymentStatus` clases to the `Deployment` and `DeploymentStatus` classes? I do get returned values for these new fields in the HTTP response, so I could map them.

ryangribble:
Yep! Sounds good

ErikSchierboom:
@ryangribble I've updated the PR with some tests.

ryangribble:
@ErikSchierboom it's looking pretty good :grinning:  There are a couple of failed convention tests if you check the build results, due to mutable properties (a couple of fields you added have a public setter rather than protected) which need to be tidied up.

And other than that, it'd be really awesome to see the existing integration tests for [Depoloyments](https://github.com/octokit/octokit.net/blob/master/Octokit.Tests.Integration/Clients/DeploymentsClientTests.cs) and [DeploymentStatus](https://github.com/octokit/octokit.net/blob/master/Octokit.Tests.Integration/Clients/DeploymentStatusClientTests.cs) updated/added to ensure these new fields are functioning correctly against the real github API.  If you havent already set yourself up for integration tests, check the details in the [CONTRIBUTING](https://github.com/octokit/octokit.net/blob/master/CONTRIBUTING.md#slow-tests) guide (namely, create a separate test account on github.com so you don't impact your real account, and use the provided powershell script to configure your integration test settings).

Looks like there are already a range of integration tests so it could just be as simple as enhancing some of them to assert that key fields are actually populated rather than just `Assert.NotNull(deployment)` and so on

ErikSchierboom:
I'll get on it!

ErikSchierboom:
@ryangribble The builds are now successful :) I have also added integration tests for the newly added properties, except for the `AutoInactive` property which isn't returned by the API.

ryangribble:
> except for the AutoInactive property which isn't returned by the API.

Just to clarify this statement - are you saying this is an issue, or just meaning that it isnt meant to be returned (since it sounds like it''s more a field on the update request, not on an actual deployment or DeploymentStatus itself?)

ErikSchierboom:
@ryangribble As far as I can tell, it makes sense to not return that field as it appears to be intended for use only when creating/updating.

ryangribble:
Yep agreed, just wanted to clarify you meant that too :)

Ive done a final review and also pulled your branch and run the integration tests etc... 
This is :gem: and ready to go... congratulations on your first octokit contribution!

ryangribble:
LGTM

ErikSchierboom:
Great! Thanks for the help.

Hurray

PR #1369, opened 2016-06-12
Add repositoryId overloads to methods on I(Observable)TreesClient

dampir:
As part of my work on #1120 I've added new overloads on I(Observable)TreesClient to get access by repository id.

- [x] **Add overloads to ITreesClient.**

- [x] **Add overloads to IObservableTreesClient.**

- [x] **Add unit tests.**

/cc @shiftkey, @ryangribble

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: csharp
+mono:
+  - 4.2.3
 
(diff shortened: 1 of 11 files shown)


dampir:

shiftkey:
Tests and code look good. Let's just :fire: the `<returns>` statements in here.

PR #1370, opened 2016-06-12
Add repositoryId overloads to methods on I(Observable)ReleasesClient

dampir:
As part of my work on #1120 I've added new overloads on I(Observable)ReleasesClient to get access by repository id.

- [x] **Update XML documentation of interface methods of clients (also synchronize XML docs of IReleasesClient and IObservableReleasesClient).**

	  There is some divergence between XML documentation of methods in IReleasesClient and IObservableReleasesClient. So I've decided 

- [x] **Add overloads to IReleasesClient.**

- [x] **Add overloads to IObservableReleasesClient.**

- [x] **Add unit tests.**

/cc @shiftkey, @ryangribble


dampir:

ryangribble:
Some of the integration tests are failing for me, I think because you have hardcoded the RepositoryId to 252774 but since the tests are creating a repository on the fly I think you will need to use `_context.Repository.Id`

dampir:
@ryangribble I've fixed integration tests and removed <returns> tags, please check this out :+1:

Octokit.Tests.Integration/Clients/ReleasesClientTests.cs
```diff
@@ -645,7 +645,7 @@ public TheUploadAssetMethod()
             {
                 PageSize = 1,
                 PageCount = 1,
-                StartPage = 2
+                StartPage = 1
```
ryangribble: if this is changed to 1 here, should it also be changed in the existing owner/name test? `ReturnsCorrectCountOfReleasesWithStart()`

ryangribble: hmmm confusing, it says this is an outdated diff but i cant see where it was changed back :confused:

Octokit.Tests.Integration/Clients/ReleasesClientTests.cs
```diff
@@ -372,6 +639,21 @@ public TheUploadAssetMethod()
         }
 
         [IntegrationTest]
+        public async Task ReturnsCorrectCountOfReleasesWithStartWithRepositoryId()
+        {
+            var options = new ApiOptions
+            {
+                PageSize = 1,
+                PageCount = 1,
+                StartPage = 1
```
ryangribble: Ah here it is!  This is set to `StartPage = 1` here, but in the non repository Id test it is `2`.  It would be good to have these consistent (eventhough it doesnt really affect the result)

dampir: @ryangribble yep, sure, I'm going to fix this for now!

dampir:
@ryangribble  I've fixed all remarks 👍

PR #1372, opened 2016-06-13
Add repositoryId overloads to methods on I(Observable)WatchedClient

dampir:
As part of my work on #1120 I've added new overloads on I(Observable)WatchedClient to get access by repository id.

- [x] **Update XML documentation of interface methods of clients (also synchronize XML docs of IWatchedClient and IObservableWatchedClient).**

	  There is some divergence between XML documentation of methods in IWatchedClient and IObservableWatchedClient. So I've decided 

- [x] **Add overloads to IWatchedClient.**

- [x] **Add overloads to IObservableWatchedClient.**

- [x] **Add unit tests.**

/cc @shiftkey, @ryangribble

### DIFF
--- a/Octokit.Reactive/Clients/IObservableWatchedClient.cs
+++ b/Octokit.Reactive/Clients/IObservableWatchedClient.cs
@@ -103,7 +103,7 @@ namespace Octokit.Reactive
         /// <param name="owner">The owner of the repository to star</param>
         /// <param name="name">The name of the repository to star</param>
         /// <param name="newSubscription">A <see cref="NewSubscription"/> instance describing the new subscription to create</param>
-        /// <returns>A <c>bool</c> representing the success of starring</returns>
+        /// <returns>A <see cref="Subscription"/> representing the subscription on specified repository.</returns>
         IObservable<Subscription> WatchRepo(string owner, string name, NewSubscription newSubscription);
 
         /// <summary>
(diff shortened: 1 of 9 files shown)


dampir:
@ryangribble please rerun Travis whenever possible :+1:

dampir:

Octokit.Reactive/Clients/IObservableWatchedClient.cs
```diff
@@ -11,24 +17,41 @@ public interface IObservableWatchedClient
         /// <param name="owner">The owner of the repository</param>
         /// <param name="name">The name of the repository</param>
         /// <exception cref="AuthorizationException">Thrown if the client is not authenticated</exception>
-        /// <returns>A <see cref="IObservable{User}"/> of <see cref="User"/>s watching the passed repository</returns>
```
Haacked: Why are we removing the return description? If that's a direction we want to take, might as well remove the entire `/// <returns></returns>` line.

dampir: @Haacked We've discussed this [here](https://github.com/octokit/octokit.net/pull/1344/files/ddb34265bbab715c5e151f4a8e51d05095bc8987#diff-cda94b5dbadd28f0d27ebaff90d53faaL33) with @shiftkey. And yep. I'll remove `/// <returns></returns>` in next commits when other things will be reviwed :+1:

dampir:
@shiftkey I've removed <returns> tags here.

PR #1379, opened 2016-06-14
Add repositoryId overloads to methods on I(Observable)IssueCommentsClient

dampir:
As part of my work on #1120 I've added new overloads on I(Observable)IssueCommentsClient to get access by repository id.

- [x] **Update XML documentation of interface methods of clients (also synchronize XML docs of IIssueCommentsClient and IObservableIssueCommentsClient).**

	  There is some divergence between XML documentation of methods in IIssueCommentsClient and IObservableIssueCommentsClient. So I've decided 

- [x] **Add overloads to IIssueCommentsClient.**

- [x] **Add overloads to IObservableIssueCommentsClient.**

- [x] **Add unit tests.**

I've deleted class ObservableIssueCommentsClientTests beacuse it is useless. All test cases are covered in non-reactive IssueCommentsClientTests class.

/cc @shiftkey, @ryangribble


dampir:

Octokit.Tests.Integration/Clients/IssueCommentsClientTests.cs
```diff
@@ -1,53 +1,476 @@
...
-using System.Text;
-using System.Threading.Tasks;
-using Octokit;
-using Octokit.Tests.Integration;
+﻿using System.Threading.Tasks;
+using Octokit.Tests.Integration.Helpers;
 using Xunit;
 
-public class IssueCommentsClientTests
+namespace Octokit.Tests.Integration.Clients
```
shiftkey: Can we revert this namespace addition to simplify the diff?

Octokit.Tests/Reactive/ObservableIssueCommentsClientTests.cs
```diff
@@ -104,38 +152,73 @@ public void RequestsCorrectUrl()
...
                     PageCount = 1
                 };
+
                 client.GetAllForIssue("fake", "repo", 3, options);
 
                 gitHubClient.Connection.Received(1).Get<List<IssueComment>>(
-                    new Uri("repos/fake/repo/issues/3/comments", UriKind.Relative), 
-                    Arg.Any<Dictionary<string, string>>(),
-                    "application/vnd.github.squirrel-girl-preview");
+                    new Uri("repos/fake/repo/issues/3/comments", UriKind.Relative), Arg.Is<IDictionary<string, string>>(d => d.Count == 2), null);
```
shiftkey: I think you need to bring back the custom `Accept` header here to get that test passing...

Octokit.Tests.Integration/Reactive/ObservableIssueCommentsClientTests.cs
```diff
@@ -1,126 +0,0 @@
-﻿using System;
```
shiftkey: Is there a reason why we've removed this file?

shiftkey:
This one also needs to :fire: the `<return>` statements...

Octokit.Tests.Integration/Reactive/ObservableIssueCommentsClientTests.cs
dampir: I've removed this file because it is fully similar to non-reactive implementation. and usually we have integration tests for non-reactive clients, because reactive ones use non-reactive clients internally.If it not sounds reasonable I going to revert it back :+1:

dampir:
@shiftkey I've worked on your remarks, please review it whenever possible :+1:

PR #1387, opened 2016-06-15
Add repositoryId overloads to methods on I(Observable)RepoCollaboratorsClient

dampir:
_This PR created instead of #1305 (I've renamed branch :smile: )_

As part of my work on #1120 I've added new overloads on I(Observable)RepoCollaboratorsClient to get access by repository id.

- [x] **Rename parameter "repo" to "name" (in order to deliver more consistency into the codebase).**

	 I've found out that there is the pair (owner, name) is used to identify repository. But some clients use pair (owner, repo) and it's some type of inconsistency.
	  So I've decided to fix these cases during my work on #1120.
- [x] **Update XML documentation of interface methods of clients (also synchronize XML docs of IRepoCollaboratorsClient and IObservableRepoCollaboratorsClient).**

	  There is some divergence between XML documentation of methods in IRepoCollaboratorsClient and IObservableRepoCollaboratorsClient. So I've decided 
- [x] **Add overloads to IRepoCollaboratorsClient.**

- [x] **Add overloads to IObservableRepoCollaboratorsClient.**

- [x] **Add unit tests.**

/cc @shiftkey, @ryangribble

### DIFF
--- a/Octokit.Reactive/Clients/IObservableRepoCollaboratorsClient.cs
+++ b/Octokit.Reactive/Clients/IObservableRepoCollaboratorsClient.cs
@@ -92,7 +92,7 @@ namespace Octokit.Reactive
         /// </remarks>
         /// <param name="owner">The owner of the repository</param>
         /// <param name="name">The name of the repository</param>
-        /// <param name="user">Username of the prospective collaborator</param>
+        /// <param name="user">Username of the new collaborator</param>
         /// <exception cref="ApiException">Thrown when a general API error occurs.</exception>
         /// <returns><see cref="Unit"/></returns>
         IObservable<Unit> Add(string owner, string name, string user);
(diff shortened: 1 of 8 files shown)


dampir:

shiftkey:
No tests to worry about here. Just need to remove the empty `<returns></returns>` elements.

ryangribble:
> No tests to worry about here

I can't remember why there aren't any integration tests with this change.  The commit history shows they were added but then reverted but I can't find any comment/discussion as to why.  Can someone jog my memory?

shiftkey:
@ryangribble excellent point! I left that note in here to just indicate whether I'd looked at the tests as part of reviewing. Perhaps @dampir has found something which means those tests aren't possible?

ryangribble:
Yeah I know there were some cases (like the activity one requiring  user other than the one running the tests etc) but cant find mention of anything specific to the collaborators stuff.  Will wait for @dampir to advise

dampir:
@shiftkey, @ryangribble not sure why I did that (revert integration tests), it was made accidentaly  I guess. Anyway, now it is fixed. Also, I removed <returns> tags.

PR #1389, opened 2016-06-16
Add repositoryId overloads to methods on I(Observable)StarredClient

dampir:
As part of my work on #1120 I've added new overloads on I(Observable)StarredClient to get access by repository id.

- [x] **Update XML documentation of interface methods of clients (also synchronize XML docs of IStarredClient and IObservableStarredClient).**

	  There is some divergence between XML documentation of methods in IStarredClient and IObservableStarredClient. So I've decided 
- [x] **Add overloads to IStarredClient.**

- [x] **Add overloads to IObservableStarredClient.**

- [x] **Add unit tests.**

### DIFF
--- a/Octokit.Tests/Clients/StarredClientTests.cs
+++ b/Octokit.Tests/Clients/StarredClientTests.cs
@@ -1,9 +1,9 @@
-﻿using System.Collections.Generic;
-using Octokit.Internal;
-using System;
+﻿using System;
+using System.Collections.Generic;
 using System.Net;
 using System.Threading.Tasks;
 using NSubstitute;
+using Octokit.Internal;
 using Xunit;
 
 namespace Octokit.Tests.Clients
@@ -15,8 +15,7 @@ namespace Octokit.Tests.Clients
             [Fact]
             public void EnsuresNonNullArguments()
             {
-                Assert.Throws<ArgumentNullException>(
-                    () => new StarredClient(null));
+                Assert.Throws<ArgumentNullException>(() => new StarredClient(null));
             }
         }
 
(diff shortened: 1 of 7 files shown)


dampir:

Octokit.Tests/Clients/StarredClientTests.cs
```diff
@@ -346,19 +345,31 @@ public class TheGetAllForUserMethod
         public class TheGetAllStargazersForRepoMethod
         {
             [Fact]
-            public void RequestsCorrectUrl()
+            public async Task RequestsCorrectUrl()
```
ryangribble: Is this required?  I had a look at most `RequestsCorrectUrl()` tests and they aren't `async Task<>` they are just `void`

ryangribble: Actually it looks like there is inconsistency in the code base with both `async` and regular examples... but im interested whether you were trying to standardise to always being `async` or whether you actually had a failing test here?

shiftkey: While it's technically correct to `await` in these tests, I've not seen this affect the test behaviour - as we're interested in what was invoked, rather than the result returned. Although, saying it out loud like that makes me feel like we're just asking for some neat race condition to come along and cause us issues...

dampir: @ryangribble Yes, I tried to standardise all tests to use `async` and there is no failed tests without `async Task` return type. I've just tried be consiistent in sense that awaitable methods have to be called with usage of `await` keyword.

@shiftkey I don't understood your last sentence, do you mean that this change from `void` to `async Task` is going to produce some troubles with multithreading?

ryangribble: he is saying that using `await` is the "correct" way, and we may just be lucky that we havent seen a timing issue when not using `await`

ryangribble:
This one still has the empty `<returns>` tags, but once they're removed it should be good to merge.

On that note, if you are able to fixup those `<returns>` tags on ALL your PR's it will make it faster to get these merged

Octokit.Tests/Clients/StarredClientTests.cs
dampir: Oh, yeah! Thank you @ryangribble 👍  So, there are no any issues with it because we don't use return results for any purposes 👍

dampir:
@ryangribble done! I'll try to clear all returns tags in RepositoryId PRs.

PR #1392, opened 2016-06-16
Add repositoryId overloads to methods on I(Observable)MilestonesClient

dampir:
As part of my work on #1120 I've added new overloads on I(Observable)MilestonesClient to get access by repository id.

- [x] **Add overloads to IMilestonesClient.**

- [x] **Add overloads to IObservableMilestonesClient.**

- [x] **Add unit tests.**


dampir:
no need to clear <returns> tags, they are clean.

ryangribble:
LGTM

PR #1398, opened 2016-06-17
Updated git data commit response with signature verification object

Sarmad93:
Fixes #1254  
Hi @ryangribble @shiftkey  could you plz review this code :smiley: Regarding Integration test i am little confused where should i put this method, is it fine here in `CommitsClientTests` ? And what do you think of this integration test, right now this test is passing on repository which is of my test account. Is it fine here to hard code(string literals) repository parameter? Your feedback on this is much needed so that we can make improvement to this.

PR #1401, opened 2016-06-20
Signature Verification with commits.

### LINKED ISSUE #1255: Get Signature Verification When Viewing Repository Commits
API documentation: https://developer.github.com/v3/repos/commits/#commit-signature-verification

An enhancement as part of the Commit Signing API is to surface the signature details in the commit.

It looks like these are the new properties:

``` json
    "verification": {
      "verified": true,
      "reason": "valid",
      "signature": "-----BEGIN PGP MESSAGE-----\n...\n-----END PGP MESSAGE-----",
      "payload": "tree 6dcb09b5b57875f334f61aebed695e2e4193db5e\n..."
    }
```

These fields should be available on the `GitHubCommit` response object (perhaps in a class called `Verification` or something), and the `RepositoryCommitsClient.Get` and `RepositoryCommitsClient.GetAll` actions  should be updated to use the new `Accepts` header:

> application/vnd.github.cryptographer-preview+sha

I'd also like to see some new integration tests added to verify this behaviour is working as expected.

devkhan:
Resolves #1255.

A `Verification` object inside `GitHubCommit.Commit`.

To-do:
- [x] Models
- [x] Tests
- [x] Integration tests


devkhan:
@shiftkey I've added the Verification object inside Commit and not directly under GitHubCommit according to the API.

Also, I don't know whether we should create a class for the `signature` and `payload` fields, so I've just marked them as strings for now. Please let me know if they should get their own classes.

Octokit.Tests/Clients/RepositoriesClientTests.cs
```diff
@@ -771,7 +771,7 @@ public void GetsCorrectUrl()
                 client.Get("owner", "name", "reference");
 
                 connection.Received()
-                    .Get<GitHubCommit>(Arg.Is<Uri>(u => u.ToString() == "repos/owner/name/commits/reference"));
+				          .Get<GitHubCommit>(Arg.Is<Uri>(u => u.ToString() == "repos/owner/name/commits/reference"), null, AcceptHeaders.SignatureVerificationPreview);
```
ryangribble: looks like a tab/whitespace issue here

ryangribble: Also, we don't want to use implementation logic as part of the test, so rather than using the `AcceptHeaders` helper, can you hardcode the expected header in the tests?

ryangribble:
@devkhan looks like you need to adjust your tab settings as most files are showing whitespace/alignment issues

devkhan:
@ryangribble Yeah, I recently started using Xamarin, so didn't noticed the tabs/spaces. Will fix it after #1398 is merged. Thanks.

shiftkey:
#1398 has been merged

ryangribble:
Still got merge conflicts on this one, also dont forget about the tab/space issues :grinning:

PR #1402, opened 2016-06-21
Fix serialization of enum value attributes

maddin2016:
On my work on reactions payload i have found that `ParameterAttribute` for enum values are not be serialized. For example
```
public enum ReactionType
    {
        [Parameter(Value = "+1")]
        Plus1,
        [Parameter(Value = "-1")]
        Minus1,
        Laugh,
        Confused,
        Heart,
        Hooray
    }
```
`Plus1` is always serialized to `{content: plus1}`. So i always get an error from api.github.com because it expected `{content: +1}`
You can test this behavior in Integrationtest `CommitCommentReactionsClientTests` `CanCreateReaction` if you set `ReactionType.Plus1`


maddin2016:
Any Ideas why .net core does not have method `GetField()`?
[https://github.com/dotnet/corefx/blob/master/src/System.Reflection.TypeExtensions/src/System/Reflection/TypeExtensions.CoreCLR.cs/#L66](https://github.com/dotnet/corefx/blob/master/src/System.Reflection.TypeExtensions/src/System/Reflection/TypeExtensions.CoreCLR.cs/#L66)

maddin2016:
I actually can not test if this work with .net core.

ryangribble:
It's weird though as this should have worked I thought...  Surely this isn't the first time we have an enum with an overriden value on a request message? 

Maybe it's `[Parameter(Key=)]` that is required, rather than `[Parameter(Value=)]`

It would be good to make that integration test loop through all enum values and "create reaction" for each one of them, that would have picked this up in the beginning. I thought we already did that, or at least made the integration test use the enum value that did have the custom override value, but perhaps I'm thinking of the "Get reaction" payload and not the create reaction one

maddin2016:
I have tested all variations `[Parameter(Key = "+1")]`. But if you step trough the serialization process 
```
  protected virtual bool TrySerializeKnownTypes(object input, out object output)
        {
            bool returnValue = true;
            if (input is DateTime)
                output = ((DateTime)input).ToUniversalTime().ToString(Iso8601Format[0], CultureInfo.InvariantCulture);
            else if (input is DateTimeOffset)
                output = ((DateTimeOffset)input).ToUniversalTime().ToString(Iso8601Format[0], CultureInfo.InvariantCulture);
            else if (input is Guid)
                output = ((Guid)input).ToString("D");
            else if (input is Uri)
                output = input.ToString();
            else
            {
                Enum inputEnum = input as Enum;
                if (inputEnum != null)
                    output = SerializeEnum(inputEnum);
                else
                {
                    returnValue = false;
                    output = null;
                }
            }
            return returnValue;
        }
```
and then
```
rotected override object SerializeEnum(Enum p)
            {               
                return p.ToString().ToLowerInvariant();
            }
```
so maybe this is really the first time this happens or i'm wrong :confused:

maddin2016:
There are also problems when converting back to enum. In `SimpleJsonSerializer.cs` `-1` becomes `1`
because of 
```
                        stringValue = RemoveHyphenAndUnderscore(stringValue);
                        return Enum.Parse(type, stringValue, ignoreCase: true);
```
And `+1` is converted to `Minus1` :weary:

maddin2016:
Maybe we should use hardcoded strings for `ReactionType`?

ryangribble:
> so maybe this is really the first time this happens or i'm wrong :confused:

So there are a fair few other occurences of `Enum`'s on request classes that have `[Parameter(Value = "xxx")]` attribute.  Did you want to test these through the serializer to see if they also have the problem

[`MilestoneSort.DueDate`](https://github.com/octokit/octokit.net/blob/9231286fedce0c7a5f986d72cf1012a95941c62f/Octokit/Models/Request/MilestoneRequest.cs#L42)
[`SortDirection.Ascending`](https://github.com/octokit/octokit.net/blob/9231286fedce0c7a5f986d72cf1012a95941c62f/Octokit/Models/Request/IssueRequest.cs#L189)
[`SortDirection.Descending`](https://github.com/octokit/octokit.net/blob/9231286fedce0c7a5f986d72cf1012a95941c62f/Octokit/Models/Request/IssueRequest.cs#L195)
[`DeployTask.DeployMigrations`](https://github.com/octokit/octokit.net/blob/9231286fedce0c7a5f986d72cf1012a95941c62f/Octokit/Models/Request/NewDeployment.cs#L99)

maddin2016:
Here is maybe another solution. For example if you have `+1`  how `Enum.Parse` should parse it to the
right value?

maddin2016:
In think i know where the issue comes from. All the enums with parameters are used Requests which inherits from `Requestparameter`. For example MilestoneRequest
https://github.com/octokit/octokit.net/blob/master/Octokit/Models/Request/MilestoneRequest.cs#L11

Here you have the `MilestoneRequest` which use `SortProperty` which is enum `MilestoneSort`

In the Apiconnection call you see `request.ToParametersDictionary()` which internally use reflections to get the parameter attribute
```
/// <summary>
        /// Converts the derived object into a dictionary that can be used to supply query string parameters.
        /// </summary>
        /// <returns></returns>
        public virtual IDictionary<string, string> ToParametersDictionary()
        {
            var map = _propertiesMap.GetOrAdd(GetType(), GetPropertyParametersForType);
            return (from property in map
                    let value = property.GetValue(this)
                    let key = property.Key
                    where value != null
                    select new { key, value }).ToDictionary(kvp => kvp.key, kvp => kvp.value);
        }
```
In `ReactionsClient` we pass `reactions` as a normal parameter
```
return ApiConnection.Post<Reaction>(ApiUrls.CommitCommentReactions(owner, name, number), reaction, AcceptHeaders.ReactionsPreview);
```
which is serialized by
```
 async Task<IApiResponse<T>> Run<T>(IRequest request, CancellationToken cancellationToken)
        {
            _jsonPipeline.SerializeRequest(request);
            var response = await RunRequest(request, cancellationToken).ConfigureAwait(false);
            return _jsonPipeline.DeserializeResponse<T>(response);
        }
```
So the solution for this first problem is to inherit `NewReactiion` from `RequestParameters` and change the Api call for all reaction clients. I have tested this and it works as expected.
```
public class NewReaction : RequestParameters
```
```
return ApiConnection.Post<Reaction>(ApiUrls.CommitCommentReactions(owner, name, number), reaction.ToParametersDictionary(), AcceptHeaders.ReactionsPreview);
```

The second problem is if you get a payload which has to serialze. This only happens at one place.
https://github.com/octokit/octokit.net/blob/master/Octokit/Http/SimpleJsonSerializer.cs#L79
So here i think is the only solution from my last commit
https://github.com/octokit/octokit.net/pull/1402/commits/4c05be81d43c38ae447db0e90d85fc44233a9d3d#diff-2e183b0cbda11827ea1db237ac05ce00R106

Please correct me if i'm wrong

ryangribble:
Although the `ParametersDictionary` approach does fix things, I guess my thoughts are that since a change is needed in `SimpleJsonSerializer` on the `Deserialize` side, we may as well also fix it in the `Serialize` side.  But I don't think it's necessary to write that logic yourself, as there is a `.ToParameter()` extension method in `EnumExtensionMethods` that already handles retrieving the custom parameter if present, otherwise uses lowercase string.  So something like this:

``` csharp
protected override object SerializeEnum(Enum p)
{
    return p.ToParameter();
}
```

Octokit.Tests.Integration/Clients/PullRequestReviewCommentReactionsClientTests.cs
```diff
@@ -40,15 +40,42 @@ public PullRequestReviewCommentReactionsClientTests()
...
-
-        Assert.NotNull(pullRequestReviewCommentReaction);
-
-        Assert.IsType<Reaction>(pullRequestReviewCommentReaction);
-
-        Assert.Equal(ReactionType.Heart, pullRequestReviewCommentReaction.Content);
-
-        Assert.Equal(commentFromGitHub.User.Id, pullRequestReviewCommentReaction.User.Id);
+        var newReactionConfused = new NewReaction(ReactionType.Confused);
+        var newReactionHeart = new NewReaction(ReactionType.Heart);
```
ryangribble: Although I definitely want integration tests to try every enum value, I dont think we need to be this verbose (plus in the future if more reactions are added, we'd need to remember to add them to this test).

So whatabout soething like this:

``` csharp
foreach (ReactionType reactionType in Enum.GetValues(typeof(ReactionType)))
{
    var newReaction = new NewReaction(reactionType);

    var reaction = await _github.Reaction.CommitComment.Create(_context.RepositoryOwner, _context.RepositoryName, result.Id, newReaction);

    Assert.IsType<Reaction>(reaction);
    Assert.Equal(reactionType, reaction.Content);
    Assert.Equal(result.User.Id, reaction.User.Id);
}
```

maddin2016: That's great!

maddin2016:
@ryangribble `p.ToParameter()` works perfectly :tada:

Octokit/Http/SimpleJsonSerializer.cs
```diff
@@ -92,6 +92,17 @@ public override object DeserializeObject(object value, Type type)
                 {
                     if (ReflectionUtils.GetTypeInfo(type).IsEnum)
                     {
+                        //first try to get all custom attributes
```
ryangribble: Would be good for @shiftkey or @Haacked to comment on whether they're happy with this change to deserializer.  Is this the most elegant way of doing this and would there be any concerns around performance?  Just wondering, since it's looping for every value of every enum and checking for custom attribute, when the majority of enums/values dont have one.

maddin2016: Maybe we should hardcoded strings. Because i have no idea how we can deserialize enum properly without to loop trough all the enum values and check if it has an attribute or not.

maddin2016: Or we create a customAttribute class `DeserializeProperlyAttribute` for example. And then
```
public override object DeserializeObject(object value, Type type)
            {
                var stringValue = value as string;
                var jsonValue = value as JsonObject;
                if (stringValue != null)
                {
                    if (ReflectionUtils.GetTypeInfo(type).IsEnum)
                    {
                        if(Attribute.IsDefined(type, typeof(DeserializeProperlyAttribute)))
                        {
                            //first try to get all custom attributes
                            var fields = type.GetRuntimeFields();
                            foreach (var field in fields)
                            {
                                var attribute = (ParameterAttribute)field.GetCustomAttribute(typeof(ParameterAttribute));
                                if (attribute != null)
                                {
                                    if (attribute.Value.Equals(value))
                                        return field.GetValue(null);
                                }
                            }
                        }
                        stringValue = RemoveHyphenAndUnderscore(stringValue);
                        return Enum.Parse(type, stringValue, ignoreCase: true);
                    }
```

ryangribble: yeah im not saying there is a better way of doing it or that it does/doesn't have performance impacts, just asking the question and looking for more opinions.  

Im not a an of hardcoded strings or a new attribute, if anything had to be done, I would probably do similar to how the serializer pre-caches all the constructors of the model objects, and do some sort of pre-cache of enum items and their value (custom or default), then it would be a quick look up of that list here in the deserializer.  

It may not even be necessary though, which is why i refer to @shiftkey

Haacked: It's very unlikely that an enum is going to have a huge number of elements. And we only need to iterate once and cache the result. An enum change would require recompilation. So I'm happy with ENUMeration. :stuck_out_tongue:

Octokit/Http/SimpleJsonSerializer.cs
```diff
@@ -92,6 +92,17 @@ public override object DeserializeObject(object value, Type type)
...
+                        var fields = type.GetRuntimeFields();
+                        foreach (var field in fields)
+                        {
+                            var attribute = (ParameterAttribute)field.GetCustomAttribute(typeof(ParameterAttribute));
+                            if (attribute != null)
+                            {
+                                if (attribute.Value.Equals(value))
+                                    return field.GetValue(null);
+                            }
+                        }
```
Haacked: We should cache the result of this in a dictionary of dictionaries (probably `Dictionary<Type, Dictionary<Enum, string>>`.) That way, the next time we try to serialize an enum value, we don't have to do this iteration. It'll be a very fast lookup.

ryangribble: cool so to paraphrase/clarify you are saying we dont need to upfront cache everything, but we may as well keep a running cache of anything we've already evaluated so we dont have to do it more than once?

maddin2016:
Can you look on my last commit. Maybe someone has another solution. (Maybe simpler).

Octokit/Http/SimpleJsonSerializer.cs
Haacked: Yup!

Octokit/Http/SimpleJsonSerializer.cs
```diff
@@ -24,6 +24,7 @@ public T Deserialize<T>(string json)
         class GitHubSerializerStrategy : PocoJsonSerializerStrategy
         {
             readonly List<string> _membersWhichShouldPublishNull = new List<string>();
+            Dictionary<Type, Dictionary<object, object>> _cachedEnums = new Dictionary<Type, Dictionary<object, object>>();
```
Haacked: This could be declared `readonly`

Octokit/Http/SimpleJsonSerializer.cs
```diff
@@ -88,24 +89,62 @@ public override object DeserializeObject(object value, Type type)
...
+                            return attributeValue;
+                        }
+                        else
+                        {
+                            //Fallback for any reason we have no match. This sould also happens once
+                            // remove '-' from values coming in to be able to enum utf-8
+                            stringValue = RemoveHyphenAndUnderscore(stringValue);
+                            var parsed = Enum.Parse(type, stringValue, ignoreCase: true);
+                            _cachedEnums[type].Add(value, parsed);
+                            return parsed;
```
Haacked: This fallback block is the same as the one earlier (line 106). Maybe restructure this so there's no duplication here.

ryangribble: yeah I would probably suggest a couple of tidy ups such as

- When you loop through each value of the enum you are only storing those with the [Parameter] attribute in the dictionary, you may as well store those that dont have the attribute (using the fallback code) here too.  That way you can just have an approach similar to this (pseudocode)

```
if Type Is Enum
{
   if Type is not in cache
      add it to cache (all fields both custom attribute and non attribute)

   if Type is now in cache (it should be at this point!)
      return the value from cache
   else
     throw exception (we shouldnt really ever get here!)
}
```

maddin2016: I've ended with 
https://github.com/octokit/octokit.net/pull/1402/files#diff-2e183b0cbda11827ea1db237ac05ce00R120
Because we need to store the value that comes from the api like `blob` or `open` and not the name of the enum field `Blob`, `Open`. If we store the name of the field we every time have to parse between these two values. Please correct me if i'm wrong.

Haacked: So one point of confusion is that the bug states the bug is in serialization. But this is changing deserialization code. Could you add some tests to `SimpleJsonSerializerTests.cs` that demonstrate the problem and are fixed by this code?

Haacked: Nevermind. I see that you fixed both directions. I wrote some tests I'll add.

Haacked:
selfie-4

PR #1404, opened 2016-06-21
Code to test URL for an enterprise instance.

Haacked:
Based on code we use in GitHub Desktop for Windows.

This is take two of #1097. In this iteration, we don't check for the following header `Server: GitHub.com` because a reverse proxy could rewrite it.

Instead, we only check for the header `X-GitHub-Request-Id` to know that it's a GitHub server.


shiftkey:
Code looks good, a couple of questions:

 - do we want to make this available under `GitHubClient`?
 - is it worth writing up some docs around this? Even just a quick sample?

Haacked:
> do we want to make this available under GitHubClient?

Under the current design of the code, turns out to be not so easy. And you might not want the probe to use the same credentials registered with the GitHubClient anyways. We could try it the way it is for now and change it if we hear anything from people.

> is it worth writing up some docs around this? Even just a quick sample?

Sure!

Octokit/Clients/Enterprise/EnterpriseProbe.cs
```diff
@@ -0,0 +1,100 @@
...
+    /// <summary>
+    /// Service used to check if a GitHub Enterprise instance lives at a given URL.
+    /// </summary>
+    public class EnterpriseProbe : IEnterpriseProbe
+    {
+        static readonly Uri endPoint = new Uri("/site/sha", UriKind.Relative);
+        readonly ProductHeaderValue productHeader;
+        readonly IHttpClient httpClient;
+
+        public EnterpriseProbe(ProductHeaderValue productHeader, IHttpClient httpClient)
```
ryangribble: Can we add an overload that doesn't take the `httpClient` and instead defaults it to `new HttpClientAdapter(HttpMessageHandlerFactory.CreateDefault)` 

This is consistent with the way most normal users would be using octokit, e.g they dont really have any idea about `IHttpClient` or the fact that a default one can be created using the `HttpMessageHandlerFactory`.  And yes I do acknowledge the documentation example does show how to use the handler factory, but I think it would be more inline with the overloads on `GitHubClient` (well, really it's the `Connection` class) to not have to specify it for regular use

Haacked:
@ryangribble @shiftkey I finally got around to adding the proposed ctor.

ryangribble:
Thanks @Haacked , i've just pushed a couple of commits to this branch to suppress the code analysis rule violation that failed the build, and also tidied up the `ctor` doc and getting-started example.

PR #1408, opened 2016-06-30
Add some unit tests for enum serialization

Haacked:


Haacked:
These test failures on Mono don't seem to be related to anything I've done. @shiftkey thoughts?

ryangribble:
We often have inconsistent travis failures on linux and/or OSX. I've restarted them to see if they pass this time, but it would be good for someone who knows that ecosystem to figure out the root cause and if there's something we can do to avoid the failures.  We were tracking them here #1076

Haacked:
Ah, thanks for the info @ryangribble and thanks for merging @shiftkey!

Issue #1101, opened 2016-02-09
Add user as a collaborator with permission parameter


snadon:
Should we be able to set permissions when adding a collaborator to a repository?

The GitHub API supports it:
https://developer.github.com/v3/repos/collaborators/#add-user-as-a-collaborator

But Octokit.Net does not:
https://github.com/octokit/octokit.net/blob/master/Octokit%2FClients%2FIRepoCollaboratorsClient.cs#L44

Haacked:
Yeah, that looks like an oversight on our part. It think we should add an overload that accepts an enum with the available permission options.

Haacked:
Want to submit a PR?

M-Zuber:
And here I was talking bad about you.
![image](https://cloud.githubusercontent.com/assets/4746537/12930800/15d71fc8-cf83-11e5-8110-19b580e11c1f.png)
Please accept my apologies :smile:

Haacked:
> Please accept my apologies

LOLOLOL! No apologies necessary. You're usually correct. I happen to be taking a break to try and respond to all the OSS projects I maintain so the timing was lucky.

ryangribble:
Just a question on this out of interest... 

> I think we should add an overload that accepts an enum with the available permission options.

Are you saying @Haacked that this body parameter should actually be explicitly in the overloaded function, or should it be what appears to be the "usual" Octokit way where a Request object is passed in?

Just trying to get a feel for the octokit.net conventions, most I had come across seem to use the request object approach, so was surprised to see you mention adding an overload directly taking the enum paramter (unless i misunderstood)

Enum directly in Ad function
```
enum CollaboratorPermission { Read, Write, Admin }

Task Add(string owner, string repo, string user, CollaboratorPermission permission);
```

vs

Using a request object
```
enum CollaboratorPermission { Read, Write, Admin }

public class CollaboratorUpdate
{
   public CollaboratorPermission Permission { get; set; }
}

Task Add(string owner, string repo, string user, CollaboratorUpdate update);
```

Haacked:
@ryangribble you're exactly right, it should be a request object. I was being lazy with my words. :smile:

ryangribble:
Sweet, my understanding remains intact 😁

ryangribble:
[...]

PR #1410, opened 2016-07-01
Repository invitations changes

### LINKED ISSUE #1399: Repository Invitations Changes
### Summary
Recently announced on this [blog post](https://developer.github.com/changes/2016-06-14-repository-invitations/), Repository Invitations are now in preview in the GitHub API with changes in the following areas:

[Repository Collaborators](https://developer.github.com/v3/repos/collaborators/#add-user-as-a-collaborator) API
[Repository Invitations](https://developer.github.com/v3/repos/invitations/) API (new)

### Repository Collaborators
[Adding a collaborator](https://developer.github.com/v3/repos/collaborators/#add-user-as-a-collaborator) can now send an invite to the collaborator rather than directly adding them.  

In octokit.net this could be implemented by copying the [`Add()`](https://github.com/octokit/octokit.net/blob/26cc074d743612fc1f0c9268dad1e4c2a15c3fc4/Octokit/Clients/IRepoCollaboratorsClient.cs#L59) to create an `Invite()` function (and setting the new preview accepts header on the Http Post call in the `Invite()` call).

### Managing Invitations
A new `Invitations` API has been added to `Repositories`

In octokit.net this should be implemented as `IRepositoryInivtationsClient` and be a sub client of the `IRepositoriesClient` accessed by `GitHubClient.Repository.Invitation` with methods
- `GetAll()`
`GET /repositories/:repo_id/invitations`
- `Delete()` 
`DELETE /repositories/:repo_id/invitations/:invitation_id`
- `Update()` 
`PATCH /repositories/:repo_id/invitations/:invitation_id`
- `GetAllForCurrent()` 
`GET /user/repository_invitations`
- `Accept()` 
`PATCH /user/repository_invitations/:invitation_id`
- `Decline()` 
`DELETE /user/repository_invitations/:invitation_id`

Reminder that like all functionality in "preview mode" a custom accepts header needs to be specified (see AcceptHeaders.cs)

> To access the API during the preview period, you must provide a custom media type in the Accept header:
> `application/vnd.github.swamp-thing-preview+json`

### Working on this issue
This issue has been flagged as `up-for-grabs` as it is a reasonably easy issue for a new contributor to pick up, and be guided through the process of contributing to octokit
(that said, if an existing contributor wants to work on this, by all means go for it!)

Please comment here if you want to take this issue.

Please read and follow the [CONTRIBUTING guide](https://github.com/octokit/octokit.net/blob/master/CONTRIBUTING.md) guide and if you need any help, just ask!  Don't be afraid to open a `[WIP]` (work in progress) Pull Request once you've started implementing the changes

maddin2016:
fixed #1399 

add a new `RepositoryInvitationsClient` to invite a user to collaborate. I have to create some new overloads and classes. 

Todo
- add unit and integration tests


ryangribble:
Build is failing convention tests due to missing unit test class for `TheCtor` in `RepositoryInvitationsClientTests.cs`

Octokit.Reactive/Clients/IObservableRepositoryInvitationsClient.cs
```diff
@@ -0,0 +1,66 @@
...
+        IObservable<RepositoryInvitation> GetAllForCurrent();
+
+        /// <summary>
+        /// Gets all the invitations on a repository.
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="https://developer.github.com/v3/repos/invitations/#list-invitations-for-a-repository">API documentation</a> for more information.
+        /// </remarks>        
+        /// <param name="id">The id of the repository</param>        
+        IObservable<RepositoryInvitation> GetAllForRepository(int id);
```
ryangribble: I think for consistency, this variable should be named `repositoryId`

Octokit.Tests.Integration/Clients/RepositoryCollaboratorClientTests.cs
```diff
@@ -136,4 +136,24 @@ public class TheIsCollaboratorMethod
             }
         }
     }
+
+    public class TheInviteNewCollaboratorMethod
```
ryangribble: this should be named as per the method being tested, eg `TheInviteMethod`

Octokit.Tests/Clients/RepositoryInvitationsClientTests.cs
```diff
@@ -0,0 +1,95 @@
...
+            var connection = Substitute.For<IApiConnection>();
+            var client = new RepositoryInvitationsClient(connection);
+
+            await client.Delete(1, 2);
+
+            connection.Received().Delete(Arg.Is<Uri>(u => u.ToString() == "repositories/1/invitations/2"), Arg.Any<object>(), "application/vnd.github.swamp-thing-preview+json");
+        }
+    }
+
+    public class TheEditMethod
```
ryangribble: should probably have a unit test for when `InvitiationUpdate` parameter is `null`

Octokit.Tests/Reactive/ObservableRepoCollaboratorsClientTests.cs
```diff
@@ -170,6 +170,71 @@ public void CallsCreateOnRegularDeploymentsClient()
...
+                Assert.Throws<ArgumentNullException>(() => _client.Invite("owner", "repo", null));
+            }
+
+            [Fact]
+            public void EnsuresNonEmptyArguments()
+            {
+                SetupWithNonReactiveClient();
+
+                Assert.Throws<ArgumentException>(() => _client.Invite("", "repo", "user"));
+                Assert.Throws<ArgumentException>(() => _client.Invite("owner", "", "user"));
```
ryangribble: how about when the `user` parameter is empty string?

Octokit.Tests/Reactive/ObservableRepoCollaboratorsClientTests.cs
```diff
@@ -170,6 +170,71 @@ public void CallsCreateOnRegularDeploymentsClient()
...
+
+            [Fact]
+            public async Task EnsuresNonWhitespaceArguments()
+            {
+                SetupWithNonReactiveClient();
+
+                await AssertEx.ThrowsWhenGivenWhitespaceArgument(
+                    async whitespace => await _client.Invite(whitespace, "repo", "user"));
+                await AssertEx.ThrowsWhenGivenWhitespaceArgument(
+                    async whitespace => await _client.Invite("owner", whitespace, "user"));
```
ryangribble: Add case for `whitespace` being passed as `user`

Octokit.Tests/Reactive/ObservableRepoCollaboratorsClientTests.cs
```diff
@@ -170,6 +170,71 @@ public void CallsCreateOnRegularDeploymentsClient()
...
+            }
+
+            [Fact]
+            public void CallsCreateOnRegularDeploymentsClient()
```
ryangribble: typo?  Assumedly this should be named `CallsInviteOnRegularDeploymentsClient`

Octokit/Clients/IRepoCollaboratorsClient.cs
```diff
@@ -59,6 +59,16 @@ public interface IRepoCollaboratorsClient
         Task Add(string owner, string repo, string user);
 
         /// <summary>
+        /// Invites a new collaborator to the repo
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="http://developer.github.com/v3/repos/collaborators/#add-collaborator">API documentation</a> for more information.
+        /// </remarks>
+        /// <exception cref="ApiException">Thrown when a general API error occurs.</exception>
+        /// <returns><see cref="Task"/></returns>
```
ryangribble: Incorrect return type for this function

Octokit/Models/Response/RepositoryInvitation.cs
```diff
@@ -0,0 +1,54 @@
+﻿using System;
+using System.Diagnostics;
+using System.Globalization;
+
+namespace Octokit
+{
+    public enum InvitationPermissionType
```
ryangribble: There is already a `RepositoryPermissions` enum that should be used for these values

ryangribble:
Nice work @maddin2016 !  :grinning:  I've made a few comments on various diffs, and also have some overall comments:

### API Calls that return no payload
With the calls that dont return a payload, you'll note the API docs do actually indicate that a "success" is a Http 204 (NoContent).  So that means we can actually implement these calls (`Accept`, `Decline`, `Delete`) return a `Task<bool>` and then you can actually get the `HttpStatusCode` and return `true` or `false` accordingly.  There are several examples of this in the code base already, one such example is `TeamsClient.RemoveMembership()` if you want to take a look at how it's done

### XmlDoc comments
There are some cases where the XmlDoc comments are not consistent or mention incorrect return types.  I realise some of this is because the comment you may have copied was, itself, incorrect, but we should aim to have correct comments at leats for new work we are adding.  So instead of `IReadOnlyPagedCollection` you should be saying `IReadonlyList` or whatever the actual return type is.  Also we have been discussing the `<returns>` doc tags and came to the decision that unless something specific needs to be called out about the function, it isn't really adding any benefit to document "the obvious" eg "returns a collection of repository invites".  So you also have the option to completely remove `<returns>` tag from any functions where it is not adding value by having it :+1:

Octokit/Clients/IRepoCollaboratorsClient.cs
```diff
@@ -59,6 +59,16 @@ public interface IRepoCollaboratorsClient
...
+        Task<RepositoryInvitation> Invite(string owner, string repo, string user);
```
ryangribble: According to the API docs, this function should actually take a request object stating the `permissions` for the user to  have.  This is missing from the `Add()` method which you would have copied for the `Invite()` method (an issue is raised for that but hasn't been fixed yet).  But we may as well at least include the `permissions` request on the `Invite()` method

Octokit/Models/Request/InvitationUpdate.cs
```diff
@@ -0,0 +1,30 @@
...
+
+namespace Octokit
+{
+    /// <summary>
+    /// Used to update a invitation.
+    /// </summary>
+    /// <remarks>    
+    /// </remarks>
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
+    public class InvitationUpdate
```
ryangribble: I made a previous comment that the `permissions` are actually provided on the initial `Invite()` as well as on this update (and infact will also be added to the `Add()` method on the collaborators client, when that issue is fixed), so this class could probably be re-used for all cases.  This means it should be renamed to something like `CollaboratorRequest` rather than `InvitationUpdate`

Octokit/Models/Response/RepositoryInvitation.cs
maddin2016: That's right. But the `RepositoryPermission` enum uses `admin, push, pull` and we need `admin, write, read`. So because of that i created a new enum

ryangribble: Ah indeed you are right :grinning: the List Invite and Edit Invite calls do have this admin/write/read options.  But the Add and Invite methods need the Admin/Push/Pull.  So i guess you should keep what you have here, but additionally add a `CollaboratorRequest` class that uses the `RepositoryPermission` enum, and use that on the `Invite()` method

Octokit.Reactive/Clients/IObservableRepositoryInvitationsClient.cs
```diff
@@ -0,0 +1,69 @@
...
+        [SuppressMessage("Microsoft.Design", "CA1024:UsePropertiesWhereAppropriate")]
+        IObservable<RepositoryInvitation> GetAllForCurrent();
+
+        /// <summary>
+        /// Gets all the invitations on a repository.
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="https://developer.github.com/v3/repos/invitations/#list-invitations-for-a-repository">API documentation</a> for more information.
+        /// </remarks>        
+        /// <param name="id">The id of the repository</param>
```
ryangribble: Missed a rename here

ryangribble:
There is some inconsistency with variable naming for the invitation Id (`id` on the Accept/Decline endpoints, but `invitationId` on the GetAll endpoints).  It would be good to have consistent naming for this - either always `id` or always `invitationId`

The `<returns>` tags still have some tweaks required to be correct/useful - I'd actually suggest just completely removing the entire `<returns>xxxx</returns>` doc comment, since they really aren't adding any value by repeating what the function return value is anyway

maddin2016:
I think i'm ready. appveyor test is failing :sweat:

ryangribble:
It seems like it timed out on a build step.  I dont have AppVeyor access so unless you have another commit to push we will need to wait for @shiftkey to give it a kick

So earlier you mentioned you had issues where the invitation email was never received.  Is that still the case or is everything working now?

Octokit.Reactive/Clients/IObservableRepoCollaboratorsClient.cs
```diff
@@ -41,6 +41,16 @@ public interface IObservableRepoCollaboratorsClient
         IObservable<Unit> Add(string owner, string repo, string user);
 
         /// <summary>
+        /// Invites a user as a collaborator to a repository.
+        /// </summary>
+        /// <param name="owner">The owner of the repository</param>
+        /// <param name="repo">The name of the repository</param>
+        /// <param name="user">The username of the prospective collaborator</param>
+        /// <param name="permission">The permission to set</param>
+        /// <returns></returns>
```
ryangribble: remove `<returns>` line as it's empty anyway

Octokit.Reactive/Clients/IObservableRepositoryInvitationsClient.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// <param name="invitationId">The id of the invitation.</param>        
+        IObservable<bool> Delete(int repositoryId, int invitationId);
+
+        /// <summary>
+        /// Gets all invitations for the current user.
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="https://developer.github.com/v3/repos/invitations/#list-a-users-repository-invitations">API documentation</a> for more information.
+        /// </remarks>       
+        /// <returns>A <see cref="IReadOnlyList{RepositoryInvitation}"/> of <see cref="RepositoryInvitation"/>.</returns>
```
ryangribble: remove `<returns>` line as it's incorrect (should be `Observable` not `IReadOnlyList`) but also it's not adding any value to simply say what the return type is, so just :fire: the whole line

Octokit.Reactive/Clients/IObservableRepositoryInvitationsClient.cs
```diff
@@ -0,0 +1,69 @@
...
+        IObservable<RepositoryInvitation> GetAllForCurrent();
+
+        /// <summary>
+        /// Gets all the invitations on a repository.
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="https://developer.github.com/v3/repos/invitations/#list-invitations-for-a-repository">API documentation</a> for more information.
+        /// </remarks>        
+        /// <param name="repositoryId">The id of the repository</param> 
+        /// <returns>A <see cref="IReadOnlyList{RepositoryInvitation}"/> of <see cref="RepositoryInvitation"/>.</returns>
```
ryangribble: :fire: `<returns>` line

Octokit.Reactive/Clients/ObservableRepoCollaboratorsClient.cs
```diff
@@ -73,6 +73,24 @@ public IObservable<Unit> Add(string owner, string repo, string user)
         }
 
         /// <summary>
+        /// Invites a user as a collaborator to a repository.
+        /// </summary>
+        /// <param name="owner">The owner of the repository</param>
+        /// <param name="repo">The name of the repository</param>
+        /// <param name="user">The username of the prospective collaborator</param>
+        /// <param name="permission">The permission to set</param>
+        /// <returns></returns>
```
ryangribble: :fire: `<returns>` line

Octokit.Reactive/Clients/ObservableRepositoryInvitationsClient.cs
```diff
@@ -0,0 +1,98 @@
...
+            return _client.Delete(repositoryId, invitationId).ToObservable();
+        }
+
+        /// <summary>
+        /// Updates a repository invitation.
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="https://developer.github.com/v3/repos/invitations/#update-a-repository-invitation">API documentation</a> for more information.
+        /// </remarks>
+        /// <param name="repositoryId">The id ot the repository.</param>
```
ryangribble: typo
"o~~t~~f the repository"

Octokit/Clients/IRepoCollaboratorsClient.cs
```diff
@@ -54,18 +54,30 @@ public interface IRepoCollaboratorsClient
...
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="http://developer.github.com/v3/repos/collaborators/#add-collaborator">API documentation</a> for more information.
+        /// </remarks>
+        /// <param name="owner">The owner of the repository.</param>
+        /// <param name="repo">The name of the repository.</param>
+        /// <param name="user">The name of the user to invite.</param>
+        /// <param name="permission">The permission to set.</param>
+        /// <exception cref="ApiException">Thrown when a general API error occurs.</exception>
+        /// <returns><see cref="RepositoryInvitation"/></returns>
```
ryangribble: :fire: `<returns>` line

Octokit/Clients/IRepositoryInvitationsClient.cs
```diff
@@ -0,0 +1,82 @@
...
+
+        /// <summary>
+        /// Decline a repository invitation.
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="https://developer.github.com/v3/repos/invitations/#decline-a-repository-invitation">API documentation</a> for more information.
+        /// </remarks>        
+        /// <param name="invitationId">The id of the invitation</param>
+        /// <exception cref="ApiException">Thrown when a general API error occurs.</exception>        
+        /// <returns><see cref="Task"/></returns>
```
ryangribble: :fire: `<returns>` line

Octokit/Clients/IRepositoryInvitationsClient.cs
```diff
@@ -0,0 +1,82 @@
...
+        Task<bool> Delete(int repositoryId, int invitationId);
+
+        /// <summary>
+        /// Gets all invitations for the current user.
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="https://developer.github.com/v3/repos/invitations/#list-a-users-repository-invitations">API documentation</a> for more information.
+        /// </remarks>        
+        /// <exception cref="ApiException">Thrown when a general API error occurs.</exception>
+        /// <returns>A <see cref="IReadOnlyList{RepositoryInvitation}"/> of <see cref="RepositoryInvitation"/>.</returns>
```
ryangribble: :fire: `<returns>` line

Octokit/Clients/IRepositoryInvitationsClient.cs
```diff
@@ -0,0 +1,82 @@
...
+
+        /// <summary>
+        /// Gets all the invitations on a repository.
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="https://developer.github.com/v3/repos/invitations/#list-invitations-for-a-repository">API documentation</a> for more information.
+        /// </remarks>        
+        /// <param name="repositoryId">The id of the repository</param>
+        /// <exception cref="ApiException">Thrown when a general API error occurs.</exception>
+        /// <returns>A <see cref="IReadOnlyList{RepositoryInvitation}"/> of <see cref="RepositoryInvitation"/>.</returns>
```
ryangribble: :fire: `<returns>` line

Octokit/Clients/IRepositoryInvitationsClient.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// Updates a repository invitation.
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="https://developer.github.com/v3/repos/invitations/#update-a-repository-invitation">API documentation</a> for more information.
+        /// </remarks>
+        /// <param name="repositoryId">The id ot the repository</param>
+        /// <param name="invitationId">The id of the invitation</param>
+        /// <param name="permissions">The permission for the collsborator</param>
+        /// <exception cref="ApiException">Thrown when a general API error occurs.</exception>
+        /// <returns><see cref="RepositoryInvitation"/></returns>
```
ryangribble: :fire: `<returns>` line

Octokit/Clients/RepoCollaboratorsClient.cs
```diff
@@ -109,14 +108,35 @@ public Task Add(string owner, string repo, string user)
...
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="http://developer.github.com/v3/repos/collaborators/#add-collaborator">API documentation</a> for more information.
+        /// </remarks>
+        /// <param name="owner">The owner of the repository.</param>
+        /// <param name="repo">The name of the repository.</param>
+        /// <param name="user">The name of the user to invite.</param>
+        /// <param name="permission">The permission to set.</param>
+        /// <exception cref="ApiException">Thrown when a general API error occurs.</exception>
+        /// <returns><see cref="RepositoryInvitation"/></returns>
```
ryangribble: :fire: `<returns>` line

Octokit/Clients/RepositoryInvitationsClient.cs
```diff
@@ -0,0 +1,131 @@
...
+
+        /// <summary>
+        /// Accept a repository invitation.
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="https://developer.github.com/v3/repos/invitations/#accept-a-repository-invitation">API documentation</a> for more information.
+        /// </remarks>        
+        /// <param name="invitationId">The id of the invitation</param>
+        /// <exception cref="ApiException">Thrown when a general API error occurs.</exception>        
+        /// <returns><see cref="Task"/></returns>
```
ryangribble: :fire: `<returns>` line

Octokit/Clients/RepositoryInvitationsClient.cs
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        /// <summary>
+        /// Gets all invitations for the current user.
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="https://developer.github.com/v3/repos/invitations/#list-a-users-repository-invitations">API documentation</a> for more information.
+        /// </remarks>        
+        /// <exception cref="ApiException">Thrown when a general API error occurs.</exception>
+        /// <returns>A <see cref="IReadOnlyList{RepositoryInvitation}"/> of <see cref="RepositoryInvitation"/>.</returns>
```
ryangribble: :fire: `<returns>` line

Octokit/Clients/RepositoryInvitationsClient.cs
```diff
@@ -0,0 +1,131 @@
...
+
+        /// <summary>
+        /// Gets all the invitations on a repository.
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="https://developer.github.com/v3/repos/invitations/#list-invitations-for-a-repository">API documentation</a> for more information.
+        /// </remarks>        
+        /// <param name="repositoryId">The id of the repository</param>
+        /// <exception cref="ApiException">Thrown when a general API error occurs.</exception>
+        /// <returns>A <see cref="IReadOnlyList{RepositoryInvitation}"/> of <see cref="RepositoryInvitation"/>.</returns>
```
ryangribble: :fire: `<returns>` line

Octokit/Clients/RepositoryInvitationsClient.cs
```diff
@@ -0,0 +1,131 @@
...
+        /// Updates a repository invitation.
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="https://developer.github.com/v3/repos/invitations/#update-a-repository-invitation">API documentation</a> for more information.
+        /// </remarks>
+        /// <param name="repositoryId">The id ot the repository</param>
+        /// <param name="invitationId">The id of the invitation</param>
+        /// <param name="permissions">The permission for the collsborator</param>
+        /// <exception cref="ApiException">Thrown when a general API error occurs.</exception>
+        /// <returns><see cref="RepositoryInvitation"/></returns>
```
ryangribble: :fire: `<returns>` line

Octokit/Helpers/ApiUrls.cs
```diff
@@ -2959,5 +2959,45 @@ public static Uri Reactions(int number)
...
+        {
+            return "repositories/{0}/invitations".FormatUri(repositoryId);
+        }
+
+        /// <summary>
+        /// Returns the <see cref="Uri"/> for repository invitations.
+        /// </summary>
+        /// <param name="repositoryId">The id of the repository</param>
+        /// <param name="invitationId">The id of the invitation</param>
+        /// <returns>The <see cref="Uri"/> for repository invitations.</returns>
```
ryangribble: This should probably say "for a single repository invitation"

Octokit/Helpers/ApiUrls.cs
```diff
@@ -2959,5 +2959,45 @@ public static Uri Reactions(int number)
...
+        public static Uri UserInvitations()
+        {
+            return "user/repository_invitations".FormatUri();
+        }
+
+        /// <summary>
+        /// Returns the <see cref="Uri"/> for invitations for the current user.
+        /// </summary>
+        /// <param name="invitationId">The id of the invitation</param>
+        /// <returns>The <see cref="Uri"/> for invitations for the current user.</returns>
```
ryangribble: "for a single invitation of the current user"

Octokit/Models/Request/CollaboratorRequest.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System.Diagnostics;
+using System.Globalization;
+
+namespace Octokit
+{
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
+    public class CollaboratorRequest
```
ryangribble: I'm actually wondering whether we should just be using the existing `RepositoryPermissionRequest` request class for this (currently used in `TeamsClient.AddRepository()`).  Not sure what @shiftkey thinks about re-using a request model in different clients (eg TeamsClient and RepoCollaboratorsClient) vs keeping them separate, incase they ever diverge

Octokit.Tests.Integration/Clients/RepositoryCollaboratorClientTests.cs
```diff
@@ -136,4 +136,26 @@ public class TheIsCollaboratorMethod
...
+            var github = Helper.GetAuthenticatedClient();
+            var repoName = Helper.MakeNameWithTimestamp("public-repo");
+
+            using (var context = await github.CreateRepositoryContext(new NewRepository(repoName)))
+            {
+                var fixture = github.Repository.Collaborator;
+                var permission = new CollaboratorRequest();
+
+                // invite a collaborator
+                var response = await fixture.Invite(context.RepositoryOwner, context.RepositoryName, "maddin2016", permission);
```
maddin2016: This test always fails because context is trying to delete this repo and you cannot delete a repo until there is a pending invitation. And yes, it still not sending any email. Next question is, how should i write integration tests for invitations. Is there any test client where user can send invitations to?

ryangribble: >  you cannot delete a repo until there is a pending invitation
Did you mean cannot delete *when* there is a pending invite?  If so then I guess you could `Delete` the invitation before trying to delete the repo?

Im not sure about why the email is not received.  I will try some invites myself when I can...

maddin2016: youre right. **When** there is a pending invite. Ok. Then i use my `lrz-hal` account.

maddin2016: @ryangribble there is some problem with deleting the invitation
```c#
using (var context = await github.CreateRepositoryContext(new NewRepository(repoName)))
            {
                var fixture = github.Repository.Collaborator;
                var permission = new CollaboratorRequest();

                // invite a collaborator
                var response = await fixture.Invite(context.RepositoryOwner, context.RepositoryName, "lrz-hal", permission);

                Assert.Equal("lrz-hal", response.Invitee.Login);
                Assert.Equal(InvitationPermissionType.Write, response.Permissions);

                var deleteInvite = await github.Repository.Invitation.Delete(context.Repository.Id, response.Id);
            }
```
The problem is when `fixture.invite` is calling, in the same moment `context` is trying to delete the repo. So i cannot delete the invitation, because i don't reach that call!!??

maddin2016: `Dispose` of `RepositoryContext` is calling too early.

ryangribble: Are you sure an exception isnt happening in the fixture.Invite() method?  When an exception is thrown, that will cause the `using (var context) { }` to go out of scope and then if you are debugging you WILL see it doint the disposal/deletion.  But I'd say the true problem is your `fixture.Invite()` has thrown an exception

maddin2016: It breakes [here](https://github.com/octokit/octokit.net/blob/master/Octokit/SimpleJson.cs#L1472) with `RepositoryInvitation` :sweat:

ryangribble: Ah right!  

The deserializer creates a "constructor cache" of all the constructors for the request/response objects, which it uses to instantiate the class, then it uses the protected setter properties to populate any fields it finds in the json payload, on the object it created.

But it can't use constructors that take parameters so all model classes actually require a parameterless constructor, which we have missed on the `RepositoryInvitation` class!

maddin2016: :tada: That's awesome!! So then my comment 

> you cannot delete a repo until there is a pending invitation
is wrong. It deletes the repo and the invitation.

Octokit/Models/Response/RepositoryInvitation.cs
```diff
@@ -14,6 +14,10 @@ public enum InvitationPermissionType
     [DebuggerDisplay("{DebuggerDisplay,nq}")]
     public class RepositoryInvitation
     {
+        public RepositoryInvitation()
```
ryangribble: for consistency with all the other response class ctors, can you put this on a single line?

`public RepositoryInvitation() { }`

maddin2016:
Is there something more to do before merging this request?

Octokit/Helpers/ApiUrls.cs
ryangribble: I think there is just this comment to fix up still

maddin2016:
i'm now writing the unit tests

maddin2016:
Is it ok to use the `octokat` for the invitations?

Octokit/Models/Request/CollaboratorRequest.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        public CollaboratorRequest()
+        {
+            Permission = Permission.Push;
```
ryangribble: Im not sure if we should have a paramaterless constructor and a default value of Push here... let's :fire: this and just leave the constructor that forces the user to state explictly what permission they want the invitee to have

maddin2016: In the [api](https://developer.github.com/v3/repos/collaborators/#add-user-as-a-collaborator) the default value is push. So i thought it would be correct.

Octokit/Models/Request/CollaboratorRequest.cs
```diff
@@ -0,0 +1,29 @@
...
+        }
+
+        public CollaboratorRequest(Permission permissions)
```
ryangribble: Since this is a request object, we should put some XmlDoc on this constructor, to give the users intellisense.  Check out `RepositoryPermissionRequest` for an example

Octokit/Models/Request/CollaboratorRequest.cs
```diff
@@ -0,0 +1,29 @@
...
+        {
+            Permission = permissions;
+        }
+
+        public Permission Permission { get; private set; }
```
ryangribble: Since this is a request object, we should put some XmlDoc on this field.  Check out `RepositoryPermissionRequest` for an example

maddin2016:
`Used to set the permission for a collaborator.`
`The permission to grant the collaborator on this repository.`
??

Octokit/Models/Request/CollaboratorRequest.cs
ryangribble: I guess the API default only really applies if we dont pass the field at all?
I dont really have a problem with having to explicitly specify the permission when constructing the object, but if you feel strongly enough about it feel free to keep the default constructor, just make sure to add an XmlDoc comment to state that it defaults to Push access

maddin2016:
What's with the default constructor. Should we let the user decide?

Octokit/Models/Request/CollaboratorRequest.cs
maddin2016: Ah. That's right. Failure of mine. So if i understand it correct, i we don't set the `CollaboratorRequest` the api sets `push` as the default value. Right?

maddin2016: So then we should have two overloads for `Add` and `Invite` in `RepoCollaboratorsClient`!?

ryangribble: I think so.  but from a consistency point of view we dont do optional parameters that much.  I think in this case it's ok to always pass the request object and only have the `ctor` to explicitly state the permission required.  Nothing wrong with being explicit, particularly when its something like granting access to someone

ryangribble: > So then we should have two overloads for Add and Invite in RepoCollaboratorsClient!? Because the docs say Only valid on organization-owned repositories.

I missed that in the docs.  Im confused whether it is saying the "permissions" attribute is only applicable to org repositories, or the whole "Add collaborator" endpoint.

ryangribble:
So Im running the integration tests from my test account, invitnig my real account and i get no email and no notification on github about being invited.  yet if i explicitly visit the `HtmlUrl` property in the response then it does say ive been invited and i can accept/decline.

Yet when i interactively use my test account and invite my real account to a repo, i get an email and the github notifications section shows it there... :confused:

This is all on a personal account of my tes repo.  I guess given that comment in api docs about organization repos, we should see if anything is different when using the test user's organization rather than personal repo

ryangribble:
Also confirming that yes the permissions only applies to organization repos.  so probably should update the Xml doc comments on the parameter to the `Invite()` method to specify that

Personal repo
image

Org repo
image




But still doesnt explain why the email and github notifications arent received for invites done through the API preview (but they are actually created as evidenced by the invite link working for the invitee, and the collaborators section showing the pending invite from the inviters side)

maddin2016:
For summary:
 `Add` and `Invite` Methods should have the `permissions` attribute with note `Only valid on organization-owned repositories`. Right?

ryangribble:
Oh nice, you decided to tackle #1101 while you're in this area? Thanks! 

On the topic of the email/notifications not occuring, I contacted github support and they confirm they can reproduce it, they are logging an internal issue, so we shouldn't need to worry about that aspect at least

maddin2016:
Ah. Right. This I have forget to tell you :grin:

Octokit/Clients/RepositoryInvitationsClient.cs
```diff
@@ -0,0 +1,127 @@
...
+        public async Task<bool> Delete(int repositoryId, int invitationId)
+        {
+            var endpoint = ApiUrls.RepositoryInvitations(repositoryId, invitationId);
+
+            try
+            {
+                var httpStatusCode = await ApiConnection.Connection.Delete(endpoint, new object(), AcceptHeaders.InvitationsApiPreview).ConfigureAwait(false);
+                return httpStatusCode == HttpStatusCode.NoContent;
+            }
+            catch
```
ryangribble: For consistency with other simliar clients/endpoints, this should be 
``` csharp
catch (NotFoundException)
```

rather than catching any type of exception

Octokit/Clients/RepositoryInvitationsClient.cs
```diff
@@ -0,0 +1,127 @@
...
+        public async Task<bool> Decline(int invitationId)
+        {
+            var endpoint = ApiUrls.UserInvitations(invitationId);
+
+            try
+            {
+                var httpStatusCode = await ApiConnection.Connection.Delete(endpoint, new object(), AcceptHeaders.InvitationsApiPreview).ConfigureAwait(false);
+                return httpStatusCode == HttpStatusCode.NoContent;
+            }
+            catch
```
ryangribble: For consistency with other simliar clients/endpoints, this should be 
``` csharp
catch (NotFoundException)
```

rather than catching any type of exception

Octokit/Clients/RepositoryInvitationsClient.cs
```diff
@@ -0,0 +1,127 @@
...
+        public async Task<bool> Accept(int invitationId)
+        {
+            var endpoint = ApiUrls.UserInvitations(invitationId);
+
+            try
+            {
+                var httpStatusCode = await ApiConnection.Connection.Patch(endpoint, AcceptHeaders.InvitationsApiPreview).ConfigureAwait(false);
+                return httpStatusCode == HttpStatusCode.NoContent;
+            }
+            catch
```
ryangribble: For consistency with other simliar clients/endpoints, this should be 
``` csharp
catch (NotFoundException)
```

rather than catching any type of exception

Octokit/Clients/RepositoryInvitationsClient.cs
```diff
@@ -0,0 +1,127 @@
...
+        public async Task<bool> Accept(int invitationId)
+        {
+            var endpoint = ApiUrls.UserInvitations(invitationId);
+
+            try
+            {
+                var httpStatusCode = await Connection.Patch(endpoint, AcceptHeaders.InvitationsApiPreview).ConfigureAwait(false);
+                return httpStatusCode == HttpStatusCode.NoContent;
+            }
+            catch
```
ryangribble: For consistency with other simliar clients/endpoints, this should be 
``` csharp
catch (NotFoundException)
```

rather than catching any type of exception

Octokit/Clients/RepositoryInvitationsClient.cs
```diff
@@ -0,0 +1,127 @@
...
+        public async Task<bool> Decline(int invitationId)
+        {
+            var endpoint = ApiUrls.UserInvitations(invitationId);
+
+            try
+            {
+                var httpStatusCode = await Connection.Delete(endpoint, new object(), AcceptHeaders.InvitationsApiPreview).ConfigureAwait(false);
+                return httpStatusCode == HttpStatusCode.NoContent;
+            }
+            catch
```
ryangribble:
``` csharp
catch (NotFoundException)
```

Octokit/Clients/RepositoryInvitationsClient.cs
```diff
@@ -0,0 +1,127 @@
...
+        public async Task<bool> Delete(int repositoryId, int invitationId)
+        {
+            var endpoint = ApiUrls.RepositoryInvitations(repositoryId, invitationId);
+
+            try
+            {
+                var httpStatusCode = await Connection.Delete(endpoint, new object(), AcceptHeaders.InvitationsApiPreview).ConfigureAwait(false);
+                return httpStatusCode == HttpStatusCode.NoContent;
+            }
+            catch
```
ryangribble:
``` csharp
catch (NotFoundException)
```

ryangribble:
Nice work on the overloads for `Add()` and `Invite()` to make the permissions parameter optional, since it doest apply to non organization repositories :+1:

Looking at that though, since #1387 added an overload to the `Add()` method that takes `repositoryId` rather than `owner` and `repositoryName` parameters, we probably need the same on the new `Invite()` method (as long as it does actually work).

ryangribble:
Also as part of the repositoryId work, @dampir was trying to be consistent with always using `name` for the repositoryName parameter rather than previously a mix of `name` `repo` `repoName` `repositoryName` etc

So since he standardised on `name`, are you able to change `repo` to `name` in your `Add()` method?

Octokit.Tests.Integration/Clients/RepositoryInvitationsClientTests.cs
```diff
@@ -0,0 +1,111 @@
...
+                var response = await fixture.Invite(context.RepositoryOwner, context.RepositoryName, context.RepositoryOwner);
+
+                Assert.Equal(context.RepositoryOwner, response.Invitee.Login);
+                Assert.Equal(InvitationPermissionType.Write, response.Permissions);
+
+                var declineResponse = await github.Repository.Invitation.Decline(response.Id);
+
+                Assert.True(declineResponse);
+
+                //Assert.True(invitations.Count >= 1);
```
shiftkey: Can we :fire: this commented-out code?

Octokit/Models/Request/CollaboratorRequest.cs
shiftkey: I think my philosophy here is trending towards "keep them separate", so this is fine...

Octokit.Tests.Integration/Clients/RepositoryInvitationsClientTests.cs
maddin2016: This is still WIP. I actually don't know how to implement integration tests for Repository Invitations because i need two contexts. One for sending the invitation and one for accept and decline the invitation. Any ideas?

shiftkey: Cool, I'll update the PR title so I know - please remove it when you're ready to go around again...

shiftkey: > Any ideas?

I'm happy to punt on the receiving end of the request and leave an open issue to track that, because we don't really have the infrastructure right now to support multiple accounts...

maddin2016:
I reduce the integration tests to only receive invitations. Tests for accept and decline not possible yet. Anything more to do?

ryangribble:
> Anything more to do?

Integration tests for `Delete()` and `Edit()` ?

ryangribble:
Also it seems with the existing integration tests you are inviting YOURSELF to a test repository you created.  And this actually seems to work, with the invite showing up in the system.

So given it seems to work to invite yourself, then you probably CAN have tests for `Accept()` and `Decline()`  ?

maddin2016:
Unfortunately not. I always catch an NotFoundexception if try to decline and accept with the same account. I have tested it with different accounts. There i can accept and decline the invitiations.

Date: Sat, 16 Jul 2016 01:04:24 -0700

ryangribble:
> Unfortunately not. I always catch an NotFoundexception if try to decline and accept with the same account. I have tested it with different accounts. There i can accept and decline the invitiations.

It seems to work for me?  Eg this tets:

``` csharp
        [IntegrationTest]
        public async Task CanDeclineInvitation()
        {
            var repoName = Helper.MakeNameWithTimestamp("public-repo");

            using (var context = await github.CreateRepositoryContext(new NewRepository(repoName)))
            {
                var fixture = github.Repository.Collaborator;
                var permission = new CollaboratorRequest(Permission.Push);

                // invite a collaborator
                var response = await fixture.Invite(context.RepositoryOwner, context.RepositoryName, context.RepositoryOwner, permission);

                Assert.Equal(context.RepositoryOwner, response.Invitee.Login);
                Assert.Equal(InvitationPermissionType.Write, response.Permissions);

                // Decline the invitation
                var declined = await github.Repository.Invitation.Decline(response.Id);

                Assert.True(declined);
            }
        }
```

maddin2016:
I still don't get this test working :confused: 
```
[IntegrationTest]
        public async Task CanDeclineInvitation()
        {
            var repoName = Helper.MakeNameWithTimestamp("public-repo");

            using (var context = await github.CreateRepositoryContext(new NewRepository(repoName)))
            {
                var fixture = github.Repository.Collaborator;
                var permission = new CollaboratorRequest(Permission.Push);

                // invite a collaborator
                var response = await fixture.Invite(context.RepositoryOwner, context.RepositoryName, context.RepositoryOwner, permission);

                Assert.Equal(context.RepositoryOwner, response.Invitee.Login);
                Assert.Equal(InvitationPermissionType.Write, response.Permissions);

                // Decline the invitation
                var declined = await github.Repository.Invitation.Decline(response.Id);

                Assert.True(declined);
            }
        }
```
Should i though write these tests although they don't work with my account?

ryangribble:
:+1: write them and I'll check them my side. What error do you get?

maddin2016:
In that try catch block I get an 404 NotFoundException. Later I can send you detailed informations.





On Tue, Jul 19, 2016 at 1:15 PM +0200, "Ryan Gribble" <notifications@github.com> wrote:





:+1: write them and I'll check them my side. What error do you get?

---

ryangribble:
Your test works fine on my side :+1: 

image



So if you can also add an "Accept" method integration test I will double check that too and then I think we are good!

maddin2016:
If i login i can also see that invitation. But i can't accept or decline it with octokit. But if it works with your account then maybe its not an issue. Maybe someone third should test it.

ryangribble:
Yep. that Accept test works for me too.

> Maybe someone third should test it.

Yep probably a good idea

shiftkey:
> > Maybe someone third should test it.
> Yep probably a good idea

:eyes:

ryangribble:
Hey @maddin2016 I was pondering why accepting/declining a self invite might work for me and not you... one thing that came to mind is that I currently have my integration test configuration setup to us username and password (and have not specified an apikey/personal access token).  How have you got yours setup?  If you have an access token defined, try removing it just to see if that makes any difference.

maddin2016:
winner
100 points :tada: That's it.

ryangribble:
Nice one!

OK so I reckon this is good to :ship: but let's get a :+1: from @shiftkey

shiftkey:
Lets do this

ryangribble:
LGTM

maddin2016:
:joy:  That last gif

ryangribble:
Just a quick update on

>On the topic of the email/notifications not occuring, I contacted github support and they confirm they can reproduce it, they are logging an internal issue, so we shouldn't need to worry about that aspect at least

GitHub support notified me recently that they had pushed a fix for this, and I can confirm that emails are now being received when creating invites with octokit.

Issue #874, opened 2015-08-28
Patch on redirected repo fails


Red-Folder:
This is being picked up by the integration test: Octokit.Tests.Integration.RedirectTests.CanCreateIssueOnRedirectedRepository

The test runs until it tries to update the issue.

From stepping through:

* The patch to the original repo is sent and a redirect is received (as expected)
* The new request (against the new repo) is created and submitted.  This then sits on "var response = await base.SendAsync(request, cancellationToken);" until we get an exception that the task was cancelled (I assume timeout).

It looks like the new request doesn't have its content set-up correctly.

I'll have more of a dig when I get a chance, but wanted to post while still fresh in my head.

Red-Folder:
Seems to be something odd ...

If I run from VS, it seems to time out on the send of the new request (following the redirect).

Yet, I run Fiddler, and it all works.  Disable Fiddler Traffic Capture and it times out.  Re-enable and it works.

Red-Folder:
@shiftkey Any thoughts on this?  Do you see the same problem?

shiftkey:
@Red-Folder digging my way out from under a heap of psychic debt at the moment, so i'm not quite sure when I'll get to investigating this. But thanks for reporting it!

alfhenrik:
I can confirm I'm seeing the same thing as @Red-Folder, running the test in VS fails with a timeout. Start Fiddler, then run the test in VS and the test passes. I'll see if I can dig a bit deeper and figure out what is going on.

shiftkey:
Bumping this and opening it up for investigating, because I've seen it while testing #989

Red-Folder:
@shiftkey Sorry - I didn't get much further in my investigation than noted above.

I seem to remember that I felt it was something to do with a connection being held open and it causing blocking (maybe some reference not being released correctly) - but it is likely to be something quite low level if the problem "disappears" when Fiddler is involved.

I have no scientific proof for my theory ;)

I'll have a another go when I get a chance (maybe wireshark or something similar will be less obtrusive than Fiddler and give some indication if the connection is still open)

shiftkey:
@Red-Folder it's all good. I wish I had a better answer for you about what might be happening but I've got many Windmills™ to Tilt At™* at the moment...

\* [citation](https://en.wikipedia.org/wiki/Tilting_at_windmills)

Red-Folder:
[...]

PR #1411, opened 2016-07-05
Fix timeout getting multiple repositories

### LINKED ISSUE #1396: Timeout when getting multiple repositories that have been renamed
The following test fails for me with a `TaskCancelledException` for the [`HttpClient.SendAsync`](https://github.com/octokit/octokit.net/blob/49f246f15b3c78db72b64666cbd1e5ca8f67c029/Octokit/Http/HttpClientAdapter.cs#L46) call, inside the second call to get a repository:

```C#
[IntegrationTest]
public async Task ReturnsRenamedRepository()
{

    var repository = await github.Repository.Get("michael-wolfenden", "Polly");

    Assert.Equal("https://github.com/App-vNext/Polly.git", repository.CloneUrl);
    Assert.False(repository.Private);
    Assert.False(repository.Fork);
    //Assert.Equal(AccountType.User, repository.Owner.Type);

    repository = await github.Repository.Get("fsprojects", "FSharp.Atom");

    Assert.Equal("https://github.com/ionide/ionide-fsharp.git", repository.CloneUrl);
    Assert.False(repository.Private);
    Assert.False(repository.Fork);

    repository = await github.Repository.Get("cabbage89", "Orchard.Weixin");

    Assert.Equal("https://github.com/cabbage89/Orchard.WeChat.git", repository.CloneUrl);
    Assert.False(repository.Private);
    Assert.False(repository.Fork);
}
```

It does not fail if I instead use repos that have not been renamed:

```C#
[IntegrationTest]
public async Task ReturnsRenamedRepository()
{

    var repository = await github.Repository.Get("dotnet", "corefx");

    Assert.Equal("https://github.com/dotnet/corefx.git", repository.CloneUrl);
    Assert.False(repository.Private);
    Assert.False(repository.Fork);
    //Assert.Equal(AccountType.User, repository.Owner.Type);

    repository = await github.Repository.Get("dsplaisted", "PCLStorage");

    Assert.Equal("https://github.com/dsplaisted/PCLStorage.git", repository.CloneUrl);
    Assert.False(repository.Private);
    Assert.False(repository.Fork);

    repository = await github.Repository.Get("paulcbetts", "splat");

    Assert.Equal("https://github.com/paulcbetts/splat.git", repository.CloneUrl);
    Assert.False(repository.Private);
    Assert.False(repository.Fork);
}
```

It also does not fail if I have Fiddler running.

I have no idea what I'm doing

Full exception spew:

```
Test Name:	RepositoriesClientTests+TheGetMethod.ReturnsRenamedRepository
Test FullName:	RepositoriesClientTests+TheGetMethod.ReturnsRenamedRepository
Test Source:	C:\git\octokit.net\Octokit.Tests.Integration\Clients\RepositoriesClientTests.cs : line 513
Test Outcome:	Failed
Test Duration:	0:01:41.313

Result StackTrace:	
at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.ConfiguredTaskAwaitable`1.ConfiguredTaskAwaiter.GetResult()
   at Octokit.Internal.HttpClientAdapter.<Send>d__2.MoveNext() in C:\git\octokit.net\Octokit\Http\HttpClientAdapter.cs:line 46
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.ConfiguredTaskAwaitable`1.ConfiguredTaskAwaiter.GetResult()
   at Octokit.Connection.<RunRequest>d__52.MoveNext() in C:\git\octokit.net\Octokit\Http\Connection.cs:line 537
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.ConfiguredTaskAwaitable`1.ConfiguredTaskAwaiter.GetResult()
   at Octokit.Connection.<Run>d__51`1.MoveNext() in C:\git\octokit.net\Octokit\Http\Connection.cs:line 528
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.ConfiguredTaskAwaitable`1.ConfiguredTaskAwaiter.GetResult()
   at Octokit.ApiConnection.<Get>d__9`1.MoveNext() in C:\git\octokit.net\Octokit\Http\ApiConnection.cs:line 83
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at RepositoriesClientTests.TheGetMethod.<ReturnsRenamedRepository>d__3.MoveNext() in C:\git\octokit.net\Octokit.Tests.Integration\Clients\RepositoriesClientTests.cs:line 523
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
Result Message:	System.Threading.Tasks.TaskCanceledException : A task was canceled.
```

maddin2016:
fixed #1396 

Put logic for redirects outside delegatinghandler in `HttpClientAdapter`.


Octokit/Http/HttpClientAdapter.cs
```diff
@@ -42,10 +44,99 @@ public HttpClientAdapter(Func<HttpMessageHandler> getHandler)
...
+                        newRequest.Content = new StreamContent(stream);
+                    }
+                    else
+                    {
+                        newRequest.Content = null;
+                    }
+                }
+                newRequest.RequestUri = response.Headers.Location;
+
+                if (string.Compare(newRequest.RequestUri.Host, newRequest.RequestUri.Host, StringComparison.OrdinalIgnoreCase) != 0)
```
ryangribble: this is comparing the same variable.  the old code was `newRequest` vs `request`

Octokit/Http/HttpClientAdapter.cs
```diff
@@ -42,10 +44,99 @@ public HttpClientAdapter(Func<HttpMessageHandler> getHandler)
...
+                        || response.StatusCode == HttpStatusCode.Found
+                        || response.StatusCode == HttpStatusCode.SeeOther
+                        || response.StatusCode == HttpStatusCode.TemporaryRedirect
+                        || (int)response.StatusCode == 308)
+            {
+                var newRequest = CopyRequest(request);
+
+                if (response.StatusCode == HttpStatusCode.SeeOther)
+                {
+                    request.Content = null;
```
ryangribble: Is this intentional?  the old code was `newRequest` here rather than `request`

Octokit/Http/HttpClientAdapter.cs
```diff
@@ -42,10 +44,99 @@ public HttpClientAdapter(Func<HttpMessageHandler> getHandler)
...
+                    request.Method = HttpMethod.Get;
```
ryangribble: Is this intentional?  the old code was `newRequest` here rather than `request`

Octokit/Http/HttpClientAdapter.cs
```diff
@@ -42,10 +44,99 @@ public HttpClientAdapter(Func<HttpMessageHandler> getHandler)
...
+                    newRequest.Headers.Authorization = null;
+                }
+
+                response = await GetSuccessfulResponse(newRequest, cancellationToken).ConfigureAwait(false);
+            }
+
+            return response;
+        }
+
+        private async Task<HttpResponseMessage> SendRequest(HttpRequestMessage request, CancellationToken cancellationToken)
```
ryangribble: this method isnt really needed as it just `awaits` the call then returns the result without doing anything with it.
I think you could just move the actual `await _http.SendAsync()` call to line #55 where `SendRequest()` is called currently

Octokit/Http/HttpClientAdapter.cs
```diff
@@ -42,10 +44,99 @@ public HttpClientAdapter(Func<HttpMessageHandler> getHandler)
 
             using (var requestMessage = BuildRequestMessage(request))
             {
-                // Make the request
-                var responseMessage = await _http.SendAsync(requestMessage, HttpCompletionOption.ResponseContentRead, cancellationTokenForRequest).ConfigureAwait(false);
-                return await BuildResponse(responseMessage).ConfigureAwait(false);
+                var response = await GetSuccessfulResponse(requestMessage, cancellationTokenForRequest);
```
ryangribble: missing a `.ConfigureAwait(false)` here i think

ryangribble:
I have just been troubleshooting a problem with the GitHub Enterprise management console stuff I'm implementing at the moment, where I was getting hanging tasks in my integration tests... it turns out it's the same problem you are fixing here!  :grinning: (management console uses redirects which as we've seen seems to trigger this behaviour).

The good news is, with your fix in place (modified a bit) I no longer get the deadlock :grinning:

Your code changes were pretty hard to diff due to moving the functions around in the file.  I actually re-organised it so the `GetSuccessfulResponse` and `CopyRequest` functions were last in the class, which then made the diff line up much nicer and revealed only a few lines of difference.  This let me spot a few places where you are not doing exactly what the old code did which I've highlighted in the review comments.  

I fixed these locally, plus fixed up a few async/await type issues and renamed `GetSuccessfulResponse` to `SendAsync` and everything is working :gem: for me now.  I added what I ended up with to [this gist](https://gist.github.com/ryangribble/571072a97d77baf27999a15be5be308f) if you wanted to see

maddin2016:
I rewrite the failed unit tests

shiftkey:
Just throwing myself on as a reminder to sit down and properly understand this change 🤘

shiftkey:
@maddin2016 have you had a chance to look at those failing tests?

maddin2016:
Not really. Maybe in the next few days. What i have seen is that we have to completely rewright tests for redirects. Yet they are in `RedirectHandlerTests` but i think we should move them to `HttpClientAdapterTests` because redirecting is now handled directly in there. What do you think?

shiftkey:
@maddin2016 I'll look at those failing tests and see if I can help

maddin2016:
Hi @shiftkey. I have sit down and think about these failing redirect tests. Here is an approach in which i have move one test into `HttpClientAdapterTests`. Because redirecting is handled by this class. But to translate tests i have to create a new property `RequestMessage` for class `Response` to get access to the `RequestMessage`. What do you think? So this is really a first approach!!

maddin2016:
I have try to copy these other tests and have find that is not possible to copy them 1 to 1. Because i always have to create new properties for `IResponse` and `IRequest` like `Authorization`. What is the intent for these both classes instead of using `HttpRequestMessage` and `HttpResponse`. Because you anyway translate them into `HttpRequestMessage` like [here](https://github.com/octokit/octokit.net/blob/master/Octokit/Http/HttpClientAdapter.cs#L105)

ryangribble:
Yeah I dont think we want to add these properties to `IResponse` just so we can "test" what request was used.

I had a go at re-jigging the tests with the new redirect handling like you've got in this PR.  I found that inside the redirect method, I had to pre-save off the original request's `content` so it could be later added to the request for the redirected Url, as by that stage a `Object has been disposed` exception was thrown when trying to access the original request's content directly.  With this change, and some modifications to the test/mocks I was able to get all the tests passing, except for the `Assert` that tries to verify that same field (`response.RequestMessage.Content`).  If you wanted to take a look at what I did, the commit is here: https://github.com/TattsGroup/octokit.net/commit/c6f6654b87a1c1f6a66c5f99e1cd1654e10dc288

maddin2016:
Oh man, time for holidays :dizzy_face: i have rollback my commits and add ryans changes.

maddin2016:
Oh man, it's time for holidays :dizzy_face:  I rollback all my comments and add the changes form ryan. Now all should be fine, expect that one failure ryan described

> Assert that tries to verify that same field (response.RequestMessage.Content).

maddin2016:
One thing I would like to comment is that we now have two public send methods in `HttpClientAdapter`. Both are asynchronous but one is called `Send` and the other `SendAsync`. Is this ok?

Octokit/Http/HttpClientAdapter.cs
```diff
@@ -238,18 +245,28 @@ internal class RedirectHandler : DelegatingHandler
...
             }
 
-            return newrequest;
+            return newRequest;
+        }
+    }
+
+    internal class RedirectHandler : DelegatingHandler
+    {
+        public bool Enabled { get; set; }
```
shiftkey: I don't see this value being used anywhere (I suspect it also wasn't used before).

Octokit/Http/HttpClientAdapter.cs
```diff
@@ -238,18 +245,28 @@ internal class RedirectHandler : DelegatingHandler
...
+
+        protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+        {
+            return base.SendAsync(request, cancellationToken);
```
shiftkey: This override also doesn't seem necessary now.

shiftkey:
> One thing I would like to comment is that we now have two public send methods in `HttpClientAdapter`. Both are asynchronous but one is called `Send` and the other `SendAsync`. Is this ok?

This is fine. I'd rather not break this API now, and come back to it when we want to let people provide their own `HttpClient` implementation.

Octokit/Http/HttpClientAdapter.cs
```diff
@@ -168,18 +170,30 @@ protected virtual void Dispose(bool disposing)
...
+                var stream = await request.Content.ReadAsStreamAsync().ConfigureAwait(false);
+                if (stream.CanSeek)
+                {
+                    stream.Position = 0;
+                    savedContent = new StreamContent(stream);
+                }
+                else
+                {
+                    unbuffered = true;
+                    //throw new Exception("Cannot redirect a request with an unbuffered body");
```
shiftkey: :lipstick: let's :fire: these commented lines

shiftkey:
@maddin2016 i like where this is heading, and I think this is close to merging in.

I'd like to push this out as a release on it's own so we can verify it in isolation from other features being rolled out :metal:

ryangribble:
From memory the main issue i found with these changes was that you can no longer "reach in" and access the original request message (response.Request. Content)  as it's been disposed. 

I'm not sure if anything really needs to do this (it was only the integration test that was doing it in our code)

shiftkey:
@ryangribble I'm okay with locking that down if it means resolving this redirect issue - it was supposed to be a black box and I'm not really hearing anyone reimplement our `IHttpClient` without significant pain - and they'll likely be not using this stuff anyway.

If someone comes up with a good use case for needing to poke the request content after we lock this down, then I'm all ears...

maddin2016:
@shiftkey this means these changes from @ryangribble are ok?

shiftkey:
@maddin2016 yep, those comments are fine. Just a couple of tweaks I think to get this ready to merge.

maddin2016:
Something I can do to get ready?

dampir:
@maddin2016 GUESS

maddin2016:
Thanks for that hint @dampir. I thought there where unresolved things because of
>I'm not sure if anything really needs to do this (it was only the integration test that was doing it in our code)

I will asap fix those comments from @shiftkey.

ryangribble:
Hey so I pulled these changes down and was going through things one final time and I decided to unskip the failing test from #874 now that we "should" be handling redirects properly.  And unfortunately it failed :rage:   The problem was something related to the way I had saved off the content stream from the original request, and the new content still had a reference to that old stream.

So anyway, I did some more digging and have come up with a "better" implementation which fully clones the request and looks tidier too.  With this fix all the tests are passing including unskipping the one from #874.

The commit is https://github.com/TattsGroup/octokit.net/commit/168ea4133080575ea41e3b7d84edaf3fe4ef57c6, if you wanted to cherrypick it @maddin2016

maddin2016:
Hi @ryangribble, do you mean test `CanCreateIssueOnRedirectedRepository`? Maybe it is better to clone the request after we get the respone? Only for design
```C#
            // Send initial response
            var response = await _http.SendAsync(request, HttpCompletionOption.ResponseContentRead, cancellationToken).ConfigureAwait(false);

            // Can't redirect without somewhere to redirect to.
            if (response.Headers.Location == null)
            {
                return response;
            }

// Clone the request/content incase we get a redirect
var clonedRequest = await CloneHttpRequestMessageAsync(request);
```

ryangribble:
That's the whole problem, you can't get the content stream ad it's been disposed

maddin2016:
Ok, that makes sense

ryangribble:
@maddin2016 were you happy with that change? Can you confirm all the related integration tests are working for you? Including the one I unskilled?

maddin2016:
All tests including that one works for me. So i think we're done, right?

ryangribble:
Awesome stuff, I'll merge this in then

ryangribble:
Thanks @maddin2016 

LGTM

shiftkey:
Thanks for wrapping this up while I was off sick last week.

maddin2016:
noprob

PR #1414, opened 2016-07-05
Add convention test to ensure all response models have a public parameterless ctor

### LINKED ISSUE #1413: Add convention test to ensure response models have appropriate constructor
The Json deserializer requires a public parameterless ctor in order to create response objects.

This [catches people out](https://github.com/octokit/octokit.net/pull/1410#discussion_r69558986) from time to time, so it would be good to add a convention test that enforces this

ryangribble:
Fixes #1413

`PunchCard` and `RepositoryLanguage` classes were missing the `ctor` and failing this test.  Technically they don't actually need it since they are not being deserialized from json, but populated from basic data types but in the interest of consistency and ease of testing I've added the `ctor` to these 2 classes


shiftkey:
A++ WOULD REVIEW AGAIN

PR #1415, opened 2016-07-05
Enterprise ManagementConsole API - Authorized Ssh Keys

ryangribble:
#### Summary

This PR implements the Authorized Ssh Key management for GitHub Entperprise Management Console
- [Retrieve Authorized Keys](https://developer.github.com/v3/enterprise/management_console/#retrieve-authorized-ssh-keys)
- [Add Authorized Key](https://developer.github.com/v3/enterprise/management_console/#add-a-new-authorized-ssh-key)
- [Remove Authorized Key](https://developer.github.com/v3/enterprise/management_console/#remove-an-authorized-ssh-key)

This continues to complete functionality for #1024 and builds on top of #1269 (which is pending feedback/help from some experts!)
#### Notes

Not as many curve balls as #1269 but a couple of points of note
- The `Delete` method actually returns a response (list of remaining authorized keys) so this necessitated implementing `Delete<T>()` methods on `IConnection` and `IApiConnection`
- The GHE Management Console errors don't seem to conform with the normal GitHub ApiErrors as they just have a simple Json property "error" and then a string.  But this actually revealed an interesting issue with the Json deserializer where it always returns an instantiated object of the type requested, eventhough calling code (like the `ApiError` [deserialization attempt](https://github.com/octokit/octokit.net/blob/d7d03fafa3c306e20706ee321e98c79411672a4f/Octokit/Exceptions/ApiException.cs#L131)) is checking for `null` and then falling back to providing the `string` payload as the error message.  So I made a tweak to the deserializer such that it only instantiates the object when at least one field has been found in the payload response.  This means if it completely fails to deserialize an object to the requested type (ie no fields matched) it now returns `null` (which is what calling code seemed to expect anyway)
  https://github.com/octokit/octokit.net/pull/1415/files#diff-7e26fb1257de97b32ec85c7d5d84dbc9
  Note: this causes a failed test which was written way back in #714.  Need to check with @shiftkey if this is OK to be changed or not.  I have skipped it for the time being
- I needed to fix the `UriExtensions.ApplyParameters()` function to work with non absolute Uris that already had query parameters present.  Without this fix the new+existing parameters are appended to the end of the url string which already included the existing parameters, causing an invalid Uri
  https://github.com/octokit/octokit.net/pull/1415/files#diff-80cf41f33db170a6fcd471657e4f224a
- The integration test to add/delete authorized keys is currently failing, due to the redirect await deadlock currently being discussed in this issue #1396 and this PR #1411
#### Task List
- [x] Regular client implementation
- [x] Regular client unit tests
- [x] Regular client integration tests
- [x] Observable client methods
- [x] Observable client unit tests
- [x] Observable client integration tests
- [x] Tidy up XmlDocs
- [ ] Wait for #1269 to be merged
- [ ] Get ok from @shiftkey regarding comments above


Octokit.Tests/Clients/UsersClientTests.cs
```diff
@@ -92,7 +92,7 @@ public void SendsUpdateToCorrectUrl()
 
         public class SerializationTests
         {
-            [Fact]
+            [Fact(Skip = "Change to serializer means null is now returned")]
```
ryangribble: @shiftkey can you shed any light on the behaviour this test is asserting?  It seems that when you added support for Organizatino/User "type" on Account, you made it nullable and then have this test here to prove that it's null when deserializing a json fragment that is not a user/org/account at all.

Meanwhile in this PR I found that the deserializer instantiating the requested class even when the json had no matching fields seemed to go against what other parts of the code were expecting (linked in PR body) so i changed it to only instantiate class when it finds at least 1 expected field in the json, otherwise it returns a null object.  This test here is the only place (we have test coverage for at least) that seems to be impacted by this.

shiftkey:
Moving this to #2011

PR #1420, opened 2016-07-13
Updated Git Data Tag With Signature Verification

### LINKED ISSUE #1256: Get Signature Verification When Viewing Git Data Tags
API documentation: https://developer.github.com/v3/git/tags/#tag-signature-verification

An enhancement as part of the Commit Signing API is to surface the signature details in the tag.

It looks like these are the new properties:

```json
    "verification": {
      "verified": true,
      "reason": "valid",
      "signature": "-----BEGIN PGP MESSAGE-----\n...\n-----END PGP MESSAGE-----",
      "payload": "tree 6dcb09b5b57875f334f61aebed695e2e4193db5e\n..."
    }
```

These fields should be available on the `GitTag` response object (perhaps in a class called `Verification` or something), and the `TagsClient.Get` action should be updated to use the new `Accepts` header:

> application/vnd.github.cryptographer-preview+sha

I'd also like to see some new integration tests added to verify this behaviour is working as expected.

Sarmad93:
Fixes #1256 
Updated `GitTag` Response class.
Updated `TagsClient` Class with new Accept Header.
Added Integration Test.

### DIFF
--- a/Octokit.Tests.Integration/Clients/TagsClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/TagsClientTests.cs
@@ -107,6 +107,26 @@ namespace Octokit.Tests.Integration.Clients
                 Assert.Equal(gitTag.Message, "Hello");
                 Assert.Equal(gitTag.Object.Sha, sha);
             }
+
+            [IntegrationTest]
+            public async Task DeserializeTagSignatureVerification()
+            {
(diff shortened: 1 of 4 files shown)


shiftkey:
@Sarmad93 looks good to me!

Sarmad93:
@shiftkey  
awesome

PR #1432, opened 2016-07-23
v0.21 - Which Puppy

shiftkey:
- [x] release notes
- [x] version bump
- [x] update linqpad scripts to fix regressions
- [x] :thumbsup: from @Haacked or @ryangribble 
- [x] tag and publish to NuGet
- [x] :shipit:

### DIFF
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -35,7 +35,7 @@ exist in other parts of the API surface:
  - `I(Observable)GitHubClient.Repository.RepoCollaborators`
  - `I(Observable)GitHubClient.Repository.Commits`
 
-This method has also been removed, but is no longer supported through the API: 
+This method is no longer supported through the API and has been removed from Octokit.net.: 
 
  - `I(Observable)GitHubClient.Authorization.RevokeAllApplicationAuthentications()`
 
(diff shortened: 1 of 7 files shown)


shiftkey:
@ryangribble @haacked @dampir any feedback on the release notes before I :ship: this tomorrow?

ReleaseNotes.md
```diff
@@ -1,3 +1,44 @@
...
+ - `I(Observable)GitHubClient.Release`
+ - `I(Observable)GitHubClient.Notification`
+ - `I(Observable)GitHubClient.GitDatabase`
+ - `I(Observable)GitHubClient.SshKey`
+ - `I(Observable)GitHubClient.Repository.RepositoryComments`
+ - `I(Observable)GitHubClient.Repository.CommitStatus`
+ - `I(Observable)GitHubClient.Repository.RepoCollaborators`
+ - `I(Observable)GitHubClient.Repository.Commits`
+
+This method has also been removed, but is no longer supported through the API:
```
Haacked: The `, but` here feels like the second clause contrasts with the first, but it really goes hand in hand with it. How about,

"This method is no longer supported through the API and has been removed from Octokit.net." or something like that?

Haacked:
One minor nit, otherwise :+1: to the release notes.

shiftkey:
This has shipped:

 - https://www.nuget.org/packages/Octokit/0.21.0
 - https://www.nuget.org/packages/Octokit.Reactive/0.21.0

Haacked:
:party:

PR #1435, opened 2016-08-03
Implement Issue Timeline preview API

### LINKED ISSUE #1436: Issue timeline preview API
The issue timeline API is in [preview](https://developer.github.com/changes/2016-05-23-timeline-preview-api/) so we should add that to Octokit.

alfhenrik:
Adding the Issue Timeline [API](https://developer.github.com/v3/issues/timeline/) client as it is now in preview (and we have a need for it for an internal project).

- [x] Timeline API client
- [x] Observable Timeline API client
- [x] XML Doc comments
- [x] Unit tests
- [x] Observable unit tests
- [x] Integration tests
- [x] Observable integration tests
- [x] Repository Id overloads
- [x] Repository Id overload tests
- [x] Paging support
- [x] Paging support tests

Fixes #1436

### DIFF
--- a/Octokit.Reactive/Clients/IObservableIssueTimelineClient.cs
+++ b/Octokit.Reactive/Clients/IObservableIssueTimelineClient.cs
@@ -6,7 +6,7 @@ namespace Octokit.Reactive
     /// A client for GitHub's Issue Timeline API.
     /// </summary>
     /// <remarks>
-    /// See the <a href="http://developer.github.com/v3/issues/timeline/">Issue Timeline API documentation</a> for more information.
+    /// See the <a href="https://developer.github.com/v3/issues/timeline/">Issue Timeline API documentation</a> for more information.
     /// </remarks>
     public interface IObservableIssueTimelineClient
     {
--- a/Octokit.Reactive/Clients/IObservableIssueTimelineClient.cs
+++ b/Octokit.Reactive/Clients/IObservableIssueTimelineClient.cs
@@ -2,8 +2,23 @@
 
 namespace Octokit.Reactive
 {
+    /// <summary>
+    /// A client for GitHub's Issue Timeline API.
+    /// </summary>
+    /// <remarks>
+    /// See the <a href="http://developer.github.com/v3/issues/timeline/">Issue Timeline API documentation</a> for more information.
+    /// </remarks>
     public interface IObservableIssueTimelineClient
     {
+        /// <summary>
+        /// Gets all the various events that have occurred around an issue or pull request.
+        /// </summary>
+        /// <remarks>
+        /// https://developer.github.com/v3/issues/timeline/#list-events-for-an-issue
+        /// </remarks>
+        /// <param name="owner">The owner of the repository</param>
+        /// <param name="repo">The name of the repository</param>
+        /// <param name="number">The issue number</param>
         IObservable<TimelineEventInfo> GetAllForIssue(string owner, string repo, int number);
     }
 }
--- a/Octokit.Reactive/Clients/IObservableIssueTimelineClient.cs
+++ b/Octokit.Reactive/Clients/IObservableIssueTimelineClient.cs
@@ -1,0 +1,9 @@
+﻿using System;
+
+namespace Octokit.Reactive
+{
+    public interface IObservableIssueTimelineClient
+    {
+        IObservable<TimelineEventInfo> GetAllForIssue(string owner, string repo, int number);
+    }
+}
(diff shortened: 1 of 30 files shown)


alfhenrik:
All tests are now passing 🎉 so this is ready for someone to cast their 👀 over 😁

ryangribble:
This is all looking rather 💎 to me! Ive run the integration tests too 👍 

One thing I noticed was a quick count of the "events" listed in the timeline API doc (21) is a couple more than values declared in the `EventInfoState` enum (19), so is there possibly another couple of values that need to be added to this enum in addition to the cross referenced one you added?

Octokit.Reactive/Clients/IObservableIssueTimelineClient.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+
+namespace Octokit.Reactive
+{
+    /// <summary>
+    /// A client for GitHub's Issue Timeline API.
+    /// </summary>
+    /// <remarks>
+    /// See the <a href="http://developer.github.com/v3/issues/timeline/">Issue Timeline API documentation</a> for more information.
```
ryangribble: Minor nitpick but the doc comments added in this PR have a mix of http and https on the URLs... 😉

dampir:
@alfhenrik since version 0.21.1 every "repo/name" method has overload that use repository id instead of repo/name pair. Could you add it in your PR?

Octokit.Reactive/Clients/IObservableIssueTimelineClient.cs
alfhenrik: That's very odd as I'm pretty sure I copied those from the API documentation...I'll fix it and check the docs if there might something pointing to just `http`

alfhenrik:
Thanks @dampir, I'll add in the necessary overloads and tests

ryangribble:
> @ryangribble not sure, but I think that now every new clients have to have repostoryId overload for each owner/name method, aren't they?

Good pickup, I actualy realised this myself overnight (the things you think about while going to sleep, lol!)

Yes we should be adding `repositoryId` overload of any method that takes `owner/name` (as long as it actually works of course!).

I also realised, we should be adding paging support to any `GetAllxxx()` type methods (ie an overload taking `ApiOptions` class, with the default method passing `ApiOptions.None` through to it).  Again, assuming paging is actually supported on that end point of course.

alfhenrik:
> I also realised, we should be adding paging support to any GetAllxxx() type methods (ie an overload taking ApiOptions class, with the default method passing ApiOptions.None through to it). Again, assuming paging is actually supported on that end point of course.

Looking at the API docs, it doesn't look like the endpoint supports paging, but I'll do some Fiddlering to see if it might actually support it, just being undocumented.

alfhenrik:
OK, this should now be good to go

ryangribble:
Nice work!  Everything looking good, ill just pull down your branch and run through the integration tests  :+1:

Octokit/Helpers/ApiUrls.cs
```diff
@@ -341,6 +341,16 @@ public static Uri IssueReactions(int repositoryId, int number)
             return "repositories/{0}/issues/{1}/reactions".FormatUri(repositoryId, number);
         }
 
+        public static Uri IssueTimeline(string owner, string repo, int number)
```
ryangribble: XmlDoc comment on this function

Octokit/Helpers/ApiUrls.cs
```diff
@@ -341,6 +341,16 @@ public static Uri IssueReactions(int repositoryId, int number)
...
+        {
+            return "repos/{0}/{1}/issues/{2}/timeline".FormatUri(owner, repo, number);
+        }
+
+        public static Uri IssueTimeline(int repositoryId, int number)
```
ryangribble: And here

ryangribble:
I just noticed missing XmlDoc comments on the 2 new functions added to `ApiUrls.cs` helper class which we probably should add, since every other function in that class has doc tags currently...

Once that's done I'm good to merge :grinning:

Octokit/Helpers/ApiUrls.cs
alfhenrik: Done

Octokit/Helpers/ApiUrls.cs
alfhenrik: And done

alfhenrik:
**Pretty** sure that TravisCI failure isn't because of me...looks like that random build failure we've seen before...

ryangribble:
LGTM

PR #1445, opened 2016-08-15
Change repository id from int to long

shana:
Repository ids are global across github. On a completely noob assessment of the growth of the number of repositories over the last 6 months, if growth remains constant, we'll reach max int and overflow around September 2017. Since this is an identifier field, it should really be at least an `uint` (they can't be negative, can they?), but since we're already using a signed type, might as well go `long`.

Thoughts, concerns, emojis?

### DIFF
--- a/Octokit.Reactive/Clients/IObservableAssigneesClient.cs
+++ b/Octokit.Reactive/Clients/IObservableAssigneesClient.cs
@@ -21,7 +21,7 @@ namespace Octokit.Reactive
         /// Gets all the available assignees (owner + collaborators) to which issues may be assigned.
         /// </summary>
         /// <param name="repositoryId">The Id of the repository</param>
-        IObservable<User> GetAllForRepository(int repositoryId);
+        IObservable<User> GetAllForRepository(long repositoryId);
 
         /// <summary>
         /// Gets all the available assignees (owner + collaborators) to which issues may be assigned.
@@ -36,7 +36,7 @@ namespace Octokit.Reactive
         /// </summary>
         /// <param name="repositoryId">The Id of the repository</param>
         /// <param name="options">The options to change API's behaviour.</param>
-        IObservable<User> GetAllForRepository(int repositoryId, ApiOptions options);
+        IObservable<User> GetAllForRepository(long repositoryId, ApiOptions options);
 
         /// <summary>
         /// Checks to see if a user is an assignee for a repository.
@@ -51,6 +51,6 @@ namespace Octokit.Reactive
         /// </summary>
         /// <param name="repositoryId">The Id of the repository</param>
         /// <param name="assignee">Username of the prospective assignee</param>
-        IObservable<bool> CheckAssignee(int repositoryId, string assignee);
+        IObservable<bool> CheckAssignee(long repositoryId, string assignee);
     }
 }
--- a/Octokit.Reactive/Clients/IObservableBlobsClient.cs
+++ b/Octokit.Reactive/Clients/IObservableBlobsClient.cs
@@ -31,7 +31,7 @@ namespace Octokit.Reactive
         /// <param name="repositoryId">The Id of the repository</param>
         /// <param name="reference">The SHA of the blob</param>
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1716:IdentifiersShouldNotMatchKeywords", MessageId = "Get")]
-        IObservable<Blob> Get(int repositoryId, string reference);
+        IObservable<Blob> Get(long repositoryId, string reference);
 
         /// <summary>
         /// Creates a new Blob
@@ -52,6 +52,6 @@ namespace Octokit.Reactive
         /// </remarks>
         /// <param name="repositoryId">The Id of the repository</param>
         /// <param name="newBlob">The new Blob</param>
-        IObservable<BlobReference> Create(int repositoryId, NewBlob newBlob);
+        IObservable<BlobReference> Create(long repositoryId, NewBlob newBlob);
     }
 }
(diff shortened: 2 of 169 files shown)


ryangribble:
Sounds reasonable 

It should be changed in [all the places](https://github.com/octokit/octokit.net/search?utf8=✓&q=RepositoryId) we use RepositoryId as a parameter too though

shana:
Ah, indeed! I'll hunt those down and update.

shiftkey:
I'll pull this in to the next release (not the one currently being prepared).

shiftkey:
> Since this is an identifier field, it should really be at least an `uint` (they can't be negative, can they?)

And no, [`uint` cannot be negative](https://msdn.microsoft.com/en-AU/library/x0sksh43.aspx)

ryangribble:
changing to `uint` would be a breaking change for every single place changed, whereas `long` will implicitly allow an `int` to be passed so isnt a breaking change in the majority of the places touched in this PR.

The only place I can see currently that would be a breaking change is the change to `long` for the `Repository.Id` on the response class and even then only in some cases depending on the consuming code.  We'll need to include something in the release notes on that.

shiftkey:
@ryangribble  to be clear - I'm :thumbsup: on using `long` here due to the implicit behaviour of `int` saving a lot of breaking change. Just wanted to make the behaviour of `uint` clear here 😀

shana:
Right, uint can't be negative, but I was wondering more about the identifier field itself. The API doesn't return `-1` at any time in identifier fields, does it? Nor does octokit do that to signify some invalid state condition?

I'm just curious, we can't change the type sign now since code assumes it for conversion purposes, just wondering of anything is taking advantage of the sign for something.

shiftkey:
> The API doesn't return `-1` at any time in identifier fields, does it?

Not that I've encountered.

> Nor does octokit do that to signify some invalid state condition?

No

shana:
✨ 

BTW, thank you for being so consistent about naming things, it was way easier than I thought to fix all the tests! ❤️

dampir:
@shana I knew that consistent naming in repository Id implementation could be useful someday, but I don't even supposed that it would be so soon :smile: 
Best code practices in action :+1:

ryangribble:
the release is done now so should be good to rebase this and merge it in.  There might be a couple of new methods added in recent PRs that would have missed the initial `int => long` conversion too....

shana:
PR updated against latest master.

I was wondering, it's not like I want to have more work, but... this is technically an ABI change. Should I be obsoleting the existing methods that take an int and adding overloads that take a long, and redirecting the int implementations to long as appropriate?

shiftkey:
> I was wondering, it's not like I want to have more work, but... this is technically an ABI change.

This is true, and after the whole discussion in #1464 I'm pondering how often I want to break the API as broadly as this (even if it is still pre 1.0 in semver land, this stuff sucks).

Given numeric ids are not going to be a part of the [GraphQL API](http://githubengineering.com/the-github-graphql-api/) and I think we'll be supporting the v3 API for a while, that looming September 2017 deadline is my only concern about getting this in...

shana:
> This is true, and after the whole discussion in #1464 I'm pondering how often I want to break the API as broadly as this (even if it is still pre 1.0 in semver land, this stuff sucks).

Yes, that one got me thinking about the breakage too. Also all the latest obsolete methods that are now gone and caused a bunch of conflicts on my rebase :-P

> Given numeric ids are not going to be a part of the GraphQL API and I think we'll be supporting the v3 API for a while, that looming September 2017 deadline is my only concern about getting this in...

Loooom. Well, I've only got myself to blame on this one, really. I'll add overloads for all the APIs and redirect as needed to avoid breakage. How many versions do obsolete methods usually last for?

shiftkey:
>  How many versions do obsolete methods usually last for?

We usually leave them around for at least one or two releases.

ryangribble:
yep 2 releases is the current agreed "standard" for obsoleting things.

Just on this - im not getting why the change from `int` to `long` is such a "breaking" change.  Won't the majority of cases happily continue to work without any changes even once they take this update?  I think given pre 1.0 we could be flexible on this.  The `Async` method suffixes was a totally different proposition requiring nearly every line of code interacting with octokit to have to be changed.  Conversely with this repositoryId change, there would be nowhere near as many code changes, if any at all?

M-Zuber:
I agree with @ryangribble. 
To add to that, these overloads where added in [0.21](https://github.com/octokit/octokit.net/releases/tag/v0.21.0) which was only the release before last.

shana:
It's up to you peeps. As far as the runtime is concerned, int->long is a breaking change at the ABI level, i.e., it'll throw methodnotfoundexception if you get the long version and you compiled against the int version. Since you don't sign the assembly, the runtime is free to provide whatever is already loaded in the domain, which might or might not be the actual specific version that is ABI compatible with what you built against.

But, since this is a < 0.x library and you don't sign it, there's no stringent compatibility requirement and it's going to blow up in people's faces anyway in shared appdomain scenarios, so ¯\_(ツ)_/¯

ryangribble:
Given its delivered via nuget there wouldn't be any supported way to update to the new version WITHOUT also compiling against it though right? 

Obviously @shiftkey can have the final say but based on these points I think we should not worry about it 
- not a breaking change for almost all use cases 
- amount of work/churn to obsolete int methods is large 
- int methods were only recently added anyway, further reducing the target audience that could be affected

shana:
> Given its delivered via nuget there wouldn't be any supported way to update to the new version WITHOUT also compiling against it though right?

Weeeeell actually (:trollface:), bla bla bla signing bla versioning bla runtime loader bla plugin model bla horrible explosion. There's a whole signing/versioning/plugin model discussion on that one and it's not relevant for this case 😄 

When updating the nuget package and compiling against the new version, things should not break at all at compile time because implicit casts for wider numeric types are a thing. There might be silly people out there invoking octokit via reflection, that'll break them (but they probably deserve it? ¯\_(ツ)_/¯

shiftkey:
Excellent points @ryangribble. Let's do this.

Issue #1446, opened 2016-08-16
Organization can fail deserialization when user is not an org owner.


grokys:
According to the [API docs](https://developer.github.com/v3/orgs/#get-an-organization):

> Note: To receive values for private_gists, disk_usage, collaborators, and billing_email in the Organization response, the authenticated user must be an organization owner and have authorized the admin:org scope.

What this means in practice is that if the user is a *member* of the org but not an owner. the resulting json has `null` for the values of these fields. However the `PrivateGists`, `DiskUsage` and `Collaborators` fields are not null in the `Organization` model meaning that the deserialization fails.

PR to follow.

khellang:
BTW, this was recently announced on the [dev blog](https://developer.github.com/changes/2016-08-09-breaking-change-removed-sensitive-fields-from-organization-api-responses-for-owner/) as well.

ryangribble:
> BTW, this was recently announced as a breaking change on the dev blog as well.

I saw this thismorning but incorrectly assumed we'd be fine since they were just making some fields not returned in some cases, which happens quite commonly.  Obviously being non nullable int's wasn't too happy though :grin:

PR #1448, opened 2016-08-16
Fix issue #1446: Organization can fail deserialization when user is not an org owner.:

grokys:
Make Collaborators, DiskUsage and PrivateGists properties nullable as they may be returned as null when fetching an organization of which the user is a member by not an owner.


ryangribble:
Thanks for the quick fix!

LGTM

PR #1454, opened 2016-08-31
Adds explicit guidance around Personal Accees Tokens

dalpert-korewireless:
- makes relationship between Personal Access tokens and OAuth tokens more explicit

/cc @Haacked

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -16,7 +16,7 @@ The API will reject you if you don't provide a `User-Agent` header (more details
 
 If you want to access private repositories or perform actions on behalf of a user, you need to pass credentials to the client.
 
-There are two options supported by the API - basic and OAuth authentication.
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -16,24 +16,31 @@ The API will reject you if you don't provide a `User-Agent` header (more details
 
 If you want to access private repositories or perform actions on behalf of a user, you need to pass credentials to the client.
 
-There are two options supported by the API - basic and OAuth authentication.
+There are three options supported by the API - Basic, Personal Access Token, and OAuth authentication.
```
Haacked: A personal access token _is_ an OAuth token.

How about we just say

> There are two options supported by the API - basic and OAuth (ex. personal access token) authentication.

Then there's no need to add anything else.

ryangribble:
Just having a new year's clean of open PR's - I'd like to either get the wording fixed up as @Haacked requested or drop this PR

PR #1457, opened 2016-09-02
[WIP] Add repository traffic preview

### LINKED ISSUE #1447: Implement Repository Traffic preview API
As per [this announcement](https://developer.github.com/changes/2016-08-15-traffic-api-preview/) there is a new "Repository Traffic" API

I'll flesh this issue out later, but just logging it now so we don't miss it.

maddin2016:
fix #1447 

Implement client for the new repository traffic api.


maddin2016:
Maybe someone has a better naming for `TrafficeDayOrWeek` 😸

maddin2016:
@ryangribble, what about integration tests for this feature. The only thing we can test is that the result is not null :expressionless:. Should i do that?

ryangribble:
I guess integration tests still prove everything is wired up correctly, even if there aren't any meaningful stats...  

I take it we can't point the calls at an existing repository due to permissions?

maddin2016:
Yeap, you need to have push access.

maddin2016:
image

maddin2016:
Hi @ryangribble, is this ready for 🚢?

ryangribble:
Hi @maddin2016 

Do you think you could add some integration tests then flag them as `Skipped due to requiring admin permissions`.  This way I can at least run these locally (after changing the repostory to something I do have admin permissions on and unskipping the tests) to ensure the functionality is working as intended?

ryangribble:
Hey @maddin2016 

I've had a look over the tests and found a couple of issues.  

- RepositoryId was recently changed from `int` to `long` in #1445 so the `repositoryId` parameters in this new API need to be `long` now too

- The `timestamp` field in the Api responses is in unix utc time, and doesn't deserialize into a `DateTimeOffset` directly, causing exceptions on the View and Clone methods when the repository actually had some data to return for these calls

- I wasnt a fan of naming the item level classes `View` and `Clone` since they are a bit too generic IMO

- Integration tests were added for owner/name methods but not for repositoryId methods

As I was working through things I ended up making local fixes anyhow, so I've pushed a commit that addresses these already if you want to cherry pick it in: https://github.com/TattsGroup/octokit.net/commit/b145bb30d1a8848a160bd13bc0f9818f22be9b7f


If you are interested in how we handle the timestamp thing, basically in other areas of the codebase we have a pattern already, where we add a `long TimestampAsUtcEpochSeconds` field to take the deserialized value (using the `Parameter` override to map it to `timestamp`) and then add a `DateTimeOffset Timestamp` getter property that uses a helper function to convert the unix time into `DateTimeOffset`.  it looks something like this (and you'll see these fixes in my commit):

```
        [Parameter(Key = "ignoreThisField")]
        public DateTimeOffset Timestamp
        {
            get { return TimestampAsUtcEpochSeconds.FromUnixTime(); }
        }

        [Parameter(Key = "timestamp")]
        public long TimestampAsUtcEpochSeconds { get; protected set; }
```


Let me know what you think of my changes

maddin2016:
Hi @ryangribble, thanks for your review. I have read on the format of `timestamp`. Is there any explanation why it's this format? And yes, `View` is realy a bit to common 😁. The renaiming of `GetAllxxx()` makes sense. I already have this in mind but i don't know wheter it woul be correct. So many thanks for your fixes 👍

maddin2016:
And the renaming of the repsitory id's i have forgotten, because i already did this in another PR :sweat_smile:

ryangribble:
Thanks for another great contribution @maddin2016 

LGTM

maddin2016:
Many thanks @ryangribble 
highfive

Issue #1419, opened 2016-07-12
[Proposal] The Road to .NET Core Support


shiftkey:
I came back to look at what would be involved with getting Octokit.net running on .NET Core recently now that they had the 1.0 launch, and this is the initial brain dump for what I'm anticipating.
### Pros
- fewer projects to manage
- less project hacks (I'm still not sure how we're getting NSubstitute to work on `netcore50`)
- `.\build FixProjects` is now redundant
- we can now enable `Octokit.Reactive` on all the platforms that `System.Reactive` works on
- can simplify build scripts (`dotnet pack` and `dotnet test` can be used directly)
### Cons
- requires VS2015 Update 3 + additional tooling to contribute (you'll be prompted on first open)
- `project.json` is an interim step, unclear when we'll have to come back to it and port back to MSBuild
### Things To Investigate
- ~~we're using NSubstitute, which depends on [Castle.Core](https://www.nuget.org/packages/Castle.Core/4.0.0-alpha001) which has a pre-release package with .NET Core support - this might impact us running tests on certain platforms~~ we should try out the [pre-release NSubstitute package](https://twitter.com/nsubstitute/status/762118809784229888)
- PCL `Profile259` maps to `netstandard1.0` as per the docs [https://github.com/dotnet/corefx/blob/master/Documentation/architecture/net-platform-standard.md] but `System.Net.Http` has a minimum version of `netstandard1.1` - this might need to be obsoleted during this transition
- Xamarin support is also there (it's `netstandard1.6`) but I'm not familiar enough with how this should transition...
- need to look at what this means for all our Mono build scripts (may simplify things as `dotnet` runs on those platforms too)

alfhenrik:
Looks like `NSubstitute

ryangribble:
Whether in conjunction with this or not, i'd also like to look at moving the builds over to CAKE rather than FAKE...

shiftkey:
@ryangribble here you go #1440

mderriey:
Is anything blocking this issue at the moment? I'd like to help if that's OK.

ryangribble:
No, nothing blocking it... it looks like all the dependencies are supporting dotnet core now

mderriey:
Great. I'll start with something to kick off a discussion around it.

In the meantime, if you think of anything I should be aware of, please let me know. Thanks.

hahmed:
> requires VS2015 Update 3 + additional tooling to contribute (you'll be prompted on first open)

[...]

PR #1462, opened 2016-09-09
Get a .NET Core project to build in AppVeyor and Travis

mderriey:
First step of #1419.

It looks like AppVeyor has the dotnet CLI installed in the image so no need to download it.  
Also, that's really as simple as it gets. Maybe you wanted something more ... realistic.

Made some comments in the F# code where I'm really not sure if that's The Way To Do Things.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,18 +2,13 @@ language: csharp
 
 matrix:
   include:
-    - os: linux
-      dist: trusty
-      sudo: required
-      mono: 4.2.3
-      dotnet: 1.0.0-preview2-003121
     - os: osx
       osx_image: xcode8
       mono: 4.2.3
       dotnet: 1.0.0-preview2-003121
 
-before_install:  
-  - if test "$TRAVIS_OS_NAME" == "osx"; then ln -s /usr/local/Cellar/openssl/1.0.2h/lib/libcrypto.1.0.0.dylib /usr/local/lib/; ln -s /usr/local/Cellar/openssl/1.0.2h/lib/libssl.1.0.0.dylib /usr/local/lib/; fi
+before_install:
+  - ln -s /usr/local/opt/openssl/lib/libcrypto.1.0.0.dylib /usr/local/lib/; ln -s /usr/local/opt/openssl/lib/libssl.1.0.0.dylib /usr/local/lib/;
 
 install:
   - nuget restore Octokit-Mono.sln
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,9 @@ matrix:
       mono: 4.2.3
       dotnet: 1.0.0-preview2-003121
 
+before_install:  
+  - if test "$TRAVIS_OS_NAME" == "osx"; then ln -s /usr/local/Cellar/openssl/lib/libcrypto.1.0.0.dylib /usr/local/lib/; ln -s /usr/local/Cellar/openssl/lib/libssl.1.0.0.dylib /usr/local/lib/; fi
+
 install:
   - nuget restore Octokit-Mono.sln
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,13 +2,18 @@ language: csharp
 
 matrix:
   include:
+    - os: linux
+      dist: trusty
+      sudo: required
+      mono: 4.2.3
+      dotnet: 1.0.0-preview2-003121
     - os: osx
       osx_image: xcode8
       mono: 4.2.3
       dotnet: 1.0.0-preview2-003121
 
 before_install:
-  - ln -s /usr/local/opt/openssl/lib/libcrypto.1.0.0.dylib /usr/local/lib/; ln -s /usr/local/opt/openssl/lib/libssl.1.0.0.dylib /usr/local/lib/;
+  - if test "$TRAVIS_OS_NAME" == "osx"; then ln -s /usr/local/opt/openssl/lib/libcrypto.1.0.0.dylib /usr/local/lib/; ln -s /usr/local/opt/openssl/lib/libssl.1.0.0.dylib /usr/local/lib/; fi
 
 install:
   - nuget restore Octokit-Mono.sln
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ matrix:
       dotnet: 1.0.0-preview2-003121
 
 before_install:  
-  - if test "$TRAVIS_OS_NAME" == "osx"; then ln -s /usr/local/Cellar/openssl/lib/libcrypto.1.0.0.dylib /usr/local/lib/; ln -s /usr/local/Cellar/openssl/lib/libssl.1.0.0.dylib /usr/local/lib/; fi
+  - if test "$TRAVIS_OS_NAME" == "osx"; then ln -s /usr/local/Cellar/openssl/1.0.2h/lib/libcrypto.1.0.0.dylib /usr/local/lib/; ln -s /usr/local/Cellar/openssl/1.0.2h/lib/libssl.1.0.0.dylib /usr/local/lib/; fi
 
 install:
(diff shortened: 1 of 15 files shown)


build.fsx
```diff
@@ -107,6 +107,15 @@ Target "BuildApp" (fun _ ->
         |> DoNothing
 )
 
+Target "BuildDotNetCore" (fun _ ->
+    Fake.DotNetCli.Restore (fun p ->
+      {p with
+          WorkingDir = "./Octokit.Next" })
+
+    !! "./Octokit.Next/project.json"
+    |> Fake.DotNetCli.Build (fun p -> p)
```
mderriey: maybe there's an easier way than `(fun p -> p)` to return the input parameter as is

build.fsx
```diff
@@ -107,6 +107,15 @@ Target "BuildApp" (fun _ ->
         |> DoNothing
 )
 
+Target "BuildDotNetCore" (fun _ ->
+    Fake.DotNetCli.Restore (fun p ->
+      {p with
+          WorkingDir = "./Octokit.Next" })
+
+    !! "./Octokit.Next/project.json"
```
mderriey: since `Fake.DotNetCli.Build` accepts a `seq<string>`, I went this way, taking example on other tasks. I don't know if it's right to do it this way since there's no globbing involved

mderriey:
I'm struggling a bit to gauge how simple the .NET Core has to be.

Is the intent realy only to have a .NET Core, no matter how small it is, to build on AppVeyor? Or do we want to try tackling the porting to the existing projects to .NET Core?

Thanks!
/cc @shiftkey

shiftkey:
@mderriey the reason why I suggest getting a trivial .NET Core thing building is that I fear there's lots of integration hassles waiting for us getting this running on both Travis and Appveyor.

mderriey:
For the sake of simplicity, I went with a separate VS solution containing a very simple class library and a unit tests project.

It took a few trial and error tries to get it working on Travis CI.

mderriey:
Is that a good enough baseline to start with the port? How should we proceed?

.travis.yml
```diff
@@ -1,14 +1,17 @@
...
+      sudo: required
+      mono: latest
+      dotnet: 1.0.0-preview2-003121
+    - os: osx
+      osx_image: xcode7.2
+      mono: latest
+      dotnet: 1.0.0-preview2-003121
+
+script:
+  - dotnet --info
```
shiftkey: Can we put the existing build steps in here, so they both run?

appveyor.yml
```diff
@@ -1,10 +1,7 @@
 init:
   - git config --global core.autocrlf input
 build_script:
-  - cmd: build.cmd BuildApp
-  - cmd: build.cmd UnitTests
-  - cmd: build.cmd ConventionTests
-  - cmd: build.cmd CreatePackages
+  - cmd: build.cmd UnitTestsDotNetCore
```
shiftkey: Same thing here, let's keep the existing tasks around

shiftkey:
@mderriey I like where this is heading. If it's too hard to run both the original build and this new build, then let's verify we can test the .NET Core project from the command line at least

mderriey:
Sure, I'll bring the existing build targets back.

> If it's too hard to run both the original build and this new build (for example, if we hit timeouts with CI), then let's verify we can test the .NET Core project from the command line at least

I didn't get this part, sorry. The new build does execute tests, but I'm sure you saw that, so I must miss something.

Also, a couple of random questions:

 - for the time being, are we happy to duplicate source files in the new projects? Having both .csproj and project.json seems to be causing issues with MSBuild wanting more target frameworks in the project.json
 - if it turns out `netstandard1.1` is all we need, are we happy to have it as the only target?

shiftkey:
> The new build does execute tests, but I'm sure you saw that, so I must miss something.

Yes, I wasn't clear. Am I able to run the same command locally? Or are there other dependencies I need.

> for the time being, are we happy to duplicate source files in the new projects? Having both .csproj and project.json seems to be causing issues with MSBuild wanting more target frameworks in the project.json

I'd rather not do this, and the MSBuild issues are similar to what I was seeing a while ago. 

> if it turns out `netstandard1.1` is all we need, are we happy to have it as the only target?

I'm fine with targeting this initially as a new platform, as adding in extra platforms is rather straightforward. My main focus initially is ensuring we support the current platforms in `project.json` as well...

mderriey:
> Yes, I wasn't clear. Am I able to run the same command locally? Or are there other dependencies I need.

As long as you have the dotnet CLI, then yeah, you can execute the build target to run .NET Core tests.

> I'd rather not do this, and the MSBuild issues are similar to what I was seeing a while ago.

Understood. Will investigate on this a bit more by having a look at the branch you mentioned before.

> I'm fine with targeting this initially as a new platform, as adding in extra platforms is rather straightforward. My main focus initially is ensuring we support the current platforms in project.json as well...

That's consistent with what you said just before.

Finally, the build timed out on Travis with both the existing and the new build. Can we kick off a new one without pushing changes to that branch just in case?

shiftkey:
> As long as you have the dotnet CLI, then yeah, you can execute the build target to run .NET Core tests.

Do you think it's worth installing the dotnet CLI locally, like we do for [Rx.NET](https://github.com/Reactive-Extensions/Rx.NET/blob/master/build.ps1#L1-L21)?

mderriey:
Hmm.

I'd say no. Let's say you're willing to work on the port to .NET Core, you'll need to install .NET Core runtime and tooling anyway. Another argument for not installing it as part of the build is that both AppVeyor and Travis support it, which saves us from installing it ourselves on the build agents.

Does that answer your question?

shiftkey:
@mderriey all good points 👍

mderriey:
Cool.

I've just noticed another build started on Travis for [this commit (`use mono 4.2.3 as latest successful builds show this version was used`)](https://github.com/octokit/octokit.net/pull/1462/commits/b3c1c80585b297875aa5903aa157b9169e866349).

Did you start this explicitly or does Travis have a retry mechanism when a build times out?

shiftkey:
@mderriey nope, this is me restarting the build :rage1:

shiftkey:
This seems to be an upstream problem 😞 

```
Unhandled Exception: System.TypeInitializationException: The type initializer for 'Crypto' threw an exception. ---> System.TypeInitializationException: The type initializer for 'CryptoInitializer' threw an exception. ---> System.DllNotFoundException: Unable to load DLL 'System.Security.Cryptography.Native': The specified module could not be found.
```

mderriey:
Well that's weird, the last build passed but this one didn't, the only change being to bring back linux again and adding a condition around if the OS is OSX to link libs.

@shiftkey could you please restart the build for that last commit when you have some time? And just in case, could you check [the last commit](https://github.com/octokit/octokit.net/pull/1462/commits/766d3e21d39024a20253e130abc436d82a6f89a5) see if I did something wrong? My bash-fu is not very good.

mderriey:
Great, thanks! 👍  
Let's just hope the build won't be too flakey.

I'll start working on moving existing platforms to `project.json`. If I understand this right, and if all the current platforms are supported by `project.json`, we could end up with only one `Octokit` project in the solution.

Some posts I want to keep as a reference here:

 - https://oren.codes/2016/02/08/project-json-all-the-things/
 - https://docs.nuget.org/consume/projectjson-intro

Thanks again!

shiftkey:
@mderriey that's fine, we can keep an eye on it over time.

Thanks for getting started on this!

mderriey:
@shiftkey Thanks!

I could have squashed all these commits, it's a bit embarrassing to be honest.

shiftkey:
> I could have squashed all these commits, it's a bit embarrassing to be honest.

Eh, it's fine. Onward and upward.

PR #1463, opened 2016-09-12
Fix of bug with search of C# code

### LINKED ISSUE #1459: Can't search C# code
When I specify Language.CSharp to [SearchCodeRequest.Language] (https://github.com/octokit/octokit.net/blob/master/Octokit/Models/Request/SearchCodeRequest.cs#L90), [SearchCode] (https://github.com/octokit/octokit.net/blob/master/docs/search.md#search-code) request is failed.
Because '#' character contains in request URL and subsequent part is not passed to the server.

The request URL contains the results of [Language.ToParameter] (https://github.com/octokit/octokit.net/blob/master/Octokit/Models/Request/SearchCodeRequest.cs#L164) method.
And the value of ParameterAttribute of Language.CSharp is ["C#"] (https://github.com/octokit/octokit.net/blob/master/Octokit/Models/Request/SearchRepositoriesRequest.cs#L449-L450).

I think should not use Language.ToParameter method in SearchCodeRequest.MergedQualifiers, because it is not used in [SearchRepositoriesRequest] (https://github.com/octokit/octokit.net/blob/master/Octokit/Models/Request/SearchRepositoriesRequest.cs#L146).

dampir:
fix #1459 

Fixed value of `Parameter` attribute. 
Added new integration test and refined unit tests.

### DIFF
--- a/Octokit.Tests.Integration/Clients/SearchClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/SearchClientTests.cs
@@ -1,5 +1,4 @@
 ﻿using System.Collections.Generic;
-using System.Collections.ObjectModel;
 using System.Linq;
 using System.Threading.Tasks;
 using Octokit;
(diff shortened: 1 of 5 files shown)


shiftkey:
@dampir nice work!

PR #1469, opened 2016-09-21
Added a get all organizations method

malamour-work:
Hi,

I noticed that the list all organizations  `GET /organizations` listed here [https://developer.github.com/v3/orgs/](https://developer.github.com/v3/orgs/) was missing from the OrganizationClient. I did not find any issue related to this in the backlog.

I added the implementation in the two client that i found (Base and Observable) with the documentation and added some unit tests.

Let me know if i forgot something or if something need to be changed.

### DIFF
--- a/Octokit.Tests/Clients/OrganizationsClientTests.cs
+++ b/Octokit.Tests/Clients/OrganizationsClientTests.cs
@@ -44,7 +44,7 @@ namespace Octokit.Tests.Clients
             }
         }
 
-        public class TheGetAllMethod
+        public class TheGetAllForUserMethod
         {
             [Fact]
             public async Task RequestsTheCorrectUrl()
@@ -130,7 +130,48 @@ namespace Octokit.Tests.Clients
                 await Assert.ThrowsAsync<ArgumentNullException>(() => client.GetAllForCurrent(null));
             }
         }
+        
+        public class TheGetAllOrganizationsMethod
+        {
+            [Fact]
+            public async Task RequestsTheCorrectUrl()
+            {
+                var connection = Substitute.For<IApiConnection>();
+                var client = new OrganizationsClient(connection);
 
+                await client.GetAllOrganizations();
+
+                connection.Received().GetAll<Organization>(Arg.Is<Uri>(u => u.ToString() == "organizations"), Args.ApiOptions);
+            }
+
+            [Fact]
+            public async Task RequestsTheCorrectUrlWithApiOptions()
+            {
+                var connection = Substitute.For<IApiConnection>();
+                var client = new OrganizationsClient(connection);
+
+                var options = new ApiOptions
+                {
+                    StartPage = 1,
+                    PageCount = 1,
+                    PageSize = 1
+                };
+
+                await client.GetAllOrganizations(options);
+
+                connection.Received().GetAll<Organization>(Arg.Is<Uri>(u => u.ToString() == "organizations"), options);
+            }
+
+            [Fact]
+            public async Task EnsuresNonNullArguments()
+            {
+                var connection = Substitute.For<IApiConnection>();
+                var client = new OrganizationsClient(connection);
+
+                await Assert.ThrowsAsync<ArgumentNullException>(() => client.GetAllOrganizations(null));
+            }
+        }
+        
         public class TheUpdateMethod
         {
             [Fact]
--- a/Octokit.Tests/Clients/OrganizationsClientTests.cs
+++ b/Octokit.Tests/Clients/OrganizationsClientTests.cs
@@ -44,7 +44,7 @@ namespace Octokit.Tests.Clients
             }
         }
 
-        public class TheGetAllForUserMethod
+        public class TheGetAllMethod
         {
             [Fact]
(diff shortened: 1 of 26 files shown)


ryangribble:
Hi @malamour-work thanks heaps for your contribution!  :grinning: 

A couple of things for you:

### Breaking changes to public interfaces/methods
Being a public API we do need to be careful about not breaking existing users and the renamed methods in `ApiUrls` class would currently be a "breaking" change for people.  Whilst they are just helper methods, given they are `public` we could have users that are using them and as such I believe they would fall under our obligation to not make breaking changes without going through the normal process (see below).

It makes sense to rename the existing `ApiUrls.Organizations()` to `ApiUrls.UserOrganizations()` as you've done but technically that means the old name can't be used for the new "all organizations" URL since it needs to be obsoleted.  You'll need to use a different name for that one such as `ApiUrls.AllOrganizations()`.


### Maintaining naming consistency
We try to follow a naming convention where `GetAll` gets the given resource, `GetAllForCurrent()` is for the current user, `GetAllForUser()` is for a nominated user etc.  So in this case, `GetAllOrganizations()` should really just be named `GetAll()` since they are in the organizations client already.  I realize you went with this option because the other methods were already just called `GetAll` but infact they SHOULD have been called `GetAllForUser(string login)` all this time... So it would be cool if you could rename `GetAll(string user) => GetAllForUser(string login)` as part of this change, and implement `GetAll()` as your new method rather than calling it `GetAllOrganizations()`.  Any renames would be done following the "normal process" outlined below.


#### Process for breaking changes:
Create the new method/implementation/classes and mark any old ones as 
`[Obsolete("Please use OrganizationsClient.GetAllForUser() instead.  This method will be removed in a future version")]`

We wait at least 2 releases before removing the obsolete items from the code base.  This gives users some notice (in the form of compiler warnings) before things change on them :+1: 


Let me know if you need any help with this and thanks for stepping up to correct this missing functionality!

Octokit/Helpers/ApiUrls.cs
```diff
@@ -85,10 +86,19 @@ public static Uri Organizations()
...
         /// <param name="login">The login for the user</param>
         /// <returns></returns>
-        public static Uri Organizations(string login)
+        public static Uri UserOrganizations(string login)
         {
             return "users/{0}/orgs".FormatUri(login);
         }
+        
+        /// <summary>
+        /// Returns the <see cref="Uri"/> that returns all of the organizations for the currently logged in user.
```
ryangribble: Extra text here from cut+paste :grinning:

Returns the <see cref="Uri"/> that returns all of the organizations ~~for the currently logged in user~~

ryangribble:
Another thing I just realized is that the API docs mentions this "List organizations" API method is not paginated like normal methods (pagesize, count etc), and instead just uses the `since` parameter.  So you wont be able to take `ApiOptions` in your method signature, instead you'll need a custom request class with the `Since` parameter etc.  You can have a look at the `Repositories.GetAllPublic(PublicRepositoryRequest request)` method for some ideas on how to tackle that

malamour-work:
Hi @ryangribble 

Thanks for your time and for the review. I did the modification you have requested.

Octokit.Tests/Clients/OrganizationsClientTests.cs
```diff
@@ -130,7 +176,43 @@ public class TheGetAllForCurrentMethod
...
+                var connection = Substitute.For<IApiConnection>();
+                var client = new OrganizationsClient(connection);
+
+                await client.GetAll();
 
+                connection.Received().GetAll<Organization>(Arg.Is<Uri>(u => u.ToString() == "organizations"));
+            }
+
+            [Fact]
+            public async Task RequestsTheCorrectUrlWithApiOptions()
```
ryangribble: Since this isn't actually using `ApiOptions` it should be named something else, perhaps `RequestsTheCorrectUrlWithRequestParameter`

malamour-work:
Hi @ryangribble 

I have obsoleted the old `ApiUrls.Organizations()` and created the integration tests like you requested.

If anything else let me know.

Thanks

ryangribble:
Thanks @malamour-work sorry I hadnt had a chance over the weekend to look at this.  I will try and do it tonight.  

Have you actually run the integration tests?  I notice they are flagged as GitHubEnterprise (because they create organizations to then run the tests against, and create organizations is not avaialble on github.com only github enterprise).  For something like a GET call, we could potentially run the test against github.com but I guess that becomes hard since we dont know what organization ID to use as the `since` parameter (and as time marches on, any ID used in the test will start to become further and further back in time, leading to the test retrieving larger amounts of orgs until it eventually times out!).  So probably leaving the tests as `GitHubEnterprise` tests is fine, but we do just need to make sure they actually run succesfully of course :grinning:  

I wll pull this down and run the tests etc and get back to you

ryangribble:
Sorry for the delay on this one 😭 

Ive run the integration tests and made a couple of tweaks and pushed them back up.

This LGTM now - many thanks for your PR and particularly working through all the review comments 😀 


LGTM

PR #1473, opened 2016-09-25
Add missing fields to Repository class - HasPages, SubscribersCount, Size

### LINKED ISSUE #1470: No "Watchers" available in Octokit.net?
Using
`client.Search.SearchRepo(new SearchRepositoriesRequest(searchTerm));`

I don't seem to find "watchers" listed from the repos that I pulled. Is this available/implemented in Octokit?

ryangribble:
As discussed in #1470 this PR adds back the `SubscribersCount` property on `Repository` which is the number of "watchers".  This field was removed in the past however it does contain the correct value and would avoid additional API calls to the `WatchedClient` to retrieve this info.

Also added some other fields that are missing in octokit.net such as `Size` and `HasPages`

Fixes #1470

### DIFF
--- a/Octokit/Models/Response/Repository.cs
+++ b/Octokit/Models/Response/Repository.cs
@@ -14,7 +14,7 @@ namespace Octokit
             Id = id;
         }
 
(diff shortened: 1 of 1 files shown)


ryangribble:
Is some one able to give this one a :+1: ?  I'd like to merge it in

alfhenrik:
Do I count? 😃 Code looks good to me 👍

PR #1477, opened 2016-09-28
Add repository settings to configure PR merge methods

### LINKED ISSUE #1475: Support repository settings to configure the allowed PR merge methods
As per the [recent announcement](https://developer.github.com/changes/2016-09-26-pull-request-merge-api-update/) the repository settings to enable/disable the 3 Pull Request merge options (Merge, Squash, Rebase) are now configurable via the API 🙌 

There are 3 new properties to accomplish this, which will need to be added to our `Repository` response class and also the Create and Edit requests will be updated with new request parameters

API Docs
~~- [Create](https://developer.github.com/v3/repos/#pull-request-merge-configuration-settings)~~ (this turned out to be a documentation error, and the `Create` method does not support these new fields at this time)
``` json
{
  "name": "Hello-World",
  "allow_rebase_merge": false,
  "allow_squash_merge": true,
  "allow_merge_commit": false
}
```

- [Get](https://developer.github.com/v3/repos/#pull-request-merge-configuration-settings-1)
``` json
{
  "id": 2278524,
  "name": "hubot",
  "full_name": "github/hubot",

  "allow_rebase_merge": true,
  "allow_squash_merge": true,
  "allow_merge_commit": true,

}
```

- [Edit](https://developer.github.com/v3/repos/#pull-request-merge-configuration-settings-2)

``` json
{
  "name": "Hello-World",
  "allow_rebase_merge": false,
  "allow_squash_merge": true,
  "allow_merge_commit": false
}
```

ryangribble:
Fixes #1475 

- [x] Add new options to Get Repository methods
- [x] Poke API to figure out whether any of the GetAll (for Current, Public, For Organization, etc) methods return these new fields (despite API docs not mentioning it) and implement if so

(confirmed no other methods like GetAllForCurrent, GetAllPublic, GetAllForOrg return the new fields even when accepts header is provided, so only the individual Get Repository call needs to be implemented)
- [x] Add new options to Create Repository request
- [x] Add new options to Edit Repository request
- [x]  Waiting for github support to advise regarding Create Repository call still not appearing to support the new fields

(support confirmed this was a documentation error and Create Repository call does not support these parameters)

I've had to contact GitHub support as the new fields aren't actually being returned but I'm pretty sure everything is correct on my side...

### DIFF
--- a/Octokit.Tests.Integration/Clients/RepositoriesClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/RepositoriesClientTests.cs
@@ -268,6 +268,36 @@ public class RepositoriesClientTests
                 Task.WhenAll(deleteRepos).Wait();
             }
         }
+
+        [IntegrationTest]
+        public async Task CreatesARepositoryWithRequestedMergeMethod()
+        {
+            var github = Helper.GetAuthenticatedClient();
+            var repoName = Helper.MakeNameWithTimestamp("public-repo");
+
+            var newRepository = new NewRepository(repoName)
+            {
+                AllowMergeCommit = false,
+                AllowSquashMerge = true,
+                AllowRebaseMerge = false
+            };
+
+            using (var context = await github.CreateRepositoryContext(newRepository))
+            {
+                var createdRepository = context.Repository;
+
+                Assert.Equal(repoName, createdRepository.Name);
+                Assert.False(createdRepository.AllowMergeCommit);
+                Assert.True(createdRepository.AllowSquashMerge);
+                Assert.False(createdRepository.AllowRebaseMerge);
+
+                var repository = await github.Repository.Get(Helper.UserName, repoName);
+                Assert.Equal(repoName, repository.Name);
+                Assert.False(repository.AllowMergeCommit);
+                Assert.True(repository.AllowSquashMerge);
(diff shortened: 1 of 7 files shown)


ryangribble:
Github support replied and confirmed there was some upstream problems which they've pushed fixes for. 

With those foxes, the Get and Update integration tests I added are passing :+1:

Still seems to be an issue with Create call which I've asked support to look into

ryangribble:
Support reply regarding create method:

> Turns out that this isn't a problem in the code, but the documentation -- it was originally intended for the feature to be supported by the endpoints to fetch and modify a repository, even though the documentation says it should be supported by the endpoint for creating a repository (the documentation shouldn't have said that). We'll be updating the documentation to reflect this.

So I'll remove the create method tomorrow then this PR can be merged

ryangribble:
OK this is now good to go, I've removed the `CreateRepository` changes, but left the [commit in the history](https://github.com/octokit/octokit.net/pull/1477/commits/cf4c78c9001da4f08dcb3adbb3a84578b2401d4a) incase github do end up supporting it, we can resurrect it

shiftkey:
I'm impatient and Travis is having issues

`#yolo`

PR #1486, opened 2016-10-03
Fixing Async Deadlock

### LINKED ISSUE #1483: Missing ConfigureAwait(false)
In the issue #1111, we discussed on missing ConfigureAwait(false) which has been fixed in the pull #1248 

However, I tried again your solution today, but there is still some ConfigureAwait missing like in the file HttpClientAdapter.SendAsync

```csharp
var clonedRequest = await CloneHttpRequestMessageAsync(request);
```

should be

```csharp
var clonedRequest = await CloneHttpRequestMessageAsync(request).ConfigureAwait(false);
```

If you want, I can create a Pull Request. However, it only takes five mins to fix them all, so I'm not sure if you prefer to verify/fix it yourself.

zzzprojects:
Adding missing ConfigureAwait + put on single line await &
ConfigureAwait

Fixes #1483

### DIFF
--- a/Octokit/Clients/Enterprise/EnterpriseProbe.cs
+++ b/Octokit/Clients/Enterprise/EnterpriseProbe.cs
@@ -77,7 +77,7 @@ namespace Octokit
             IResponse response;
             try
             {
-                response = await httpClient.Send(request);
+                response = await httpClient.Send(request).ConfigureAwait(false);
             }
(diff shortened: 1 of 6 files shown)


zzzprojects:
Pull request for issue #1483

shiftkey:
@zzzprojects thanks!

PR #1488, opened 2016-10-06
Allow base64 content for create/update file

### LINKED ISSUE #1143: Repository Contents API doesn't support Base64 content directly
As raised in #925, we're making the assumption here that you're interested in publishing a UTF-8 string as your content when you create or update a file. That's not entirely true.

As raised in #925 there's a workaround which is to use the Git Data API - see sample here https://github.com/octokit/octokit.net/issues/925#issuecomment-147390168 - but I'll leave this open to indicate an area where we don't currently support it.

laedit:
Fixes #1143.
Since 1.0 has not been reached yet, I made the fix minimal and as I think the method should have been originally, but there is a breaking change.

If you want to mitigate the breaking change I propose three ways:
1. `Content` contains base64 content. Breaking change on `Content.Get()`. Could add corresponding ctors with bool if convert to base64.
2. Insert base class `Base64CreateFileRequest` with a virtual `Content` property wich contains base64 content. `CreateFileRequest` overrides it and provides decoded content.
3. Add `SerializeIgnore` and `SerializeName` attributes, add first on `Content` and second on new property `Base64Content`. When set, `Content` convert value to base64 and set `Base64Content`.

### DIFF
--- a/Octokit/Models/Request/CreateFileRequest.cs
+++ b/Octokit/Models/Request/CreateFileRequest.cs
@@ -106,11 +106,33 @@ namespace Octokit
         /// Creates an instance of a <see cref="CreateFileRequest" />.
         /// </summary>
         /// <param name="message">The message.</param>
-        /// <param name="content">The content encoded in base64.</param>
-        public CreateFileRequest(string message, string content) : base(message)
+        /// <param name="content">The content.</param>
+        public CreateFileRequest(string message, string content) : this(message, content, true)
+        { }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="CreateFileRequest"/> class.
+        /// </summary>
+        /// <param name="message">The message.</param>
+        /// <param name="content">The content.</param>
+        /// <param name="branch">The branch the request is for.</param>
+        public CreateFileRequest(string message, string content, string branch) : this(message, content, branch, true)
+        { }
+
+        /// <summary>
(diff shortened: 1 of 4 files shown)


Octokit/Models/Request/CreateFileRequest.cs
```diff
@@ -106,7 +106,7 @@ public class CreateFileRequest : ContentRequest
         /// Creates an instance of a <see cref="CreateFileRequest" />.
         /// </summary>
         /// <param name="message">The message.</param>
-        /// <param name="content">The content.</param>
+        /// <param name="content">The content encoded in base64.</param>
```
shiftkey: @ryangribble @Haacked how do we feel about this breaking change? Should we instead introduce overloads?

Haacked: Yeah, that's a very subtle breaking change. Probably best to use an overload.

Someday I'd like to write a `Base64Content` type. :smile: rather than using strings for everything.

laedit:
@shiftkey @Haacked is that what you have in mind? Or do you prefer another way?

ryangribble:
+1 on not making a breaking change

The overload approach you've gone with seems ok to me :+1:

ryangribble:
LGTM

laedit:
@ryangribble  No problem, thanks for the merge :smile:

Issue #1358, opened 2016-06-10
Missing Properties in Milestone Class


shaggygi:
Why doesn't the Milestone Class provide a HtmlUrl property similar to the Repository Class?  I noticed there is a Url property and the JSON returned includes a tag for html_url.  Is there another approach to get this information?

ryangribble:
Any missing properties are not likely to be intentional, more so just historic or accidental. 

Would you be happy to send a pull request that adds the property?

shaggygi:
Yes, I'll look into the PR.

ryangribble:
Great!  Let us know if you need any help :+1:

SeanKilleen:
I believe this was closed via #1489.

PR #1489, opened 2016-10-09
Missing Milestone fields

StanleyGoldman:
Adding missing Milestone fields as per #1358

### DIFF
--- a/Octokit/Models/Response/Milestone.cs
+++ b/Octokit/Models/Response/Milestone.cs
@@ -14,11 +14,10 @@ namespace Octokit
             Number = number;
         }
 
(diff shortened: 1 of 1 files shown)


ryangribble:
Thanks for the PR 👍 

Adding the `HtmlUrl` is 💎 - this is the url to this milestone on github.com and definitely should be included 

But fields like `LabelsUrl` generally wouldn't be added to octokit.net, since we aim to provide a c# model over the github API, we don't normally expose the internal API hypermedia url links to the end user. 

Instead we provide access to the [associated API endpoint] (https://developer.github.com/v3/issues/labels/#get-labels-for-every-issue-in-a-milestone) through the object model, in this case it's at [`client.Issues.Labels.GetAllForMilestone()`](https://github.com/octokit/octokit.net/blob/master/Octokit/Clients/IssuesLabelsClient.cs#L154)

Given the above, are you cool to drop off that labels field from the PR? If not let us know your use case and we can discuss 😀

StanleyGoldman:
Done

ryangribble:
Sorry for the delay, have been having a bit of no home computer time this week :)

Thanks for the PR!

LGTM

Issue #1251, opened 2016-04-07
New activity events not being deserialized correctly


shiftkey:
Found while testing #1240 - the lack of `type` on these is why the test is failing due to us incorrectly deserializing the payload:

<img width="928" src="https://cloud.githubusercontent.com/assets/359239/14354978/7764e796-fcad-11e5-8a04-2ac16f38f695.png">
<img width="924" src="https://cloud.githubusercontent.com/assets/359239/14354980/776a2d78-fcad-11e5-807c-52aa84699d90.png">
<img width="922" src="https://cloud.githubusercontent.com/assets/359239/14354982/776e222a-fcad-11e5-88d1-a823b743c343.png">
<img width="955" src="https://cloud.githubusercontent.com/assets/359239/14354981/776a5bd6-fcad-11e5-8b97-dfae08662cd7.png">
<img width="924" src="https://cloud.githubusercontent.com/assets/359239/14354979/77656720-fcad-11e5-853e-5219b42ccb41.png">

shiftkey:
Here's some example events that I've found which break our current code for parsing activities...

 - A subscription event:

```json
{
  "id": 691197253,
  "url": "https:\/\/api.github.com\/repos\/octokit\/octokit.net\/issues\/events\/691197253",
  "actor": {
    "login": "ryangribble",
    "id": 5425163,
    "avatar_url": "https:\/\/avatars.githubusercontent.com\/u\/5425163?v=3",
    "gravatar_id": "",
    "url": "https:\/\/api.github.com\/users\/ryangribble",
    "html_url": "https:\/\/github.com\/ryangribble",
    "followers_url": "https:\/\/api.github.com\/users\/ryangribble\/followers",
    "following_url": "https:\/\/api.github.com\/users\/ryangribble\/following{\/other_user}",
    "gists_url": "https:\/\/api.github.com\/users\/ryangribble\/gists{\/gist_id}",
    "starred_url": "https:\/\/api.github.com\/users\/ryangribble\/starred{\/owner}{\/repo}",
    "subscriptions_url": "https:\/\/api.github.com\/users\/ryangribble\/subscriptions",
    "organizations_url": "https:\/\/api.github.com\/users\/ryangribble\/orgs",
    "repos_url": "https:\/\/api.github.com\/users\/ryangribble\/repos",
    "events_url": "https:\/\/api.github.com\/users\/ryangribble\/events{\/privacy}",
[...]

PR #1490, opened 2016-10-09
Correcting return type in ObservableEventsClient

StanleyGoldman:
#1251 threw me for a loop for a bit.
It turns out this API deals in `IssueEvent` not `Activity`.

### DIFF
--- a/Octokit.Reactive/Clients/IObservableEventsClient.cs
+++ b/Octokit.Reactive/Clients/IObservableEventsClient.cs
@@ -76,7 +76,7 @@ namespace Octokit.Reactive
         /// </remarks>
         /// <param name="owner">The owner of the repository</param>
         /// <param name="name">The name of the repository</param>
-        IObservable<Activity> GetAllIssuesForRepository(string owner, string name);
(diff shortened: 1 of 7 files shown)


StanleyGoldman:
Hmm, I might need some help understanding why the integration tests fail.
I supplied a username, organization, and token in `Octokit.Tests.Integration.Helper` and the rest seemed to work for me.

shiftkey:
@StanleyGoldman these failures appear related to the unit tests - check the mocks locally as they may need to be updated.

Also, have you checked the [Slow Tests](https://github.com/octokit/octokit.net/blob/master/CONTRIBUTING.md#slow-tests) section in the CONTRIBUTING docs? That should help you setup your environment for testing.

StanleyGoldman:
Hey @shiftkey I didn't see the Slow Tests documentation, I will definitely give it a try tonight.
It's just odd because I did what I feel was an equivalent locally, which passed during tests.
I'll let you know how it goes.

StanleyGoldman:
@shiftkey Good to go.

shiftkey:
@StanleyGoldman thanks!

PR #1492, opened 2016-10-10
Updated monotouch projects to unified Xamarin.iOS projects

jsauvexamarin:
When I tried to add the Octokit.net Nuget package to any of my Xamarin.iOS projects (10.0.1.10), they wouldn't install because the package couldn't target the platform. Xamarin.iOS 10+ requires that the Unified Xamarin iOS APIs be used, which is why I expect the installation was failing.

I've fixed the iOS projects in-place. This is probably not the best solution, because it means that the package will break things for anyone using Xamarin.iOS that is less than version 10. I don't presume to know how you want to structure your library around those changes. A better solution would probably be to have a new set of iOS projects that target Xamarin.iOS 10+ and leave the old monotouch projects as they are. But like I said, I won't presume to know how you want to deal with that.

There were also many files that were referenced in the iOS projects, but that were missing from the repo altogether. I have updated the projects to omit these files.

### DIFF
--- a/Octokit.Reactive/Octokit.Reactive-Monotouch.csproj
+++ b/Octokit.Reactive/Octokit.Reactive-Monotouch.csproj
@@ -6,7 +6,7 @@
     <ProductVersion>8.0.30703</ProductVersion>
     <SchemaVersion>2.0</SchemaVersion>
     <ProjectGuid>{F4ADA431-8344-4B36-9A0B-C4D96AF53908}</ProjectGuid>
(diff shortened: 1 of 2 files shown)


jsauvexamarin:
Instead of modifying the existing projects, I think I _will_ just add two new projects that target the Xamarin unified iOS API: 
- Octokit-Xamarin.iOS
- Octokit.Reactive-Xamarin.iOS

I'll submit a PR soon when it's ready.

shiftkey:
> I'll submit a PR soon when it's ready.

So this PR is no longer needed?

jsauvexamarin:
Correct. I'll submit a new one.

Issue #1500, opened 2016-11-09
Issue Comments - need an overload to take in optional arguments


shiftkey:
Documentation: https://developer.github.com/v3/issues/comments/#list-comments-in-a-repository

Existing methods: https://github.com/octokit/octokit.net/blob/master/Octokit/Clients/IIssueCommentsClient.cs#L36-L66

There should be overloads for each of these methods which accept an parameter which `IssueCommentRequest` and has those fields from the documentation.

SeanKilleen:
I believe this is now closed via #1501.

PR #1501, opened 2016-11-11
Fix IssueCommentClient #1500

pjc0247:
Original Issue : https://github.com/octokit/octokit.net/issues/1500

Hi,
I added 4 methods to `IssueCommentsClient.cs`.

But there are two duplicated options which named `IssueCommentSort` and `PullRequestReviewCommentSort`. I think these types can be merged like a `SortDirection` option.

Thanks.

### DIFF
--- a/Octokit.Reactive/Clients/ObservableIssueCommentsClient.cs
+++ b/Octokit.Reactive/Clients/ObservableIssueCommentsClient.cs
@@ -87,7 +87,7 @@ namespace Octokit.Reactive
             Ensure.ArgumentNotNullOrEmptyString(name, "name");
             Ensure.ArgumentNotNull(options, "options");
 
-            return GetAllForRepository(owner, name, new IssueCommentRequest(), ApiOptions.None);
+            return GetAllForRepository(owner, name, new IssueCommentRequest(), options);
         }
 
         /// <summary>
@@ -100,7 +100,7 @@ namespace Octokit.Reactive
         {
             Ensure.ArgumentNotNull(options, "options");
 
-            return GetAllForRepository(repositoryId, new IssueCommentRequest(), ApiOptions.None);
+            return GetAllForRepository(repositoryId, new IssueCommentRequest(), options);
         }
 
         /// <summary>
--- a/Octokit.Reactive/Clients/ObservableIssueCommentsClient.cs
+++ b/Octokit.Reactive/Clients/ObservableIssueCommentsClient.cs
@@ -87,7 +87,7 @@ namespace Octokit.Reactive
             Ensure.ArgumentNotNullOrEmptyString(name, "name");
             Ensure.ArgumentNotNull(options, "options");
 
-            return _connection.GetAndFlattenAllPages<IssueComment>(ApiUrls.IssueComments(owner, name), null, AcceptHeaders.ReactionsPreview, options);
+            return GetAllForRepository(owner, name, new IssueCommentRequest(), ApiOptions.None);
         }
 
         /// <summary>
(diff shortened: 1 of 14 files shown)


ryangribble:
Hi @pjc0247 sorry for the delay in responding.

The builds are currently failing unit test let me know if you need help with it

Octokit.Tests/Reactive/ObservableIssueCommentsClientTests.cs
```diff
@@ -89,10 +95,7 @@ public void RequestsCorrectUrlWithApiOptions()
 
                 client.GetAllForRepository("fake", "repo", options);
 
-                gitHubClient.Connection.Received(1).Get<List<IssueComment>>(
-                    new Uri("repos/fake/repo/issues/comments", UriKind.Relative), 
-                    Arg.Any<Dictionary<string, string>>(),
-                    "application/vnd.github.squirrel-girl-preview");
+                gitHubClient.Received().Issue.Comment.GetAllForRepository("fake", "repo", request, options);
```
ryangribble: Id say this is the one failing, if you have a look at what the previous code was

Octokit.Reactive/Clients/ObservableIssueCommentsClient.cs
```diff
@@ -87,7 +87,7 @@ public IObservable<IssueComment> GetAllForRepository(string owner, string name,
             Ensure.ArgumentNotNullOrEmptyString(name, "name");
             Ensure.ArgumentNotNull(options, "options");
 
-            return _connection.GetAndFlattenAllPages<IssueComment>(ApiUrls.IssueComments(owner, name), null, AcceptHeaders.ReactionsPreview, options);
+            return GetAllForRepository(owner, name, new IssueCommentRequest(), ApiOptions.None);
```
ryangribble: This should be passing through the `options` variable rather than `ApiOptions.None`

Octokit.Reactive/Clients/ObservableIssueCommentsClient.cs
```diff
@@ -100,7 +100,69 @@ public IObservable<IssueComment> GetAllForRepository(long repositoryId, ApiOptio
         {
             Ensure.ArgumentNotNull(options, "options");
 
-            return _connection.GetAndFlattenAllPages<IssueComment>(ApiUrls.IssueComments(repositoryId), options);
+            return GetAllForRepository(repositoryId, new IssueCommentRequest(), ApiOptions.None);
```
ryangribble: This should be passing through the options variable rather than ApiOptions.None

pjc0247: Ah... thanks.

Octokit.Tests/Reactive/ObservableIssueCommentsClientTests.cs
ryangribble: These tests aren't right.  They arent failing but they also aren't correct.  It may be something to do with the multiple nested subclients on the end of the `.Received()` call

but this observable method doesnt actually call into the `IGitHubClient` method, it calls the `_connection.GetAndFlattenPages()` method and thus this test should be checking for `gitHubClient.Connection.Received(1).Get....`

Octokit.Tests/Reactive/ObservableIssueCommentsClientTests.cs
```diff
@@ -101,17 +106,22 @@ public void RequestsCorrectUrlWithRepositoryIdWithApiOptions()
...
                     PageSize = 1,
                     PageCount = 1
                 };
 
-                client.GetAllForRepository(1, options);
+                client.GetAllForRepository(1, request, options);
 
-                gitHubClient.Connection.Received(1).Get<List<IssueComment>>(
-                    new Uri("repositories/1/issues/comments", UriKind.Relative), Arg.Is<IDictionary<string, string>>(d => d.Count == 2), null);
+                gitHubClient.Received().Issue.Comment.GetAllForRepository(1, request, options);
```
ryangribble: and this one too

pjc0247: @ryangribble sry, I'm not famillar with unit testing. I'll make a fix.

pjc0247:
@ryangribble finally understood. they already have 2 parameters which contain pagenation data....
sorry for messy commits.

ryangribble:
Sorry for the delay, this looks good to merge now!  

Many thanks for the contribution, and for your patience :)

LGTM

PR #1476, opened 2016-09-28
[WIP] Create a new build framework for Octokit using Cake.Frosting

ryangribble:
I've started work on #1440 to investigate moving to [CAKE](https://cakebuild.net/) for our build system.

I've decided to go down the road of using their latest [`Cake.Frosting`](https://github.com/cake-build/frosting) offering, which is essentially a .NET Core host/loader around the cake build framework.  This means we have full intellisense support of the build scripts AND the ability to debug the build scripts natively in Visual Studio!

So far it is building the `Octokit.Next` and `Octokit.Next.Tests` projects and running the unit tests.  I'm also experimenting with using GitVersion to calculate version numbers with the ultimate goal of publishing alpha packages to nuget on each merged PR to master as well as still supporting the official releases off a release branch

Everything lives under a `/build` sub directory.  To execute the build we simply `dotnet restore` and `dotnet run` in this directory.

The interesting parts are `BuildContext` which is essentially our build settings, `BuildLifetime` where we populate our settings from commandline options or other builtin helpers.  Then in the `/build/Tasks/` directory is where the tasks are actually implemented

In order to run via AppVeyor/Travis we will need `.ps1/.sh` bootstrap scripts, similar to the approach of the existing FAKE implementation

So far I have the following tasks planned:
- [x] Add Cake.Frosting build for Octokit.Next projects
- [ ] Add README.md file to `/build` directory to provide some info about the build script
- [x] Get running in AppVeyor
- [ ] ~~Get running TravisCI~~
- [ ] Get sourcelink replacement (GitLinkRunner) working
- [ ] Play around with GitVersion settings some more
- [x] Play around with nuget publishing
- [x] Decide whether to tackle the legacy octokit projects or only have this for dotnet core projects
- [ ] Get feedback from maintainers and contributors regarding the use of Cake and specifically Cake.Frosting

Issue #1440, opened 2016-08-08
migrate build scripts over to CAKE


shiftkey:
I've had some interest from various corners about moving our build scripts over to [Cake](https://github.com/cake-build/cake). The process seems very similar to what we're doing currently, so I'm not really going to stand in the way.

Things I'd like to verify are possible with Cake (mostly due to emulating our current flow for releasing things):
- use `dotnet` CLI tools to build, rather than MSBuild
- `AssemblyInfo` generation from our `ReleaseNotes.md` file
- SourceLink support (there's a nice helper for FAKE, but I gather there's a way to do it directly for our scenario)

Also note that these things in the current scripts will be made redundant by the migration to using the `dotnet` CLI, so I wouldn't worry about investing too much time in porting them over:
- generating NuGet packages (although we might need to do some manual unpacking/repacking to get it working with NuGet) 
- running xUnit tests

ryangribble:
> SourceLink support (there's a nice helper for FAKE, but I gather there's a way to do it directly for our scenario)

Just on this point, I stumbled upon [GitLink](https://github.com/GitTools/GitLink) which sounds like it may do the same thing as what SourceLink currently does.  CAKE has [built in aliases](http://cakebuild.net/api/cake.common.tools.gitlink) for GitLink

shiftkey:
@ryangribble if GitLink gives us the same end result, then I'm fine with switching. I forget the backstory and why they didn't team up way back in the day...

ryangribble:
Hmmm so it turns out both SourceLink and GitLink dont support dotnetcore project.json and .xproj files which is a bit of a bummer in terms of our own dotnetcore move (which isnt strictly related to moving to CAKE however I was kind of intending to do it all at the same time)

GitLink [aren't planning on doing anything](https://github.com/GitTools/GitLink/issues/83) until project.json goes away (although Im not sure when that actually will occur), and SourceLink were [waiting for upstream depedencies](https://github.com/ctaggart/SourceLink/issues/121) as of a few months ago which are hopefully now in place so there could be hope there...

ghuntley:
Yo @shiftkey I'd suggest using `gitversion` to automatically version your releases and generating release notes using `GitReleaseManager` which automatically builds RELEASENOTES.md from closed github issues + github tags. 

See https://github.com/reactiveui/ReactiveUI/blob/develop/GitReleaseManager.yaml  and https://github.com/reactiveui/ReactiveUI/blob/develop/GitVersion.yml

[...]

PR #1503, opened 2016-11-17
Port to .NET Core

mderriey:
As mentioned in #1419 I'm trying to port Octokit to .NET Core.

Very early stage at the moment but I'd rather PR and get feedback quickly.
Right now only targeting `netstandard1.1` and `net45`.

Any comment/feedback more than welcome.


mderriey:
I'm struggling to get the FAKE script running. If a kind soul comes by and spots the error 🤗 Otherwise I'll take another look at it a bit later.

ryangribble:
I can't help on the Fake front so much, but I'd like to combine the CAKE changes I was working on  (#1440)  with this branch since it seems logical to me to do the cake switch as part of the dotnet core work. I should be able to do some work on this this weekend

mderriey:
Sweet, that'd be great!

ryangribble:
I merged your changes into my branch for #1476 and built+tested everything fine on AppVeyor using the Cake build 😁 

Basically the cake build is doing GitVersion to calculate version number, package restore, build, test and then packaging nuget packages, setup as artifacts in AppVeyor.  I also removed the Octokit.Next projects since they were just a sample before you got stuck into the real thing.

As youll see from the commands in the AppVeyor.yml file - to run the build you just change into `./build` directory and `dotnet restore` `dotnet run` the Cake.Frosting build project.  Or if you want to debug the build you simply open up `/build/build.sln` in Visual Studio and hit F5.  Ultimately I will make it easier to run a build (eg a bootstrap `build.ps1` in the root directory) but for now its just the 2 commands, until we actually delete the Fake f# and associated scripts

Take a look at it on branch https://github.com/TattsGroup/octokit.net/tree/port-to-dotnet-core and merge it into your branch if you think it's a go :+1:

mderriey:
Wow, great, I'll take a look at it soon. Thanks :+1:

mderriey:
Thank you so much @ryangribble!
Next steps:

 - get the integration test project ported too, and restored/built/tested in CI
 - bump `System.Reactive` to v3.x

@shiftkey based on https://github.com/octokit/octokit.net/issues/1419#issuecomment-263150006, should I just get rid of `net45` and keep just `netstandard1.1`?

shiftkey:
@mderriey I'd leave `net45` in there unless you're having specific issues with it

mderriey:
@shiftkey / @ryangribble 

The integration test project has been ported to `netstandard1.1` and it turns out `System.Reactive` had already been bumped to `3.x` since it's the only version compatible with `netstandard`.

What I think would be good is for you to have a look at some specific methods that were ported to .NET Core and see if you're happy with the implementation:

 - [`FormatUserAgent` of `Connection`](https://github.com/mderriey/octokit.net/blob/091248bd1580153baecde6e701efa891bc434a1e/Octokit/Http/Connection.cs#L645)
 - [A fix on a test to check the body contains expected values](https://github.com/octokit/octokit.net/pull/1503/commits/e69169c8720962903f0b1d337d538287839e56a5). The original implementation was trowing a `'object' does not contain a definition for 'fingerprint'`
 - [A change in SimpleJson since `SerializationException` doesn't exist in `netstandard` land](https://github.com/mderriey/octokit.net/commit/ad561d70b0d96e840d241d14f7fddd3e0c93f8e9#diff-7e26fb1257de97b32ec85c7d5d84dbc9)

There might be others, I'll have deeper look.
Could you also let me know what you think is missing / what we have to focus on to get closer to a merge?

Thanks for the support, it's really appreciated 👍

Octokit/Http/Connection.cs
```diff
@@ -644,10 +647,22 @@ internal static TwoFactorType ParseTwoFactorType(IResponse restResponse)
...
+#endif
+
             return string.Format(CultureInfo.InvariantCulture,
+                format,
+#if !HAS_ENVIRONMENT
+                productInformation,
+                RuntimeInformation.OSDescription,
+                RuntimeInformation.OSArchitecture.ToString().ToLowerInvariant(),
+#elif NETFX_CORE
                 "{0} ({1} {2}; {3}; {4}; Octokit {5})",
```
ryangribble: since `format` is being set above and included in all cases, I think this line may be spurious?

infact do we even really need this hardcoded WindowsRT/Windows8 etc stuff below @shiftkey? 

Can this `NETFX_CORE` case here just be removed and either user `Environment` or `RuntimeInformation` according to the `HAS_ENVIRONMENT` preprocessor definition now added?

Octokit/Clients/MiscellaneousClient.cs
```diff
@@ -3,9 +3,8 @@
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Threading.Tasks;
-#if NET_45
 using System.Collections.ObjectModel;
-#endif
+
```
ryangribble: extra whitespace line (there are a few similar occurences on other files).  this is definitely sraping the bottom of the barrel in terms of what I can pick up in review!

Octokit/Http/ApiInfoParser.cs
```diff
@@ -8,7 +8,7 @@ namespace Octokit.Internal
     internal static class ApiInfoParser
     {
         const RegexOptions regexOptions =
-#if NETFX_CORE
+#if !HAS_REGEX_COMPILED_OPTIONS
```
ryangribble: this block could be written the same as [here](https://github.com/octokit/octokit.net/pull/1503/files#diff-9b75446537d00a96e613eefafd903662R126) for consistency

Octokit/Octokit.xproj
```diff
@@ -7,11 +7,11 @@
 
   <Import Project="$(VSToolsPath)\DotNet\Microsoft.DotNet.Props" Condition="'$(VSToolsPath)' != ''" />
   <PropertyGroup Label="Globals">
-    <ProjectGuid>4052af53-4c1b-48d1-b873-bfd0351481c3</ProjectGuid>
-    <RootNamespace>Octokit.Next</RootNamespace>
+    <ProjectGuid>104e8324-c2b9-43be-8040-36b736a64d45</ProjectGuid>
+    <RootNamespace>OctokitBidon</RootNamespace>
```
ryangribble: OctokitBidOn again

ryangribble:
In terms of the #defines 

- Some code is referencing `HAS_REGEX_COMPILED_OPTIONS` but this isnt defined for any project.  Im thinking the `net45` one should set this?  There are also perhaps other #defines in a similar boat (eg code that uses them but no projects define them, or projects that define something which isnt actually used anywhere in code)

- The `HAS_TYPEINFO` define is used to wrap most calls to `GetTypeInfo()` but there are some in the code base that are "unguarded".  Is this a problem?  If it's not a problem, then why do we need to guard any of them?

ryangribble:
This is really great stuff @mderriey I can't spot anything too serious and everything is locally working for me including builds in VS2015 and cake script, unit tests, integration tests etc.  

I guess the next steps are to actually prove things are working (eg a running dotnet core app on on mac/linux/windows using octokit.core library) as well as a traditional full framework dotnet app on windows using the newly built library.  We can also work on our nuget packages (AppVeyor should be spitting them out) to see if they are working etc...  

The last piece of the puzzle was source linking in the PDB files but I dont think either of the 2 options mentioned (SourceLink and GitLink) will work for us at this time.  I found another fork/derivative recently that Im going to look at [PdbGit](https://github.com/AArnott/PdbGit)

Octokit/Http/Connection.cs
mderriey: noted, thanks. for now I removed the WindowsRT/Windows8 specific stuff since I don't think the target is to create platform-specific packages.

Octokit.Tests.Conventions/Octokit.Tests.Conventions.xproj
```diff
@@ -6,11 +6,11 @@
   </PropertyGroup>
   <Import Project="$(VSToolsPath)\DotNet\Microsoft.DotNet.Props" Condition="'$(VSToolsPath)' != ''" />
   <PropertyGroup Label="Globals">
-    <ProjectGuid>2740af44-cb2e-4fd8-a221-248693978a0c</ProjectGuid>
-    <RootNamespace>Octokit.Next.Tests</RootNamespace>
+    <ProjectGuid>a04124dc-d057-4db0-8e5f-ecd114fa371a</ProjectGuid>
+    <RootNamespace>OctokitBidon</RootNamespace>
```
mderriey: sorry, some hacky stuff I must have done at the beginning and forgot to remove them.

Octokit/Clients/MiscellaneousClient.cs
mderriey: 🔥 them all!

Octokit/Http/ApiInfoParser.cs
mderriey: done, thank you.

mderriey:
Thanks for the review @ryangribble, much appreciated. I pushed some changes according to your comments.

You were right for the `HAS_REGEX_COMPILED_OPTIONS`, it was meant for `net45` but I forgot it.

I made some changes to the reflection code based on a trick @Haacked used, which is to define the `GetTypeInfo` method for platforms that don't have it. This way, you can use `GetTypeInfo` consistently in your code without having to guard for it. A few other extensions methods were created to create missing APIs in `netstandard1.1`  that exist in `net45`, like `GetProperty` or `GetMember`.

mderriey:
@ryangribble regarding making sure everything works, do you want to publish a pre-release package on NuGet and asking people who are keen to try to update and make sure their apps still work? Or were you leaning towards us making sure it looks OK?

build.fsx
```diff
@@ -125,12 +127,18 @@ Target "RestoreDotNetCore" (fun _ ->
...
     |> Fake.DotNetCli.Build id
 )
 
 Target "UnitTestsDotNetCore" (fun _ ->
-    !! "./Octokit.Next.Tests"
-    |> Fake.DotNetCli.Test id
+    [ "./Octokit.Tests"
+      "./Octokit.Tests.Conventions" ]
+    |> Seq.iter (fun d ->
+        Fake.DotNetCli.Test (fun p ->
```
shiftkey: There seems to be a signature issue with this function:

```
Error:

        build.fsx(138,9): error FS0001: This expression was expected to have type
            unit
        but here has type
            'a -> unit
```

Throwing a `|> ignore` on the end seems to do the trick :trollface:

```fsharp
Fake.DotNetCli.Test (fun p ->
    { p with
        WorkingDir = d }) |> ignore
```

ryangribble: if we are going to CAKE from fake anyway this F# voodoo may not be required anymore :)

shiftkey: Fair point - I just pulled it down to do a `.\build` but I guess that's now handled by something else...

mderriey: I'm happy to remove the FAKE file and edit the `build.cmd` and `build.sh` to mimic what happens on AppVeyor. Sorry about the confusion.

mderriey:
My bash-fu is not great, I'll have a look at what happens later

You can now use `.\build.cmd` and it will invoke CAKE. Parameters can be specified with the `--name=value` syntax.

Here are the parameters I found in the build project and their default values:

| Parameter name  | Default value | Other possible values                           |
|-----------------|---------------|-------------------------------------------------|
| `target`        | `Default`     | `Dotnet-Unit-Tests`, `Dotnet-Integration-Tests` |
| `configuration` | `Release`     | `Debug`                                         |
| `forcepublish`  | `false`       | `true`                                          |

mderriey:
@shiftkey / @ryangribble I updated the build files so they use the CAKE build.

How do you think you'll integrate PdbGit in the process? It looks to me that we need to hook this into the CAKE pipeline since NuGet packages are created as part of it. A very quick search leads me to think that we can't just execute an .exe from CAKE, we'd need to create some code that wraps that execution. Is this correct?

Cheers!

mderriey:
Alright, I took a shot at integrating PdbGit in the CAKE pipeline. It was quickly put together and possibly not very robust, but it works on both my local and AppVeyor.

ryangribble:
Wow that's awesome! Have you actually tried source debugging the built dll/pdb?

mderriey:
I haven't :confounded:. The instructions look pretty simple but every time I tried it in the past it would never work...

Sorry about that. Do you think you could give it a go? If not I'll try my luck once again :+1:

ryangribble:
Im also not very across this stuff either by the way.  My basic understanding is that source linking allows stepping into the source of a nuget loaded dll without requiring symbol servers etc.  It somehow embeds the links of the files on github into the PDB file.

SourceLink and GitLink both scraped csproj files and so on, so they werent compatible with dotnet core (due to project.json) but then even with the "new" csproj for dotnet core now, I suspect they will still need to be changed since the csproj doesnt actually contain all files anymore and just has globs etc.  PdbGit was another one I found which seemed to take a different approach and you pointed at the actual PDB files rather than the source project file, although I have no idea if that means it would work with dotnet core PDB files or not etc...  It's pretty much stuff that needs to be researched... and in the meantime I believe one or both of SourceLink/GitLink are making moves to support dotnet core now that the new project structure is here.

In terms of PdbGit it probably makes sense that an AppVeyor build from a PR in this repo wouldnt know to reference files/commits from the contributors fork/branch and is instead trying to hit the official repo... although it's weird that the commit hash doesnt seem to exist anywhere :grining: 

I guess we can always look at the PdbGit source code or reach out to the author to answer some of these questions.  It also could be possible to pass parameters to PdbGit for the SHA and repository to use (but then Im not sure if AppVeyor even knows those details).

It's not really important to build source linked PDB files from pending PR's though anyway... we really just want the "official" builds from the octokit repo to have that capability.  What I'd like to do when we cut over to this dotnet core and CAKE stuff, is have every build from master branch publish a pre-release package to nuget.org, and then periodically do the "official" releases like we do now.

mderriey:
Alright, after cloning the repo [like AppVeyor does](https://ci.appveyor.com/project/github-windows/octokit-net/build/0.23.1-PullRequest.1503+76.build.1899#L3), posh-git shows me that `326116e` commit SHA that [PdbGit uses to link sources](https://ci.appveyor.com/project/github-windows/octokit-net/build/0.23.1-PullRequest.1503+76.build.1899#L600). I don't know where it comes from, though.

So what I did is a fresh clone of the repository that lives under my account, built it, created a console application pointing to the newly built assembly with the modified pdb next to it, and gave it a go.

It partially works, in the sense that I can F11 into the code from the Octokit assembly, but it's off by a line in the file that I tried. Trying to set a breakpoint on a specific line of code would have VS show a popup saying `The following breakpoint cannot be set [...] The breakpoint failed to bind`.

mderriey:
Unfortunately this doesn't work. As far as AppVeyor is concerned, the repo comes from `octokit.net/octokit`. I know it's not a big deal but I wanted to see if I could get this working.

Now here's a crazy: how about we publish a prerelease package of Octokit targeting `netstandard1.1` and use it in this PR to get the correct repo and commit information since AppVeyor gives us the PR number? I know it's a very small test and won't cover much, though.

What are your thoughts?

ryangribble:
Yep I think we can publish a prerelease alpha package off this branch (without source linking in the PDB) so we can play around with it further in test projects and get something in the hands of our users so they can kick the tyres too.  

If you also want to play with writing a CAKE helper that uses that octokit dotnetcore library to find out the correct URL to pass to PdbGit so source linking of pending PR builds works from the contributors source branch that's cool from a "nailed it!" perspective... although not "necessary" if you dont want to, since I dont think we would ever publish packages from a pending PR.  

I do plan to step up the release frequency this year, including automatic publishing by the CAKE build script of every merged PR to master as a prerelease package to nuget, but I dont think packages from inflight PR's would be a thing wanted or needed.  Still, for "science" it could be cool to see if you can get it working 😛

ryangribble:
@mderriey and I have made a few more tweaks and all the integration tests are now passing.

I'm pleased to announce we now have Pre-Release packages up on NuGet!

https://www.nuget.org/packages/Octokit/0.24.1-alpha0001
https://www.nuget.org/packages/Octokit.Reactive/0.24.1-alpha0001

ryangribble:
Im going to merge this PR into the `dotnetcore` branch I just created.

We can then correct any further issues via specific targeted PR's for better tracking.

Ive also created a Milestone to track everything related to dotnetcore

ryangribble:
@mderriey - you sir are a LEGEND :+1:

shiftkey:
Congrats to the both of you on getting this preview out! Keep up the awesome work 💖!

Issue #1512, opened 2016-12-12
ProtectedBranches API changes for Required Review Enforcement


ryangribble:
As per this [blog anouncement](https://developer.github.com/v3/repos/branches/#update-branch-protection) the ProtectedBranches API has a new `required_pull_request_reviews` object 

- On the existing `UpdateBranchProtection` method (add to our `BranchProtectionSettingsUpdate` request object, which will require a few `ctor`s to be added to cater for the different combinations of the 3 fields being provided/not provided)

- Returned by the existing GetBranchProtection method (add to our `BranchProtectionSettings` response object)

- On new methods to [Get](https://developer.github.com/v3/repos/branches/#get-pull-request-review-enforcement-of-protected-branch), [Update](https://developer.github.com/v3/repos/branches/#update-pull-request-review-enforcement-of-protected-branch), [Remove](https://developer.github.com/v3/repos/branches/#remove-pull-request-review-enforcement-of-protected-branch) the required review enforcement
  - GET /repos/:owner/:repo/branches/:branch/protection/required_pull_request_reviews
  - PATCH /repos/:owner/:repo/branches/:branch/protection/required_pull_request_reviews
  - DELETE /repos/:owner/:repo/branches/:branch/protection/required_pull_request_reviews

M-Zuber:
I'd like to try and do this one if it's still available

ryangribble:
All yours @M-Zuber !

ryangribble:
Unless you prefer the organization ones I also just raised #1513 and #1514 in which case Im happy to knock over this protected branches one since Im pretty familiar with this area.  

Up to you though, if you want to do this one that's cool

M-Zuber:
I want something small in order to get back into things.
Lets see if I can get this one done fast enough to start tackling those also 😀

BigAlInTheHouse:
existen datos ahora.

ALG Services

M-Zuber:
How exactly do you want the additional ctors to look?
Currently there are 2: one that takes only a `BranchProtectionRequiredStatusChecksUpdate`, and the other that takes a `BranchProtectionPushRestrictionsUpdate` in addition.

Based on my understanding of the docs: `You can, and should, pass required_pull_request_reviews as well. While this is not required currently, it should be treated as such and will be soon.`,  the `BranchProtectionRequiredPullRequestReviewsUpdate` object is different in that it can not be disabled by passing in null.

This would lead to a total of 3 ctors:
[...]

PR #1523, opened 2016-12-22
ProtectedBranches API changes for Required Review Enforcement

M-Zuber:
Edited by @ryangribble

Hope you don't mind @M-Zuber but I've taken over this PR 😀 

I realized that [we don't have much time](https://developer.github.com/changes/2017-06-16-loki-preview-ending-soon/) before we have to pass the new `required_pull_request_reviews` field in all branch protection calls, otherwise they will stop working when the API becomes official in less than a week

There were also many changes to this API during it's preview period, and it now has functionality to control the dismissal of pull request reviews

I've implemented the extra fields required as well as the new explicit GET/UPDATE/REMOVE "Review Enforcement" client methods.

- [x] Add new dismissal related fields to `BranchProtectionRequiredReviews` and `BranchProtectionRequiredReviewsUpdate`
- [x] Update integration tests for parent branch protection calls to use/validate the new fields
- [x] Implement new explicit `GetReviewRestrictions`, `UpdateReviewRestrictions` and `RemoveReviewRestrictions` client methods\
- [x] Unit tests
- [x] Integration tests
- [x] Rinse and repeat for the `ObservableRepositoryBranchesClient`
---

~~This is on hold until #1598 lands, as that has a decent sized impact on the code~~
~~- [ ] Unit Tests~~
~~- [ ] Integration Tests~~

~~- [ ] Test if @ryangribble is correct [here](https://github.com/octokit/octokit.net/issues/1512#issuecomment-266735950)~~
~~- if yes~~
~~- [ ] Update Contsructors~~
~~- [ ] Update Unit tests~~
~~- [ ] Update integration tests~~


ryangribble:
Looking pretty good so far :+1:

The names are getting pretty long though!  I wonder if instead of `BranchProtectionRequiredPullRequestReviewsUpdate` we can get away with `BranchProtectionRequiredReviewsUpdate` instead.  What do you think?

In terms of the "test whether @ryangribble is right" thing, Im pretty positive it's going to be how I said. 
eg if you look at the webUI below, passing the item as null controls the "require reviews before merging" checkbox, whilst passing the `IncludeAdmins` as true/false controls the Include Administrators checkbox.  If you always pass up the object, how will you ever configure branch protection to NOT require a review?

image

Tests will reveal soon enough I guess 😛

M-Zuber:
> The names are getting pretty long though! I wonder if instead of `BranchProtectionRequiredPullRequestReviewsUpdate` we can get away with `BranchProtectionRequiredReviewsUpdate` instead. What do you think?

`BranchProtectionRequiredReviewsUpdate` sounds good enough to me

I am also pretty sure that you are correct, I just made my initial code based off of my understanding of the docs. I was a bit too daft to go check the UI 😆

M-Zuber:
@ryangribble while I am fixing the naming for the new object, should I shorten the other names, or make that a separate pr?

mkchandler:
@M-Zuber Any updates on this?

M-Zuber:
Thank you for the nudge

Had a rough spot in my personal life, and then my computer needed to be repaved. 
I will do my best to finish this up by the end of the week. 
A decent amount of the code is already written, I just have to polish and push

mkchandler:
@M-Zuber No worries! Life happens, I definitely understand :-)

ryangribble:
As is the case with preview API's there is also the possibility of shifting goal posts! 

Some changes were recently announced to this that will need to be taken into account too 

https://developer.github.com/changes/2017-03-22-protected-branches-admin-improvements/

@m-zuber did you want to push up the code you do have so far?

M-Zuber:
Based on the tests I tried updating, and the UI, any `IncludeAdmin` is ignored if it is not at the top level of the `BranchProtectionSettings`
Should I just make those changes as part of this PR, or make a new one that updates the existing code in master, and then rebase this PR on top of it?

M-Zuber:
To be more accurate, it seems that (as per docs I suppose) if any of the update objects pass in `true` for `includeAdmins` the top level setting is set as `true`

M-Zuber:
@shiftkey or @ryangribble any tips on how to proceed with this?
The main issue is that we need to obsolete all ctors that accept the `include_admins` param, as setting it in one object can be overriden by another, or by the top level param (which still needs to be added)

ryangribble:
Hope you don't mind @M-Zuber but I've taken over this PR 😀

I realized that [we don't have much time](https://developer.github.com/changes/2017-06-16-loki-preview-ending-soon/) before we have to pass the new `required_pull_request_reviews` field in all branch protection calls, otherwise they will stop working when the API becomes official in less than a week

First post has been updated...

ryangribble:
OK this should be good!  All integration tests are passing.  

Just needs a sanity check for typos etc

M-Zuber:
Thanks for picking this one up, and my apologies for dropping the ball

Octokit/Models/Request/BranchProtectionUpdate.cs
```diff
@@ -304,28 +293,146 @@ internal string DebuggerDisplay
...
+        {
+            if (enabled)
+            {
+                Teams = new BranchProtectionTeamCollection();
+                Users = new BranchProtectionUserCollection();
+            }
+            else
+            {
+                // API requires an empty array to be passed, to disable these checks
+                Teams = null;
```
M-Zuber: Is there a way to make it clearer what is happening here?
as a newcomer to the codebase might be confused by `null` being equated with an empty array

M-Zuber:
One little thought, but otherwise LGTM
3oEdva9BUHPIs2SkGk

Octokit/Models/Request/BranchProtectionUpdate.cs
ryangribble: Yep I agree, this is not great but it's the best I could come up with 😡 

Im not sure why this API endpoint is different to the other branch protection ones!  Normally to "disable" a protection setting, you can send a `null` for the object but in this case to disable the item, you have to send an empty bodied member.

So basically to be disabled, as per the following screen shot, we have to send an instance of the class but with no members (ie the Teams/Users array are `null`)
image

To enable the dismiss restriction and have only admins (as per next screenshot) you send empty Teams/Users arrays.
image

To specify teams/users you populate them
image



So given that we need to somehow have either the empty bodied class or empty array or populated arrays, what i came up with was the `ctor` thqat takes a `bool` for whether the restriction is enabled (admin only, thus empty team/user list)... or disabled (`null` arrays, so the class has an empty body).

I tried to explain the usage in the xmldoc comments which users hopefully see in intellisense.

If you have any better ideas for implementation let me know!

M-Zuber: From the perspective of an external user, I think what you have is a nice solution.
My worry is more on someone coming along later trying to contribute, and not understanding the code comment
Maybe something like `API requires an object with empty members to be passed in order to disable these checks`?

ryangribble: Ah right, sorry if misunderstood. Yes that code comment is poorly worded, I will update thanks!

ryangribble:
release_notes:  Implement `RequiredPullRequestReviews` support in `IRepositoryBranchesClient` BranchProtection methods, required when [preview mode ends on 1st September 2017](https://developer.github.com/changes/2017-06-16-loki-preview-ending-soon/)

ryangribble:
advisories: This release contains the necessary Octokit changes to ensure that we send the `required-pull-request-reviews` field which becomes mandatory  when the Protected Branches API [graduates from preview mode](https://developer.github.com/changes/2017-06-16-loki-preview-ending-soon/) on the 1st September

PR #1493, opened 2016-10-13
Add missing `SecurityCritical` attribute

M-Zuber:
Fixes #1471

PR #1528, opened 2017-01-09
Introduce AbuseException

### LINKED ISSUE #1527: Add "retry-after" information to abuse detection exceptions
Offering a suggestion that I'm willing to look at implementing as well.

I was having a little too much fun exploring Github with some concurrent actor processes and tripped the abuse mechanism. I received a `ForbiddenException` with the reason and pointing me to [Github's Abuse Rate Limits Guide](https://developer.github.com/v3/#abuse-rate-limits).

There, I see that a header exists for `Retry-After` -- an integer representing the number of seconds to wait. Would be awesome to be able to retrieve this information so that I could respect it in the event that an abuse-style exception is triggered. 

### Current way
I think I could currently achieve this, catching the `ForbiddenException`, and then getting the headers directly from the `IResponse` passed in as part of the `HttpResponse` field. Is that correct?

### A better way?
Would it make sense to inherit this exception type from Forbidden exception and supply the `Retry-After` value directly in the Exception?

Please let me know if there's interest in this. Thanks!

SeanKilleen:
Resolves #1527.

This aims to Add and `AbuseException` that inherits from `ForbiddenException`. 

This exception will be thrown in place of the `ForbiddenException` when an abuse condition occurs. 

Additionally, the `AbuseException` will contain the `Retry-After` header field as an `int`. 

## Work Remaining
* [x] Create AbuseException
* [x] Tests for AbuseException
* [x] Throw AbuseException correctly within `Connection.cs`
* [x] Ensure AbuseException displays default message when response message is mal-formed.
* [x] Additional tests for `Connection.cs`
* [x] Add `Retry-After` field to AbuseException
* [x] Mark all methods as static that can be (per build failure)
* [x] Add `GetObjectData` for AbuseException? Comment: https://github.com/octokit/octokit.net/pull/1528#issuecomment-271296108

## Feedback to Address
* [x] Re-introduce and skip failing test with notes & pointer to new issue.
* [x] Make `RetryAfterSeconds` into an `int?` and use `null` instead of a default int value.
* [x] Consolidate / compact the private method for parsing the response.
* [x] Fix whitespace issues
* [x] Move test names / formatting toward standard xUnit conventions (nested classes, etc.)
* [x] Remove redundant code from the parse method
* [x] Demonstrate real-world usage

### DIFF
--- a/Octokit.Tests/Exceptions/AbuseExceptionTests.cs
+++ b/Octokit.Tests/Exceptions/AbuseExceptionTests.cs
@@ -12,7 +12,7 @@ namespace Octokit.Tests.Exceptions
             public class Message
             {
                 [Fact]
-                public void PostsAbuseMessageFromApi()
+                public void ContainsAbuseMessageFromApi()
                 {
                     const string responseBody = "{\"message\":\"You have triggered an abuse detection mechanism. Please wait a few minutes before you try again.\"," +
                                                 "\"documentation_url\":\"https://developer.github.com/v3/#abuse-rate-limits\"}";
@@ -54,7 +54,7 @@ namespace Octokit.Tests.Exceptions
                     }
 
                     [Fact]
-                    public void ZeroHeaderValue_RetryAfterSecondsIsZero()
+                    public void WithZeroHeaderValue_RetryAfterSecondsIsZero()
                     {
                         var headerDictionary = new Dictionary<string, string> { { "Retry-After", "0" } };
 
--- a/Octokit.Tests/Exceptions/AbuseExceptionTests.cs
+++ b/Octokit.Tests/Exceptions/AbuseExceptionTests.cs
@@ -1,0 +1,82 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net;
+using System.Text;
+using System.Threading.Tasks;
+using Octokit.Internal;
+using Xunit;
+
+namespace Octokit.Tests.Exceptions
+{
+    public class AbuseExceptionTests
+    {
+        [Fact]
+        public void PostsAbuseMessageFromApi()
+        {
+
+            const string responseBody = "{\"message\":\"You have triggered an abuse detection mechanism. Please wait a few minutes before you try again.\"," +
+                                        "\"documentation_url\":\"https://developer.github.com/v3/#abuse-rate-limits\"}";
+
+            var response = new Response(
+                HttpStatusCode.Forbidden,
+                responseBody,
+                new Dictionary<string, string>(),
+                "application/json");
+            var abuseException = new AbuseException(response);
+
+            Assert.Equal("You have triggered an abuse detection mechanism. Please wait a few minutes before you try again.", abuseException.ApiError.Message);
+        }
+
+        [Fact]
+        public void HasDefaultMessage()
+        {
+            var response = new Response(HttpStatusCode.Forbidden, null, new Dictionary<string, string>(), "application/json");
+            var abuseException = new AbuseException(response);
+
+            Assert.Equal("Request Forbidden - Abuse Detection", abuseException.Message);
+        }
+
+        public class RetryAfterHeaderHandling
+        {
+            [Fact]
+            public void WithRetryAfterHeader_PopulatesRetryAfterSeconds()
+            {
+                var headerDictionary = new Dictionary<string, string>();
+                headerDictionary.Add("Retry-After", "60");
+
+                var response = new Response(HttpStatusCode.Forbidden, null, headerDictionary, "application/json");
+                var abuseException = new AbuseException(response);
+
+                Assert.Equal(60, abuseException.RetryAfterSeconds);
+            }
+
+            [Fact]
+            public void NoHeader_RetryAfterSecondsDefaultsTo60()
+            {
+                throw new NotImplementedException();
+            }
+
+            [Theory]
+            [InlineData(null)]
+            [InlineData("")]
+            [InlineData("    ")]
+            public void EmptyHeaderValue_RetryAfterSecondsDefaultsTo60()
+            {
+                throw new NotImplementedException();
+            }
+
+            [Fact]
+            public void NonParseableHeaderValue_RetryAfterSecondsDefaultsTo60()
+            {
+                throw new NotImplementedException();
+            }
+
+            [Fact]
+            public void NegativeHeaderValue_RetryAfterSecondsDefaultsTo60()
+            {
+                throw new NotImplementedException();
+            }
+        }
+    }
+}
--- a/Octokit.Tests/Exceptions/AbuseExceptionTests.cs
+++ b/Octokit.Tests/Exceptions/AbuseExceptionTests.cs
@@ -7,116 +7,113 @@ namespace Octokit.Tests.Exceptions
 {
     public class AbuseExceptionTests
     {
-        [Fact]
-        public void PostsAbuseMessageFromApi()
+        public class TheConstructor
         {
-            const string responseBody = "{\"message\":\"You have triggered an abuse detection mechanism. Please wait a few minutes before you try again.\"," +
-                                        "\"documentation_url\":\"https://developer.github.com/v3/#abuse-rate-limits\"}";
-
-            var response = new Response(
-                HttpStatusCode.Forbidden,
-                responseBody,
-                new Dictionary<string, string>(),
-                "application/json");
-
-            var abuseException = new AbuseException(response);
-
-            Assert.Equal("You have triggered an abuse detection mechanism. Please wait a few minutes before you try again.", abuseException.ApiError.Message);
-        }
-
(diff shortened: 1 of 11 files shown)


SeanKilleen:
@ryangribble / @shiftkey ready for review at your leisure.

SeanKilleen:
@ryangribble one thing I wanted to check on here. I see a build failure telling me to add an override for `GetObjectData` to the exception. I see that `ApiException` does this.

However, `AbuseException` inherits from `ForbiddenException`, and `ForbiddenException` does not have this override, yet does not throw the error.

Should I:

* Add overrides to both `ForbiddenException` and `AbuseException` as part of this PR?
* Suppress this warning and open an issue to fix later?
* Suppress this warning and fuggedaboudit?

ryangribble:
So to me it looks like you only need to override GetObjectData if you have extra properties on the class.  `ForbiddenException` doesn't have any properties of it's own so it doesn't have to do this, but your class does have an extra property.

So to correctly implement the `[Serializable]` attribute (and satisfy the code analysis CA2240 error), it sounds like you do need to provide a `GetObjectData()` override on `AbuseException`.  And it also needs to be marked as `[SecurityCritical]` due to what was mentioned in #1493 (essentially it needs to match the base class method being overridden)

SeanKilleen:
@ryangribble ahhh, I see. Thanks! This is likely due to the RetryAfterSeconds property, which makes sense. Will get that cleaned up ASAP and then it should be good to g.

ryangribble:
Also regarding the test that you removed because it was failing due to #1529 im thinking so we dont lose the test maybe add it back in and skip it for now.  That way we have a test that captures the problem and a good way to test a fix by unskipping that test?

SeanKilleen:
@ryangribble changes have been made to present the value as a nullable int and default to null when it can't be parsed.

SeanKilleen:
It appears the TravisCI build is failing on the XCode build job -- an error related to `Octokit.StatisticsClient` AFAIK.

Not sure if this is transient or actually caused by something I did (it seems like the former).

Advice? Not sure I can kick off another build manually.

Octokit.Tests/Exceptions/AbuseExceptionTests.cs
```diff
@@ -0,0 +1,124 @@
...
+                {
+                    { "Retry-After", "0" }
+                };
+
+                var response = new Response(HttpStatusCode.Forbidden, null, headerDictionary, "application/json");
+                var abuseException = new AbuseException(response);
+
+                Assert.Equal(0, abuseException.RetryAfterSeconds);
+            }
+
```
ryangribble: Extra whitespace line

Octokit/Exceptions/AbuseException.cs
```diff
@@ -0,0 +1,109 @@
...
+        {
+        }
+
+        [SecurityCritical]
+        public override void GetObjectData(SerializationInfo info, StreamingContext context)
+        {
+            base.GetObjectData(info, context);
+            info.AddValue("RetryAfterSeconds", RetryAfterSeconds);
+        }
+
```
ryangribble: Extra whitespace line

Octokit/Exceptions/AbuseException.cs
```diff
@@ -0,0 +1,109 @@
...
+    /// Containts the amount of seconds after which it's safe to retry the request.
+    /// </summary>
+#if !NETFX_CORE
+    [Serializable]
+#endif
+    [SuppressMessage("Microsoft.Design", "CA1032:ImplementStandardExceptionConstructors",
+        Justification = "These exceptions are specific to the GitHub API and not general purpose exceptions")]
+    public class AbuseException : ForbiddenException
+    {
+        private readonly int? RetrySecondsDefault = null;
```
ryangribble: Missing a whitespace line after this line

Octokit/Exceptions/AbuseException.cs
```diff
@@ -0,0 +1,109 @@
...
+        public AbuseException(IResponse response, Exception innerException)
+            : base(response, innerException)
+        {
+            Debug.Assert(response != null && response.StatusCode == HttpStatusCode.Forbidden,
+                "AbuseException created with wrong status code");
+
+            SetRetryAfterSeconds(response);
+        }
+
+        private void SetRetryAfterSeconds(IResponse response)
```
ryangribble: To me it seems a bit overkill to have a property, default property and 2 private functions, all related to pulling the value out of the headers and parsing it.  I think maintainability would be higher if it was a single function called from the `ctor` that took the `HttpResponse` as input and returned the `int?` parsed value.  It also eliminates the need to declare a `default` value, as the function could just return `null` at the end if all the `TrypGetHeader` `TryParse` and `>= 0` tests didnt satisfy

Im happy to discuss if you feel things are more granularly testable with the current implementation but this was my initial thought...

Octokit.Tests/Exceptions/AbuseExceptionTests.cs
```diff
@@ -0,0 +1,124 @@
+﻿using System.Collections.Generic;
+using System.Net;
+using Octokit.Internal;
+using Xunit;
+
+namespace Octokit.Tests.Exceptions
+{
+    public class AbuseExceptionTests
```
ryangribble: The style/verbage used in the naming of the test fixtures isn't too consistent with (some/most) of the other octokit tests.  

We normally word the test fixture naming with sub classes for the functions eg `TheConstructor` `TheParseRetryHeaderMethod` etc, and then the name of the `[Fixture]` test functions is like `HandlesEmptyHeader` `HandlesNullHeader` `CorrectlyParsesValue` and so on. 

So that way we end up with common english sounding

`AbuseExceptionTests=>TheConstructor=>CorrectlyDeterminesRetryAfterValue` 

and so on

Octokit/Exceptions/AbuseException.cs
SeanKilleen: @ryangribble sure, makes sense. I'll clean it up and compact it. For you to take a second pass.

Octokit.Tests/Exceptions/AbuseExceptionTests.cs
SeanKilleen: @ryangribble makes sense. I'll reformat. Sorry, should've paid more attention to the existing conventions. They tend to be my preference anyway; I think I was likely being hasty.

ryangribble: > should've paid more attention to the existing conventions. 

We have a fairly large code base and some of it has not needed to be touched for quite a while, so it's certainly not all consistent! As we go forward (and as we work with new contributors) we are taking a greater focus on ensuring consistency at least in anything new we implement, so thanks for jumping through these hoops with me :+1:

SeanKilleen:
@ryangribble mind giving the code a second look? Should be much more compact now.

I inverted some if statements which led to some negatives inside the ifs, but I don't think it compromised on clarity.

Octokit/Exceptions/AbuseException.cs
```diff
@@ -0,0 +1,89 @@
...
+                "AbuseException created with wrong status code");
+
+            RetryAfterSeconds = ParseRetryAfterSeconds(response);
+        }
+
+        private static int? ParseRetryAfterSeconds(IResponse response)
+        {
+            string secondsValue;
+            if (!response.Headers.TryGetValue("Retry-After", out secondsValue)) { return null; }
+            if (string.IsNullOrWhiteSpace(secondsValue)){ return null; }
```
ryangribble: I think this line is redundant since the following `int.TryParse()` will handle `null` or `whitespace` strings and return `false`, right?

SeanKilleen: @ryangribble nice catch; fixed.

SeanKilleen:
@ryangribble OK, I think we should be good to go now? Give it another pass.

ryangribble:
OK yep this is all looking 💎 

One final question - have you actually managed to get a runtime test performed, just to try and ensure that the exception was correctly thrown and the retry value was present and parsed?  Or is it too hard to actually trigger the abuse conditions?

SeanKilleen:
Oh, I can trigger it. :)

Stand by. I'll build & drop this into a project and send some screenshots / output.

SeanKilleen:
@ryangribble OK, I built my branch locally. I then referenced it directly from test project. This project creates 10 actors in parallel issuing github requests for the same auth token, thus triggering an abuse mechanism. 

Here it is in the logs throwing the `Octokit.AbuseException`:

> image

Then I had the actor's supervisor detect an `AbuseException` and log out to the console the number of seconds before trying again -- this shows that `RetryAfterSeconds` is performing as expected: 

> image

ryangribble:
Thanks for that 😀 

I think we're done here!
LGTM

SeanKilleen:
@ryangribble thanks for the guidance along the way!

PR #1530, opened 2017-01-09
Add docs snippet for label coloring

### LINKED ISSUE #790: Colors
I know that the official api returns colors without the leading `#` but what do you think about adding it in? In most use cases in `C#` it is useful to necessary to have it

SeanKilleen:
Resolves #790.


M-Zuber:
👍 Thank you!
full-house-12SXVd8bmXdSg0

PR #1536, opened 2017-01-11
Add Missing Event Types

lynnfaraday:
Fixes Issue #1502.

Updates the EventInfo list to the latest values published in [GitHub API](https://developer.github.com/v3/issues/events/), plus an undocumented one we stumbled across.

PR #1504 is a much better long-term solution, but in the mean time this is blocking us from parsing Git diffs using OctoKit.  Thought we'd share the fix.

### DIFF
--- a/Octokit/Models/Response/EventInfo.cs
+++ b/Octokit/Models/Response/EventInfo.cs
@@ -194,7 +194,7 @@ namespace Octokit
         ReviewRequestRemoved,
 
         /// <summary>
-        /// Not documented in GitHub API.
+        /// Base branch of the pull request was changed.
         /// </summary>
         BaseRefChanged,
 
--- a/Octokit/Models/Response/EventInfo.cs
+++ b/Octokit/Models/Response/EventInfo.cs
@@ -152,10 +152,10 @@ namespace Octokit
         /// </summary>
         Unlocked,
 
-        /// <summary>
-        /// The pull request’s branch was deleted.
-        /// </summary>
-        HeadRefDeleted,
+		/// <summary>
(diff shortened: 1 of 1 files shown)


Octokit/Models/Response/EventInfo.cs
```diff
@@ -179,6 +179,26 @@ public enum EventInfoState
...
+        /// </summary>
+        ReviewDismissed,
+
+        /// <summary>
+        /// The actor removed the review request for the subject on this pull request.
+        /// </summary>
+        ReviewRequestRemoved,
+
+        /// <summary>
+        /// Not documented in GitHub API.
```
ryangribble: I think we can advise upstream github support this value is missing from their docs, so rather than saying this I think we can just give it an actual description? 

Eg `Base branch of the pull request was changed`

ryangribble: Cool I was able to do this myself!  Thanks 😀

ryangribble:
Ive edited that comment myself since you had allowed contributors to push to your branch :grinning: and Ive also notified github support of the undocumented value so they can consider updating the docs

Thanks @lynnfaraday and congratulations on your first contribution to octokit.net 🎉 

LGTM

lynnfaraday:
@ryangribble Oh cool, thanks for fixing that.  Beat me to it :)

ryangribble:
(just waiting to get Travis CI happy before hitting the button)

ryangribble:
release_notes: Fix `IssueTimelineClient` deserialization exceptions by adding new `EventInfoState` values

Issue #1514, opened 2016-12-12
Organization Membership Preview API Changes - existing endpoints


ryangribble:
As per this [blog post](https://developer.github.com/changes/2016-11-28-preview-org-membership/) several changes are in preview for the Organization Membership API

This issue is to cover changes to existing endpoints

- The `default_repository_permission` and `members_can_create_repositories` settings for an organization can now be retrieved and set through the [Get an organization](https://developer.github.com/v3/orgs/#get-an-organization) and [Edit an organization endpoints](https://developer.github.com/v3/orgs/#edit-an-organization).

- The [Create team](https://developer.github.com/v3/orgs/teams/#create-team) endpoint now offers the `maintainers` parameter which will allow you to add team maintainers upon creation.

- You can now use the `affiliation` parameter with the [List collaborators](https://developer.github.com/v3/repos/collaborators/#list-collaborators) endpoint to filter a repository's collaborators by their affiliation type.

NOTE: A new preview accepts header needs to be specified `application/vnd.github.korra-preview+json` in `AcceptHeaders.cs`

alfhenrik:
And while I'm in the area, I'd like to put my face on this one too

hnrkndrssn:
> You can now use the `affiliation` parameter with the List collaborators endpoint to filter a repository's collaborators by their affiliation type.

There is already a `CollaboratorRequest` class that is used when creating/updating a collaborator to set their permissions and I don't think I can re-use this class for the `affiliation` parameter (looking at the code it'd add all the properties of the class to the query string)? 

Should I add a new class `ListCollaboratorRequest`(?) with the `Affiliation` property and use that for the List collaborators endpoint?

ryangribble:
Yeah it looks like the naming conventions have been a bit out of whack with this one.  "normally" we would have `XxxRequest` on a `GetAll` endpoint, and `NewXxx`/`UpdateXxx` on any create/update endpoints... but in this case the `CollaboratorRequest` name has aleady been used 😬 

So yeah I would say :+1: to this
> Should I add a new class ListCollaboratorRequest(?) with the Affiliation property and use that for the List collaborators endpoint?

PR #1537, opened 2017-01-12
add Id for PullRequest.cs

YunLi1988:
added the Id property for PullRequest model

### DIFF
--- a/Octokit.Tests.Integration/Clients/PullRequestsClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/PullRequestsClientTests.cs
@@ -59,6 +59,7 @@ public class PullRequestsClientTests : IDisposable
 
         Assert.Equal(1, pullRequests.Count);
         Assert.Equal(result.Title, pullRequests[0].Title);
+        Assert.NotNull(pullRequests[0].Id);
     }
 
     [IntegrationTest]
--- a/Octokit.Tests.Integration/Clients/PullRequestsClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/PullRequestsClientTests.cs
@@ -59,7 +59,7 @@ public class PullRequestsClientTests : IDisposable
 
         Assert.Equal(1, pullRequests.Count);
         Assert.Equal(result.Title, pullRequests[0].Title);
-        Assert.NotNull(pullRequests[0].Id);
+        Assert.True(pullRequests[0].Id > 0);
     }
 
     [IntegrationTest]
--- a/Octokit/Models/Response/PullRequest.cs
+++ b/Octokit/Models/Response/PullRequest.cs
@@ -49,6 +49,7 @@ namespace Octokit
         }
 
         /// <summary>
+        /// The internal Id for this pull request (not the pull request number)
         /// </summary>
         public long Id { get; protected set; }
 
--- a/Octokit/Models/Response/PullRequest.cs
+++ b/Octokit/Models/Response/PullRequest.cs
@@ -15,7 +15,7 @@ namespace Octokit
             Number = number;
         }
 
-        public PullRequest(int id, Uri url, Uri htmlUrl, Uri diffUrl, Uri patchUrl, Uri issueUrl, Uri statusesUrl, int number, ItemState state, string title, string body, DateTimeOffset createdAt, DateTimeOffset updatedAt, DateTimeOffset? closedAt, DateTimeOffset? mergedAt, GitReference head, GitReference @base, User user, User assignee, IReadOnlyList<User> assignees, bool? mergeable, User mergedBy, int comments, int commits, int additions, int deletions, int changedFiles, Milestone milestone, bool locked)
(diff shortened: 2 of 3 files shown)


Octokit/Models/Response/PullRequest.cs
```diff
@@ -48,6 +49,11 @@ public PullRequest(Uri url, Uri htmlUrl, Uri diffUrl, Uri patchUrl, Uri issueUrl
         }
 
         /// <summary>
+        /// The universal Id (different from the pull request number) this pull request.
+        /// </summary>
+        public int Id { get; protected set; }
```
ryangribble: It would be good to use `long` data type, since github could well have more than "max int" pull requests!

YunLi1988: Got it. You did suggest long. I checked other class like issue, comment, they all using int type id. used int for consistency issue. As you mentioned we have more pullrequest than int can support already, I am going to change it to long again. 

BTW, one of the check Travis CI build failed. I tried to read the log, but did not found any intuitive information. Do you mind to share me some tips on looking at that log?

YunLi1988: Another thing is when I run the integration test, there is an authorizationException : bad credentials been thrown. Does that mean I need to change certain credentials to run the tests or I did not set some configurations right?

Octokit/Models/Response/PullRequest.cs
```diff
@@ -48,6 +49,11 @@ public PullRequest(Uri url, Uri htmlUrl, Uri diffUrl, Uri patchUrl, Uri issueUrl
         }
 
         /// <summary>
+        /// The universal Id (different from the pull request number) this pull request.
```
ryangribble: Can we please reword this to "The internal Id for this pull request (not the pull request number)." ?

And could you edit the comment on `Issue.Id` to be similar?  "The internal Id for this issue (not the issue number)."

Thanks 😀

Octokit/Models/Response/PullRequest.cs
ryangribble: In terms of running the integration tests, you need to configure your development environment (and is strongly recomended you setup a separate github account to use for the integration tests).  More info is documented [here in the contributors guide](https://github.com/octokit/octokit.net/blob/master/CONTRIBUTING.md#slow-tests)

ryangribble: And in terms of the TravisCI build sometimes we get an unrelated problem where it blows up in TravisCI and is not related to your changes.  We have to kick off another build (or the next commit pushed will obviously trigger a new build as well).  I see your latest commit, the builds have all passed :+1:

YunLi1988:
I am not familiar with git hub. somehow I cannot find the comments I received in my email.

Once you are done with this one, can you assign another entrylevel task?

Octokit.Tests.Integration/Clients/PullRequestsClientTests.cs
```diff
@@ -59,6 +59,7 @@ public PullRequestsClientTests()
 
         Assert.Equal(1, pullRequests.Count);
         Assert.Equal(result.Title, pullRequests[0].Title);
+        Assert.NotNull(pullRequests[0].Id);
```
ryangribble: This `int` field can never be `null` so I'm thinking we will need to assert something else here (eg `Assert.True(pullRequests[0].Id > 0)`

YunLi1988: Thanks for the catch. I am going to update that. I am assuming all internal ids  are larger than 0.

ryangribble:
> I am not familiar with git hub. somehow I cannot find the comments I received in my email.

If the comments were on some lines of code that have since been replaced, then github collapses it into a "outdated" section.  You can click that to see the comments again:

image

image


> Once you are done with this one, can you assign another entrylevel task?

That's awesome!  All of our issues with a tag of "up for grabs" are meant to be easy enough for new contributors to tackle.  You can choose anything that takes your fancy but if you really want a suggestion how about https://github.com/octokit/octokit.net/issues/1514 😀   When you find an issue you want to work on, Just comment on that issue that you want to take a look at it (this makes sure we dont have multiple people trying to fix the same thing!).  Create a branch on your fork of the repo (I notice you used your master branch for this PR, instead you should create a branch and do the work there, so your master branch can just be a identical copy of this repo's master branch) push that branch up and raise a pull request.  Put [WIP] ( meaning work in progress) in the PR title, and we can then work together as you make the change.  It's totally fine to raise the [WIP] pull request early in the process so we can collaborate rather than waiting until you've done lots of coding!

ryangribble:
Many thanks @YunLi1988 for this enhancement, and congratulations on your first octokit.net contribution!

ryangribble:
PS: I noticed the commits on this PR are from a different email address than your github profile, so they arent linked together.  You should [add this email address](https://github.com/settings/emails) to your profile so the commits are associated with your github account!  👍

PR #1540, opened 2017-01-15
Update ApiErrorMessageSafe to return null for empty and whitespace strings

### LINKED ISSUE #1529: Exceptions don't display default exception messages when ApiError.Message is empty
Currently, many exceptions take `ApiErrorMessageSafe` and display the default text if it's null.

However, this should probably also apply if `ApiErrorMessageSafe` is empty so that the user gets something.

I came across this while attempting to write out tests for #1528. I didn't want to change too many things in that PR, so figured I'd create this as an off-shoot.

## A failing test
Removed this from my previous PR. Expected the default error message (e.g. generic "Request Forbidden") -- actual was an empty string. This was part of `ConnectionTests.cs` in #1528 before I ripped it out:

```csharp
[Fact]
public async Task ThrowsAbuseExceptionWithDefaultMessageForUnsafeAbuseResponse()
{
	string messageText = string.Empty;

	var httpClient = Substitute.For<IHttpClient>();
	IResponse response = new Response(
		HttpStatusCode.Forbidden,
		 "{\"message\":\"" + messageText + "\"," +
		"\"documentation_url\":\"https://developer.github.com/v3/#abuse-rate-limits\"}",
	   new Dictionary<string, string>(),
		"application/json");
	httpClient.Send(Args.Request, Args.CancellationToken).Returns(Task.FromResult(response));
	var connection = new Connection(new ProductHeaderValue("OctokitTests"),
		_exampleUri,
		Substitute.For<ICredentialStore>(),
		httpClient,
		Substitute.For<IJsonSerializer>());

	var exception = await Assert.ThrowsAsync<AbuseException>(
		() => connection.GetResponse<string>(new Uri("endpoint", UriKind.Relative)));

	Assert.Equal("Request Forbidden - Abuse Detection", exception.Message);
}
```

## Potential Fixes
* The fix for this would likely be updating the `Message` property of the affected exceptions to do an `IsNullOrWhitespace` check on `ApiErrorMessageSafe`. 
* Or, `ApiErrorMessageSafe` could be changed to return null if the Message string is empty or whitepsace, rather than just returning the message property itself.

SeanKilleen:
Resolves #1529.

## Background
If an API Error is returned and the message is empty, the default error message is not shown, and instead an empty string is shown.

##Fix
Per discussion on #1529, we'll resolve this by modifying ApiErrorMessageSafe to return null when an empty or whitespace string is provided.

## Work
* [x] Un-skip the failing test
* [x] Modify ApiErrorMessageSafe code

### DIFF
--- a/Octokit.Tests/Http/ConnectionTests.cs
+++ b/Octokit.Tests/Http/ConnectionTests.cs
@@ -349,7 +349,7 @@ namespace Octokit.Tests.Http
                 Assert.Equal(45, exception.RetryAfterSeconds);
             }
 
-            [Fact(Skip = "Fails due to https://github.com/octokit/octokit.net/issues/1529. The message is empty but the default message isn't displayed. However, this isn't due to the introduction of AbuseException, but rather something else.")]
+            [Fact]
             public async Task ThrowsAbuseExceptionWithDefaultMessageForUnsafeAbuseResponse()
             {
                 string messageText = string.Empty;
--- a/Octokit/Exceptions/ApiException.cs
+++ b/Octokit/Exceptions/ApiException.cs
@@ -178,7 +178,12 @@ namespace Octokit
         {
             get
             {
-                return ApiError != null ? ApiError.Message : null;
+                if (ApiError != null && !string.IsNullOrWhiteSpace(ApiError.Message))
+                {
+                    return ApiError.Message;
+                }
+
+                return null;
             }
         }
 


SeanKilleen:
@ryangribble ready for your review.

I couldn't think of too many more tests that needed adding, and `ApiErrorMessageSafe` is already covered by 50+ tests so I think we're good there.

ryangribble:
This is good, thanks @SeanKilleen!

Not sure what is up with AppVeyor, all our builds seem to be stuck "queued" today.  Will see if it's fixed by tomorrow

ryangribble:
release_notes: Unparseable ApiErrors should now fall back to better default error messages

ryangribble:
LGTM

PR #1541, opened 2017-01-16
Add new AccountType "Bot" to prevent deserialization errors

ryangribble:
The GitHub API is apparently now returning "Bot" account types.  This was causing a deserialization exception when encountering repositories that had a "bot" user (perhaps related to the new API integrations features) performing actions.  Example here: https://github.com/depfutest/devfu/issues/2


ryangribble:
🤔 I initially fixed this as a commit on the release branch while fixing a broken integration test...  But then I thought this should probably be in release notes which means I needed it on a PR 😀

PR #1544, opened 2017-01-21
[.NET Core] - additional work

mderriey:
Hey,

I started doing a bit of work to make the .NET Core solution tidier and on par with the existing one.

This includes:

 - remove anything platform-specific that we don't use anymore
 - bring back the StyleCop rule set to both `Octokit` and `Octokit.Reactive`
 - a build task running the LINQPad samples before packaging, as I think @shiftkey is keen on keeping them up-to-date and working
 - a build task to format code

There's some not-so-nice things going on in the formatting task. `CodeFormatter` doesn't support xproj/project.json, so the task creates a temp `.csproj` listing all the `.cs` files and feeds it to the formatter.

Let me know if you'd like things made different.


mderriey:
Closing this PR as we decided to break them into several PRs so we don't end up with unrelated work in a PR.

New issues were created: #1546, #1547, #1548.

PR #1549, opened 2017-01-24
Remove unneeded files for .NET Core

### LINKED ISSUE #1546: [.NET Core] - Remove platform-specific/unneeded files
As we're shifting to `netstandard`, we can now safely remove platform-specific artifacts that were left before #1503 was merged. Some files are also not needed anymore, like `.nuspec` files since NuGet packages metadata is now described in the `project.json` files.

mderriey:
Close #1546


ryangribble:
Considering the builds are passing this must be good 😀 

While looking through this I noticed that in the Octokit.ruleset file is this comment 
> This ruleset only includes the rules we care about. I'll be adding new ones as we fix our codebase.

But if Im understanding the actual ruleset file correctly, I think actually we are actually opting in to ALL rules except 7 specific exemptions.  If that's correct, perhaps while we are 👀 at this part of the project we can update that comment.  Something like "This ruleset enables all of the Microsoft CodeAnalysis rules, with a few deliberate exceptions".

Thoughts?

mderriey:
Opening the ruleset file in VS 2015 shows you're right. All rules should raise an error, except the 7 we explicitly disabled.

ryangribble:
LGTM

PR #1551, opened 2017-01-24
Add a build task to validate LINQPad samples

### LINKED ISSUE #1548: [.NET Core] - Add CAKE task to run LINQPad samples
As we shifted from FAKE to CAKE in #1503, we need to port the task that runs LINQPad samples against the built code to make sure we don't break them/keep them up-to-date.

mderriey:
Closes #1548


ryangribble:
Ive tested this and it's :gem:

In terms of consistency with the older FAKE builds, this task was called `ValidateLINQPadSamples` rather than `RunLinqPadSamples`.  Do you think we should rename it?  If not then can you double check there are no docs anywhere that need to be updated?

Also, from an OCD point of view (and I realise it was like this under FAKE anyway!) it would be nice if the samples were done in logical 1,2,3...10 order rather than alphabetical (1,10,2,3).  Or do you think it would be too ugly to add an `.OrderBy()` lambda to the query (considering it has to split the string, check if it's numeric, etc etc)? 😛  Or perhaps we could just rename the sample files to be `01-xxxxxx`.  Thoughts?

mderriey:
I'm happy to rename it to what it was before. *Validate* is much closer to what the task is doing, as running them doesn't add any value per-se.

For the samples names, I'd lean towards renaming the files so they all have two digits - so `01-xxxxx`. It keeps the task focused on executing the samples and it's something we can check in the future in PRs if we add some more.

ryangribble:
Love your work! 

Just waiting to see what is up with travisCI builds...

mderriey:
Hmm, I remember adding a command to the TravisCI config for macOS to link openssl libs where the `dotnet` CLI was looking for them. Now it fails saying the files already exist, so maybe they fixed it. I removed it and we'll see what happend 😉

ryangribble:
Ah nice, I couldnt see anything that correlated with Travis failures but as you say I guess they fixed their OSX image and the workaround is now no longer required

This is ready to 🚢

ryangribble:
LGTM

PR #1553, opened 2017-01-27
add membership in an organization - feature added

### LINKED ISSUE #1533: Missing Organization Membership API
Hi, in the OrganizationMembersClient, you have got the Delete task/function. However I think you the Add/Invite task/function is missing.

guilhermecorsino:
This commit fixes #1533

### DIFF
--- a/Octokit.Reactive/Clients/IObservableOrganizationsClient.cs
+++ b/Octokit.Reactive/Clients/IObservableOrganizationsClient.cs
@@ -94,5 +94,18 @@ namespace Octokit.Reactive
         /// <exception cref="AuthorizationException">Thrown if the client is not authenticated.</exception>
         /// <returns>A <see cref="Organization"/></returns>
         IObservable<Organization> Update(string organizationName, OrganizationUpdate updateRequest);
+
+
+        /// <summary>
+        /// Adds a <see cref="User"/> to a <see cref="Organization"/>.
+        /// </summary>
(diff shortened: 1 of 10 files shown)


ryangribble:
Hi @guilhermecorsino many thanks for taking up this feature enhancement!

The progress looks good so far!  

A couple of comments:

If we look at the official [API docs](https://developer.github.com/v3/orgs/members/#add-or-update-organization-membership) I can see that the request to "Add or Update Organization Membership" actually takes a parameter to indicate the role the user should have (eg Mmber, Admin).  So that means you would create a new Request model object (eg `OrganizationMembershipUpdate`) that has a `Role` property of that enum type.

And looking at the response model returned, currently you have this defined as a single `enum` however it is actually a response object that includes `State`, `Role`, `Oragnization` and `User` fields.  So you will actually want your `OrganizationMembership` be a response class that has these fields on it, and rename your enum to something like `OrganizationMembershipStatus`

If you add some integration tests it will be easy to test this call against the real github.com - make sure you follow the details in our contributors section about how to configure the integration test settings (and of course, make sure you have setup a test github account and organization so you arent using your real account, if you havent already)

Let me know if you want more detailed help or if you are happy to have a crack at addressing these. :+1:

guilhermecorsino:
Hi @ryangribble ! 
I've got it and I'm going to start with these alters, thank you so much.   👍👍👍

Issue #1545, opened 2017-01-21
Create a draft release when building a PR from a release branch


mderriey:
As per the current release process, maintainers create a `release` branch to tidy things up, get release notes, etc...

Now that the .NET Core version of Octokit uses CAKE, we can automate the creation of this draft release. We can also take advantage of GitVersion if the release branch is named `release/{version}` where `{version}` is `{major.minor.patch}`.

The goal is for the build to verify is a matching draft release exists, and create it if it doesn't.

PR #1555, opened 2017-01-29
add GitVersion configuration file

### LINKED ISSUE #1554: [.NET Core] - add GitVersion config file
This will have the benefit of making the configuration and behaviour explicit.

mderriey:
Closes #1554.

Right now we use GitVersion with the default settings.

`master`, by default, uses the [Continuous Delivery](https://gitversion.readthedocs.io/en/latest/reference/continuous-delivery/) mode which generates the same version for all the commits until we tag the repo with that version.

Since we might want to publish pre-release packages in the future, we could take advantage of the other mode named [Continuous Deployment](https://gitversion.readthedocs.io/en/latest/reference/continuous-deployment/). This mode generates a new version for each commit. To avoid the version going through the roof, a pre-release tag is forced with this mode. By default, the tag for `master` is `ci`. As we used the `alpha` one for the pre-release NuGet package, I thought we might use the same.

Also, right now GitVersion increments the `minor` after the repo has been tagged (see [that AppVeyor build](https://ci.appveyor.com/project/github-windows/octokit-net/build/0.24.1-dotnetcore.1+4.build.1969)). It looks more natural to me to go with the assumption that the next release will contain a new feature. We can always force a bugfix release by using either a release branch - `release/0.24.1` - or use the convention-based commit messages - `+semver: minor` - to instruct GitVersion of the version we want it to generate.

At the moment the file contains configuration for both `master` and `dotnetcore` since it's the branch we're working on. When the branch gets merged into `master` we can get rid of `dotnetcore`

### DIFF
--- a/GitVersion.yml
+++ b/GitVersion.yml
@@ -1,0 +1,11 @@
+branches:
+  master:
+    mode: ContinuousDeployment
+    increment: Minor
+    tag: alpha
+  dotnetcore:
+    mode: ContinuousDeployment
+    increment: Minor
+    tag: alpha
+ignore:
+  sha: []
--- a/GitVersion.yml
+++ b/GitVersion.yml
@@ -2,7 +2,7 @@ branches:
   master:
     mode: ContinuousDeployment
     increment: Minor
-    tag: alpha
+    tag: beta
   dotnetcore:
     mode: ContinuousDeployment
     increment: Minor


ryangribble:
Awesome stuff @mderriey 

So going forward I'd like to have each pull request merged to master automatically push a pre release package to nuget. I think we could use beta tag in this case rather than alpha (which I used for the dotnetcore stuff since it was pretty experimental and also not in master yet) 

Mapping things out as to how I could see this working, do you think this sounds good? 

0.24 last release 
Merge a PR to master : 0.25.0-beta0001 to nuget 
Merge next PR to master: 0.25.0-beta0002 to nuget 
Merge another PR to master: 0.25.0-beta0003 to nuget 
Create release branch release/0.25.0: create draft release in repo (and nothing to nuget yet) 
Update release notes, tweak tests etc on release branch: update draft release with latest release notes (nothing to nuget)
Publish draft release (tags master): 0.25.0 to nuget 
Merge PR to master: 0.26.0-beta0001 to nuget 

etc

mderriey:
@ryangribble `beta` sounds fine!

As to the workflow you described, it matches what I had in mind, too.  
Maybe we don't need to go into that much detail right now, but with the sentence *Update release notes, tweak tests etc on release branch: update draft release with latest release notes (nothing to nuget)*, does it mean the CAKE build should update the draft release? If so, maybe we can discuss that outside of this PR, but I'd be keen to know more about it.

I'll make the change straigth away.

ryangribble:
Yeah we can discuss the release creation/updating etc on #1545 

In terms of this change though, does the configuration achieve the numbering that I mentioned in my previous comment?

mderriey:
Yes, it will. the `increment: minor` will increment the minor version by default, so after a tag `v0.24.0`, the next version will be `0.25.0`. Since we use `ContinuousDeployment`, a new version will be created after every commit, but the new version will affect the pre-release tag, not the minor.

I hope this makes sense.

ryangribble:
Yep, I'm guessing we have to tag the repo with -beta0001 in order for the next merge to master to get - beta0002 and so on? Not that it's a problem, just that we currently don't tag

mderriey:
Ah, sorry, I must not be clear enough.

There's no need to tag the repo. The reason is that `ContinuousDeployment` always uses pre-release tags when it calculates a version. It will get the latest base version - most probably from a tag - then applies the increment specified - in our case Minor - then uses the specified pre-release tag so that each commit has a different version.

The thing to keep in mind is that the increment will be applied only once between two tags.

In our case:

 - `master`, tagged with `v0.24.0`
 - a PR is merged to `master`; the base version comes from the tag, it's `0.24.0`, the increment is Minor, so the new version is `0.25.0`. but as we use `ContinuousDeployment`, we apply the `beta` tag, so final version is `0.25.0-beta0001`
 - another PR is merged to `master`. all the steps from above are exactly the same, except the final version is `0.25.0-beta0002` because it's the second commit since the tag
 - another PR to `master`. same thing, version is now `0.25.0-beta0003`

Does this make things clearer?

ryangribble:
Ok yeah so I was wondering how it kept track of the 1,2,3 if there was nothing in the repo 😀  so it counts the commits, clever! Does that mean if a PR was merged instead of squashed (thus there might be 5 commits to master rather than 1) it would jump from beta0001 to beta0006?

mderriey:
I just tried, and yes, it will be incremented by how many commits the PR branch contains. I hadn't thought about that. Is this a problem? I saw you squashed my previous PRs, do you do this for PRs that target `master`, too?

ryangribble:
Yeah we squash PR's normally since it keeps the history cleaner. I think it's ok 👌

PR #1556, opened 2017-02-11
fix .sln file for linux

ArsenShnurkov:
paths in .sln file should match actual file names


ryangribble:
Hi @ArsenShnurkov 

Thanks for the correction 😀 

I may as well merge this (eventhough all these projects will be going away once we merge the `dotnetcore` branch)

PR #1562, opened 2017-03-09
Add merge_commit_sha to PullRequest model

alexperovich:


ryangribble:
Hi @alexperovich 

There is a bit of history behind this field, but it looks like it has been changed again (without an api change notification being posted on the api blog).

Originally, this field did not actually include the merge commit of a PR but was the internal "test commit" of the PR merged with the base branch, and wasnt updated to the resultant SHA when thr PR was actually merged.  It was then deprecated from the github API, and we [removed it from octokit.net](https://github.com/octokit/octokit.net/commit/2da609a724a46cad13625cc22934b5edee6eaa9d) to avoid confusion.

This was the deprecation comments back then, from upstream:
```
The merge_commit_sha attribute holds the SHA of the test merge commit; however, this attribute is deprecated and is scheduled for removal in the next version of the API. The Boolean mergeable attribute will remain to indicate whether the pull request can be automatically merged.
```

But looking at the API docs now, it appears this field is BACK and actually does contain a useful value! 🎉 

https://developer.github.com/v3/pulls/#get-a-single-pull-request
```
Each time the pull request receives new commits, GitHub creates a merge commit to test whether the pull request can be automatically merged into the base branch. (This test commit is not added to the base branch or the head branch.)

The value of the merge_commit_sha attribute changes depending on the state of the pull request. Before a pull request is merged, the merge_commit_sha attribute holds the SHA of the test merge commit. After a pull request is merged, the attribute changes depending on how the pull request was merged:

- If the pull request was merged as a merge commit, the attribute represents the SHA of the merge commit.
- If the pull request was merged via a squash, the attribute represents the SHA of the squashed commit on the base branch.
- If the pull request was rebased, the attribute represents the commit that the base branch was updated to.
```

I'm happy to merge this change and get the field added again, however could you please update the field comment to contain a synopsis of the above information.  My suggestion (not particularly elegant, if you can do better!) would be something like:

```
/// <summary>
/// The test/merge/squash/rebase commit hash, depending on the state of the pull request and how it was merged
/// </summary>
```

ryangribble:
Nailed it!

LGTM

Thanks @alexperovich for picking this up 👍

ryangribble:
release_notes: Add support for the newly resurrected `PullRequest.MergeCommitSha` property

PR #1564, opened 2017-03-13
[StringExtensions] - Display correct variable name when throwing.

ivandrofly:


ryangribble:
Thanks @ivandrofly for the correction... Out of interest did you actually run into a situation where exceptions were being thrown by these extension methods?

ivandrofly:
> did you actually run into a situation where exceptions were being thrown by these extension methods?

Nope, just trying to be consistent with methods present inside **StringExtensions.cs** :)

ryangribble:
Whoops forgot about this one!

Cheers @ivandrofly

PR #1567, opened 2017-03-15
Migrate dotnetcore to vs2017 tooling

ryangribble:
Now that VS2017 is released, the time seems right to get our `dotnetcore` branch over to that tooling, before we actually merge and release it.  SourceLink support for dotnetcore is also now possible when using the msbuild tooling (thanks to @ctaggart  :love:) which is great as we've always included sourcelink in our libraries but dropped it when we did our alpha release for dotnetcore, will be great to have it back!

### Working
- [x] Projects converted to VS2017 csproj format
- [x] CAKE build script updated to the latest cake.frosting versions and remodelled based on the new official "dotnet build templates" for cake frosting, using [their own reworked build](https://github.com/cake-build/frosting/tree/develop/build) as inspiration 😉 
- [x] Builds, Tests and build script are working OK, with the exception of the nuget package versioning (mentioned below)

### TODO
- [x] nuget pack step is getting version 1.0.0 - need to figure out the best way for the GitVersion calculated version numbers to be used.  Someoptions appear to be
 -- add nuget package information/fields to the csproj files and have a build task that replaces the correct version numbers (like we used to patch assemblyinfo files and project.json files)
-- see if we can pass commandline parameters to the DotNetCorePackage command
-- use the GitVersion msbuild target inside our csproj so the versioning stuff happens magically, rather than us explicitly running GitVersion and using the output values ourself
-- disable automatic AssemblyInfo generation via an msbuild property in csproj files, and maintain/generate our own AssemblyInfo files

- [ ] ~~Implement sourcelink 2.0 support and change our msbuild setup to embed sourcelinked portable PDBs into the assemblies.  This is essentially the adoped "best practice" which is being rolled out across the official framework.  See [here](https://github.com/ctaggart/SourceLink#enable-source-link-support) for the doco, and [here](https://github.com/aspnet/Mvc/pull/5905) for an example PR showing how it is being added to the `aspnet/Mvc` project~~
To be done on separate PR

- [ ] Get TravisCI linux/mac builds working, or remove it from builds going forward

@mderriey can you please validate everything is working OK for you, and feel free to help tackle the TODO items!


mderriey:
Sweet! Will definitely take a loot at it.

I also think xUnit now supports cross-targetting test projects, so we could run them in the `net45` and `netstandard1.1` contexts.

Thanks for that.

ryangribble:
@patriksvensson @gep13 did you guys have any guidance on "the best" way to get the DotNetCorePack Cake task to use the desired version number on the nuget package.  As mentioned in the PR there are a number of possible ways, interested in what "the standard" should be

patriksvensson:
@ryangribble Here is an example of setting the desired version number with the new dotnet SDK (1.0.0).

```csharp
DotNetCorePack(project.FullPath, new DotNetCorePackSettings 
        {
            Configuration = config,
            OutputDirectory = "./.artifacts",
            VersionSuffix = version.Suffix,
            NoBuild = true,
            Verbose = false,
            ArgumentCustomization = args => args
                .Append("/p:Version={0}", version.GetSemanticVersion())
                .Append("--include-symbols --include-source") // For symbols
        });
```

patriksvensson:
@ryangribble I also noticed that you do not include a `NuGet.config` in your `build` directory which you should since Cake.Frosting only exists on MyGet at the moment.

Example here: https://github.com/cake-build/frosting/blob/develop/build/nuget.config

ctaggart:
Once you upgrade to the MSBuild 15 tooling, adding in SourceLink v2 is a piece of cake. I recommend it just being a follow-up pull request. Ping me when this is merged and I can assist. Rx.NET just merged source link support and is [a good example](https://github.com/Reactive-Extensions/Rx.NET/pull/338/files).

ryangribble:
@patriksvensson thanks for advising to go with option 2 for the dotnet core pack 😀 Although it seems when providing the full version via the `/p:Version=` argument, there isn't any need to also keep the `VersionSuffix` property.

We actually [do have a nuget.config already](https://github.com/octokit/octokit.net/blob/dotnetcore/build/nuget.config), as we have been using cake.frosting for a while on our previous project.json setup... perhaps you were looking at our `master` branch rather than `dotnetcore` branch?

Thanks for the help and again - thanks for cake.frosting it's been going well for us :+1:


@ctaggart thanks for the offer of assistance, as you say it's super easy to add sourcelink 2.0 support with the new msbuild tooling - pretty much just plan on copying one of the other PRs you've already done, but we'll certainly let you know if we can't sort it!

ryangribble:
OK so appveyor is now building OK but ive been battling with travisCI unsuccesfully, running into a permissions error.  

I just realised though that im not even sure if the cake tasks we have would work on linux... we include tools/nuget.exe in our repo and it seems this is picked up/used by the ToolInstaller class in cake script, to install our GitVersion and OctokitCodeFormatter tools.  So if we wanted that to work i think we would need to get the linux compatible nuget.exe and run it via `mono nuget install` on the TravisCI linux and mac boxes... GitVersion.exe is the same thing (can be run via mono i beleive).  our code formatter tool doesnt support non windows platform, but its not actually used in the default build target so would be ok.

Potentially with dotnetcore, we dont even need to worry about TravisCI and could just use Appveyor instead...  Thoughts @shiftkey  ?

mderriey:
I also think that we can ditch TravisCI. Maybe it was needed before because some specific platform variants had to be built with Mono but it's not the case anymore with .NET Core.

shiftkey:
> Potentially with dotnetcore, we dont even need to worry about TravisCI and could just use Appveyor instead... Thoughts @shiftkey ?

:thumbsup: especially to drop macOS - not being dependent on those will speed up things significantly!

mderriey:
I checked out the PR base branch, and running `.\build.ps1` worked as expected. The `Package` task created the expected version of NuGet packages. I think you can check this one off.

I had a look at how source linking was enabled in both the Mvc repo and [Rx.NET](https://github.com/Reactive-Extensions/Rx.NET/commit/30b92d3fb21367b27ec1c8473fcb161bd22cf3e6). Looks very simple. I'll have to dig a bit deeper to understand where the `IsTestProject` property being used.

Thanks for that, good effort!

mderriey:
Hey @ryangribble what's left to get this one merged?

Do you want to get SourceLink in this one or in a separate one as @ctaggart suggested?

ryangribble:
> Hey @ryangribble what's left to get this one merged?
>
> Do you want to get SourceLink in this one or in a separate one as @ctaggart suggested?

Yeah a separate PR for SourceLink sounds good.

In terms of this one, the only thing left was to either get TravisCI working or drop it.  I guess we did have agreement to drop it, but a part of me still wants to pursue it "for science" if nothing else 😛 

I just pushed a commit that adjusted the NuGet.exe that is in our repo's permissions and this actually has seen the Travis linux build able to install tools (Cake's internals appear to correctly handle running NuGet.exe via `mono nuget.exe` when it needs to) and it's now failing trying to run GitVersion commandline.  Again, GitVersion does actually support running via `mono` but it could be that Cake doesn't handle this internally.  Assuming we can get GitVersion working, the only other thing that may not be possible on linux/osx is the LinqPad samples (we could skip them on these platforms perhaps) and the code formatter (which doesnt run as part of a CI build anyway).

Let me know if you think im flogging a dead horse!

mderriey:
I don't see the value in getting the build working on Travis, but I understand you want to give it a go. Plus, that way we'll find out if CAKE's internals support using mono for all the tools when on Linux/macOS.

I also found that page in the docs talking about troubleshooting issues on Travis locally with Docker containers: https://docs.travis-ci.com/user/common-build-problems/#Troubleshooting-Locally-in-a-Docker-Image.

I'll give it a look later today.

M-Zuber:
https://github.com/vadimdemedes/trevor might also be useful in debugging the travis issues

gep13:
@ryangribble that is an interesting approach with running GitVersion on non-windows!  Great thinking!

ryangribble:
Haha yeah although my current debug method (push commit and wait for TravisCI) has some pretty poor feedback loops 😝   I really should look into the docker image stuff!

mderriey:
I'm certainly not a bash expert, but if I run `echo == test ==` in Bash on Windows, it throws an error. Adding double-quotes around works, like `echo "== test =="`.

Also, is it possible to set the permissions of the wrapper and make it executable? it doesn't say on the PR what kind of permissions are associated with it.

mderriey:
@ryangribble amazing, that last idea was genius!

The only blocker now is the GitVersion runner [captures the standard output and deserializes it into a `GitVersion` instance](https://github.com/cake-build/cake/blob/acc5ee013272da3a40c11efc7e961902fbc31f8e/src/Cake.Common/Tools/GitVersion/GitVersionRunner.cs#L58).

The first `echo` statement in the wrapper confuses the JSON deserializer which throws. Removing it should help.

Great workaround 👍 !

ryangribble:
woot so GitVersion is now running on TravisCI via Mono hehehe

Next problem is the projects trying to build for net45

ryangribble:
@mderriey 👏 👏

rjk40:
Go directly to jail do not pass  go

ryangribble:
Nice, hopefully that last commit makes the OSX builds more stable 

@mderriey I think we're done here, did you want to take a final look over and give me a 👍 to hit the button?

rjk40:
Anyone involved in the act that has sabotaging innocent people and
company's names with your lies is a active of terrorism

rjk40:
You are definitely gonna need a leave your identity number 45 behind  I
hope you get caught for this by the near or Friday someone has to do
something

Octokit.Tests.Conventions/Octokit.Tests.Conventions.csproj
```diff
@@ -0,0 +1,45 @@
...
+    <ProjectReference Include="..\Octokit.Reactive\Octokit.Reactive.csproj" />
+    <ProjectReference Include="..\Octokit.Tests\Octokit.Tests.csproj" />
+  </ItemGroup>
+
+  <PropertyGroup Condition=" '$(TargetFramework)' == 'netcoreapp1.0' ">
+    <DefineConstants>$(DefineConstants);NO_SERIALIZABLE;HAS_TYPEINFO</DefineConstants>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.0.0-preview-20170106-08" />
```
mderriey: It seems there's a new, stable version of this package [on NuGet](https://www.nuget.org/packages/Microsoft.NET.Test.Sdk/15.0.0). Any specific reason to use this pre-release one?

Octokit.Tests.Integration/Octokit.Tests.Integration.csproj
```diff
@@ -0,0 +1,42 @@
...
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp1.0' ">
+    <ProjectReference Include="..\Octokit\Octokit.csproj" />
+    <ProjectReference Include="..\Octokit.Reactive\Octokit.Reactive.csproj" />
+    <ProjectReference Include="..\Octokit.Tests\Octokit.Tests.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.0.0-preview-20170106-08" />
```
mderriey: Same as above

Octokit.Tests/Octokit.Tests.csproj
```diff
@@ -0,0 +1,44 @@
...
+    <ProjectReference Include="..\Octokit.Reactive\Octokit.Reactive.csproj" />
+    <PackageReference Include="NSubstitute" Version="2.0.0-rc" />
+  </ItemGroup>
+
+  <PropertyGroup Condition=" '$(TargetFramework)' == 'netcoreapp1.0' ">
+    <DefineConstants>$(DefineConstants);NO_SERIALIZABLE;HAS_TYPEINFO</DefineConstants>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.0.0-preview-20170106-08" />
```
mderriey: Same as above

build/Lifetime.cs
```diff
@@ -0,0 +1,87 @@
...
+            context.IsOriginalRepo = StringComparer.OrdinalIgnoreCase.Equals("octokit/octokit.net", buildSystem.AppVeyor.Environment.Repository.Name);
+            context.IsMasterBranch = StringComparer.OrdinalIgnoreCase.Equals("master", buildSystem.AppVeyor.Environment.Repository.Branch);
+        }
+        else if (context.TravisCI)
+        {
+            context.IsPullRequest = !string.IsNullOrEmpty(buildSystem.TravisCI.Environment.Repository.PullRequest);
+            context.IsOriginalRepo = StringComparer.OrdinalIgnoreCase.Equals("octokit/octokit.net", buildSystem.TravisCI.Environment.Repository.Slug);
+            context.IsMasterBranch = StringComparer.OrdinalIgnoreCase.Equals("master", buildSystem.TravisCI.Environment.Build.Branch);
+        }
+
```
mderriey: nit: extra empty line

build/Lifetime.cs
```diff
@@ -0,0 +1,87 @@
...
+            new Project { Name = "Octokit.Tests", Path = "./Octokit.Tests/Octokit.Tests.csproj", UnitTests = true },
+            new Project { Name = "Octokit.Tests.Conventions", Path = "./Octokit.Tests.Conventions/Octokit.Tests.Conventions.csproj", UnitTests = true },
+            new Project { Name = "Octokit.Tests.Integration", Path = "./Octokit.Tests.Integration/Octokit.Tests.Integration.csproj", IntegrationTests = true }
+        };
+
+        // Install tools
+        context.Information("Installing tools...");
+        ToolInstaller.Install(context, "GitVersion.CommandLine", "3.6.2");
+        ToolInstaller.Install(context, "Octokit.CodeFormatter", "1.0.0-preview");
+
```
mderriey: nit: extra empty line

Octokit.Tests.Conventions/Octokit.Tests.Conventions.csproj
ryangribble: I just converted the project.json to csproj with VS2017 so I guess this is historic.  Ill update it and see how we go!

mderriey:
I thought GUIDs in `.csproj` were a thing of the past, but Visual Studio still comes with surprises :wink:.

Sweet, thank you, :shipit:!

PR #1571, opened 2017-03-27
Revert "fix .sln file for linux"

cokochiaki:
Reverts octokit/octokit.net#1556


ryangribble:
Hi @cokochiaki can you please explain why this should be reverted? Thanks

ryangribble:
dotnetcore is now merged, so these projects/solutions no longer exist

PR #1574, opened 2017-03-30
Add sourcelink support

ryangribble:
For our 2 library projects,
- Set PDB option to embedded
- add sourcelink MSBuild tools (should run automatically)
- add sourcelink test MSBuild tool (should verify the sourcelink worked ok)

### DIFF
--- a/Octokit.Reactive/Octokit.Reactive.csproj
+++ b/Octokit.Reactive/Octokit.Reactive.csproj
@@ -37,7 +37,6 @@
     <PackageReference Include="SourceLink.Create.GitHub" Version="2.1.0" PrivateAssets="All" />
     <DotNetCliToolReference Include="dotnet-sourcelink-git" Version="2.1.0" />
     <DotNetCliToolReference Include="dotnet-sourcelink" Version="2.1.0" />
-    <PackageReference Include="SourceLink.Test" Version="2.1.0" PrivateAssets="all" />
   </ItemGroup>
 
 </Project>
--- a/Octokit.Reactive/Octokit.Reactive.csproj
+++ b/Octokit.Reactive/Octokit.Reactive.csproj
@@ -34,10 +34,10 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="SourceLink.Create.GitHub" Version="2.0.2" PrivateAssets="All" /> 
-    <DotNetCliToolReference Include="dotnet-sourcelink-git" Version="2.0.2" />
-    <DotNetCliToolReference Include="dotnet-sourcelink" Version="2.0.2" />
-    <PackageReference Include="SourceLink.Test" Version="2.0.2" PrivateAssets="all" />
+    <PackageReference Include="SourceLink.Create.GitHub" Version="2.1.0" PrivateAssets="All" />
+    <DotNetCliToolReference Include="dotnet-sourcelink-git" Version="2.1.0" />
+    <DotNetCliToolReference Include="dotnet-sourcelink" Version="2.1.0" />
+    <PackageReference Include="SourceLink.Test" Version="2.1.0" PrivateAssets="all" />
   </ItemGroup>
 
 </Project>
--- a/Octokit.Reactive/Octokit.Reactive.csproj
+++ b/Octokit.Reactive/Octokit.Reactive.csproj
@@ -12,6 +12,7 @@
     <GenerateAssemblyProductAttribute>false</GenerateAssemblyProductAttribute>
     <GenerateAssemblyVersionAttribute>false</GenerateAssemblyVersionAttribute>
     <GenerateAssemblyFileVersionAttribute>false</GenerateAssemblyFileVersionAttribute>
+    <DebugType>embedded</DebugType>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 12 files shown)


ryangribble:
OK so this worked 😀 

image

image


But I'd like to make a couple of enhancements 

- [ ] Ability to run sourcelink on local builds
By default sourcelink is running only on CI server.  Ill add a cake parameter allowing us to run sourcelink locally if required.  When this parameter is passed to Cake, we can set `/p:ci=true` parameter on the DotNetBuild task which will cause sourcelink to run.  Not sure what other side effects of setting `ci=true` are though...  Alternatively I can investigate how to control running sourcelink or not with a specific parameter maybe...

- [ ] Show sourcelink output in build script output
Currently the log doesnt show any of this output unless you set verbosity (passing `/v:n` to the build task).  But that of course shows HEAPS of other output when ideally there would be a way to just see the sourcelink info (or even just the result of `dotnet sourcelink test`). Let me know if there is anything like that @ctaggart or whether I have to go with verbose mode?

mderriey:
That's great @ryangribble!

ctaggart:
> Ability to run sourcelink on local builds
By default sourcelink is running only on CI server. Ill add a cake parameter allowing us to run sourcelink locally if required. When this parameter is passed to Cake, we can set /p:ci=true parameter on the DotNetBuild task which will cause sourcelink to run. Not sure what other side effects of setting ci=true are though... Alternatively I can investigate how to control running sourcelink or not with a specific parameter maybe

You could set the environment variables or msbuild properties of `CI` or `SourceLinkCreate` to `true`. They are not case sensitive. See [this code](https://github.com/ctaggart/SourceLink/blob/v2/SourceLink.Create.GitHub/SourceLink.Create.GitHub.targets#L9-L11) for details.

> Show sourcelink output in build script output
Currently the log doesnt show any of this output unless you set verbosity (passing /v:n to the build task). But that of course shows HEAPS of other output when ideally there would be a way to just see the sourcelink info (or even just the result of dotnet sourcelink test). Let me know if there is anything like that @ctaggart or whether I have to go with verbose mode?

2.1 will be a lot less verbose. It will not print all the source files. I'll probably publish to NuGet Gallery the [latest prerelease](https://ci.appveyor.com/project/ctaggart/sourcelink/history) soon if it passes tests.

ctaggart:
2.1.0 is out :-)

ryangribble:
> 2.1 will be a lot less verbose. It will not print all the source files. I'll probably publish to NuGet Gallery the latest prerelease soon if it passes tests.

I didnt mind the verbosity of the sourcelink output itself, the main problem was since it's part of the "build" task, when I set verbosity I got loads of build related output from MSBuild (when i really just wanted the sourelink output).

For the time being ive actually removed the builtin MSBuild sourcelink test, and running my own cake task to do a sourcelink test on the produced assemblies, so our build output shows that sourcelink was verified.

ryangribble:
Builds are green, I'm happy if you are :+1: @mderriey 

From the appveyor build log:
image

TravisCI:
image

ctaggart:
You are the first to get it working on Travis! Thanks. I'll look to see how you got msbuild 15 running there. With 2.1, you can also just test nupkg files.

ryangribble:
Oh nice!  I'll change my step to run after packaging and use the nupkg files instead, as the way I was getting the dll assemblies was assuming a few things! 

Regarding travis yeah we had a few hoops to jump  but between @mderriey and I we got it working! Most of the magic will be in travis.yml and build.sh and a little in our cake build (running nuget and gitversion via mono)

mderriey:
I don't explain it, but running `dotnet sourcelink test` as part of the Cake build is much much longer than running it directly; When I tried before pushing, it took a good 2 minutes as part of the Cake build, and maybe 4-5 seconds from a PowerShell console.

@ryangribble do you see the same behaviour? Is it worth looking into it?

mderriey:
Hmm, must be something with my machine, it took ~40 seconds on AppVeyor and Travis CI.

ctaggart:
It is literally downloading all the source files 1 by 1. I added an `up for grabs` to [speed this up](https://github.com/ctaggart/SourceLink/issues/190). How fast is your internet connection compared to AppVeyor and Travis CI?

ryangribble:
Yeah it did take a noticable amount of time but as @ctaggart said it is checking each file on github's checksum against that stored in the embedded pdb so it was expected.  it seems to be faster on the subsequent assemblies so I'm guessing some caching must be involved?

Also when comparing to previous builds Im not sure if the SourceLink.Test msbuild target was even actually running once @mderriey introduced the `SourceLinkCreate` msbuild parameter setting.  From the brief look I had, the `CI` parameter would run create and test actions, but the `SourceLinkCreate` parameter only controlled creation, and a separate `SourceLinkTest` parameter controlled the testing.  [See here](https://github.com/ctaggart/SourceLink/blob/v2/SourceLink.Test/SourceLink.Test.targets#L9-L11).

mderriey:
@ctaggart in this case, I'm using a 100Mb/s link.
@ryangribble you're right, I don't think it was, and this was intentional. I like the fact that we're explicitly testing that source linking works. Do you prefer running it as part of the `dotnet build`?

If not, I'm happy with the current state of things 😃

ryangribble:
> Do you prefer running it as part of the dotnet build?

I prefer running it separately as I liked the idea of seeing output/validation of the sourcelink data.  Perhaps in the future if the msbuild sourcelink test target can log the "sourcelink test on xxxx succesful" when msbuild logging verbosity is the default, then we could remove the explicit step, but for now im happy!

ryangribble:
One other thing I just thought about... when you run sourcelink test "before pushing" it gets 404 failures because that commit hash doesnt exist on github.com.  So perhaps that is the reason for seeing execution time take longer locally

mderriey:
I thought about that, too, I pushed before running it locally as the sourcelink test passed. It's not a big deal, really, I thought it would also take a long time on CI, but it turns out it's not, so it's not really an issue IMO.

ryangribble:
release_notes: Provide sourcelink capability for Octokit and Octokit.Reactive assemblies

PR #1576, opened 2017-03-31
Surface exceptions in RepoCollaboratorsClient.Add method

### LINKED ISSUE #1575: Calling RepoCollaboratorsClient.Add with CollaboratorRequest param swallows exceptions
If I call the `RepoCollaboratorsClient.Add(string owner, string name, string user)` without specifying credentials I get an "Octokit.AuthorizationException: Bad credentials". That's expected.

If I call the `RepoCollaboratorsClient.Add(string owner, string name, string user, CollaboratorRequest permission)` overload the same exception is swallowed and instead a false boolean is returned, making it impossible to know why the operation is failing.

I would expect both methods to behave in the same way, preferably throwing exceptions.

Is there a reason for the difference, that I'm just not seeing?

joensindholt:
Fixes #1575

This PR removes the "swallowing" of exceptions in the overloaded `RepoCollaboratorsClient.Add` method making it possible to know what went wrong in case of an exception happening.

### DIFF
--- a/Octokit.Tests/Clients/RepoCollaboratorsClientTests.cs
+++ b/Octokit.Tests/Clients/RepoCollaboratorsClientTests.cs
@@ -227,6 +227,28 @@ namespace Octokit.Tests.Clients
                 await Assert.ThrowsAsync<ArgumentException>(() => client.Add("owner", "test", ""));
                 await Assert.ThrowsAsync<ArgumentException>(() => client.Add(1, ""));
             }
+
+            [Fact]
+            public async Task SurfacesAuthorizationException()
+            {
+                var connection = Substitute.For<IApiConnection>();
+                var client = new RepoCollaboratorsClient(connection);
+
+                connection.Put(Arg.Any<Uri>()).Returns(x => { throw new AuthorizationException(); });
+
+                await Assert.ThrowsAsync<AuthorizationException>(() => client.Add("owner", "test", "user1"));
+            }
+
+            [Fact]
+            public async Task SurfacesAuthorizationExceptionWhenSpecifyingCollaboratorRequest()
+            {
+                var connection = Substitute.For<IApiConnection>();
+                var client = new RepoCollaboratorsClient(connection);
+
+                connection.Connection.Put<object>(Arg.Any<Uri>(), Arg.Any<object>()).ThrowsAsync(new AuthorizationException());
+
+                await Assert.ThrowsAsync<AuthorizationException>(() => client.Add("owner", "test", "user1", new CollaboratorRequest(Permission.Pull)));
+            }
         }
 
         public class TheInviteMethod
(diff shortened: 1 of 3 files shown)


ryangribble:
Im not sure of the historic reasons, but it looks like there are 12 places where the `response.IsTrue()` is guarded by a try/catch that eats the exception and returns false instead.  For consistency, if they are to be changed I'd like to see them all changed.

Im also not sure how to treat this behavioural change as some consumers could be impacted by exceptions being thrown when previously they received a `false` return.  Although it does seem the exceptions would only occur in true error situations, since a response of 404 or 204 will cause `false`/`true` to be returned.

Any thoughts @shiftkey  ?

ryangribble:
Actually of those 12 instances it looks like all but 2 are actually using `catch (NotFoundException) { }` as opposed to `catch { }` 

The former is "correct" because it will have the logic that a HTTP 404 (NotFound) will return `false` a 204 (NoContent) means `true` and any other status (like your authorization exception) will be thrown.

So rather than how you've changed it in this PR, if you just make the `catch` be for `NotFoundException` only, that should achieve the desired result.  If you could also fix up the [the other one](https://github.com/octokit/octokit.net/blob/master/Octokit/Clients/RepoCollaboratorsClient.cs#L223) while you're at it, that would be swell!


FYI there are also 40-50 matches of other code/methods where they check for a HTTP 204/404 for success only they dont use that handy `HttpResponse.IsTrue()` extension method :grinning:.  Looking at most of those it looks like they are also behaving "correctly" in that a `true`/`false` return results on a 204/404 and an exception on any other status code...

joensindholt:
Sounds fair - I'll change it.

shiftkey:
I'm not across the history of the `.IsTrue()` extension method - it does seem to look functionally correct:

https://github.com/octokit/octokit.net/blob/c9b2c1260bc87b7782d4fd9645d43a8885203923/Octokit/Helpers/ApiExtensions.cs#L105

But it does overlook that the underlying connection will raise an exception before the caller has had a chance to look at the response:

https://github.com/octokit/octokit.net/blob/3c818934b8b3b4cdc5472bf01e869c07343c3928/Octokit/Http/Connection.cs#L597

So the try-catch is still necessary, but it should be more granular and only look for the `NotFoundException`.

joensindholt:
I have reintroduced the try...catch blocks but this time only for NotFoundException. I did it in both the Add methods as requested.

ryangribble:
Great!  Thanks @joensindholt

ryangribble:
LGTM

ryangribble:
ugh so I was giong to merge this but the builds are actually red 😢

joensindholt:
Yes, I saw that but I convinced myself it wasn't due to the changes in this PR. Partly because #1571 fails the exact same way, partly because it loos like something related to .net core?

ryangribble:
Yeah that was just a demonstration project anyway and it seems the appveyor image and dotnet core tooling doesn't like it anymore. Will be superseded by our proper dotnet core port soon anyway, so ive removed it from master for now 

Thanks!

joensindholt:
No - thank you :)

PR #1580, opened 2017-04-11
Cloning ApiInfo object should work when some fields are null

### LINKED ISSUE #1509: GetLastApiInfo always returns null
Using octokit v0.23.0 from nuget I am trying to capture rate limit information from github but whenever I call `gitHubClient.GetLastApiInfo()` it is returning null

ryangribble:
Fixes #1509 

`ApiInfo.Clone()` previously returned `null` if any of the child properties of the object to clone were `null`.  

I encountered a situation where there was no ETag header in an API response, which caused the "Last Api Info" object to be `null` due to this logic.

With this change, the `Clone()` method will clone whatever fields do exist, which means we should now be correctly providing the "Last Api Info" in more cases.


ryangribble:
Anyone able to take a 👀 to check what I've done seems sensible?

@101shipit @shiftkey

ryangribble:
LGTM

ryangribble:
release_notes: Fix `connection.GetLastApiInfo()` was returning `null` in some situations

Issue #1582, opened 2017-04-14
*Url response properties are inconsistent


shiftkey:
Before going 1.0 we should do a scan and ensure all response fields suffixed with `Url` are `Uri`s instead of strings.

@nickcraver pointed these examples out to me:

 - https://github.com/octokit/octokit.net/blob/master/Octokit/Models/Response/Author.cs#L38-L58
 - https://github.com/octokit/octokit.net/blob/master/Octokit/Models/Response/GitHubCommit.cs#L45
 - https://github.com/octokit/octokit.net/blob/master/Octokit/Models/Response/Account.cs#L99

shiftkey:
Or maybe they should just go back to `string` if we're not getting anything from using `Uri` due to the increased allocations?

NickCraver:
+1 on consistency, and a selfish +1 on strings (for performance reasons). I'd value consistency more overall, though.

mderriey:
@shiftkey I'm happy to do this.
Would you rather have them as `string` or `Uri`?

So far, 43 classes have at least one property named or suffixed with `Url` that are not `Uri`. 27 classes have at least one property named or suffixed with `Url` that are all typed as `Uri`.

shiftkey:
@mderriey I really don't think we're getting much out of returning `Uri`s - and they're boring enough that the caller can `new Uri()` the result if they need it explicitly.

Let's make them all `string`.

ryangribble:
Sounds good. You could also add a convention test @mderriey,  to keep us consistent going forward :D

mderriey:
Cool, will do!

NickCraver:
\o/ thanks @mderriey and @ryangribble!

PR #1585, opened 2017-04-16
Change response models 'Url' properties from `Uri` to `string`

mderriey:
See #1582.

### DIFF
--- a/Octokit.Tests.Conventions/Exception/InvalidUrlPropertyTypeException.cs
+++ b/Octokit.Tests.Conventions/Exception/InvalidUrlPropertyTypeException.cs
@@ -13,7 +13,7 @@ namespace Octokit.Tests.Conventions
 
         static string CreateMessage(Type modelType, IEnumerable<PropertyInfo> propertiesWithInvalidType)
         {
-            return string.Format("Model type '{0}' contains the following properties that are named or suffixed with 'Url' but are not of type String: {2}{3}",
+            return string.Format("Model type '{0}' contains the following properties that are named or suffixed with 'Url' but are not of type String: {!}{2}",
                 modelType.FullName,
                 Environment.NewLine,
                 string.Join(Environment.NewLine, propertiesWithInvalidType.Select(x => x.Name)));
--- a/Octokit.Tests.Conventions/Exception/InvalidUrlPropertyTypeException.cs
+++ b/Octokit.Tests.Conventions/Exception/InvalidUrlPropertyTypeException.cs
@@ -7,15 +7,14 @@ namespace Octokit.Tests.Conventions
 {
     public class InvalidUrlPropertyTypeException : Exception
     {
-        public InvalidUrlPropertyTypeException(Type modelType, Type expectedUrlPropertyType, IEnumerable<PropertyInfo> propertiesWithInvalidType)
-            : base(CreateMessage(modelType, expectedUrlPropertyType, propertiesWithInvalidType))
+        public InvalidUrlPropertyTypeException(Type modelType, IEnumerable<PropertyInfo> propertiesWithInvalidType)
+            : base(CreateMessage(modelType, propertiesWithInvalidType))
         { }
 
-        static string CreateMessage(Type modelType, Type expectedUrlPropertyType, IEnumerable<PropertyInfo> propertiesWithInvalidType)
+        static string CreateMessage(Type modelType, IEnumerable<PropertyInfo> propertiesWithInvalidType)
         {
-            return string.Format("Model type '{0}' contains the following properties that are named or suffixed with 'Url' but are not of type {1}: {2}{3}",
+            return string.Format("Model type '{0}' contains the following properties that are named or suffixed with 'Url' but are not of type String: {2}{3}",
                 modelType.FullName,
-                expectedUrlPropertyType.Name,
                 Environment.NewLine,
                 string.Join(Environment.NewLine, propertiesWithInvalidType.Select(x => x.Name)));
         }
--- a/Octokit.Tests.Conventions/Exception/InvalidUrlPropertyTypeException.cs
+++ b/Octokit.Tests.Conventions/Exception/InvalidUrlPropertyTypeException.cs
@@ -13,7 +13,7 @@ namespace Octokit.Tests.Conventions
 
         static string CreateMessage(Type modelType, IEnumerable<PropertyInfo> propertiesWithInvalidType)
         {
-            return string.Format("Model type '{0}' contains the following properties that are named or suffixed with 'Url' but which type is not String: {1}{2}",
+            return string.Format("Model type '{0}' contains the following properties that are named or suffixed with 'Url' but are not of type String: {1}{2}",
                 modelType.FullName,
                 Environment.NewLine,
                 string.Join(Environment.NewLine, propertiesWithInvalidType.Select(x => x.Name)));
--- a/Octokit.Tests.Conventions/Exception/InvalidUrlPropertyTypeException.cs
+++ b/Octokit.Tests.Conventions/Exception/InvalidUrlPropertyTypeException.cs
@@ -7,14 +7,15 @@ namespace Octokit.Tests.Conventions
 {
     public class InvalidUrlPropertyTypeException : Exception
     {
-        public InvalidUrlPropertyTypeException(Type modelType, IEnumerable<PropertyInfo> propertiesWithInvalidType)
-            : base(CreateMessage(modelType, propertiesWithInvalidType))
+        public InvalidUrlPropertyTypeException(Type modelType, Type expectedUrlPropertyType, IEnumerable<PropertyInfo> propertiesWithInvalidType)
+            : base(CreateMessage(modelType, expectedUrlPropertyType, propertiesWithInvalidType))
         { }
 
-        static string CreateMessage(Type modelType, IEnumerable<PropertyInfo> propertiesWithInvalidType)
+        static string CreateMessage(Type modelType, Type expectedUrlPropertyType, IEnumerable<PropertyInfo> propertiesWithInvalidType)
         {
-            return string.Format("Model type '{0}' contains the following properties that are named or suffixed with 'Url' but are not of type String: {1}{2}",
+            return string.Format("Model type '{0}' contains the following properties that are named or suffixed with 'Url' but are not of type {1}: {2}{3}",
                 modelType.FullName,
+                expectedUrlPropertyType.Name,
                 Environment.NewLine,
                 string.Join(Environment.NewLine, propertiesWithInvalidType.Select(x => x.Name)));
         }
--- a/Octokit.Tests.Conventions/Exception/InvalidUrlPropertyTypeException.cs
+++ b/Octokit.Tests.Conventions/Exception/InvalidUrlPropertyTypeException.cs
@@ -1,0 +1,22 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+
+namespace Octokit.Tests.Conventions
+{
+    public class InvalidUrlPropertyTypeException : Exception
+    {
+        public InvalidUrlPropertyTypeException(Type modelType, IEnumerable<PropertyInfo> propertiesWithInvalidType)
+            : base(CreateMessage(modelType, propertiesWithInvalidType))
+        { }
+
+        static string CreateMessage(Type modelType, IEnumerable<PropertyInfo> propertiesWithInvalidType)
+        {
+            return string.Format("Model type '{0}' contains the following properties that are named or suffixed with 'Url' but which type is not String: {1}{2}",
+                modelType.FullName,
+                Environment.NewLine,
+                string.Join(Environment.NewLine, propertiesWithInvalidType.Select(x => x.Name)));
+        }
+    }
+}
--- a/Octokit.Tests.Conventions/ModelTests.cs
+++ b/Octokit.Tests.Conventions/ModelTests.cs
@@ -127,6 +127,18 @@ namespace Octokit.Tests.Conventions
         [MemberData("RequestModelTypes")]
         public void RequestModelsHaveUrlPropertiesOfTypeUri(Type modelType)
         {
+            if (modelType == typeof(AuthorizationUpdate) ||
+                modelType == typeof(NewAuthorization) ||
+                modelType == typeof(NewCommitStatus) ||
+                modelType == typeof(NewDeploymentStatus))
+            {
+                // These request models types can have Url properties
+                // of type string as the API doesn't enforce a valid
+                // URI on the server side
+                // see thread from https://github.com/octokit/octokit.net/pull/1585#issuecomment-295163847
+                return;
+            }
+
             var propertiesWithInvalidType = modelType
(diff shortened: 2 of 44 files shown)


mderriey:
@shiftkey @ryangribble could you please review when you have a chance?

I assumed from the wording of the associated issue that this only had to apply to the response models. Let me know if you want to expand this to request models, too, it will be a quick change.

ryangribble:
I reckon we may as well do request models as well

Octokit.Tests.Conventions/Exception/InvalidUrlPropertyTypeException.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    public class InvalidUrlPropertyTypeException : Exception
+    {
+        public InvalidUrlPropertyTypeException(Type modelType, IEnumerable<PropertyInfo> propertiesWithInvalidType)
+            : base(CreateMessage(modelType, propertiesWithInvalidType))
+        { }
+
+        static string CreateMessage(Type modelType, IEnumerable<PropertyInfo> propertiesWithInvalidType)
+        {
+            return string.Format("Model type '{0}' contains the following properties that are named or suffixed with 'Url' but which type is not String: {1}{2}",
```
ryangribble: wording seems a little off here.  Maybe it could be "....  but are not of type String"

mderriey: You wouldn't believe it took me 5 good minutes to come up with that weird wording, would you? :wink:

ryangribble:
I kicked it and it's Green now... Travis builds seem to be frustratingly flaky 😢 

I'm having second thoughts about the Request models though.  My thinking has kind of come around to the fact that these are meant to be proper Url's so allowing people to pass "any old string" may not be great.  Thoughts?


I also had a look at other places we use `Uri` type such as some helper functions or the `GitHubClient` itself, and I don't think any of those cases need to be turned to `string`.  So I guess pending a thought on whether Request models should or shouldn't be converted to `string` this should be good to merge.

Also obviously this is a breaking change but I think just getting it in and flagging it in the release notes is preferable than waiting out 2 releases with the current properties marked `[Obsolete]` - given our pre 1.0 status I'm thinking that should be OK...

shiftkey:
> I'm having second thoughts about the Request models though. My thinking has kind of come around to the fact that these are meant to be proper Url's so allowing people to pass "any old string" may not be great. Thoughts?

Yeah, I'm less into doing this for request models now because of how invalid input might be rejected - we're not doing anything explicit here with validating, but it's still a decent constraint to make callers follow.

I can only see `NewCommitStatus` updated here - @mderriey how many would change here if we enforced `Uri` on the`*Url` request model properties?

mderriey:
@shiftkey @ryangribble Only 4 request models have `*Url` properties which are not `Uri`, so the change should be pretty quick and easy.

One more thing; The current logic to get all the request models is to analyse all the parameters of all the methods on the client interfaces. It turns out that the [`Release`](https://github.com/octokit/octokit.net/blob/master/Octokit/Models/Response/Release.cs) response model is a parameter of the [`UploadAssetData`](https://github.com/octokit/octokit.net/blob/0a9a31dea48b5af8240b3626cd41e6fa244df181/Octokit/Clients/IReleasesClient.cs#L235) method on `IReleasesClient`.

That hasn't been a problem before since the tests always include the response models and we selectively choose if we want the request models.

I see a couple of options available here:

 - a quick hack to exclude the `Release` type from the request model types - as it's done for [`IStatisticsClient`](https://github.com/octokit/octokit.net/blob/0a9a31dea48b5af8240b3626cd41e6fa244df181/Octokit.Tests.Conventions/ModelTests.cs#L116)
 - duplicate the `Release` type to make it a request model type. This would mean duplicate also the inner types.

mderriey:
I have a couple of commits ready to go that implement the first option, that is the `Release` type is excluded from request model types.

Here are the changes that'd br brought in should you decide to go ahead with this approach: https://github.com/mderriey/octokit.net/compare/1582-change-model-url-properties-to-string...mderriey:change-request-model-url-properties-to-uri

ryangribble:
Those few request models seem ok to me... 

I'd be interested in what currently happens if a user provides a strong that is NOT a properly formatted Url... If the upstream API rejects it anyway then it's a no brained to enforce the Uri on our side. If they are somehow allowed through then we'd need to think about whether we want to potentially prevent users from doing something they could be doing currently

mderriey:
Here's what I found out:

 - `AuthorizationUpdate.NoteUrl` and `NewAuthorization.NoteUrl` are both optional properties and will accept any `string`; I called the API endpoints with a value of `not a URL` and the call succeeded
 - `NewCommitStatus.TargetUrl` needs to use the `https` scheme, but doesn't necessarily require a valid Uri; the API rejected `not a URL` but accepted `https://not a URL` - which will throw if used in `Uri` ctor

I didn't know how to test the properties on `NewDeploymentStatus` as I couldn't figure how to create a deployment in the first place despite the [associated API documentation](https://developer.github.com/v3/repos/deployments)

mderriey:
@ryangribble @shiftkey quick update on this

I played around with the deployment status API.
`NewDeploymentStatus.LogUrl` and `NewDeploymentStatus.EnvironmentUrl` are enforced to start with `https`, but can be invalid Uri - like `https://not a URL`.

Going from there, I would suggest:

 - leaving `NoteUrl` on `AuthorizationUpdate` and `NewAuthorization` as `string` since the API will accept any value
 - leaving `TargetUrl` on `NewCommitStatus` as `string` since a valid Uri is not enforced - it just needs to start with `https`
 - leaving `LogUrl` and `EnvironmentUrl` of `NewDeploymentStatus` for the same reason as above

shiftkey:
@mderriey I like those suggestions 👍

mderriey:
@shiftkey Cool, thanks.

I still kept the integration test which checks that request models have `*Url` properties of type `string`, but excluded the types we talked about earlier from it.

Is that fine by you or would you rather not check request models at all?

shiftkey:
> Is that fine by you or would you rather not check request models at all?

This is fine - it's good to be explicit here.

mderriey:
I *think* the AppVeyor build failed because the NuGet v2 feed is experiencing issues

image

I don't know how the Travis one passed, though. Local NuGet cache?

Octokit.Tests.Conventions/ModelTests.cs
```diff
@@ -107,6 +107,50 @@ public void ResponseModelsHaveReadOnlyCollections(Type modelType)
             }
         }
 
+        [Theory]
+        [MemberData("ResponseModelTypes")]
```
khellang: `nameof`? Might as well change all of them if you decide to go for it.

Octokit.Tests.Conventions/ModelTests.cs
```diff
@@ -107,6 +107,50 @@ public void ResponseModelsHaveReadOnlyCollections(Type modelType)
...
+            }
+        }
+
+        [Theory]
+        [MemberData("RequestModelTypes")]
+        public void RequestModelsHaveUrlPropertiesOfTypeUri(Type modelType)
+        {
+            if (modelType == typeof(AuthorizationUpdate) ||
+                modelType == typeof(NewAuthorization) ||
+                modelType == typeof(NewCommitStatus) ||
```
khellang: Can these type have other properties that *should* be `Uri`? Is it worth doing the check on type + property instead?

khellang: Also, ProTip™: You can use `nameof` for properties on a type, i.e. `nameof(AuthorizationUpdate.NoteUrl)` :smile:

Octokit.Tests.Conventions/ModelTests.cs
mderriey: I went for consistency here. not knowing what versions of VS contributors use, even if I guess it's safe they use a recent enough version. The .NET Core port should be merged in the near future, and since the new .csproj involves using VS 2017+, these updates will be totally safe.

Octokit.Tests.Conventions/ModelTests.cs
mderriey: You're right, they can. At the moment they don't, so I felt like checking for the type was good enough.

Octokit.Tests.Conventions/ModelTests.cs
khellang: Oh, sorry. I just assumed this project used C# 6 by now (given that C# 7 has since been released), but I see the `LangVersion` elements from https://github.com/octokit/octokit.net/pull/820 are still there...

Octokit.Tests.Conventions/ModelTests.cs
mderriey: No worries at all, thanks for spotting this. I didn't even know C# 6 was explicitly disabled. I'll take a look at the places in the [`dotnetcore`](https://github.com/octokit/octokit.net/tree/dotnetcore) branch tomorrow and use new features where possible.

Octokit.Tests.Conventions/ModelTests.cs
mderriey: No worries at all, thanks for spotting this. I didn't even know C# 6 was explicitly disabled. I'll take a look at the places in the dotnetcore branch tomorrow and use new features where possible.

mderriey:
@ryangribble @ryangribble Could one of you trigger a new AppVeyor build?

If it's green and you're happy with the current state of the PR, I think this is good to go.

khellang:
You should be able to trigger a new build by rebasing this branch

shiftkey:
@mderriey i got you

mderriey:
@shiftkey Thanks! :+1: 

@khellang Forgive my ignorance, could you explain a bit more? I usually add an empty commit when I want to do this, but keen to learn if there's a better way.

khellang:
@mderriey If you rebase this branch, `mderriey/1582-change-model-url-properties-to-string`, on top of `octokit/master` and force push, it should trigger a new build. This PR would probably be a good exercise for squashing as well 😉

ryangribble:
Since GitHub introduced the "squash" action directly when merging PR button ❤️ ❤️ I never bother to explicitly squash the branch anymore 😛 

SO in terms of the Request models Im just trying to catch up here... did we end up with ANY of them needing to be `Uri` or did they all end up being `string` because the upstream api does accept non Uri's?  If there are no real examples of existing request models that must be Uri we probably shouldn't have the convention test since it would stand to reason that future api calls implemented would likely follow the same upstream behaviour (and thus we should use `string` rather than `Uri` in future too?)

mderriey:
Here's how it went:

 - After the convention test was created, 3 types were found having `string` properties
 - Calling the API endpoints associated with these types showed that they didn't need to be `Uri` as, well, the upstream API doesn't enforce valid URLs
 - The convention test was modified to add those 3 types as exceptions

In other words, all the request model types but these three have `Uri` properties. I don't know the API well enough to determine if the other request model types should have `Uri` properties.

ryangribble:
Thanks for the recap @mderriey .  So i pulled down your changes and played a bit and actually there are NO other Request models with Url properties except the ones you've excluded anyway.  So im thinking how about we actually change the test to "RequestModelsHaveUrlPropertiesOfTypeString" and enforce that we use strings for these.  Given that the only existing 4 request models with these properties are all accepted as strings upstream anyway it seems likely that future upstream API fields would be similar.

``` csharp
    [Theory]
    [MemberData("RequestModelTypes")]
    public void RequestModelsHaveUrlPropertiesOfTypeString(Type modelType)
    {
        var propertiesWithInvalidType = modelType
            .GetProperties()
            .Where(x => x.Name.EndsWith("Url"))
            .Where(x => x.PropertyType != typeof(string))
            .ToList();

        if (propertiesWithInvalidType.Count > 0)
        {
            throw new InvalidUrlPropertyTypeException(modelType, typeof(string), propertiesWithInvalidType);
        }
    }
```

mderriey:
@ryangribble Haha,that's funny. I hadn't looked that closely. Thanks for that. Maybe the tests should only test types that do have `*Url` properties...

Since it's only the excluded types, maybe we can go back to the scenario we had at some point with one convention test checking request and response model types at the same time? Would that work or would you prefer having two separate tests?

ryangribble:
> Maybe the tests should only test types that do have *Url properties...

Yeah i guess at the moment the `MemberData` is all response types which does artificially inflate the test cases needlessly...  It's fine if you want to write a new function that only gets models that DO have the property that `EndsWith("Url")` if you want to

Similarly, it's fine if you want to combine back to one test case that does Request+Response models, or to leave it as is :+1:

mderriey:
I went with:

 - Filtering the test input to get only the types that do have at least one `*Url` property
 - Merge back the tests into one

Octokit.Tests.Conventions/Exception/InvalidUrlPropertyTypeException.cs
```diff
@@ -7,15 +7,14 @@ namespace Octokit.Tests.Conventions
...
-            : base(CreateMessage(modelType, expectedUrlPropertyType, propertiesWithInvalidType))
+        public InvalidUrlPropertyTypeException(Type modelType, IEnumerable<PropertyInfo> propertiesWithInvalidType)
+            : base(CreateMessage(modelType, propertiesWithInvalidType))
         { }
 
-        static string CreateMessage(Type modelType, Type expectedUrlPropertyType, IEnumerable<PropertyInfo> propertiesWithInvalidType)
+        static string CreateMessage(Type modelType, IEnumerable<PropertyInfo> propertiesWithInvalidType)
         {
-            return string.Format("Model type '{0}' contains the following properties that are named or suffixed with 'Url' but are not of type {1}: {2}{3}",
+            return string.Format("Model type '{0}' contains the following properties that are named or suffixed with 'Url' but are not of type String: {2}{3}",
```
ryangribble: {2} and {3} should now be {1} and {2} I guess...

mderriey: thanks!

Octokit.Tests.Conventions/ModelTests.cs
```diff
@@ -142,8 +168,9 @@ private static IEnumerable<Type> GetModelTypes(bool includeRequestModels)
 
                 if (includeRequestModels)
                 {
-                    var requestModels = methods.SelectMany(method => method.GetParameters(),
-                        (method, parameter) => parameter.ParameterType);
+                    var requestModels = methods
+                        .SelectMany(method => method.GetParameters(), (method, parameter) => parameter.ParameterType)
+                        .Where(t => t != typeof(Release)); // Release is a response model used as a parameter in IReleasesClient.UploadAsset
```
ryangribble: Does this clause about `Release` type need to be here?  Eventhough it may be a Response model, if it is part of a client method parameters then it technically is a Request parameter as well, so probably shouldn't be excluded from Request model tests...

mderriey: You're right. I rushed through this.

ryangribble:
Thanks @mderriey this is looking 💎 now, Ill hit the button

ryangribble:
LGTM

PR #1588, opened 2017-04-20
Add Pull Request Review Request API.

gdziadkiewicz:
TODO:
- [x] Add interfaces.
- [x] Add models.
- [x] Add implementations.
- [x] Add tests.
- [x] Fix names.
- [x] Add missing documentation.
- [x] Use `PullRequest` instead of `PullRequestReviewRequestCreate`
- [x] Add repositoryId methods. 
- [x] Add integration tests.
- [x] Final check


Octokit.Reactive/Clients/IObservablePullRequestsClient.cs
```diff
@@ -23,6 +23,11 @@ public interface IObservablePullRequestsClient
         IObservablePullRequestReviewCommentsClient ReviewComment { get; }
 
         /// <summary>
+        /// Client for managing review requests.
+        /// </summary>
+        IObservablePullRequestReviewRequestsClient ReviewRequests { get; }
```
ryangribble: Our convention is to use plural for the client name (which you've done `PullRequestReviewRequestsClient`) but to use singular for the sub-client property accessors.  So in this case rather than `ReviewRequests` it should just be `ReviewRequest`

Octokit.Reactive/Clients/ObservablePullRequestReviewRequestsClient.cs
```diff
@@ -0,0 +1,47 @@
...
+        readonly IConnection _connection;
+
+        public ObservablePullRequestReviewRequestsClient(IGitHubClient client)
+        {
+            Ensure.ArgumentNotNull(client, "client");
+
+            _client = client.PullRequest.ReviewRequests;
+            _connection = client.Connection;
+        }
+
```
ryangribble: Just FYI that we add the xmldoc comments to the interfaces and the client implementations, so once you are done with the changes please add the xmldoc comments to all the conrete methods as well :+1:

Octokit.Reactive/Clients/ObservablePullRequestsClient.cs
```diff
@@ -26,13 +26,19 @@ public class ObservablePullRequestsClient : IObservablePullRequestsClient
         /// </summary>
         public IObservablePullRequestReviewCommentsClient ReviewComment { get; private set; }
 
+        /// <summary>
+        /// Client for managing review requests.
+        /// </summary>
+        public IObservablePullRequestReviewRequestsClient ReviewRequests { get; private set; }
```
ryangribble: `ReviewRequests` => `ReviewRequest`

Octokit/Clients/IPullRequestsClient.cs
```diff
@@ -25,6 +25,11 @@ public interface IPullRequestsClient
         IPullRequestReviewCommentsClient ReviewComment { get; }
 
         /// <summary>
+        /// Client for managing review requests.
+        /// </summary>
+        IPullRequestReviewRequestsClient ReviewRequests { get; }
```
ryangribble: and here

ryangribble:
Welcome @gdziadkiewicz - thanks for contributing to octokit!  

I'll mark up a few quick "house keeping" things in a review, but it looks like you'll need a bit of guidance about the API request/response models etc as I can see a couple of things that are going to trip things up

ryangribble:
In terms of more structural things, there are 2 main things i've spotted so far

- With the response models, it looks like you are creating a `PullRequestReviewCreate` model for the return of this call, however my take on [the api docs](https://developer.github.com/v3/pulls/review_requests/#list-review-requests) is that this call actually just returns a `PullRequest` object.  So in this case I don't believe we need a new response model, it should just use `PullRequest`.  However it looks like we DO need to include some new fields on the `PullRequest` model (namely the `IReadOnlyList<User> RequestedReviewers` property).

- There are usually "undocumented" api endpoints for each call that has a repository specified...  The traditional/documented calls are for owner and name, but there is usually another one taking the `long repositoryId` instead.  If you take a look at some other pull request related calls you should see the `repositoryId` based calls and what the ApiUrl formatting looks like.  As part of making this change we need to implement the `repositoryId` versions of these calls (unless for some reason it turns out they arent supported... the only way to tell is to probe the github api and see if they work!)

gdziadkiewicz:
RepositoryId work, I have just checked them and added methods for them in code.

I have a problem with the delete methods. According to the docs it should only return appropriate result code, but in practice it does return an object similar to the create method. Should I stick to the docs and return `Task` or should I return `Task<PullRequest>` as this is what is really returned in response?

shiftkey:
> I have a problem with the delete methods. 

This one? https://developer.github.com/v3/pulls/review_requests/#delete-a-review-request

> According to the docs it should only return appropriate result code, but in practice it does return an object similar to the create method.

This seems like an inconsistency - they should be returning `204 No Content`. I'll poke around and talk with the team to confirm this is the intended behaviour. Let's not surface the response here and just use `Task`, as long as the status codes are reliable enough.

Octokit/Models/Response/PullRequest.cs
```diff
@@ -211,6 +211,11 @@ public bool Merged
         /// </summary>
         public bool Locked { get; protected set; }
 
+        /// <summary>
+        /// Users requested for review
+        /// </summary>
+        public IReadOnlyList<User> RequestedReviewers { get; protected set; }
```
ryangribble: Dont forget to add this new property to the `ctor` of the response class as well

Octokit.Tests.Integration/Clients/PullRequestsReviewRequestClientTests.cs
```diff
@@ -0,0 +1,164 @@
...
+    const string path = "CONTRIBUTING.md";
+
+    public PullRequestsReviewRequestClientTests()
+    {
+        _github = Helper.GetAuthenticatedClient();
+
+        _client = _github.PullRequest.ReviewRequest;
+
+        _context = _github.CreateRepositoryContext("test-repo").Result;
+        _github.Repository.Collaborator.Add(Helper.UserName, _context.RepositoryName, collaboratorLogin);
```
ryangribble: When refering to the created `RepositoryContext` I'd prefer we use `_context.RepositoryOwner` rather than `Helper.UserName` - just to keep from making assumptions about the internal workings of repository context

Octokit.Tests.Integration/Clients/PullRequestsReviewRequestClientTests.cs
```diff
@@ -0,0 +1,164 @@
...
+using Octokit.Tests.Integration.Helpers;
+using Xunit;
+
+public class PullRequestsReviewRequestClientTests : IDisposable
+{
+    private readonly IGitHubClient _github;
+    private readonly IPullRequestReviewRequestsClient _client;
+    private readonly RepositoryContext _context;
+
+    const string branchName = "new-branch";
```
ryangribble: Does this need to be a class level const since it's only used in 1 place?

Octokit.Tests.Integration/Clients/PullRequestsReviewRequestClientTests.cs
```diff
@@ -0,0 +1,164 @@
...
+    const string collaboratorLogin = "m-zuber-octokit-integration-tests";
+    const string path = "CONTRIBUTING.md";
```
ryangribble: Ditto with this one, if it's not re-used multiple times it doesn't need to be a class level const

Octokit.Tests.Integration/Clients/PullRequestsReviewRequestClientTests.cs
```diff
@@ -0,0 +1,164 @@
...
+using Xunit;
+
+public class PullRequestsReviewRequestClientTests : IDisposable
+{
+    private readonly IGitHubClient _github;
+    private readonly IPullRequestReviewRequestsClient _client;
+    private readonly RepositoryContext _context;
+
+    const string branchName = "new-branch";
+    const string collaboratorLogin = "m-zuber-octokit-integration-tests";
```
ryangribble: In other integration tests we use "octocat" as our contributor user, so it'd be good to use it here too.  The added bonus is that M-Zuber would be able to run these tests lol 😛

Octokit.Tests.Integration/Clients/PullRequestsReviewRequestClientTests.cs
```diff
@@ -0,0 +1,164 @@
...
+    public void Dispose()
+    {
+        _context.Dispose();
+    }
+
+    /// <summary>
+    /// Creates the base state for testing (creates a repo, a commit in master, a branch, a commit in the branch and a pull request)
+    /// </summary>
+    /// <returns></returns>
+    async Task<PullRequestData> CreatePullRequest(RepositoryContext context, string branch = branchName)
```
ryangribble: For consistency it would be nice to create the base state using the same process/code that the `CreateTheWorld()` functions in the `PullRequestsClientTests` use:   https://github.com/octokit/octokit.net/blob/master/Octokit.Tests.Integration/Clients/PullRequestsClientTests.cs#L931-L993

That way we have consistency in how we create Pull Requests for integration tests

ryangribble:
One other thing we need to sort out with this new API is whether it supports Pagination or not.  If it does, we need to have overloads for the GetAll methods that take `ApiOptions`

Octokit.Tests.Integration/Clients/PullRequestsReviewRequestClientTests.cs
gdziadkiewicz: I'll change it :) But in my defense: I've take it from `RepositoryCollaboratorClientTests.cs` and ran the test.

Octokit.Tests.Integration/Clients/PullRequestsReviewRequestClientTests.cs
gdziadkiewicz: Inlined it.

Octokit.Tests.Integration/Clients/PullRequestsReviewRequestClientTests.cs
gdziadkiewicz: Inlined it.

Octokit.Tests.Integration/Clients/PullRequestsReviewRequestClientTests.cs
gdziadkiewicz: Changed it.

Octokit.Tests.Integration/Clients/PullRequestsReviewRequestClientTests.cs
gdziadkiewicz: I've changed the method name and it's code to be like `PullRequestClientTests.CreateTheWorld`, but without second branch creation and with baked-in pull request creation.

gdziadkiewicz:
I am investigating the pagination support in GetAll method ATM.

Octokit.Tests.Integration/Clients/PullRequestsReviewRequestClientTests.cs
M-Zuber: But, but, then I won't know when someone runs the integration tests 😁

ryangribble: Haha 😀 

yeah as I said there are a fair few inconsistencies with how integration tests are being done across the code base so I just try to focus on new integration tests being done in the preferred way, and perhaps one day we could even go back and restyle the existing ones!

Octokit.Tests.Integration/Clients/PullRequestsReviewRequestClientTests.cs
```diff
@@ -39,6 +39,17 @@ public TheGetAllMethod()
         }
 
         [IntegrationTest]
+        public async Task CanGetAllWhenNoneById()
```
ryangribble: To keep consistency, can we name these "CanxxxxWithRepositoryId" ?

Octokit.Tests.Integration/Clients/PullRequestsReviewRequestClientTests.cs
```diff
@@ -0,0 +1,248 @@
...
+            Assert.Equal(reviewers, pr.RequestedReviewers.Select(rr => rr.Login));
+        }
+
+        public void Dispose()
+        {
+            _context.Dispose();
+        }
+    }
+
+    static async Task<int> CreateTheWorldAndPullRequest(IGitHubClient github, RepositoryContext context)
```
ryangribble: Again, for consistency let's just go with `CreateTheWorld()`

Octokit.Tests/Clients/PullRequestReviewRequestsClientTests.cs
```diff
@@ -0,0 +1,125 @@
...
+                await client.GetAll("owner", "name", 7);
+
+                connection.Received().GetAll<User>(
+                    Arg.Is<Uri>(u => u.ToString() == "repos/owner/name/pulls/7/requested_reviewers"),
+                    null,
+                    "application/vnd.github.black-cat-preview+json");
+            }
+
+            [Fact]
+            public async Task EnsuresNotNullArguments()
```
ryangribble: For consistency, use `EnsuresNonNull` rather than `EnsuresNotNull`.  Despite the fact the former actually sounds better, across the codebase we have 585 cases of NonNull and only 5 of NotNull 🤔

Octokit/Models/Response/PullRequest.cs
gdziadkiewicz: Added it.

ryangribble:
This is looking 💎  @gdziadkiewicz 

Thanks for all your work on this... I'll pull it down tonight and run through the integration tests, then should be good to merge 👏

ryangribble:
Integration tests are good, although we still need to add the pagination integration tests (eg https://github.com/octokit/octokit.net/blob/master/Octokit.Tests.Integration/Clients/ReleasesClientTests.cs#L200-L310 )

Octokit.Tests.Integration/Clients/PullRequestsReviewRequestClientTests.cs
```diff
@@ -9,30 +9,40 @@
 
 public class PullRequestsReviewRequestClientTests
```
ryangribble: Looks like the class name (and filename) are incorrect here:

 PullRequest`s`ReviewRequestClientTests =>  PullRequestReviewRequest`s`ClientTests

Octokit.Tests.Integration/Clients/PullRequestsReviewRequestClientTests.cs
```diff
@@ -9,30 +9,40 @@
...
 {
-    const string collaboratorLogin = "octocat";
+    private static readonly string[] _collaboratorLogins = {
+        "octocat",
+        "gdziadkiewicz-test",
+        "ryangribble-testorg"
```
ryangribble: this is my test organization rather than a user, which is causing tests to fail.

Also it should be possible to prove pagination works with only 2 collaborators (just adjust the pagesizes in the ApiOptions in the tests to 1 rather than 2)

gdziadkiewicz:
@ryangribble Can you check my latest changes? I've reduced the number of users and adjusted the tests. How do you feel about leaving my test user in the tests? It would be nice to have some kind of `octocatPrime` user for the tests.

ryangribble:
OK the test worked for me, i made a couple of tidy ups.  Im having a heck of a time with the test accounts though.  Due to using octocat, my test account got banned/flagged for abuse!  I got in unlocked and tried to create 2 test accounts "octokitnet-test1" and "octokitnet-test2" to use for these tests but those accounts are also now flagged (eventhough they didnt have any activity).  I managed to get all the tests to pass by using my real username and a friend's real username, but I dont want to check that change in.

ryangribble:
The 2 test accounts I created (and am using in the integration tests) should be unblocked now and the tests are working for me.  

Can I get you to review the changes I pushed up @gdziadkiewicz and let me know you're happy?  Once you give the thumbs up Im good to merge this in

gdziadkiewicz:
@ryangribble LGTM 👍

ryangribble:
LGTM

Great work @gdziadkiewicz thanks for contributing this new API implementation to octokit!

ryangribble:
release_notes: Implement support for [Pull Request Review Requests API (Preview)](https://developer.github.com/v3/pulls/review_requests/)

PR #1593, opened 2017-04-26
Remove Rx-Main dependency from samples

### LINKED ISSUE #1592: LINQPad samples are requiring Reactive, but not actually using it
Many of the LINQPad samples have this prompt:
screen shot 2017-04-25 at 23 54 28

...but they don't *actually* use it. Since there's also a bug restoring Rx-Main right now, this creates a blocker for the user. I'll submit a PR soon as time allows cleaning up the samples and removing the extra namespace declarations where not needed if there are no objections.

NickCraver:
This resolves #1592 - LINQPad doesn't understand how to restore this unlisted package and it's not actually needed in the samples.

### DIFF
--- a/samples/linqpad-samples/1-introducing-octokit.linq
+++ b/samples/linqpad-samples/1-introducing-octokit.linq
@@ -1,7 +1,6 @@
 <Query Kind="Program">
   <NuGetReference>Octokit</NuGetReference>
   <NuGetReference>Octokit.Reactive</NuGetReference>
-  <NuGetReference>Rx-Main</NuGetReference>
   <Namespace>Octokit</Namespace>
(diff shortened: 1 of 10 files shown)


shiftkey:
@nickcraver so there's no need to also remove `Octokit.Reactive` from these files?

NickCraver:
I didn't hit any issues with that one, so erred on the side of minimal changes as always :) Just verified on another fresh install: no issues as far as I can see.

shiftkey:
@NickCraver cool, I'll take this in then. Thanks!

NickCraver:
<3

ryangribble:
release_notes: Remove unused Rx-Main dependency from LINQPad samples

Issue #1594, opened 2017-04-27
Timeline - missing 'AddedToProject', 'RemovedFromProject' and more events


karelz:
Looks like variant of #1563 - which doesn't seem to be fixed by looking at source code: https://github.com/octokit/octokit.net/blob/master/Octokit/Models/Response/EventInfo.cs#L74

Got it for dotnet/roslyn issues (numbers: 3 and 5).

I tried 0.24.1-alpha NuGet, which doesn't have even #1563 fix (I hit also LineCommented and CommitCommented). Will try latest master build later ...

```
System.AggregateException: One or more errors occurred. ---> System.ArgumentException: Requested value 'addedtoproject' was not found.
   at System.Enum.EnumResult.SetFailure(ParseFailureKind failure, String failureMessageID, Object failureMessageFormatArgument)
   at System.Enum.TryParseEnum(Type enumType, String value, Boolean ignoreCase, EnumResult& parseResult)
   at System.Enum.Parse(Type enumType, String value, Boolean ignoreCase)
   at Octokit.Internal.SimpleJsonSerializer.GitHubSerializerStrategy.DeserializeObject(Object value, Type type) in D:\\repos\\octokit.net\\Octokit\\Http\\SimpleJsonSerializer.cs:line 124
   at Octokit.PocoJsonSerializerStrategy.DeserializeObject(Object value, Type type) in D:\\repos\\octokit.net\\Octokit\\SimpleJson.cs:line 1478
   at Octokit.Internal.SimpleJsonSerializer.GitHubSerializerStrategy.DeserializeObject(Object value, Type type) in D:\\repos\\octokit.net\\Octokit\\Http\\SimpleJsonSerializer.cs:line 165
   at Octokit.PocoJsonSerializerStrategy.DeserializeObject(Object value, Type type) in D:\\repos\\octokit.net\\Octokit\\SimpleJson.cs:line 1505
   at Octokit.Internal.SimpleJsonSerializer.GitHubSerializerStrategy.DeserializeObject(Object value, Type type) in D:\\repos\\octokit.net\\Octokit\\Http\\SimpleJsonSerializer.cs:line 165
   at Octokit.SimpleJson.DeserializeObject(String json, Type type, IJsonSerializerStrategy jsonSerializerStrategy) in D:\\repos\\octokit.net\\Octokit\\SimpleJson.cs:line 583
[...]

PR #1595, opened 2017-04-27
Add StringEnum to handle unknown enum values returned from API

khellang:
This PR adds `StringEnum<TEnum>` to handle unknown enum values returned from the API.

I added a commit (0677683) to use it for `EventInfo.Event`. This will make it easier to handle cases like https://github.com/octokit/octokit.net/issues/1594 in the future.

This is a binary breaking change, but a lot of code will just compile without any changes (as evidenced by this codebase).

The usage pattern is something like this:

```csharp
var eventInfo = /* get info from API */;

if (eventInfo.Event.TryParse(out var state)
{
    // do something with state...
}

// this was an unknown value, do what you want with that information...
```

You can also access the `ParsedValue` property, but that will throw if the value is unknown.
And lastly, you can get the raw `string` value from the `Value` property.

I didn't get to run the integration test suite locally, so 🤞 for a green build.

### TODO List

- [x] Change serializer to only use `[Parameter()]` attributes (no magic hyphen/underscore removal)
- [x] ~~Sort out Language enum (I'm keen to not exclude it from the convention tests and even change the Request model to use `StringEnum<Language>`~~ Decided to leave this for now
- [x] Consider if we want to add Destructure extensions for c# 7 (may as well, right?)
- [ ] Write up a doc page about how octokit now deals with enums and how consumers should be accessing them going forward
- [ ] Consider whether any existing docs that refer to enums should be altered

### DIFF
--- a/Octokit.Tests.Conventions/ModelTests.cs
+++ b/Octokit.Tests.Conventions/ModelTests.cs
@@ -167,17 +167,6 @@ namespace Octokit.Tests.Conventions
             }
         }
 
-        public static IEnumerable<object[]> GetClientInterfaces()
-        {
-            return typeof(IGitHubClient)
-                .GetTypeInfo()
-                .Assembly
-                .ExportedTypes
-                .Where(TypeExtensions.IsClientInterface)
-                .Where(t => t != typeof(IStatisticsClient)) // This convention doesn't apply to this one type.
-                .Select(type => new[] { type });
-        }
-
         public static IEnumerable<object[]> ModelTypes
         {
             get { return GetModelTypes(includeRequestModels: true).Select(type => new[] { type }); }
--- a/Octokit.Tests.Conventions/ModelTests.cs
+++ b/Octokit.Tests.Conventions/ModelTests.cs
@@ -11,6 +11,8 @@ namespace Octokit.Tests.Conventions
 {
     public class ModelTests
     {
+        private static readonly Assembly Octokit = typeof(AuthorizationUpdate).GetTypeInfo().Assembly;
+
         [Theory]
         [MemberData("ModelTypes")]
         public void AllModelsHaveDebuggerDisplayAttribute(Type modelType)
@@ -183,7 +185,7 @@ namespace Octokit.Tests.Conventions
                 return GetModelTypes(includeRequestModels: true)
                     .SelectMany(type => type.GetProperties())
                     .SelectMany(property => UnwrapGenericArguments(property.PropertyType))
-                    .Where(type => type.Namespace.StartsWith("Octokit") && type.GetTypeInfo().IsEnum)
+                    .Where(type => type.GetTypeInfo().Assembly.Equals(Octokit) && type.GetTypeInfo().IsEnum)
                     .Select(type => new[] { type });
             }
         }
--- a/Octokit.Tests.Conventions/ModelTests.cs
+++ b/Octokit.Tests.Conventions/ModelTests.cs
@@ -11,6 +11,15 @@ namespace Octokit.Tests.Conventions
 {
     public class ModelTests
     {
+        private static readonly HashSet<Type> ExcludedEnumTypes = new HashSet<Type>
+        {
+            typeof(AuthenticationType),
+            typeof(EnterpriseProbeResult),
+            typeof(SearchQualifierOperator),
+            typeof(TeamMembership),
+            typeof(TwoFactorType)
+        };
+
         [Theory]
         [MemberData("ModelTypes")]
         public void AllModelsHaveDebuggerDisplayAttribute(Type modelType)
@@ -188,6 +197,7 @@ namespace Octokit.Tests.Conventions
             {
                 return typeof(IGitHubClient).GetTypeInfo().Assembly.ExportedTypes
                     .Where(type => type.GetTypeInfo().IsEnum)
+                    .Where(type => !ExcludedEnumTypes.Contains(type))
                     .Select(type => new[] { type });
             }
         }
--- a/Octokit.Tests.Conventions/ModelTests.cs
+++ b/Octokit.Tests.Conventions/ModelTests.cs
@@ -11,17 +11,6 @@ namespace Octokit.Tests.Conventions
 {
     public class ModelTests
     {
-        private static readonly HashSet<Type> ExcludedEnumTypes = new HashSet<Type>
-        {
-            typeof(AuthenticationType),
-            typeof(EnterpriseProbeResult),
-            typeof(SearchQualifierOperator),
-            typeof(TeamMembership),
-            typeof(TwoFactorType),
-
-            typeof(Language) // TODO: Add Parameter attributes and include this?
-        };
-
         [Theory]
         [MemberData("ModelTypes")]
         public void AllModelsHaveDebuggerDisplayAttribute(Type modelType)
--- a/Octokit.Tests.Conventions/ModelTests.cs
+++ b/Octokit.Tests.Conventions/ModelTests.cs
@@ -113,7 +113,7 @@ namespace Octokit.Tests.Conventions
         public void ResponseModelsUseStringEnumWrapper(Type modelType)
         {
             var enumProperties = modelType.GetProperties()
-                .Where(x => x.PropertyType.IsEnum);
+                .Where(x => x.PropertyType.GetTypeInfo().IsEnum);
 
             if (enumProperties.Any())
             {
@@ -186,8 +186,8 @@ namespace Octokit.Tests.Conventions
         {
             get
             {
-                return typeof(IGitHubClient).Assembly.ExportedTypes
-                    .Where(type => type.IsEnum)
+                return typeof(IGitHubClient).GetTypeInfo().Assembly.ExportedTypes
+                    .Where(type => type.GetTypeInfo().IsEnum)
                     .Select(type => new[] { type });
             }
         }
--- a/Octokit.Tests.Conventions/ModelTests.cs
+++ b/Octokit.Tests.Conventions/ModelTests.cs
@@ -152,6 +152,11 @@ namespace Octokit.Tests.Conventions
         [MemberData("EnumTypes")]
         public void EnumMembersHaveParameterAttribute(Type enumType)
         {
+            if (enumType == typeof(Language))
+            {
+                return; // TODO: Annotate all Language entries with a ParameterAttribute.
+            }
+
             var membersWithoutProperty = enumType.GetRuntimeFields()
                 .Where(x => x.Name != "value__")
                 .Where(x => x.GetCustomAttribute(typeof(ParameterAttribute), false) == null);
--- a/Octokit.Tests.Conventions/ModelTests.cs
+++ b/Octokit.Tests.Conventions/ModelTests.cs
@@ -17,7 +17,9 @@ namespace Octokit.Tests.Conventions
             typeof(EnterpriseProbeResult),
             typeof(SearchQualifierOperator),
             typeof(TeamMembership),
-            typeof(TwoFactorType)
+            typeof(TwoFactorType),
+
+            typeof(Language) // TODO: Add Parameter attributes and include this?
         };
 
         [Theory]
--- a/Octokit.Tests.Conventions/ModelTests.cs
+++ b/Octokit.Tests.Conventions/ModelTests.cs
@@ -197,9 +197,10 @@ namespace Octokit.Tests.Conventions
         {
             get
             {
-                return typeof(IGitHubClient).GetTypeInfo().Assembly.ExportedTypes
-                    .Where(type => type.GetTypeInfo().IsEnum)
-                    .Where(type => !ExcludedEnumTypes.Contains(type))
+                return GetModelTypes(includeRequestModels: true)
+                    .SelectMany(type => type.GetProperties())
+                    .SelectMany(property => UnwrapGenericArguments(property.PropertyType))
+                    .Where(type => type.Namespace.StartsWith("Octokit") && type.GetTypeInfo().IsEnum)
                     .Select(type => new[] { type });
             }
         }
--- a/Octokit.Tests.Conventions/ModelTests.cs
+++ b/Octokit.Tests.Conventions/ModelTests.cs
@@ -108,6 +108,19 @@ namespace Octokit.Tests.Conventions
         }
 
         [Theory]
+        [MemberData("ResponseModelTypes")]
+        public void ResponseModelsUseStringEnumWrapper(Type modelType)
+        {
+            var enumProperties = modelType.GetProperties()
+                .Where(x => x.PropertyType.IsEnum);
+
+            if (enumProperties.Any())
+            {
+                throw new ModelNotUsingStringEnumException(modelType, enumProperties);
+            }
+        }
+
+        [Theory]
         [MemberData("ModelTypesWithUrlProperties")]
         public void ModelsHaveUrlPropertiesOfTypeString(Type modelType)
         {
--- a/Octokit.Tests.Conventions/ModelTests.cs
+++ b/Octokit.Tests.Conventions/ModelTests.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using Xunit;
 using System.Collections.Generic;
 using System.Reflection;
+using Octokit.Internal;
 
 namespace Octokit.Tests.Conventions
 {
@@ -136,6 +137,20 @@ namespace Octokit.Tests.Conventions
             }
         }
 
+        [Theory]
+        [MemberData("EnumTypes")]
+        public void EnumMembersHaveParameterAttribute(Type enumType)
+        {
+            var membersWithoutProperty = enumType.GetRuntimeFields()
+                .Where(x => x.Name != "value__")
+                .Where(x => x.GetCustomAttribute(typeof(ParameterAttribute), false) == null);
+
+            if (membersWithoutProperty.Any())
+            {
+                throw new EnumMissingParameterAttributeException(enumType, membersWithoutProperty);
+            }
+        }
+
         public static IEnumerable<object[]> GetClientInterfaces()
         {
             return typeof(IGitHubClient)
@@ -165,6 +180,16 @@ namespace Octokit.Tests.Conventions
         public static IEnumerable<object[]> ResponseModelTypes
         {
             get { return GetModelTypes(includeRequestModels: false).Select(type => new[] { type }); }
+        }
+
+        public static IEnumerable<object[]> EnumTypes
+        {
+            get
+            {
+                return typeof(IGitHubClient).Assembly.ExportedTypes
+                    .Where(type => type.IsEnum)
+                    .Select(type => new[] { type });
+            }
         }
 
         private static IEnumerable<Type> GetModelTypes(bool includeRequestModels)
--- a/Octokit.Tests/Clients/BlobClientTests.cs
+++ b/Octokit.Tests/Clients/BlobClientTests.cs
@@ -122,7 +122,7 @@ namespace Octokit.Tests.Clients
             Assert.NotNull(response.Body);
             Assert.Equal(blobResponseJson, (string)response.HttpResponse.Body);
             Assert.Equal(100, response.Body.Size);
-            Assert.Equal(EncodingType.Utf8, response.Body.Encoding.ParsedValue);
+            Assert.Equal(EncodingType.Utf8, response.Body.Encoding);
         }
     }
 }
--- a/Octokit.Tests/Clients/BlobClientTests.cs
+++ b/Octokit.Tests/Clients/BlobClientTests.cs
@@ -122,7 +122,7 @@ namespace Octokit.Tests.Clients
             Assert.NotNull(response.Body);
             Assert.Equal(blobResponseJson, (string)response.HttpResponse.Body);
             Assert.Equal(100, response.Body.Size);
-            Assert.Equal(EncodingType.Utf8, response.Body.Encoding);
+            Assert.Equal(EncodingType.Utf8, response.Body.Encoding.ParsedValue);
         }
     }
 }
--- a/Octokit/Models/Request/MilestoneRequest.cs
+++ b/Octokit/Models/Request/MilestoneRequest.cs
@@ -39,8 +39,10 @@ namespace Octokit
 
     public enum MilestoneSort
     {
-        [Parameter(Value = "due_date")]
+        [Parameter(Value = "due_on")]
         DueDate,
+
+        [Parameter(Value = "completeness")]
         Completeness
     }
 }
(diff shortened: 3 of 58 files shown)


ryangribble:
This is cool!  I've pulled it down to have a play.

I think we should convert ALL response models that have an `enum` property, to wrap it in `StringEnum<>`

I've pushed a a convention test to pick them all up, and help us ensure we keep to this convention moving forward :+1:

Next, Ill push up some changes to the actual response models

ryangribble:
Hmm so there are a few curve balls when actually running this... 

In our SimpleJsonSerializer we take care of converting github api values that contain hyphens or underscores into our C# enum values, for example `utf-8` in GitHub API is `Utf8` in Octokit, and `base_ref_changed` is `BaseRefChanged` so the current implementation fails in `ParsedValue` to parse enum values from strings that contain these characters.  Our deserializer also handles having properties added to enum values to allow us to customize what string representation a given enum value should have, for example in the Reaction API, we have api value of `+1` which maps to c# enum value of `Plus1`.  

It's not nice to duplicate this "API enum strategy" code in both the `JsonSerializer` and the `StringEnum.ParseValue()` so I've rejigged the SimpleJsonSerializer to expose a `DeserializeEnum(string value, Type type)` function, that can now be used in both places.  Let me know what you think, Im about to push the commit up

ryangribble:
I also found that in the implicit conversion of `TEnum` to `StringEnum<TEnum>` the ctor needed to serialize the passed in `TEnum` value to a string using our "GitHub API strategy" rather than just `ToString()` (so that we get the "correct" string value taking into account property mappings etc)... otherwise something like this was failing in Integration Tests:

`Assert.Equal(ReactionType.Plus1, response.Content)`

This is because when `ToString()` was used, the `StringEnum.Value` property was "Plus1" when it needs to be "+1" in order to match the true `StringEnum<ReactionType>` as returned fro the API call (due to the custom `[Property(value="+1")]` attribute on those enum values, to map the GitHub api string to our eunm name).

With the DeserializeEnum and SerializeEnum changes I've just pushed up, I now have a bunch of integration tests working without needing any code changes.

Given the nature of these changes, I'll look at getting a full integration test run done as part of merging this PR (the CI builds dont run integration tests)

ryangribble:
Ive now run all the integration tests on this branch and confirmed everything is good :+1:

khellang:
Nice! Thanks for correcting my sloppy work, @ryangribble :wink: Hopefully the solution will be easier to get build (and run the full test suite) once the migration to SDK-based csproj goes through. 

Right now, it failed to load a couple of projects etc., to it wasn't straight forward to just run the full test suite.

Octokit.Tests/Clients/BlobClientTests.cs
```diff
@@ -122,7 +122,7 @@ public void CanDeserializeBlob()
             Assert.NotNull(response.Body);
             Assert.Equal(blobResponseJson, (string)response.HttpResponse.Body);
             Assert.Equal(100, response.Body.Size);
-            Assert.Equal(EncodingType.Utf8, response.Body.Encoding);
+            Assert.Equal(EncodingType.Utf8, response.Body.Encoding.ParsedValue);
```
khellang: Is this needed?

Octokit/Models/Response/StringEnum.cs
```diff
@@ -0,0 +1,147 @@
...
+
+        public StringEnum(TEnum parsedValue)
+        {
+            if (!Enum.IsDefined(typeof(TEnum), parsedValue))
+            {
+                throw GetArgumentException(parsedValue.ToString());
+            }
+
+            // Use the SimpleJsonSerializer to serialize the TEnum into the correct string according to the GitHub Api strategy
+            _value = new SimpleJsonSerializer().SerializeEnum(parsedValue as Enum);
```
khellang: This feels dirty. Does it need to be an instance method? Can we make this either a) a static method or b) a static field in `StringEnum`?

Octokit/Models/Response/StringEnum.cs
```diff
@@ -0,0 +1,147 @@
...
+            if (string.IsNullOrEmpty(Value))
+            {
+                value = default(TEnum);
+                return false;
+            }
+
+            try
+            {
+                // Use the SimpleJsonSerializer to parse the string to Enum according to the GitHub Api strategy
+                value = (TEnum)new SimpleJsonSerializer().DeserializeEnum(Value, typeof(TEnum));
```
khellang: Same here...

Octokit.Tests/Clients/BlobClientTests.cs
ryangribble: It seems it is currently, eventhough I agree the implicit conversion should mean it's not.  

Ive traced the reason:  

`EncodingType.Utf8` corresponds to an api value of `utf-8` but our response enum doesn't include a custom `[Property()]` attribute on the members.  We still deserialize the API value OK, since the deserializer removes underscores and hyphens... but when we are converting this back the other way (ie with the changes I made in `StringEnum`'s `ctor` we have no way to know that `EncodingType.Utf8` should be `utf-8`.  

To fix this, of course we could add the attribute to this enum, but there is nothing that stops this situation from existing on any other enum value that is only used in API responses.  When an enum is used in a request we HAVE to know what the correct api string value is, but when only used in responses we dont currently (need to) know the value.

One possibility would be to enforce (via convention test) that all enums (both request and response) have custom `Property()` attributes set.  Although this could be a fair few edits, it is the most explicit and obvious approach.  It would also let us remove the "magic" removal of hypens and underscores in the deserializer, instead only ever relying on explicit parameter mapping attributes.  Thoughts?

Octokit/Models/Response/StringEnum.cs
ryangribble: I was trying to avoid changing too much about the current deserializer code.  Both the underlying `GitHubSerializationStrategy` (which implements an interface) and the actual `SimpleJsonSerializer` both use instance methods rather than statics.

Octokit.Tests/Clients/BlobClientTests.cs
ryangribble: I quickly wrote the convention tests to assert this and there are 42 enums that have at least 1 member that doesnt have the `[Parameter()]` attribute specified.

Im not averse to fixing them ALL up as part of this PR, since this PR is effectively "redoing" how we do enums.

ryangribble: I guess another way to tackle this could be to change the equality operator between 2 `StringEnum<TEnum>` to compare their `ParsedValue` (if both of them succesfully can `TryParse()`) and if not, fall back to comparing the underlying string value as it currently does.

This would mean in the case of

`{ Value = "utf8", ParsedValue = EncodingType.Utf8 }` (implicitly converted from `EncodingType.Utf8`)
`{ Value = "utf-8", ParsedValue = EncodingType.Utf8 }` (deserialized from api response "utf-8")

They would correctly be seen as "equal".

Thoughts?

Although more work to change 42 enums, I think Im keen on my earlier suggestion of being explicit with `[Parameter()]` attributes on every enum member, enforced via convention test, and relied on by the deserializer

khellang: > Although more work to change 42 enums, I think Im keen on my earlier suggestion of being explicit with [Parameter()] attributes on every enum member, enforced via convention test, and relied on by the deserializer

I agree. Explicit is better. This lets us simplify the serializer as well. Want to check in the convention test? I can go through and add attributes.

khellang:
@ryangribble I rebased this branch to resolve a conflict. You should probably do a pull before you start work on it 😉

Octokit.Tests/Clients/BlobClientTests.cs
khellang: Actually, why not do both?

ryangribble:
lol I already had done some work locally but all good... ive pushed up the convention test for the Parameter attribute on enum members

khellang:
I'm thinking about changing the names of the values in `StringEnum` to `Value` (parsed) and `RawValue` or `StringValue`. What do you think?

ryangribble:
Yes I was thinking the same about having maybe `StringValue` and `Value`.  

Although I do wonder if it's too much of a surprise for a property called `Value` to potentially throw an exception?

khellang:
@ryangribble Well, that's more or less what it does today, except it'll throw when deserializing. We've essentially deferred throwing to the call site. What we _really_ want. is steer people *away* from the `Value` (and `StringValue`) properties and over to the `TryParse` method. That's the only way the API will be "future proof" when new enum values are added.

ryangribble:
Yeah agree it's better than current explosions 😁 

In terms of getting people over to using `TryParse` I guess the most we can do is update some doco/examples with some guidance (and perhaps write our integration tests that access enums in the same way we would expect others to access enums).

The problem for me is that I personally find `out` vars a real ugly pain and feel like they hamper efforts to have clean/elegant code when you have to declare the variable on a separate line etc... but that is obviously the standard way c# handles these type of things currently (in c# 7 there is an inline `out var` option i believe, and of course tuples to think about, but for now it seems TryParse is the way to go, and at least people have an option in their own control to deal with unknown API response exceptions without having to wait for a new release that added the new enum members

khellang:
Yes, `TryParse(out var value)` or, if we add `Destructure` extension methods to `StringEnum`: `var (success, value) = model.StringEnumProperty`.

Octokit/Models/Request/MilestoneRequest.cs
```diff
@@ -39,8 +39,10 @@ internal string DebuggerDisplay
 
     public enum MilestoneSort
     {
-        [Parameter(Value = "due_date")]
+        [Parameter(Value = "due_on")]
```
khellang: I think this might've been a bug:

> What to sort results by. Either `due_on` or `completeness`. Default: `due_on`
> - https://developer.github.com/v3/issues/milestones/

Octokit/Models/Request/SearchUsersRequest.cs
```diff
@@ -165,15 +168,18 @@ public enum UserInQualifier
         /// Search by the username
         /// </summary>
         [SuppressMessage("Microsoft.Naming", "CA1702:CompoundWordsShouldBeCasedCorrectly", MessageId = "Username")]
+        [Parameter(Value = "login")]
```
khellang: This might've been a bug as well:

> Qualifies which fields are searched. With this qualifier you can restrict the search to just the username (`login`), public email (`email`), full name (`fullname`), or any combination of these.
> \- https://developer.github.com/v3/search/#search-users

Octokit/Models/Response/BranchProtection.cs
```diff
@@ -84,16 +85,19 @@ public enum EnforcementLevel
         /// <summary>
         /// Turn off required status checks for this <see cref="Branch"/>.
         /// </summary>
+        [Parameter(Value = "off")]
         Off,
 
         /// <summary>
         /// Required status checks will be enforced for non-admins.
         /// </summary>
+        [Parameter(Value = "non_admins")]
```
khellang: I couldn't find this in the documentation, but I figured it was deprecated anyway, so... ¯\_(ツ)_/¯

Octokit/Models/Response/Page.cs
```diff
@@ -9,22 +10,32 @@ public enum PagesBuildStatus
...
         ///  The site has yet to be built
         /// </summary>
+        [Parameter(Value = "null")]
         Null,
 
         /// <summary>
+        /// The build has been requested but not yet begun
+        /// </summary>
+        [Parameter(Value = "queued")]
+        Queued,
```
khellang: I found this in the documentation, but not here, so just added it while I was there 😄

Octokit.Tests.Conventions/ModelTests.cs
```diff
@@ -5,11 +5,21 @@
 using Xunit;
 using System.Collections.Generic;
 using System.Reflection;
+using Octokit.Internal;
 
 namespace Octokit.Tests.Conventions
 {
     public class ModelTests
     {
+        private static readonly HashSet<Type> ExcludedEnumTypes = new HashSet<Type>
```
khellang: I excluded some enums from the test because they're not used in the actual API (I couldn't find documentation on them, at least. I think it might be worth crawling the request/response models and find enum types instead of looping through all in the assembly.

khellang:
The only enum left is `Language`. It's pretty hairy and I'm thinking about just adding it to the ignore list for now. What do you think?

ryangribble:
❤️ Awesome pickups @khellang on those enum values that have been incorrect forever 😛 

So as I see it, now we have `[Parameter()]` attributes on all the enums so now we can take the bold step of removing the magic "eat hyphens and underscores" in the deserializer to truly cement this change and make sure we are obeying only the explicit parameter definitions.

But in order to do this, literally EVERY response/request enum must have the attribute, so that means we either have to get `Language` done too, or we turn it into a string.  I guess from a search request perspective, we want to give users a helpful hand in letting them pick from nicely named languages, rather than needing to know the exact string/case the API wants, so I think there is value in providing it as an enum.  But Im thinking perhaps this would be a case where if we make it a StringEnum on the Rerquest model, users also have an option to provide a string value (if the API has a language we dont yet support)

Octokit/Models/Request/MilestoneRequest.cs
ryangribble: OK so this is really weird!  The API docs do say `due_on` however I was running some tests and these were now failing.  When I play around with things it looks like actually the docs are incorrect and our codebase was correct - a value of `due_date` provides the correct behavior!
https://api.github.com/repos/ryangribble-test/public-repo-20170506070755114/milestones?sort=due_on&direction=asc
https://api.github.com/repos/ryangribble-test/public-repo-20170506070755114/milestones?sort=due_on&direction=desc
( both of these produce the same response (milestone1, milestone2)

https://api.github.com/repos/ryangribble-test/public-repo-20170506070755114/milestones?sort=due_date&direction=asc
https://api.github.com/repos/ryangribble-test/public-repo-20170506070755114/milestones?sort=due_date&direction=desc

These ones correctly return the expected orders...

ryangribble: Ive pushed up changes to set this back to "due_date" (and also contacted GitHub support to ask them about the doc/api correctness for this call)

ryangribble:
Not sure how we missed it but `StringEnum.Equals` was comparing itself to itself, rather than itself to `other` LOL!

Luckily it was causing an integration test for Repository Migrations to fail (the test creates a migration and loops until the status is Exported (a StringEnum value) but the test was not waiting for the correct status due to the StringEnum Equals comparison not actually comparing against the "other" value)

ryangribble:
I changed the SImpleJsonSerializer to get rid of the magic hyphen/underscore eating, and have run all integration tests locally, so Im going to push those changes up.

The few remaining things I want us to tackle on this PR are
sort out Language enum (I'm keen to not exclude it from the tests and even change the Request model to use `StringEnum<Language>` 
Consider if we want to add Destructure extensions for c# 7 (may as well, right?)
Write up a doc page about how octokit now deals with enums and how consuers should be accessing them

Ill add a TODO list to the first post

khellang:
> Not sure how we missed it but `StringEnum.Equals` was comparing itself to itself, rather than itself to `other` LOL!

Ouch. That's embarrassing... Good catch! 😖

> Sort out `Language` enum (I'm keen to not exclude it from the tests and even change the request models to use `StringEnum<Language>`

Yeah, I think that makes sense. I wonder if there are other request model enums that could benefit by doing this? I guess `Language` is a good start.

> Consider if we want to add Destructure extensions for C# 7 (may as well, right?)

Yeah, I figured I'd stay away from C# 6/7 features for now. There's quite a few things that could be simplified, but I think that should be a separate PR :smile:

Octokit/Models/Response/IssueComment.cs
```diff
@@ -63,19 +63,4 @@ internal string DebuggerDisplay
             get { return string.Format(CultureInfo.InvariantCulture, "Id: {0} CreatedAt: {1}", Id, CreatedAt); }
         }
     }
-
-    public enum IssueCommentSort
```
ryangribble: I think this is a bug...   Looks like the `IssueCommentRequest` object uses `PullRequestReviewCommentSort` but should actually use `IssueCommentSort`

Octokit.Tests.Conventions/ModelTests.cs
```diff
@@ -197,9 +197,10 @@ public static IEnumerable<object[]> EnumTypes
         {
             get
             {
-                return typeof(IGitHubClient).GetTypeInfo().Assembly.ExportedTypes
-                    .Where(type => type.GetTypeInfo().IsEnum)
-                    .Where(type => !ExcludedEnumTypes.Contains(type))
+                return GetModelTypes(includeRequestModels: true)
+                    .SelectMany(type => type.GetProperties())
+                    .SelectMany(property => UnwrapGenericArguments(property.PropertyType))
+                    .Where(type => type.Namespace.StartsWith("Octokit") && type.GetTypeInfo().IsEnum)
```
ryangribble: minor nitpick but existing convention tests match the assembly to that of `IGitHubClient` rather than hardcode "Octokit".  I guess we'd never change assemblies but I feel like it's cleaner to not hardcode it... ?

mderriey:
I'm keen to help to get this one across if that's OK.

For the `Language` stuff, my understanding is as follows:

 - We have to decorate every member with a `Parameter` attribute, to follow what @khellang ha sbeen doing for the other ones
 - Change all the request/response types that use `Language` to use `StringEnum<Language>`
 - Stop treating `Language` as a special case in the convention test

Is this correct?

ryangribble:
Yep!  Sounds about right :+1:

ryangribble:
Looks like it's only the Search related requests, as other places that deal with language (eg responses for repositories, repository languages, gists) are already just using strings.

Octokit/Models/Response/IssueComment.cs
khellang: Brought it back from the dead 👍

mderriey:
Do we also have to change the serialization code to indicate how to serialize `StringEnum<TEnum>` instances?

khellang:
> Do we also have to change the serialization code to indicate how to serialize `StringEnum<TEnum>` instances?

I'd expect the serializer to just fall back to a call to `ToString`, which should give the correct value, but we have to verify that 😄

mderriey:
> I'd expect the serializer to just fall back to a call to ToString, which should give the correct value, but we have to verify that 😄

I just noticed that search request models explicitly define how they're serialised through the `MergedQualifiers` method.

ryangribble:
OK so based on our conversations in https://github.com/khellang/octokit.net/issues/1 we are going to leave the `Language` enum alone for now... so I think this PR is done from a dev point of view.  @khellang are you sure we dont want to add destructure extensions now?  

The last thing to do is write up some doc/guidance about how we handle response enums (and see whether any existing doc/samples should be changed)

khellang:
> are you sure we dont want to add destructure extensions now?

I think we can manage without them and potentially add them with a big C# 7 upgrade PR later.

Octokit/Models/Response/StringEnum.cs
```diff
@@ -0,0 +1,156 @@
...
+
+        public StringEnum(TEnum parsedValue)
+        {
+            if (!Enum.IsDefined(typeof(TEnum), parsedValue))
+            {
+                throw GetArgumentException(parsedValue.ToString());
+            }
+
+            // Use the SimpleJsonSerializer to serialize the TEnum into the correct string according to the GitHub Api strategy
+            _stringValue = new SimpleJsonSerializer().SerializeEnum(parsedValue as Enum);
```
khellang: We *really* need to get rid of these before this PR is merged. Allocating a `SimpleJsonSerializer` every time a `StringEnum` is created and parsed is hella expensive when it comes to GC. I tried making it static, but for some reason tests started failing, so it seems it actually has shared mutable state 😢

mderriey: I _think_ I found the issue. The enum cache dictionary is not filled with all the values of a specific enum, but only with the value we want to deserialize when calling the method.

After removing [this line](https://github.com/khellang/octokit.net/blob/d3daebc42b39ce5c07a9a93ac0657793aece241c/Octokit/Http/SimpleJsonSerializer.cs#L115), tests pass when `StringEnum` has a static instance of `SimpleJsonSerializer`. Am I missing something here and this is the desired behaviour?

ryangribble: Yeah that does seem like a logic flaw. The idea of the cache is to obviously "speed up" the enum deserialization by not having to evaluate the Parameter property on an enum value more than once 

But that seems to be a logic issue where only the first enum member encountered would be added to the cache...  then next time (for the same enum type) the enum type would already be in the cache (but not other members) and it would actually never look at the parameter property on other members and only do a string conversion. 

Unless I'm missing something,  this line should have actually led to bugs where we ignored the parameter attribute on any enum members other than the first one encountered? How has this not led to other problems in the past! :confused:

So yes I think removing that line is fine but now I want to understand who this hasn't already caused us problems before!

ryangribble: Yep I confirmed this was a bug, and added a failing test to repro it.  If the same `SimpleJsonSerializer` instance tried to deserialize a subsequent enum value that required the Parameter attribute (ie `EncodingType` "ascii" followed by "utf-8") it would throw an exception as it hadn't cached the other enum members, and tried to do a simple `Enum.Parse()` on "utf-8".

I've made the fix identified by @mderriey and confirmed this test now passes, plus changed the `StringEnum` to use a static `SimpleJsonSerializer` and verified the other tests are passing.

Ill push the commits up if you want to have a look @khellang and @mderriey

khellang: Awesome! LGTM ❤️

mderriey:
@ryangribble LYW! :+1:

ryangribble:
Awesome, I think we're done in terms of coding! 

 just need to get something written up in docs about the new enum usage

ryangribble:
Finally got some time to knock up a quick doc file on using `StringEnum<TEnum`, if you guys can take a look and make any suggestions/corrections!

mderriey:
LGTM @ryangribble, thanks for taking the time to do this.

khellang:
Didn't see this inntil now. Looks really good! :heart:

ryangribble:
Ok so I finally got my test account banning/flagging issues sorted and ran through all integration tests... Lucky too, as there were 2 event states that use a hyphen rather than underscore that I picked up 

Appveyor isn't posting back to github for some reason but [it did complete successfully](https://ci.appveyor.com/project/github-windows/octokit-net/build/0.25.0-PullRequest.1595+95.build.2178) so I'm going to merge this shortly :+1:

ryangribble:
Thanks @khellang for your work on this!

LGTM

khellang:
@ryangribble Thank you! ❤️

ryangribble:
release_notes: Provide a robust way to handle unknown enum values returned by GitHub API, to prevent deserialization errors until the enum values can be added to octokit

khellang:
This was a fun PR! :heart:

PR #1598, opened 2017-05-02
Add BranchProtection.EnforceAdmins object

### LINKED ISSUE #1597: Implement Protected Branches API Admin Enforcement Updates
Due to https://developer.github.com/changes/2017-03-22-protected-branches-admin-improvements/, the `BranchProtection` sub classes no longer should accept an `includeAdmins` param in the constructor.
In addition a top level property of `enforce_admins` needs to be added.

M-Zuber:
This replaces the `IncludeAdmins` property on sub-classes.
Fixes #1597 
TODO:
- [x] client methods for dealing directly with the `EnforceAdmins` setting
- [x] unit tests for the client methods
- [x] integration tests for the client methods
Breaking changes:
- `public BranchProtectionSettingsUpdate(BranchProtectionRequiredStatusChecksUpdate requiredStatusChecks, BranchProtectionPushRestrictionsUpdate restrictions)` ctor is now `public BranchProtectionSettingsUpdate(BranchProtectionRequiredStatusChecksUpdate requiredStatusChecks, BranchProtectionPushRestrictionsUpdate restrictions, bool enforceAdmins)`

### DIFF
--- a/Octokit.Reactive/Clients/ObservableRepositoryBranchesClient.cs
+++ b/Octokit.Reactive/Clients/ObservableRepositoryBranchesClient.cs
@@ -499,6 +499,54 @@ namespace Octokit.Reactive
             return _client.DeleteRequiredStatusChecksContexts(repositoryId, branch, contexts).ToObservable().SelectMany(x => x);
         }
 
+        public IObservable<EnforceAdmins> GetAdminEnforcement(string owner, string name, string branch)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
+            Ensure.ArgumentNotNullOrEmptyString(name, "name");
+            Ensure.ArgumentNotNullOrEmptyString(branch, "branch");
+
+            return _client.GetAdminEnforcement(owner, name, branch).ToObservable();
+        }
+
+        public IObservable<EnforceAdmins> GetAdminEnforcement(long repositoryId, string branch)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(branch, "branch");
+
+            return _client.GetAdminEnforcement(repositoryId, branch).ToObservable();
+        }
+
+        public IObservable<EnforceAdmins> AddAdminEnforcement(string owner, string name, string branch)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
+            Ensure.ArgumentNotNullOrEmptyString(name, "name");
+            Ensure.ArgumentNotNullOrEmptyString(branch, "branch");
+
+            return _client.AddAdminEnforcement(owner, name, branch).ToObservable();
+        }
+
+        public IObservable<EnforceAdmins> AddAdminEnforcement(long repositoryId, string branch)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(branch, "branch");
+
+            return _client.AddAdminEnforcement(repositoryId, branch).ToObservable();
+        }
+
+        public IObservable<bool> RemoveAdminEnforcement(string owner, string name, string branch)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
+            Ensure.ArgumentNotNullOrEmptyString(name, "name");
+            Ensure.ArgumentNotNullOrEmptyString(branch, "branch");
+
+            return _client.RemoveAdminEnforcement(owner, name, branch).ToObservable();
+        }
+
+        public IObservable<bool> RemoveAdminEnforcement(long repositoryId, string branch)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(branch, "branch");
+
+            return _client.RemoveAdminEnforcement(repositoryId, branch).ToObservable();
+        }
+
         /// <summary>
         /// Get the restrictions for the specified branch (applies only to Organization owned repositories)
         /// </summary>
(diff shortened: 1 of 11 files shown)


M-Zuber:
I am trying to implement this methods: https://developer.github.com/v3/repos/branches/#add-admin-enforcement-of-protected-branch, but running into trouble since there is no `IApiConnection.Post` overload that does not take data - but that **will** take an `accepts header`.
Due to that, the unit tests are failing

ryangribble:
I think in that case we normally pass an empty body eg `new object()`

M-Zuber:
The unit tests for `RemoveAdminEnforcement` are failing for some odd reason..

M-Zuber:
okaayy....
Wrote up the integration tests, and `include_admins` is still required - even though it might be ignored if the `enforce_admins` param is passed in.
Will rollback the removal and push, but do we obsolete it - or just leave it alone until it is no longer respected by the api?

M-Zuber:
I think I am done with this @ryangribble 
Just one breaking change as outlined in the pr description.

ryangribble:
Based on that API updates blog though, it sounds like as of May 2nd (even in US timezone that's mere hours away) the include_admins may start to cause an error if present?

M-Zuber:
So lets wait a day or two and give AV another kick and see what happens
At this point the property is not used anywhere - it just exists on the class so taking it off will be quick

M-Zuber:
There is also the question of those two unit tests that are failing for some reason. I am quite sure that they are implemented exactly the same way as other tests for methods with the same kind of code

ryangribble:
Sounds like a plan 👍 

Are you able to rebase your branch on master (or merge master into it) so the diffs of all the Url/string changes etc arent shown on this PR?

Octokit.Reactive/Clients/ObservableRepositoryBranchesClient.cs
```diff
@@ -499,6 +499,54 @@ public IObservable<string> DeleteRequiredStatusChecksContexts(long repositoryId,
             return _client.DeleteRequiredStatusChecksContexts(repositoryId, branch, contexts).ToObservable().SelectMany(x => x);
         }
 
+        public IObservable<EnforceAdmins> GetAdminEnforcement(string owner, string name, string branch)
```
ryangribble: The xml docs are missing in this class

M-Zuber:
Sure, did a rebase.
those commits got included by mistake when I was doing some commit cleanup

Octokit.Tests.Integration/Clients/RepositoryBranchesClientTests.cs
```diff
@@ -848,6 +846,143 @@ public TheDeleteRequiredStatusChecksContextsMethod()
...
+        public TheGetAdminEnforcementMethod()
+        {
+            var github = Helper.GetAuthenticatedClient();
+            _client = github.Repository.Branch;
+
+            _userRepoContext = github.CreateRepositoryWithProtectedBranch().Result;
+        }
+
+        [IntegrationTest]
+        public async Task GetAdminEnforcement()
```
ryangribble: Minor quibble but the method names of the tests should be worded as actions, eg Get**s**AdminEnforcement rather than GetAdminEnforcement.  
This is so that the fully qualified test case names look like:  `TheGetAdminEnforcementMethod.GetsAdminEnforcement`

Octokit/Clients/RepositoryBranchesClient.cs
```diff
@@ -538,6 +538,75 @@ public Task<IReadOnlyList<string>> DeleteRequiredStatusChecksContexts(long repos
             return ApiConnection.Delete<IReadOnlyList<string>>(ApiUrls.RepoRequiredStatusChecksContexts(repositoryId, branch), contexts, AcceptHeaders.ProtectedBranchesApiPreview);
         }
 
+        public Task<EnforceAdmins> GetAdminEnforcement(string owner, string name, string branch)
```
ryangribble: Missing xmlDoc in this class

Octokit/Models/Request/BranchProtectionUpdate.cs
```diff
@@ -29,15 +29,32 @@ public BranchProtectionSettingsUpdate(BranchProtectionRequiredStatusChecksUpdate
         /// <summary>
         /// Create a BranchProtection update request
         /// </summary>
+        /// <param name="restrictions">Specifies the requested push access restrictions (applies only to Organization owned repositories). Pass null to disable push access restrictions</param>
+        public BranchProtectionSettingsUpdate(BranchProtectionPushRestrictionsUpdate restrictions)
+        {
+            RequiredStatusChecks = null;
+            Restrictions = restrictions;
+        }
+
```
ryangribble: Should there be a `ctor` taking just the `EnforceAdmins` bool (and setting the other 2 class members to `null`) ?

M-Zuber: yes

M-Zuber:
Review comments addressed.

M-Zuber:
@ryangribble You want to try re-running AV and see what happens with re to the `include_admins` param?

ryangribble:
AppVeyor only runs unit/convention tests, so to really see what github api thinks of those fields we need to run integration tests.

Given [this anouncement](https://developer.github.com/changes/2017-05-02-adoption-of-admin-enforced/) though, I'm p[retty sure we will be good to remove include_admins now 😛

M-Zuber:
😆 I am still pedantic enough to run integration test locally
will ping in a few once I push changes

M-Zuber:
@ryangribble star-wars-haters-the-force-awakens-5xaOcLOqNmWHaLeB14I

ryangribble:
Rollin rollin rollin!

Thanks @M-Zuber 😀 

LGTM

M-Zuber:
Woot!
Thanks for working through this with me, feels so good to get back into OSS
hulu-deadbeat-hulu-original-series-3o6ZtqTJwfo4xoVrNu

ryangribble:
release_notes: Enhance `RepositoryBranchesClient` to support Admin Enforcement changes

PR #1599, opened 2017-05-02
Merge master into dotnetcore

mderriey:
I thought I'd bring the `dotnetcore` branch on par with `master` since there's been some activity on it.


mderriey:
in second thought, is that even needed? or would you rather rebase it and force push when ready to merge?

ryangribble:
I think it's fine, I prefer not to force push a branch that others may have pulled already.

PR #1620, opened 2017-06-25
Remove obsolete branch protection methods/classes

ryangribble:
These earlier Branch Protection methods have been superceded by several revisions over the past year.  

Although technically the old branch protection endpoints are still supported until ~September, the `include_admins` parameter support was removed [recently](https://developer.github.com/changes/2/#--protected-branches-api-admin-enforcement-updates) so these older implementations now no longer work.

Given that they've been flagged `[Obsolete]` for a long time now, and the newer endpoints are implemented in Octokit, we may as well remove these now


M-Zuber:
LGTM

ryangribble:
Thanks @M-Zuber and @khellang for the LGTM's 😁

PR #1623, opened 2017-06-27
Fixes #1621 NewRepositoryWebHook should not discard existing fields

### LINKED ISSUE #1621: NewRepositoryWebHook ToRequest does not transmit all information
Offending line:
https://github.com/octokit/octokit.net/blob/5ee4d6404628e43e5251eb68c85c31878b7d5b78/Octokit/Models/Request/NewRepositoryWebHook.cs#L124

As it's tossing away the existing instance and returning a new one, any of the other properties on the base `NewRepositoryHook` type that might have been set - like for instance, the `events` you'd like to receive are not persisted and never go over the wire.

I assume it does this because `Config` in `NewRepositoryHook` is `private set`, and as such this class cannot make modifications.

I don't really see any issue with making that property `protected set` but I'm not familiar enough with the code base. I can PR that change though if maintainers would prefer!

ctolkien:


ryangribble:
Thanks @ctolkien it looks OK, but I'm just wondering what happens if we call `ToRequest()` on the same  object twice? Since the first one  mutates the underlying `Config` item won't the second time throw the exception because the keys are already in the dictionary?

ctolkien:
@ryangribble  - Off the top of my head, I'm not sure of the behaviour of the Linq bits there.

Let me write some more tests to handle that scenario.

ctolkien:
@ryangribble - you're right in that subsequent `ToRequests` throw in that there is a check in place for existing entries in config...

Two options, we can modify the original code to include the `Events` and `Active` properties:

```csharp
return new NewRepositoryHook(Name, config)
{
    Active = Active,
    Events = Events
};
```

This is less invasive, but I feel is brittle. If there are new properties are added then we need to also add them to this list...

The other option is to have the explicit `NewRepositoryWebHook` _properties_ override any of the matching properties that were explicitly passed into config.

```csharp
var config = new Dictionary<string, string>
{
    { "url" = "http://urlpassedinconfig.com" }
};
var hook = new NewRepositoryWebHook("web", config, "http://urlpassedtowebhook.com");
var request = create.ToRequest();
Assert.Equal(request.Config["url"], "http://test.com/example");
```
In this case, it would use `http://urlpassedtowebhook.com`;

I prefer the latter. In this case we would remove the check for duplicate properties.

ctolkien:
@ryangribble  - FYI, I've updated this PR to reflect the latter option outlined above.

Octokit/Models/Request/NewRepositoryWebHook.cs
```diff
@@ -48,6 +48,7 @@ public class NewRepositoryWebHook : NewRepositoryHook
         /// <summary>
         /// Initializes a new instance of the <see cref="NewRepositoryWebHook"/> class.
         /// Using default values for ContentType, Secret and InsecureSsl.
+        /// <str
```
ryangribble: inadvertent?

ctolkien: Looks like it....

ryangribble:
LGTM

Thanks @ctolkien

ryangribble:
release_notes: NewRepositoryWebHook no longer discards base WebHook fields if they are set

ryangribble:
release_notes: `NewRepositoryWebHook.ToRequest()` no longer discards existing fields if they are set

Issue #1626, opened 2017-07-12
add documentation example about creating an issue and adding assignees


dvdstelt:
When creating a new issue and setting the `Assignee` property, I'm told the property is obsolete and I should use the `Assignees` property. The latter can't be set? Thank goodness for reference types! ;-)

This doesn't work. I get an obsolete warning and also get an error saying `You cannot pass both 'assignee' and 'assignees'. Only one may be provided.`
```
		var newIssue = new NewIssue("My Awesome Title")
		{
			Assignee = staffMember
		};
```

But this does work:
```
var newIssue = new NewIssue("My Awesome Title");
newIssue.Assignees.Add(staffMember);
```

ryangribble:
Using the Add and Remove methods is the intended behaviour, similar to adding repositories to search criteria or labels to issues.   Yyou can see some implementations in the integration tests here https://github.com/octokit/octokit.net/blob/master/Octokit.Tests.Integration/Clients/IssuesClientTests.cs

dvdstelt:
Possible, but at the very least update doco?! :)

dvdstelt:
I'm okay to close this then, btw.

ryangribble:
Actually there are `AddAssignee`,  `RemoveAssignee`  and `ClearAssignee` methods in `IssueUpdate` class too 

> Possible, but at the very least update doco?! :)

For sure, did you you find incorrect reference somewhere? Would you be able to send a PR to correct?

shiftkey:
Looks like `NewIssue.Assignee` has been removed, but I'm going to leave this open as this feels like a good chance to include some example code for this scenario somewhere.

PR #1628, opened 2017-07-12
Replaced `Assignee` with `Assignees`

dvdstelt:
Related to https://github.com/octokit/octokit.net/issues/1626

I've replaced `Assignee` with `Assignees` since the first is obsolete. In my case I also got an error from GitHub that I can't have both `Assignee` and `Assignees` although I didn't dive into the why.
I've also added an example in code, since the error threw me off a bit and I started searching for it in doco, but could not find anything related to `Assignees`. If it's TMI, feel free to remove the code example or anything else.

### DIFF
--- a/docs/issues.md
+++ b/docs/issues.md
@@ -54,7 +54,7 @@ var issues = await client.Issue.GetAllForCurrent(recently);
 `RepositoryIssueRequest` extends `IssueRequest` and adds these options:
 
  - `Milestone` - use `*` for any issue in a milestone, "none" for issues not assigned to a milestone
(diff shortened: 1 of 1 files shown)


ryangribble:
Thanks @dvdstelt! 

There is actually `AddAssignee()` (and Remove and Clear)  methods that are better to use than the underlying reference type's `Add() ` method if you could update the samples to use that. Same for labels

dvdstelt:
I will @ryangribble, very busy atm cleaning up work before I go on holiday. Can this stay open for a short while longer? I might do it while on vacation. :-)

PR #1630, opened 2017-07-18
Fixes #1586 - Repository license API

### LINKED ISSUE #1586: Please support repository's license API
https://developer.github.com/v3/licenses/#get-a-repositorys-license

"license": {
    "key": "mit",
    "name": "MIT License",
    "spdx_id": "MIT",
    "url": "https://api.github.com/licenses/mit"
  },

jozefizso:
Fixes #1586

* [x] license information included in existing calls that return `Repository` objects
* [x] Get a repository's license call
* [x] write unit tests
* [x] write integration tests

### DIFF
--- a/Octokit.Reactive/Clients/ObservableRepositoriesClient.cs
+++ b/Octokit.Reactive/Clients/ObservableRepositoriesClient.cs
@@ -129,7 +129,7 @@ namespace Octokit.Reactive
         /// <returns>A <see cref="IReadOnlyPagedCollection{Repository}"/> of <see cref="Repository"/>.</returns>
         public IObservable<Repository> GetAllPublic()
         {
-            return _connection.GetAndFlattenAllPages<Repository>(ApiUrls.AllPublicRepositories());
+            return _connection.GetAndFlattenAllPages<Repository>(ApiUrls.AllPublicRepositories(), null, AcceptHeaders.LicensesApiPreview);
         }
 
         /// <summary>
@@ -146,7 +146,7 @@ namespace Octokit.Reactive
 
             var url = ApiUrls.AllPublicRepositories(request.Since);
 
-            return _connection.GetAndFlattenAllPages<Repository>(url);
+            return _connection.GetAndFlattenAllPages<Repository>(url, null, AcceptHeaders.LicensesApiPreview);
         }
 
         /// <summary>
@@ -172,7 +172,7 @@ namespace Octokit.Reactive
         {
             Ensure.ArgumentNotNull(options, "options");
 
-            return _connection.GetAndFlattenAllPages<Repository>(ApiUrls.Repositories(), options);
+            return _connection.GetAndFlattenAllPages<Repository>(ApiUrls.Repositories(), null, AcceptHeaders.LicensesApiPreview, options);
         }
 
         /// <summary>
@@ -203,7 +203,7 @@ namespace Octokit.Reactive
             Ensure.ArgumentNotNull(request, "request");
             Ensure.ArgumentNotNull(options, "options");
 
-            return _connection.GetAndFlattenAllPages<Repository>(ApiUrls.Repositories(), request.ToParametersDictionary());
+            return _connection.GetAndFlattenAllPages<Repository>(ApiUrls.Repositories(), request.ToParametersDictionary(), AcceptHeaders.LicensesApiPreview);
         }
 
         /// <summary>
@@ -229,7 +229,7 @@ namespace Octokit.Reactive
             Ensure.ArgumentNotNullOrEmptyString(login, "login");
             Ensure.ArgumentNotNull(options, "options");
 
-            return _connection.GetAndFlattenAllPages<Repository>(ApiUrls.Repositories(login), options);
+            return _connection.GetAndFlattenAllPages<Repository>(ApiUrls.Repositories(login), null, AcceptHeaders.LicensesApiPreview, options);
         }
 
         /// <summary>
@@ -257,7 +257,7 @@ namespace Octokit.Reactive
             Ensure.ArgumentNotNullOrEmptyString(organization, "organization");
             Ensure.ArgumentNotNull(options, "options");
 
-            return _connection.GetAndFlattenAllPages<Repository>(ApiUrls.OrganizationRepositories(organization), options);
+            return _connection.GetAndFlattenAllPages<Repository>(ApiUrls.OrganizationRepositories(organization), null, AcceptHeaders.LicensesApiPreview, options);
         }
 
         /// <summary>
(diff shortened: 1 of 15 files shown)


ryangribble:
Hi @jozefizso just checking if you planned to progress this PR any further?

jozefizso:
I will take a look at it.

jozefizso:
This is my take on implementing `IRepositoriesClient.GetLicenseContents()`

It fixes the `LicenseMetadata` to include new attributes returned by GitHub.
Getting `Repository` will include the license information - this may be a breaking change, as I changed the API Preview header.

Octokit.Reactive/Clients/ObservableRepositoriesClient.cs
```diff
@@ -650,6 +650,36 @@ public IObservable<Repository> Edit(string owner, string name, RepositoryUpdate
...
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="https://developer.github.com/v3/licenses/#get-the-contents-of-a-repositorys-license">API documentation</a> for more details
+        /// </remarks>
+        /// <param name="owner">The owner of the repository</param>
+        /// <param name="name">The name of the repository</param>
+        /// <returns>Returns the contents of the repository's license file, if one is detected.</returns>
+        public IObservable<RepositoryContentLicense> GetLicenseContents(string owner, string name)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
```
ryangribble: Can we please update these to the newer style using `nameof(owner)` rather than hardcoded string `"owner"` ?

Octokit.Tests/Helpers/Arg.cs
```diff
@@ -72,5 +72,10 @@ public static ApiOptions ApiOptions
         {
             get { return Arg.Any<ApiOptions>(); }
         }
+
+        public static string AnyAcceptHeaders
+        {
+            get { return Arg.Any<string>(); }
```
ryangribble: I'd rather be explicit about exactly what Accept Headers are being validated in unit tests, rather than saying "any accept header" is fine.  So can we remove this and be explicit in all the tests?

jozefizso:
GitHub API supports multiple preview values in the `Accept` header separated by comma, or the API may send multiple Accept headers.

I tried this scenario, because to implement #1707 the client will need to indicate is supports License Preview and Topics Preview API.

I left the Squash Commit Preview, because I was not sure it is still required.

Octokit.Tests/Helpers/Arg.cs
jozefizso: I introduced this specifically for the `IsALukeWarmObservable` tests, because these tests already does not care about the actual URI. I think it makes this test more readable and maintainable.

If Accept headers are correct is tested in different set of tests which are specialized for this.

Octokit.Reactive/Clients/ObservableRepositoriesClient.cs
jozefizso: sure

ryangribble:
Awesome, yeah i just did a test myself and verified that `application/vnd.github.mercy-preview+json,application/vnd.github.scarlet-witch-preview+json` returns both `code_of_conduct` and `topics` on the returned repository object, so it's nice to see it does support multiple headers when required.

Leaving the old squash/merge method preview header there is good, in terms of supporting older GitHub Enterprise versions that still have that API in preview.

So :+1: on the multiple headers, but can we be explicit about them in the unit test, and also maybe do a more elegant implementation than `AcceptHeaders.Foo + "," + AcceptHeaders.Bar`?

Eg how about a a `Concat()` function on the `AcceptHeaders` class?

```
public static string Concat(params string[] headers)
{
    return string.Join(",", headers);
}
```
So the usage could be `AcceptHeaders.Concat(AcceptHeaders.Foo, AcceptHeaders.Bar)` ?

jozefizso:
Sure, I can concat it. I opted for non-elegant but more performant code, because it's compiled constant now.

Octokit.Tests/Helpers/Arg.cs
ryangribble: Yep fair enough to leave the "any" in place for these tests since they don't aim to test the actual endpoint

jozefizso: Ok, I will use AnyAcceptHeaders in observable tests.

ryangribble:
If you want to keep it as a constant maybe forget about the concat function bur declare a new const like "SquashCommitAndLicensePreviee" that can be used everywhere required?

jozefizso:
The Concat() method looks good.

jozefizso:
@ryangribble anything else I should fix here?

ryangribble:
Oh sorry, I didnt seem to get notification that you had pushed changes (maybe because you rebase/force pushed).  I'll have a final review

ryangribble:
Everything looks pretty good with what's been done so far :+1:

However I realised that we need to add the preview header to more of the calls in `IRepositoriesClient` so that license info is included in the `Repository` objects returned.  Eg `GetAllForOrg()` `GetAllForUser()` `GetAllForCurrent()` etc

jozefizso:
I put License Preview API header to more RepositoryClient calls. I did not add it to Create() method because it had no effect there. License is updated with delay after the repository is created (with AutoInit=true) and GitHub will not report the license immediately.

ryangribble:
Fair enough regarding the `Create()` call.

Changes look good but the one other thing you may not be aware of is that the Reactive implementations make their own API calls for `GetAll` methods (whereas for single item responses they just pass through the call to the normal API client)... so there are a number of places in `ObservableRepositoriesClient` that will also need the new preview header defined.
 
eg: https://github.com/octokit/octokit.net/blob/master/Octokit.Reactive/Clients/ObservableRepositoriesClient.cs#L232

ryangribble:
@jozefizso just wondering if you still wanted to finish this one off or if you'd like some help to add the `ObservableRepositoriesClient` stuff?

jozefizso:
Sorry, I don't have time now to implement `ObservableRepositoriesClient`.

M-Zuber:
I might be able to pick this one up, what would the simplest way to do that @ryangribble ?

ryangribble:
Simplest in terms of implementation or git branches etc @M-Zuber?

If you mean the latter you can just add a remote to @jozefizso's fork then create a local branch from there, that way the original commits are attributes to @jozefizso and the continuation to you. Push it up to your repo and I can update this PR (assuming @jozefizso has ticked the box to allow me to push to the branch)

jozefizso:
Yes, maintainers can update this pull request.

ryangribble:
@M-Zuber you could also do something like this

```
git checkout -b jozefizso-feature/1586-repository-license master
git pull https://github.com/jozefizso/octokit.net.git feature/1586-repository-license
```

M-Zuber:
I pushed the changes up to https://github.com/M-Zuber/octokit.net/tree/jozefizso-feature/1586-repository-license
let me know if there is anything else that needs to be done

ryangribble:
Sorry @M-Zuber I must have missed this one!  I've merged your changes into this PR and fixed up a couple of impacted unit tests.  Hopefully get a green build now...

ryangribble:
release_notes:  Implement support for Repository Licenses including adding `License` property to existing `Repository` response models, adding `SpxId` field to `LicenseMetadata` response model and a new `IRepositoriesClient.GetLicenseContents()` call

M-Zuber:
KvGkTDzBnY03m

PR #1638, opened 2017-07-29
Add global.json with .NET version pinned to 1.0.1

alfhenrik:
I ran into issue where tests would fail and test runners fail to run, it came down to that as I have the `2.0.0-*` .NET Core stuff installed it was using the latest SDK I have installed which is incompatible with `octokit.net` as it's only `1.0.1`. 

Adding `global.json` with the SDK version pinned to `1.0.1` will make sure that the correct SDK is used.


ryangribble:
Hmm that's weird, I also have the .NET core 2.0.0 stuff installed too and things seem to be fine for me using Cake and VS2017 on windows.  Haven't tried on Mac recently... What environment/method are you using?

ryangribble:
Just tried raw `dotnet test` from commandline in the `Octokit.Tests` directory and that worked too.

Not that Im against adding the global.json per se, just trying to understand why it doesnt seem necessary in my setup.

Having a read of [this overview](https://andrewlock.net/the-net-core-2-0-preview-1-version-numbers-and-global-json/) it also sounds like SDK 2 tools should be able to do everything for 1.x projects, the only slight affect of having the 2.0 stuff installed is that by default `dotnet new` would create a 2.x targeting project (unless you have a global.json saying otherwise).

alfhenrik:
Hmm...OK, running tests from command line works for me (not sure what I was doing this morning), looks like it's ReSharper's test runner that gets into a weird state...so that's fine. Closing this.

PR #1647, opened 2017-08-03
Fix broken JSON deserialization in .NET 4.5

### LINKED ISSUE #1634: Support for .NET Standard broke .NET 4.5
While updating to the new csproj format, it looks like some DefineConstants were dropped from .NET 4.5 builds.  One of these, SIMPLE_JSON_READONLY_COLLECTIONS, broke deserialization of nested arrays in at least some cases.  The CompareResult object, for example, defines its Commits and Files properties as IReadOnlyList<T>, but on .NET 4.5 this Type is excluded from the set of Types considered lists in SimpleJson.cs, preventing their deserialization.

The simple fix is to add SIMPLE_JSON_READONLY_COLLECTIONS to the net45 DefineConstants in Octokit.csproj, and it appears to work well, but it merits further investigation to make sure that this wasn't the only case where the preprocessor directives have introduced issues.

mderriey:
Fixes #1634 

One funny thing is that running tests with `dotnet test` against `net452` is *very* slow on my machine:

 - ~4 seconds for `netcoreapp1.0`
 - more than 2 minutes for `net452`

I want to see if the same thing happens on the build server.
If it does, I trialled using `dotnet xunit` and it was much faster, so that could be a workaround, but we'd have to change the Cake build script, and I don't if there's a wrapper for `dotnet xunit` already since it's pretty early days.

### DIFF
--- a/Octokit.Tests.Conventions/ClientConstructorTests.cs
+++ b/Octokit.Tests.Conventions/ClientConstructorTests.cs
@@ -9,7 +9,7 @@ namespace Octokit.Tests.Conventions
     public class ClientConstructorTests
     {
         [Theory]
-        [MemberData("GetTestConstructorClasses")]
+        [MemberData(nameof(GetTestConstructorClasses))]
         public void CheckTestConstructorNames(Type type)
         {
             const string constructorTestClassName = "TheCtor";
--- a/Octokit.Tests.Conventions/ModelTests.cs
+++ b/Octokit.Tests.Conventions/ModelTests.cs
@@ -14,7 +14,7 @@ namespace Octokit.Tests.Conventions
         private static readonly Assembly Octokit = typeof(AuthorizationUpdate).GetTypeInfo().Assembly;
 
         [Theory]
-        [MemberData("ModelTypes")]
+        [MemberData(nameof(ModelTypes))]
         public void AllModelsHaveDebuggerDisplayAttribute(Type modelType)
         {
             var attribute = modelType.GetTypeInfo().GetCustomAttribute<DebuggerDisplayAttribute>(inherit: false);
@@ -41,7 +41,7 @@ namespace Octokit.Tests.Conventions
         }
 
         [Theory]
-        [MemberData("ResponseModelTypes")]
+        [MemberData(nameof(ResponseModelTypes))]
         public void AllResponseModelsHavePublicParameterlessCtors(Type modelType)
         {
             var ctor = modelType.GetConstructor(Type.EmptyTypes);
@@ -53,7 +53,7 @@ namespace Octokit.Tests.Conventions
         }
 
         [Theory]
-        [MemberData("ResponseModelTypes")]
+        [MemberData(nameof(ResponseModelTypes))]
         public void ResponseModelsHaveGetterOnlyProperties(Type modelType)
         {
             var mutableProperties = new List<PropertyInfo>();
@@ -77,7 +77,7 @@ namespace Octokit.Tests.Conventions
         }
 
         [Theory]
-        [MemberData("ResponseModelTypes")]
+        [MemberData(nameof(ResponseModelTypes))]
         public void ResponseModelsHaveReadOnlyCollections(Type modelType)
         {
             var mutableCollectionProperties = new List<PropertyInfo>();
@@ -111,7 +111,7 @@ namespace Octokit.Tests.Conventions
         }
 
         [Theory]
-        [MemberData("ResponseModelTypes")]
+        [MemberData(nameof(ResponseModelTypes))]
         public void ResponseModelsUseStringEnumWrapper(Type modelType)
         {
             var enumProperties = modelType.GetProperties()
@@ -124,7 +124,7 @@ namespace Octokit.Tests.Conventions
         }
 
         [Theory]
-        [MemberData("ModelTypesWithUrlProperties")]
+        [MemberData(nameof(ModelTypesWithUrlProperties))]
         public void ModelsHaveUrlPropertiesOfTypeString(Type modelType)
         {
             var propertiesWithInvalidType = modelType
@@ -140,7 +140,7 @@ namespace Octokit.Tests.Conventions
         }
 
         [Theory]
-        [MemberData("EnumTypes")]
+        [MemberData(nameof(EnumTypes))]
         public void EnumMembersHaveParameterAttribute(Type enumType)
         {
             if (enumType == typeof(Language))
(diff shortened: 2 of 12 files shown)


mderriey:
Funny, I initially had an empty commit stating that tests were failing after adding the .NET 4.5 target. AppVeyor wouldn't build the PR as it found it was _non-mergeable_.

mderriey:
After asking for help on Twitter, @leastprivilege and @patriksvensson suggested to disable AppDomains when running tests.

That took the `net452` run from 2min+ to 12 seconds on my machine :+1:

ryangribble:
Nice work guys! I'm thinking we should enable the convention tests on both targets too? 

Yeah I was looking at the travis failure, didn't get too far as yet. It complains about the test runner path although we aren't even setting one. The cake guys have merged some PR's recently to add newer dotnet tooling support so we could try updating and seeing if any new options help

mderriey:
After a bit of research, it looks like using `dotnet test` against `net452` when using mono is not supported for now: https://github.com/Microsoft/vstest/issues/679

A possible workaround is to change the Cake process so that the tests are run only against `netcoreapp1.0` when not on Windows.

What do you think @ryangribble?

ryangribble:
Yep I'd been researching this on and off today too and had come across the same item you linked. 

My thoughts were to work around it as you say, by not running net45 2 targeted tests on non windows platforms

ryangribble:
Nice one @mderriey 

I noticed in the appveyor log that there are 4 extra unit tests discovered in net452 target compared to netcoreapp1.0 which is interesting

patriksvensson:
We were running tests on Mono for Cake under 4.5.2 before, so this should be possible. Have you considered using the xunit console runner for running the 4.5.2 tests on Mono?

I can't remember exactly what we were doing, but we've been cross compiling for net452 and netstandard1.6 since july last year. I can take a look later when I have some free time.

mderriey:
I think the issue is specific to `dotnet test`. Using the console runner or `dotnet xunit` should work. Do you know if a Cake wrapper exists for the latter?

mderriey:
@ryangribble I saw the 4 additional tests, too.

I found out they come from tests such as [this one](https://github.com/octokit/octokit.net/blob/master/Octokit.Tests/Exceptions/ApiExceptionTests.cs#L101) where we make sure exceptions can be hydrated from serialized data.

`BinaryFormatter` doesn't exist in .NET Core so they exist only against full .NET framework.

mderriey:
Hmm, I was trialling using `dotnet-xunit` and those serialization tests break when not using app domains:

```
Octokit.Tests.Exceptions.ApiExceptionTests+TheConstructor.CanPopulateObjectFromSerializedData [FAIL]
  System.Runtime.Serialization.SerializationException : Unable to find assembly 'Octokit, Version=0.24.0.0, Culture=neutral, PublicKeyToken=null'.
  Stack Trace:
       at System.Runtime.Serialization.Formatters.Binary.BinaryAssemblyInfo.GetAssembly()
       at System.Runtime.Serialization.Formatters.Binary.ObjectReader.GetType(BinaryAssemblyInfo assemblyInfo, String name)
       at System.Runtime.Serialization.Formatters.Binary.ObjectMap..ctor(String objectName, String[] memberNames, BinaryTypeEnum[] binaryTypeEnumA, Object[] typeInformationA, Int32[] memberAssemIds, ObjectReader objectReader, Int32 objectId, BinaryAssemblyInfo assemblyInfo, SizedArray assemIdToAssemblyTable)
       at System.Runtime.Serialization.Formatters.Binary.__BinaryParser.ReadObjectWithMapTyped(BinaryObjectWithMapTyped record)
       at System.Runtime.Serialization.Formatters.Binary.__BinaryParser.Run()
       at System.Runtime.Serialization.Formatters.Binary.ObjectReader.Deserialize(HeaderHandler handler, __BinaryParser serParser, Boolean fCheck, Boolean isCrossAppDomain, IMethodCallMessage methodCallMessage)
       at System.Runtime.Serialization.Formatters.Binary.BinaryFormatter.Deserialize(Stream serializationStream, HeaderHandler handler, Boolean fCheck, Boolean isCrossAppDomain, IMethodCallMessage methodCallMessage)
```

Some googling took me to https://github.com/xunit/xunit/issues/758. I'm already out of my league, but from what I understand, not using app domains affects assembly loading.

Another thing is I can't reproduce the slowness if I use `dotnet-xunit` with app domains enabled against `net452`.

```
λ  dotnet xunit -configuration Release -nobuild
Detecting target frameworks in Octokit.Tests.csproj...
Locating binaries for framework netcoreapp1.0...
Running .NET Core tests for framework netcoreapp1.0...
xUnit.net Console Runner (64-bit .NET Core 4.6.25211.02)
  Discovering: Octokit.Tests
  Discovered:  Octokit.Tests
  Starting:    Octokit.Tests
    Octokit.Tests.Reactive.ObservableRepositoriesClientTests+TheGetAllForCurrentMethod.StopsMakingNewRequestsWhenTakeIsFulfilled [SKIP]
      See https://github.com/octokit/octokit.net/issues/1011 for issue to investigate this further
  Finished:    Octokit.Tests
=== TEST EXECUTION SUMMARY ===
   Octokit.Tests  Total: 2790, Errors: 0, Failed: 0, Skipped: 1, Time: 3.578s
Locating binaries for framework net452...
Running desktop CLR tests for framework net452...
xUnit.net Console Runner (64-bit Desktop .NET 4.0.30319.42000)
  Discovering: Octokit.Tests
  Discovered:  Octokit.Tests
  Starting:    Octokit.Tests
    Octokit.Tests.Reactive.ObservableRepositoriesClientTests+TheGetAllForCurrentMethod.StopsMakingNewRequestsWhenTakeIsFulfilled [SKIP]
      See https://github.com/octokit/octokit.net/issues/1011 for issue to investigate this further
  Finished:    Octokit.Tests
=== TEST EXECUTION SUMMARY ===
   Octokit.Tests  Total: 2794, Errors: 0, Failed: 0, Skipped: 1, Time: 4.945s
```

patriksvensson:
Yes, assembly loading is affected. We had to remove support for things like TypeConverterAttribute in Cake when turning off AppDomains in tests.

You can use `context.Xunit2` after importing the `Cake.Common.Tools.Xunit` namespace to use the xunit console runner.

ryangribble:
Since we've got AppVeyor anyway I'm pretty happy with skipping the net452 target unit/convention tests on TravisCI until/if it can "just work" without jumping through lots of hoops

ryangribble:
Let me know if you're happy for me to :shipit: @mderriey

mderriey:
@ryangribble Alright, let's do this. I'll still investigate see if there's something we can do about it.

Thanks for your help @patriksvensson!

ryangribble:
Thanks @mderriey 

LGTM

PR #1639, opened 2017-07-29
Add client for organization outside collaborators

alfhenrik:
- [x] List outside collaborators
- [x] Remove an outside collaborator
- [x] Convert a member to an outside collaborator

Relates to #1513

PR #1649, opened 2017-08-07
Fix Uri.ApplyParameters() to handle relative Uri's that have existing parameters

ryangribble:
While investigating why pagination didn't seem to be working on #1639 I discovered a bug in the `Uri.ApplyParameters()` extension method.

When the passed in `Uri` is a relative type and already includes query parameters, the resulting `Uri` generated by this function contains the existing parameters then an additional `?` character followed by the actual set of existing+new parameters.  

This was resulting in the pagination parameters being ignored by the upstream API 
(note the 2 occurrences of `?` and both the "existing" and "new+existing" parameter sets present):

Actual: `orgs/testorg/outside_collaborators?filter=all?per_page=1&page=1&filter=all`
Expected: `orgs/testorg/outside_collaborators?per_page=1&page=1&filter=all`

Unit tests have been extended to cover relative `Uri`'s, which reveals the failure case and confirms the fix.

### DIFF
--- a/Octokit.Tests/Helpers/UriExtensionsTests.cs
+++ b/Octokit.Tests/Helpers/UriExtensionsTests.cs
@@ -23,6 +23,20 @@ namespace Octokit.Tests.Helpers
             }
 
             [Fact]
+            public void AppendsParametersAsQueryStringWithRelativeUri()
+            {
+                var uri = new Uri("issues", UriKind.Relative);
+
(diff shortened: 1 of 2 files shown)


ryangribble:
@alfhenrik once this is merged, we should be good to (re)implement pagination support for #1639

ryangribble:
release_notes: Fix pagination on API calls that use `Uri` parameters (typically for requests that include some form of filtering)

PR #1650, opened 2017-08-07
Implement pagination on Organization Outside Collaborators Client GetAll() method

ryangribble:
It appeared as though this API endpoint didnt support pagination, so `ApiOptions` overloads were dropped from the implementation in #1639 

After GitHub support confirmed pagination should be in place on this endpoint, some debugging revealed 
a bug in octokit (#1649 ) which was causing invalid Uri's that was affecting this endpoint (as it uses a query parameter to specify the filter in the `Uri`)

Once that fix is merged, pagination works OK, so this PR (re)implements the pagination support for the `GetAll()` outside collaborators function and associated integration tests

CC @alfhenrik


ryangribble:
release_notes: Implement pagination support for `OrganizationOutsideCollaboratorsClient.GetAll()` method

PR #1654, opened 2017-08-18
Fix RepositoryCommitsClient.GetSha1 API call after preview period ended

ryangribble:
While running integration tests prepping for a release, I found that the `client.Repository.Commit.GetSha1()` call was returning `null` rather than the expected `string` sha1.

This turned out to be a situation where when the API went from preview to official ([announced here](https://developer.github.com/changes/2016-04-12-commit-reference-sha-api-is-now-official/)], the old preview header is no longer able to be used (despite the blog post saying it is!).

I contacted GitHub support who informed me that the blog post was incorrect in saying the preview header would still work - unlike normal API previews where the accepts header just enables additional/changed json response fields, this particular API call actually had an entirely new media type response (a string SHA1 rather than a json payload) and therefore it required the proper header once it was out of preview.  They also allayed my fears that we need to not send preview headers once things are official, as for the "normal" cases it doesnt matter if we keep sending preview headers, since the new response formats will be returned anyway (and the preview headers help us support older GitHub Enterprise versions, where the feature is still in preview).  

Long story short, this PR fixes the `GetSha1` call to use the official media type `Accepts` header, and updates unit tests to match.  Integration test is now passing

### DIFF
--- a/Octokit.Tests/Clients/RepositoriesClientTests.cs
+++ b/Octokit.Tests/Clients/RepositoriesClientTests.cs
@@ -1062,7 +1062,7 @@ namespace Octokit.Tests.Clients
                 client.GetSha1("owner", "name", "reference");
 
                 connection.Received()
-                    .Get<string>(Arg.Is<Uri>(u => u.ToString() == "repos/owner/name/commits/reference"), null, AcceptHeaders.CommitReferenceSha1Preview);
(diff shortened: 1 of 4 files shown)


ryangribble:
release_notes: Fixed `RepositoryCommitsClient.GetSha1()` to correctly obtain the sha1 of the specified commit, after the API went from preview to official

PR #1659, opened 2017-08-30
Enable pagination convention tests

ryangribble:
Unskip and tweak the existing pagination tests so we can use them to enforce pagination conventions going forward

- ApiOptions overload pagination implemtation
For items that failed tests, probed upstream API and flagged the method as either requiring pagination implementation or API not supporting pagination.  Issues will be raised for each client/method where pagination needs to be implemented

- GetAll naming for methods returning `IReadonlyList<T>`
Renamed offending methods in `RepositoryTrafficClient` and `RepositoryBranchesClient` and flagged old ones as `[Obsolete]` so they can be removed in the future


ryangribble:
The cases where API does support pagination and we havent implemented it, have been excluded from convention tests and issues raised to fix each client.

ryangribble:
release_notes: Enable convention tests to enforce API Pagination support and naming conventions are adhered to

Issue #1617, opened 2017-06-20
Dotnet core in AWS Lambda raises exception


aqwert:
**Issue:**
I took the code from dotnetcore, built and packaged as alpha so that I could use within my dotnet core project. This works fine as a stanad alone app but fails with the following if used within a serverless AWS [Lambda environment and ASPNet](https://github.com/aws/aws-lambda-dotnet/tree/master/Libraries/src/Amazon.Lambda.AspNetCoreServer).

```
at Amazon.Lambda.AspNetCoreServer.APIGatewayProxyFunction.<ProcessRequest>d__15.MoveNext()
PlatformNotSupportedException:
System.PlatformNotSupportedException: Operation is not supported on this platform.
at System.Runtime.InteropServices.RuntimeInformation.get_OSDescription()
at Octokit.Connection.FormatUserAgent(ProductHeaderValue productInformation)
at Octokit.Connection..ctor(ProductHeaderValue productInformation, Uri baseAddress, ICredentialStore credentialStore, IHttpClient httpClient, IJsonSerializer serializer)
```

The code to format the UserAgent checks the OS Description but the call is being rejected.

```
#if !HAS_ENVIRONMENT
                RuntimeInformation.OSDescription,
                RuntimeInformation.OSArchitecture.ToString().ToLowerInvariant(),
#else
```

I suspect that the OSArchitecture call may also fail. Would it be better to exclude this information or does the Github API require it?

Using dotnetcore 1.0

Other related (non Octokit) reports with the same error:
* https://github.com/StackExchange/StackExchange.Redis/issues/582
* https://groups.google.com/forum/#!topic/mongodb-user/l7bVlblY0hU
* https://forums.aws.amazon.com/thread.jspa?messageID=754848&tstart=0 (Dec 6, 2016) (There is code that suggests that RuntimeInformation is not possible in AWS, so returning just a plain old string value)

ryangribble:
Hey @aqwert 

I'd be happy to use the approach they took in the StackExchange one you linked - catch the exception if thrown, and maybe just set the platform/architecture values to "unknown" in that case... 

It would be great if you wanted to fire a PR in to do this!

aqwert:
thanks, I will see what I can do but it won't be for a while... off on holiday for a couple of weeks from tomorrow.

ryangribble:
Just circling back on this one, is this still something you would be interested in contributing @aqwert  ?

aqwert:
[...]

PR #1660, opened 2017-09-02
Fix assembly versioning/properties and handle platform exception

ryangribble:
Thanks to @mderriey for pointing out that the Assembly/File versions are not being set automatically during the build process due to us still having `SolutionInfo.cs` and `AssemblyInfo.cs` files with hardcoded version numbers 🤦‍♂️ 

This PR gets rid of Solution/AssemblyInfo properties in favour of setting everything via the csproj, including some more of the nuget package metadata.  Our existing CAKE build already specifies the GitVersion calculated version string via the `dotnet build` task, so this should now flow through to AssemblyVersion, FileVersion, AssemlbyInformationalVersion attributes.  Similarly the nuget package metadata should be handled by the `dotnet pack` task.  I'll check out the AppVeyor build artifacts to confirm...

Also took the opportunity to tweak the user agent string we calculate, to include the informational version (so we get pre release tags if any are set) rather than assembly version (which can only be numeric and doesn't include pre release tags), and added an attempted fix for #1617 by catching any exceptions when determining the platform OS/architecture


ryangribble:
Downloading the artifacts from AppVeyor, things are looking pretty :gem:

Nuget package info/version is ✅ 
Assembly/File Version is numeric 0.27.0 ✅ 
Product version includes the pre-release version (in this case, `-PullRequest1660`) ✅ 

image

ryangribble:
release_notes: Corrected assembly versioning during build process and allow Octokit to run in environments where `PlatformNotSupported` exception is encountered when initializing the API connection (eg AWS Lambda)

ryangribble:
@mderriey can you take a 👀 and let me know what you think?

mderriey:
It looks fantastic!

One thing I noticed is that in the NuGet Package Explorer screenshot for the `Octokit.Reactive` package, the dependency on `Octokit` mentions the default `0.0.0-dev` version even though it looks these packages have been produced by the Cake build since the package itself has the correct version calculated by GitVersion. That seems a bit strange...

Also, if I had to nitpick, I'd say _dotnetcore_ in the description reads funny to me and if I was a consumer of the package, it'd make more sense for me to read _.NET Core_ or _.NET Standard_.

ryangribble:
Fixed up the wording.

The version dependency of Octokit.Reactive on Octokit is annoying... things were looking so nice and straight forward!  We will have to dig into this

mderriey:
How good would it be if that fixed it!

ryangribble:
OK sounds like for the time being we need to use a workaround of specifying the version on the dotnet restore.  This is because the `\Octokit.Reactive\obj\project.assets.json` file is used for the `dotnet pack` version dependency stuff, and it's contents are set at `dotnet restore` time. 
https://github.com/NuGet/Home/issues/4337

Testing locally looks good, let's see what ApVeyor spits out

mderriey:
Blinging 💯! 

image

ryangribble:
image

mderriey:
Thanks for the wording change.

Also, kudos for the GitHub-fu!

:shipit:

ryangribble:
release_notes: Octokit can now run in environments where `PlatformNotSupported` exception was encountered when initializing the API connection (eg AWS Lambda)

PR #1668, opened 2017-09-05
[WIP]: Share state between tests for the PR review requests tests

### LINKED ISSUE #1667: Share state between integration tests where possible
@ryangribble mentioned that it would be nice if we could share state between some integration tests where possible.

The benefits would be:

 - Faster tests as we could only set up things once for several tests instead of for each tests
 - Less chance of hitting the [API rate limits](https://developer.github.com/v3/#rate-limiting) when running tests

One area he flagged as a good candidate for this is the PR review requests tests.
Each of [these methods](https://github.com/octokit/octokit.net/blob/366ac261bf20b9e30d438b9cbd1a9a06aa6a3f25/Octokit.Tests.Integration/Clients/PullRequestReviewRequestsClientTests.cs#L66-L166) have the same setup:

 - Create a test repo
 - Add 2 collaborators to it
 - Create some files and commit them
 - Create a separate branch, make changes and commit them
 - Create a PR
 - Create a review request

These actions add up to ~15 API calls; on top of them, we have the actual tests API calls. It all adds up pretty quickly.

The idea would be to execute these actions only once and share the state between several methods. Obviously, the tests must not have side effects on the state so we don't lose the reliability we have now.

mderriey:
Fixes #1667 

That's a first go at trying to share state between some integration tests.

It got a bit crazy with the inheritance to differentiate cases where we want to create the review request and when we don't, so I'm more than happy to get feedback on the _design_.

Unfortunately, I couldn't test they worked as after inviting the test accounts as collaborators to the test repository, they need to accept the invitation before they can be added as PR reviewers.

Let's try to identify a list of areas where we can do the same thing:

 - [ ] PR review requests

### DIFF
--- a/Octokit.Tests.Integration/Clients/PullRequestReviewRequestsClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/PullRequestReviewRequestsClientTests.cs
@@ -111,31 +111,6 @@ public class PullRequestReviewRequestsClientTests
         }
     }
 
-    public class PullRequestReviewRequestClientTestsBase : IDisposable
-    {
-        internal readonly IGitHubClient _github;
-        internal readonly IPullRequestReviewRequestsClient _client;
-        internal readonly RepositoryContext _context;
-
-        public PullRequestReviewRequestClientTestsBase()
-        {
-            _github = Helper.GetAuthenticatedClient();
-
-            _client = _github.PullRequest.ReviewRequest;
-
-            _context = _github.CreateRepositoryContext("test-repo").Result;
-
-            Task.WaitAll(_collaboratorLogins.Select(AddCollaborator).ToArray());
-        }
-
-        private Task AddCollaborator(string collaborator) => _github.Repository.Collaborator.Add(_context.RepositoryOwner, _context.RepositoryName, collaborator);
-
-        public void Dispose()
-        {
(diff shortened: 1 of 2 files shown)


mderriey:
OK so things changed a bit as @ryangribble and I realised that the tests for the `GetAll` method need at least 2 review requests to test the pagination works as expected. Managing 3 accounts in the code would get a bit difficult, so it was decided to use a static repo for these tests. See [the `octokitnet-test` organisation](https://github.com/octokitnet-test) and [the specific PR](https://github.com/octokitnet-test/pr-review-request/pull/1) we use for these tests.

This change means that the state sharing between tests doesn't happen as much as the `GetAll` method would have been the one benefit from it the most, as we have 6 test methods for it.

We still take advantage of it, though, but only in [one place](https://github.com/mderriey/octokit.net/blob/f060b6d66c5faaea4425ecd314a1a9d327c5fa7b/Octokit.Tests.Integration/Clients/PullRequestReviewRequestsClientTests.cs#L133-L164).

More than happy to get feedback on this.
The whole thing feels somewhat over-engineered with the abstract class, so feel free to bounce ideas around!

mderriey:
Closing as it feels heavy compared to what it achieves.

PR #1670, opened 2017-09-06
Implement Team Membership changes

### LINKED ISSUE #1676: Ensure Membership related changes from Organization Permissions Enhancements are implemented
As per [this old blog post](https://developer.github.com/changes/2015-06-24-api-enhancements-for-working-with-organization-permissions/#team-maintainers) there were various changes made to Organization Team Membership related to the concept of Team Maintainers.

These changes have not yet been implemented in Octokit

Extract from the blog post:
- The Add team membership API accepts a `role` parameter, so that you can specify whether a given team member should be a maintainer or a regular member.
- The List team members API accepts an optional `role` parameter, allowing you to fetch only maintainers or only regular members.
- In the Get team membership and Add team membership APIs, the response includes a `role` attribute, indicating whether a user is a maintainer or a regular member of the team.

ryangribble:
Fixes #1676 

- Implement `TeamsClient.AddOrEditMembership` taking a request parameter to indicate the `role` of the member to be added/updated (Member or Maintainer)
- Implement role filter on `TeamsClient.GetAllmembers()` so that the query can be filtered based on the member roles (Member, Maintainer or All)
- Introduce new response type for Team membership which includes the `Role` (Member or Maintainer) and `Status` (Active or Pending) of the membership.

Due to needing to safely deprecate the old `GetMembership` method following our normal deprecation schedule, the new method has been named `GetMembershipDetails` as the method signature was the same except for the return type. 

The `AddMembership` method has been deprecated, replaced with `AddOrEditMembership`

### DIFF
--- a/Octokit/Models/Response/TeamMembership.cs
+++ b/Octokit/Models/Response/TeamMembership.cs
@@ -2,7 +2,7 @@
 
 namespace Octokit
 {
-    [Obsolete("This enum will be replaced with a response class TeamMembershipDetails")]
+    [Obsolete("Please use TeamMembershipDetails response class instead")]
     public enum TeamMembership
     {
         NotFound = 0,
--- a/Octokit/Models/Response/TeamMembership.cs
+++ b/Octokit/Models/Response/TeamMembership.cs
@@ -1,5 +1,8 @@
-﻿namespace Octokit
+﻿using System;
+
+namespace Octokit
 {
+    [Obsolete("This enum will be replaced with a response class TeamMembershipDetails")]
     public enum TeamMembership
     {
         NotFound = 0,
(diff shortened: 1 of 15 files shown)


ryangribble:
release_notes: Implemented the request filter for `TeamsCient.GetAllMembers()` to allow specifying `Member`, `Maintainer` or `All` roles

ryangribble:
anyone able to have a 👀 at this?  Id like to merge this and #1682 then cut a new release

Octokit/Models/Response/TeamMembership.cs
```diff
@@ -1,5 +1,8 @@
-﻿namespace Octokit
+﻿using System;
+
+namespace Octokit
 {
+    [Obsolete("This enum will be replaced with a response class TeamMembershipDetails")]
```
M-Zuber: `Will be` or already is?

ryangribble:
Thanks @M-Zuber I've updated the obsolete message to be more consistent with others

PR #1674, opened 2017-09-09
Enable generation of XML doc files

### LINKED ISSUE #1673: Missing XML documentation files
@ryangribble noticed that the new NuGet packages are missing the XML documentation files, and as a result there's no IntelliSense anymore.

It'd be great to have them again as a lot of effort is put into them to make it easier for people to consume the libraries.

mderriey:
Fixes #1673 

That post by Nate McMaster has saved my bacon a few times now: http://www.natemcmaster.com/blog/2017/01/19/project-json-to-csproj/#other-common-build-options

I think we lost the documentation files when we moved from the `project.json` project system to the new `.csproj` one.

I tested by running the `Package` build task and the XML files ended up in the `.nupkg`s which I suppose will fix this.

### DIFF
--- a/Octokit.Reactive/Octokit.Reactive.csproj
+++ b/Octokit.Reactive/Octokit.Reactive.csproj
@@ -5,6 +5,7 @@
     <AssemblyTitle>Octokit.Reactive</AssemblyTitle>
     <Authors>GitHub</Authors>
     <Version>0.0.0-dev</Version>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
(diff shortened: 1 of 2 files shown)


ryangribble:
Thanks for the fix @mderriey! 

In vs2017 alot of these csproj properties can even be set in the GUI... Just have to remember to look!

ryangribble:
release_notes: Intellisense should once again be available for Octokit libraries - sorry about that!

ryangribble:
Confirmed by consuming the nuget packages from the AppVeyor CI feed :+1:

Many thanks @mderriey 

LGTM

PR #1679, opened 2017-09-11
Provide constructors for EnforceAdmins class to allow mocking

### LINKED ISSUE #1678: Unable to mock BranchProtection calls due to missing ctor on `EnforceAdmins` response class
While writing tests for an API that uses Octokit I was unable to mock out the BranchProtection responses, as `EnforceAdmins` has no `ctor` that allows the response object to be created with the desired fields.

ryangribble:
Fixes #1678


ryangribble:
BranchProtection response class `EnforceAdmins` now provides a standard `ctor` allowing it to be mocked if required

PR #1680, opened 2017-09-11
Fix SearchRepository forks qualifier

### LINKED ISSUE #1677: Search repository fork qualifier isnt working
Need to investigate why attempting to search for a forked repository is not finding the expected result, when specifying `request.Fork = ForkQualifier.IncludeForks`

ryangribble:
Fixes #1677 

The forks qualifier should be `fork:true` or `fork:only` and was incorrectly being set to the enum member `.ToString()` rather than the decorated `[Parameter()]` values.

- Changed the `MergedQualifiers` function to use the decorated `[Parameter]` attribute value for the enum members
- Changed the values to lowercase for consistency
- Added unit and integration tests for both Forks values (Include and Only)

### DIFF
--- a/Octokit.Tests.Integration/Clients/SearchClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/SearchClientTests.cs
@@ -24,6 +24,30 @@ public class SearchClientTests
     }
 
     [IntegrationTest]
+    public async Task SearchForForkedRepositories()
+    {
+        var request = new SearchRepositoriesRequest("octokit")
+        {
+            Fork = ForkQualifier.IncludeForks
+        };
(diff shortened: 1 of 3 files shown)


ryangribble:
release_notes: `ISearchClient.SearchRepo()` now correctly handles the `Forks` search qualifiers (`Include` or `Only`)

ryangribble:
This is a pretty minor change, and the tests I added are passing, but it would be good if someone can give it a quick 👀  to make sure I havent missed anything

M-Zuber:
LGTM

ryangribble:
thanks @M-Zuber

M-Zuber:
alice-in-wonderland-thank-you-xIJLgO6rizUJi

Issue #1688, opened 2017-10-09
Unable to Set Root Path in GetAllContentsByRef That Doesn't Revert To The Default Ref


thedillonb:
`Repository.GetAllContentsByRef` has a problem where it's `reference` parameter is guarded by a Empty-Null check but it calls `ApiUtils.RepositoryContent` which contains the following formatter:

`"repos/{0}/{1}/contents/{2}?ref={3}".FormatUri(owner, name, path, reference);`

To return the root's contents I'd have to pass an empty string here because if I pass something like "/" as `path` then the following gets constructed:

`repos/bla/foo/contents//?ref=something`

Which is not only goofy, but also causes a `302` redirect that strips off the `ref` query parameter - which may be considered a bug in the site's forwarding logic but the URL really shouldn't have a double slash in it.

thedillonb:
As @ryangribble pointed out, this is unnecessary if you use the overloads of `GetAllContentsByRef` which drop the path and default to the root dir.

PR #1689, opened 2017-10-09
Allow Path To Be An Empty String

thedillonb:
Addresses #1688

### DIFF
--- a/Octokit/Clients/RepositoryContentsClient.cs
+++ b/Octokit/Clients/RepositoryContentsClient.cs
@@ -33,7 +33,7 @@ namespace Octokit
         {
             Ensure.ArgumentNotNullOrEmptyString(owner, "owner");
             Ensure.ArgumentNotNullOrEmptyString(name, "name");
-            Ensure.ArgumentNotNullOrEmptyString(path, "path");
+            Ensure.ArgumentNotNull(path, "path");
 
             var url = ApiUrls.RepositoryContent(owner, name, path);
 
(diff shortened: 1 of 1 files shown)


thedillonb:
I figured the tests would fail. Regardless, just wanted to throw something up to get discussion moving on this. Either we allow empty strings to be passed as "paths" or we allow "/" to be passed and alter the logic following the `format` of the URL to prevent a double slash.

ryangribble:
Whilst I'm not opposed to the change, were you aware that there are already overloads (that don't take a `path` parameter) which operate on the root directory of the repository?

https://github.com/octokit/octokit.net/blob/1e474f8556018ade97cc2d065aadb73aade74ca9/Octokit/Clients/IRepositoryContentsClient.cs#L48
https://github.com/octokit/octokit.net/blob/1e474f8556018ade97cc2d065aadb73aade74ca9/Octokit/Clients/IRepositoryContentsClient.cs#L58

https://github.com/octokit/octokit.net/blob/1e474f8556018ade97cc2d065aadb73aade74ca9/Octokit/Clients/IRepositoryContentsClient.cs#L96
https://github.com/octokit/octokit.net/blob/1e474f8556018ade97cc2d065aadb73aade74ca9/Octokit/Clients/IRepositoryContentsClient.cs#L108

thedillonb:
@ryangribble well thats embarrassing. Good point. Didn't think to consider the other overloads. Those will do. I'll close the PR and issue. Thanks for pointing that out.

ryangribble:
It obviuosly isn't disocverable though, given you didn't spot them...  

If we implicitly support an empty string via a different overload anyway, I dont have an issue with allowing the empty string to be explicitly passed in as `path`.  So I'd be happy to still take this PR if you got the tests working and updated the XmlDoc comment for the parameter to note that "Pass empty string for the root directory"

PR #1693, opened 2017-10-17
Add possibility to configure GitHubClient timeout (#963)

### LINKED ISSUE #963: Uploading release assets in parallel results in TaskCanceledException

pmiossec:
A first attempt to fix the problem describe in #963 by adding a possibility to extend the default timeout value (100s) that is too short to be able to post assets in github release.

Fixes #963

/ cc @ryangribble

### DIFF
--- a/Octokit/GitHubClient.cs
+++ b/Octokit/GitHubClient.cs
@@ -107,9 +107,9 @@ namespace Octokit
         /// See more information here: https://technet.microsoft.com/library/system.net.http.httpclient.timeout(v=vs.110).aspx
         /// </remarks>
         /// <param name="timeout">The Timeout value</param>
-        public void SetRequestsTimeout(TimeSpan timeout)
+        public void SetRequestTimeout(TimeSpan timeout)
         {
-            Connection.SetRequestsTimeout(timeout);
+            Connection.SetRequestTimeout(timeout);
         }
 
         /// <summary>
--- a/Octokit/GitHubClient.cs
+++ b/Octokit/GitHubClient.cs
@@ -100,6 +100,19 @@ namespace Octokit
         }
 
         /// <summary>
+        /// Set the GitHub Api request timeout.
+        /// Useful to set a specific timeout for lengthy operations, such as uploading release assets
+        /// </summary>
+        /// <remarks>
+        /// See more information here: https://technet.microsoft.com/library/system.net.http.httpclient.timeout(v=vs.110).aspx
+        /// </remarks>
+        /// <param name="timeout">The Timeout value</param>
+        public void SetRequestsTimeout(TimeSpan timeout)
+        {
+            Connection.SetRequestsTimeout(timeout);
+        }
+
+        /// <summary>
         /// Gets the latest API Info - this will be null if no API calls have been made
         /// </summary>
         /// <returns><seealso cref="ApiInfo"/> representing the information returned as part of an Api call</returns>
--- a/Octokit.Reactive/IObservableGitHubClient.cs
+++ b/Octokit.Reactive/IObservableGitHubClient.cs
@@ -14,7 +14,7 @@ namespace Octokit.Reactive
         /// See more information here: https://technet.microsoft.com/library/system.net.http.httpclient.timeout(v=vs.110).aspx
         /// </remarks>
         /// <param name="timeout">The Timeout value</param>
-        void SetRequestsTimeout(TimeSpan timeout);
+        void SetRequestTimeout(TimeSpan timeout);
         
         IObservableAuthorizationsClient Authorization { get; }
         IObservableActivitiesClient Activity { get; }
--- a/Octokit.Reactive/IObservableGitHubClient.cs
+++ b/Octokit.Reactive/IObservableGitHubClient.cs
@@ -6,6 +6,16 @@ namespace Octokit.Reactive
     {
         IConnection Connection { get; }
 
+        /// <summary>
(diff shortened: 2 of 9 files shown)


ryangribble:
Any reason why you've gone with an extra function call rather than adding a parameter to the constructors?  In your usage would you only set this larger timeout value right before you do an assett upload, and then reset it again for other calls, or would you set the timeout once globally and have all calls use the larger timeout value?

Octokit/GitHubClient.cs
```diff
@@ -100,6 +100,25 @@ public GitHubClient(IConnection connection)
         }
 
         /// <summary>
+        /// Set the timespan to wait before the requests to GitHub api times out.
+        ///
+        /// Exceptions:
```
ryangribble: We can drop these Exception details from xmldoc comments... it's not that they aren't informative but just for consistency purposes (since we don't include this detail on any/many other documented calls).  Just `<summary>` and `<param>` entries is fine

pmiossec:
>Any reason why you've gone with an extra function call rather than adding a parameter to the constructors? [...] 

You guessed right. 

In the fix I made, I just hardcoded the timeout value to fulfil our needs that was just uploading an asset ;-)

But but because that is a library, we can't enforce a use that will not fit to every use. Timeouts could be useful.

So, indeed, I provide this method to be able to change it just before uploading an asset (and set it back to a normal value just after).

But you are right, I can also add an optional parameter to the constructor to make it more convenient to use. I will do it.

Octokit/GitHubClient.cs
pmiossec: Ok I will do it.

I think I will replace it by a link towards the `HttpClient.Timeout` doc. What do you think of that?

Or completely remove? (it will be a shame because it could be useful)

pmiossec:
>But you are right, I can also add an optional parameter to the constructor to make it more convenient to use.

In fact, I had a look and I didn't find to make the API friendly to use without doubling the number of constructor of `GitHubClient` and `Connection` classes. I'm not sure to willing to do that anymore....

Octokit/GitHubClient.cs
pmiossec: I push the fix...

ryangribble:
Yeah the constructors are a bit messy... I agree perhaps leave it to the discrete method call.  Since it is more likely people would change timeout only when uploading a large asset, then change timeout back again afterwards, this should be ok

Octokit/GitHubClient.cs
```diff
@@ -100,6 +100,19 @@ public GitHubClient(IConnection connection)
         }
 
         /// <summary>
+        /// Set the timespan to wait before the requests to GitHub api times out.
+        /// Especially usefull to set a specific timeout that fit your needs to upload assets in a release
```
ryangribble: Apart from the spelling mistake (useful has one L :)) I don't like the phrasing of the sentence sorry

How abou:

> /// Set the GitHub Api request timeout.
> /// Useful to set a specific timeout for lengthy operations, such as uploading release assets

Octokit/GitHubClient.cs
```diff
@@ -100,6 +100,19 @@ public GitHubClient(IConnection connection)
...
+        /// </summary>
+        /// <remarks>
+        /// See more information here: https://technet.microsoft.com/library/system.net.http.httpclient.timeout(v=vs.110).aspx
+        /// </remarks>
+        /// <param name="timeout">the timespan to wait before a request to GitHub api times out</param>
```
ryangribble: Can we simplify to just simply:

> The timeout value

?

pmiossec:
@ryangribble I have made the changes suggested...

ryangribble:
Can you set "The timeout value" xml comment on the other cases too please? ☺

Last thing, it would be great to have a unit test that "proves" this works. 

Something like mock out an IHttpClient then create a githubclient/connection using the mock, call the new SetTimeout method and verify the mock's timeout was changed. 

I'm on my phone so can't say exactly how to do it right now, let me know if you need more help

pmiossec:
done.

ryangribble:
Thanks @pmiossec I just mad a couple of tidy ups to the xml comments and renamed the function to `SetRequestTimeout` from `SetRequestsTimeout` for consistency with the codebase :+1:

ryangribble:
release_notes: You can now use `IGitHubClient.SetRequestTimeout(TimeSpan timeout)` to set a custom timeout, particularly useful for lengthy operations such as uploading release assets.

ryangribble:
Thanks heaps @pmiossec 

LGTM

PR #1703, opened 2017-11-04
Prevent overflow exceptions when status id > int.Max

kzu:
I have a (private) repository where GitHub reports statuses with
ids like 4299360515, which fail to deserialize as it doesn't fit
on the int.

There is nothing (I could find) on the GitHub API docs that say
that this status id is limited to an int-sized value.

Fixes #1702

PR #1708, opened 2017-11-06
CommitStatus IDs are 64bit, not 32bit

alanmcgovern:
* The API returns 64bit values for commit status. For example
  '4299360515'. This is larger than uint32.MaxValue so i assume
  this type *should* be a long.


ryangribble:
Thanks @alanmcgovern, as you've realised we've already had #1703 submitted so I'll close this one :+1:

PR #1713, opened 2017-11-10
Support running in aws lambda by fixing netcoreapp1.1 vs netcoreapp1.0 confusion 

### LINKED ISSUE #1712: netcoreapp1.0 support
Octokit currently targets netstandard1.1 but did not specify a specific metapackage, so it defaults to NETStandard.Library 1.6.1. This metapackage is compatible only with netcoreapp1.1.

I'm trying to use this package in AWS Lambda, which only supports netcoreapp1.0.

The fix is simple, add `<NetStandardImplicitPackageVersion>1.6.0</NetStandardImplicitPackageVersion>` element to the package project file.

References:

- AWS Lambda limitation: https://github.com/aws/aws-sdk-net/issues/684
- Metapackage: https://github.com/dotnet/core/blob/master/release-notes/1.0/sdk/1.0-rc3-implicit-package-refs.md

ryangribble:
Fixes #1712 

AWS Lambda requires "strict" `netcoreapp1.0` support and although we target `netstandard1.1` which is compatible with `netcoreapp1.0`, the dotnet tooling automatically uses the NetStandard.Library metapackage version 1.6.1 (which is more aligned with `netcoreapp1.1`).  I'm pretty sure we are still actually `netcoreapp1.0` compatible (as evidenced by our test assemblies being `netcoreapp1.0`) but due to the way AWS Lambda tools work, they see the metapackage 1.6.1 version and refuse to deploy.

This PR forces the implicit meta package version to 1.6.0 so that the AWS Lambda tools don't complain about a lambda that is referencing octokit.

### DIFF
--- a/Octokit.Reactive/Octokit.Reactive.csproj
+++ b/Octokit.Reactive/Octokit.Reactive.csproj
@@ -17,6 +17,10 @@
     <Copyright>Copyright GitHub 2017</Copyright>
   </PropertyGroup>
 
+  <PropertyGroup Condition=" '$(TargetFramework)' == 'netstandard1.1' ">
+    <NetStandardImplicitPackageVersion>1.6.0</NetStandardImplicitPackageVersion>
+  </PropertyGroup>
+
   <ItemGroup>
     <Compile Include="..\Octokit\Helpers\Ensure.cs;..\Octokit\Helpers\Pagination.cs" />
     <None Include="app.config" />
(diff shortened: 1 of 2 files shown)


ryangribble:
Can you foresee any issues with this @mderriey @shiftkey  ?

shiftkey:
@ryangribble seems legit 🤷‍♂️

ryangribble:
release_notes: Ensure the `netstandard1.1` targeted package is compatible with `netcoreapp1.0` (eg AWS Lambda environment) by ensuring the `NetStandard.Library` reference is an appropriate version

ryangribble:
I don't think this should break anything based on my own testing, but there's only one sure way to find out...

PR #1714, opened 2017-11-12
Change integration tests configuration script test account warning color to yellow.

gdziadkiewicz:
The warning in configuration script can be easily overlooked. I changed it's color to yellow to make it more visible.

How it looks after the change:
image


ryangribble:
Thanks @gdziadkiewicz it definitely pays to have more obvious "warning"

Issue #1717, opened 2017-11-13
Support alternative payload for adding a PullRequest Comment


ryangribble:
As per [the API documentation](https://developer.github.com/v3/pulls/comments/#alternative-input) it is possible to add a PR Comment "in reply to" an existing comment, by using an alternative payload

``` json
{
  "body": "Nice change",
  "in_reply_to": 4
}
```

In octokit, we would implement this by creating a new request class (eg `PullRequestReviewCommentReply`) and adding new overloads to the regular/observable clients:

`Task<PullRequestReviewComment> Create(string owner, string name, int number, PullRequestReviewCommentReply comment)`

`Task<PullRequestReviewComment> Create(long repositoryId, int number, PullRequestReviewCommentReply comment)`

ryangribble:
This is already implemented as `CrrateReply` 😮

PR #1723, opened 2017-12-06
[WIP] Add alternative payload support for adding a PullRequest Comment.

gdziadkiewicz:


M-Zuber:
While you have been explicit that this is a WIP,  could you please explain the difference between this and https://github.com/octokit/octokit.net/blob/master/Octokit/Models/Request/PullRequestReviewCommentReplyCreate.cs ?
(best would be if you could update the PR description)

ryangribble:
Oh this is probably my fault as I raised #1717 without realising it is already implemented as `CrrateReply` rather than an overload of `Create`

gdziadkiewicz:
I didn't notice that this is already implemented.

PR #1726, opened 2017-12-12
Fix OAuthClient to work correctly with GitHub Enterprise instances

### LINKED ISSUE #1724: OauthClient.GetGitHubLoginUrl returns incorrect URL for Enterprise instances.
`OauthClient.GetGitHubLoginUrl` with an enterprise instance URL of `https://example.com` should return e.g. `https://example.com/login/oauth/authorize?client_id=secret`, and indeed that is what the [unit tests](https://github.com/octokit/octokit.net/blob/master/Octokit.Tests/Clients/OauthClientTests.cs#L25-L28) are testing.

In real usage, however, `connection.BaseAddress` will return a URL like `https://example.com/api/v3/` meaning that the returned URL will be `https://example.com/api/v3/login/oauth/authorize?client_id=secret` - this URL is wrong.

ryangribble:
Fixes #1724 and #1729 

Urls used in the `OAuthClient` are incorrect for GitHub Enterprise instances due to not stripping off the `/api/v3` from the base URL

Changed the `OAuthClient`'s stored `hostAddress` to strip any relative `Uri` from the current `baseAddress` so that it correctly uses the github website Url for OAuth actions.

This fixes the Urls in the 2 OAuthClient functions:
`OAuthClient.GetGitHubLoginUrl()`
`OAuthClient.CreateAcessToken()`

eg for an GitHub Enterprise instance we now will use:

`GitHubClient` `baseAddress`: `https://example.com/api/v3` 
`OAuthClient` `hostAddress`: `https://example.com`

`OAuthClient` Authorize Uri: `https://example.com/login/oauth/authorize`
`OAuthClient` Token Uri: `https://example.com/login/oauth/access_token`

### DIFF
--- a/Octokit.Tests/Clients/OauthClientTests.cs
+++ b/Octokit.Tests/Clients/OauthClientTests.cs
@@ -24,12 +24,13 @@ public class OauthClientTests
         [Theory]
         [InlineData("https://api.github.com", "https://github.com/login/oauth/authorize?client_id=secret")]
         [InlineData("https://github.com", "https://github.com/login/oauth/authorize?client_id=secret")]
-        [InlineData("https://example.com", "https://example.com/login/oauth/authorize?client_id=secret")]
-        [InlineData("https://api.example.com", "https://api.example.com/login/oauth/authorize?client_id=secret")]
+        [InlineData("https://example.com/api/v3", "https://example.com/login/oauth/authorize?client_id=secret")]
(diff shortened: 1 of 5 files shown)


ryangribble:
@grokys how does this look?

M-Zuber:
👍

ryangribble:
release_notes: `OAuthClient` now handles GitHub Enterprise instances correctly in `CreateAccessToken()` and `GetGitHubLoginUrl()` methods

PR #1730, opened 2017-12-28
WIP Added support for authorizing GitHub Apps

### LINKED ISSUE #1727: Extend authentication to support authenticating as installation
Docs: https://developer.github.com/apps/building-github-apps/authentication-options-for-github-apps/#authenticating-as-an-installation

I can't find any issue, docs, or code related to this and I'm not sure if it is something that Octokit should do. From my initial research, it seems that the output of the flow is a valid "auth token" and can be used in place of PAT so the existing authentication mechanism can be used as-is, we just need to pass in a different token every time.

If anyone is working or worked on this, please chime in. I'm working on an implementation as well, and it would be great to not duplicate effort.

adriangodong:
Work in progress.

Changes:
* TODO

Closes #1727

### DIFF
--- a/Octokit.Reactive/Clients/IObservableAccessTokensClient.cs
+++ b/Octokit.Reactive/Clients/IObservableAccessTokensClient.cs
@@ -1,0 +1,9 @@
+﻿using System;
+
+namespace Octokit.Reactive
+{
+    public interface IObservableAccessTokensClient
+    {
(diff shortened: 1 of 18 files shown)


ryangribble:
shall we close this one in favour of #1738 ?

adriangodong:
Yep, the other PR looks good. Closing this one.

Issue #825, opened 2015-06-24
WebHook PullRequestEventPayload


warrenbuckley:
Hello I am trying to create an API Controller that will receive the Pull Request WebHook. I want .NET model binding to bind against the JSON payload that is POSTed to my WebAPI.

I was assuming the model/class I needed to use was PullRequestEventPayload but this returns as null when the endpoint is hit when stepping through the code with ngrok as the proxy to my local machine.

```csharp
[HttpPost]
public HttpResponseMessage RecieveWebHook(PullRequestEventPayload payload)
{
            //Check the state, is it closed & merged = true
            //Means was accepeted & put back into repo
            //Thus assign contributor badge to member
            if (payload.PullRequest.State == ItemState.Closed && payload.PullRequest.Merged)
            {
                        //Do some logic here..
            }
}
```

Any pointers on what I am doing wrong that would be ace please :)

Haacked:
Hi @warrenbuckley, I wonder if the model binding is having trouble. Can you verify by sniffing the network what the data looks like that's being passed to your method? Also, as an FYI, the method name has "receive" misspelled. :smile:

One thing you could try is changing the argument type for `payload` to a string and see what the JSON looks like.

I noticed that the setters for `PullRequestEventPayload` are protected. This might be causing problems with the model binder.

shiftkey:
I must confess that I keep procrastinating on doing something to make Octokit integrate better with the ASP.NET stack. Thanks for bringing this to our attention.

> I noticed that the setters for `PullRequestEventPayload` are protected. This might be causing problems with the model binder.

This was also where my thought process went to. Perhaps we need to plug in a model binder that does things the SimpleJson way?

Haacked:
> This was also where my thought process went to. Perhaps we need to plug in a model binder that does things the SimpleJson way?

I'm starting to wonder if making the setters protected is causing more pain than it's worth.

shiftkey:
@haacked I wonder if we're making it hard/impossible for JSON.NET to deserialize with protected setters. Surely there's a quick way to test this suspicion at least...

Haacked:
Yeah, we should test these scenarios out.

warrenbuckley:
@Haacked thanks for spotting my bad typo :smile: 

I have tried this @Haacked & @shiftkey 

```csharp
[VerifyGitHubWebHook]
[HttpPost]
public HttpResponseMessage ReceiveWebHook()
{
    //From the webhook JSON payload we get POSTed to us
[...]

PR #1731, opened 2018-01-01
added support for DateTime serialized as FileTime

itaibh:


shiftkey:
@itaibh can you provide some more context for this change? Also, this might be worth upstreaming to [SimpleJson](https://github.com/facebook-csharp-sdk/simple-json) if it's useful more broadly...

itaibh:
In some cases GitHub sends wrong date format for the "created_at" and "pushed_at" fields. This is a GitHub bug, probably, but still, such a convention should be supported (and it would have saved me hours of debugging).

Back in 2015 there was an open issue on this, which I commented on today, but I can't find it right now. I'm sending this from my phone, and don't have the full attention to provide more info right now.

shiftkey:
@itaibh thanks for the extra info. I found #825 where you commented earlier.

> This is a GitHub bug, probably, but still, such a convention should be supported (and it would have saved me hours of debugging).

In your other comment you mention this:

> Also, from the timing it seems the problem had returned only recently (about two weeks ago).

If this is a recent regression then I'd love to try and chase this down, because we should be consistent with these fields. If you can point me to a repository which fired a webhook with the wrong format for `created_at` and `pushed_at` in the payload I can chase that up internally and confirm it's been resolved while we discuss this fix.

itaibh:
I'm experimenting with web hooks on my own fork of spoon-knife.
I noticed that it happens for some push events.

PR #1732, opened 2018-01-02
Add support for StatusEventPayload

itaibh:
Added payload definition for the Status Event message.

  - [x] Write basic code.
  - [x] Make sure the code design fits the rest of the library.
  - [x] Write tests.

### DIFF
--- a/Octokit.Tests/Clients/EventsClientTests.cs
+++ b/Octokit.Tests/Clients/EventsClientTests.cs
@@ -560,6 +560,7 @@ namespace Octokit.Tests.Clients
             {"PullRequestEvent", typeof(PullRequestEventPayload)},
             {"PullRequestReviewCommentEvent", typeof(PullRequestCommentPayload)},
             {"PushEvent", typeof(PushEventPayload)},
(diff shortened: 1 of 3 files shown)


ryangribble:
Hi @itaibh 

Thanks for the PR... in order to better understand your intentions could you please provide a brief description?   also to understand whether you think you are "done" or not, would could you please mark [WIP] in the title if you aren't done yet, and also provide a task list that are checked/unchecked, so we know what you still aim to do? 

Eg in this case a new payload class needs to be added to the special switch statement in the deserializer code and of course we need tests to prove it actually works 😁

ryangribble:
Looks good @itaibh , if you have finished working on it, can you remove `[WIP]` (work in progress) from the PR title so we know it's ready ffrom your perspective?

itaibh:
Hi @ryangribble!
Thank you. I consider it ready, and have removed the [WIP] tag.

ryangribble:
release_notes: Support `StatusEvent` payloads, using new response model `StatusEventPayload`

ryangribble:
Thanks @itaibh and congratulations on your first octokit contribution!

LGTM

PR #1735, opened 2018-01-09
Added support for DateTime serialized as FileTime

itaibh:
The other PR got dirty, so I created another one, and did it right this time :-)

### DIFF
--- a/Octokit.Tests/SimpleJsonSerializerTests.cs
+++ b/Octokit.Tests/SimpleJsonSerializerTests.cs
@@ -1,5 +1,6 @@
 ﻿using Octokit.Helpers;
 using Octokit.Internal;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
@@ -253,13 +254,15 @@ namespace Octokit.Tests
                   "\"followers\": 0," +
                   "\"following\": 0," +
                   "\"created_at\": \"2009-02-10T17:53:17Z\"," +
-                  "\"updated_at\": 1404691976" +
+                  "\"updated_at\": 1234288397" +
                 "}";
 
                 var result = new SimpleJsonSerializer().Deserialize<User>(json);
 
                 Assert.Equal("Mono Project", result.Name);
                 Assert.Null(result.Hireable);
+                Assert.Equal(new DateTimeOffset(2009, 02, 10, 17, 53, 17, TimeSpan.Zero), result.CreatedAt);
+                Assert.Equal(new DateTimeOffset(2009, 02, 10, 17, 53, 17, TimeSpan.Zero), result.UpdatedAt);
             }
 
             [Fact]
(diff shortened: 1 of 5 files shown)


ryangribble:
Can you add unit tests that deserialize example json with both formats of date fields?

itaibh:
@ryangribble I'm not sure I actually got it right. How does that number represents time? Is it actually a file time? or is it the number of seconds since the Unix Epoch?

ryangribble:
Can't remember off the tip of my head, I think it's Unix epoch time...  we have a helper method elsewhere in the code base

ryangribble:
Here you go :  https://github.com/octokit/octokit.net/blob/c9b2c1260bc87b7782d4fd9645d43a8885203923/Octokit/Helpers/UnixTimeStampExtensions.cs#L16

It's an extension method on `long`

itaibh:
Thanks. I made the required changes and pushed, as you can see. I also updated a bit the solution folder structure (virtual structure) so all the tests projects are bound together under the "Tests" solution folder. I hope that's okay. If you don't like it, feel free to remove it from the PR.

Octokit.Tests/SimpleJsonSerializerTests.cs
```diff
@@ -253,7 +253,7 @@ public void CanDeserializeOrganization()
                   "\"followers\": 0," +
                   "\"following\": 0," +
                   "\"created_at\": \"2009-02-10T17:53:17Z\"," +
-                  "\"updated_at\": \"2014-07-07T00:12:56Z\"" +
+                  "\"updated_at\": 1404691976" +
```
ryangribble: maybe could you add `.Assert`s that test `CreatedAt` and `UpdatedAt` fields get the values expected?

itaibh: Done.

Octokit/Models/Response/Account.cs
```diff
@@ -67,6 +67,11 @@ protected Account(string avatarUrl, string bio, string blog, int collaborators,
         /// </summary>
         public DateTimeOffset CreatedAt { get; protected set; }
 
+        /// <summary>
+        /// Date the account was updated.
+        /// </summary>
+        public DateTimeOffset UpdatedAt { get; protected set; }
```
ryangribble: Does a response to `GetOrganization` populate this field (the API docs don't show it but I'm not able to test a real response from my phone right now). If it's not on `Organization` then this field would need to be moved to `User` rather than the `Account` base class

itaibh:
Done!

itaibh:
bump

ryangribble:
As a follow up we should review other places we are using the extension method `FromUnixTime()` or otherwise dealing with `long` members for dates on API responses, and consider changing them to simply be `DateTimeOffset` now that the deserializer will automatically handle the conversion :+1:

ryangribble:
release_notes: Octokit now handles `DateTime` and `DateTimeOffset` response fields whose API response is in an unexpected Unix epoch time format

ryangribble:
Thanks @itaibh 

LGTM

PR #1738, opened 2018-01-11
Adding initial support for GitHub Apps.

### LINKED ISSUE #1727: Extend authentication to support authenticating as installation

itaibh:
Closes #1727

This is built over a PR by @adriangodong #1730, and allows to create Application Token, then use that token to get all Installations, which will allow you to authenticate your installation to allow performing actions like Create Status.


adriangodong:
@itaibh Thank you for picking this up! My original goal was only to do the bearer token authentication but stuck on some incompatibility with .NET Core 1.0. Let me know if you have any questions on my code.

itaibh:
bump

Octokit.Reactive/ObservableGitHubClient.cs
```diff
@@ -69,6 +71,8 @@ public void SetRequestTimeout(TimeSpan timeout)
 
         public IObservableAuthorizationsClient Authorization { get; private set; }
         public IObservableActivitiesClient Activity { get; private set; }
+        public IObservableApplicationClient Application { get; private set; }
+        public IObservableInstallationsClient Installations { get; private set; }
```
ryangribble: We use singular for client property names, so can this be `Installation` ?

Octokit.Reactive/IObservableGitHubClient.cs
```diff
@@ -18,6 +18,8 @@ public interface IObservableGitHubClient : IApiInfoProvider
 
         IObservableAuthorizationsClient Authorization { get; }
         IObservableActivitiesClient Activity { get; }
+        IObservableApplicationClient Application { get; }
+        IObservableInstallationsClient Installations { get; }
```
ryangribble: We use singular for client property names, so can this be `Installation` ?

Octokit/Authentication/BearerTokenAuthenticator.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+using System.Globalization;
+
+namespace Octokit.Internal
+{
+    class BearerTokenAuthenticator: IAuthenticationHandler
+    {
+        public void Authenticate(IRequest request, Credentials credentials)
+        {
+            Ensure.ArgumentNotNull(request, "request");
```
ryangribble: for any `Ensure` checks can we please use `nameof(request)` rather than `"request"` ?

Octokit/Clients/ApplicationClient.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System.Threading.Tasks;
```
ryangribble: We use plural for client class names, so can this be `ApplicationsClient` ?

Octokit/Clients/IApplicationClient.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.Threading.Tasks;
```
ryangribble: `IApplicationsClient`

Octokit/GitHubClient.cs
```diff
@@ -171,6 +173,9 @@ public Uri BaseAddress
         /// </remarks>
         public IActivitiesClient Activity { get; private set; }
 
+        public IInstallationsClient Installations { get; private set; }
+
+
```
ryangribble: whitespace

Octokit/GitHubClient.cs
```diff
@@ -171,6 +173,9 @@ public Uri BaseAddress
         /// </remarks>
         public IActivitiesClient Activity { get; private set; }
 
+        public IInstallationsClient Installations { get; private set; }
```
ryangribble: `Installation`

Octokit/Http/Connection.cs
```diff
@@ -238,12 +238,12 @@ public Task<IApiResponse<T>> Post<T>(Uri uri)
             return SendData<T>(uri, HttpMethod.Post, null, null, null, CancellationToken.None);
         }
 
-        public Task<IApiResponse<T>> Post<T>(Uri uri, object body, string accepts, string contentType)
+        public Task<IApiResponse<T>> Post<T>(Uri uri, object body, string accepts, string contentType, IDictionary<string, string> parameters = null)
```
ryangribble: since this is an existing public method, we want to avoid making a breaking change if possible, so typically we would add another method with the new signature

Octokit/Http/Credentials.cs
```diff
@@ -13,23 +13,23 @@ private Credentials()
...
-        {
-            Ensure.ArgumentNotNullOrEmptyString(token, "token");
-
-            Login = null;
-            Password = token;
-            AuthenticationType = AuthenticationType.Oauth;
-        }
-
-        public Credentials(string login, string password)
+        public Credentials(string login, string password, AuthenticationType authenticationType = AuthenticationType.Basic)
```
ryangribble: again, to avoid a breaking change (even a parameter with a default value can still be a breaking change binarily speaking) you should add new `ctor`/s for these changes

Octokit/IGitHubClient.cs
```diff
@@ -38,6 +38,10 @@ public interface IGitHubClient : IApiInfoProvider
         /// </remarks>
         IActivitiesClient Activity { get; }
 
+        IApplicationClient Application { get; }
+
+        IInstallationsClient Installations { get; }
```
ryangribble: `Installation`

Octokit/Models/Response/Installation.cs
```diff
@@ -0,0 +1,53 @@
...
+        public int Id { get; protected set; }
+        public User Account { get; protected set; }
+        public string AccessTokensUrl { get; protected set; }
+        public string RepositoriesUrl { get; protected set; }
+        public string HtmlUrl { get; protected set; }
+
+        public int AppId { get; protected set; }
+        public int TargetId { get; protected set; }
+        public StringEnum<AccountType> TargetType { get; protected set; }
+        // TODO - add permissions
```
ryangribble: are these TODO comments going to be actioned in this PR?

Octokit.Reactive/IObservableGitHubClient.cs
itaibh: But I see there is `AhtorizationsClient` and `ActivitiesClient`, in fact only `ApplicationClient` is in singular form...

Octokit/Authentication/BearerTokenAuthenticator.cs
itaibh: Already did in some places. Will change that one too.

Octokit.Reactive/IObservableGitHubClient.cs
itaibh: Oh, sorry, I see now what you meant. Will fix.

Octokit/Http/Credentials.cs
itaibh: Are you sure? AFAIK the run time takes care about it...

Octokit/Models/Response/Installation.cs
itaibh: Not at the moment. They are merely placeholders so I won't forget where I need to add them.

Octokit/Http/Credentials.cs
ryangribble: See this link, for example
https://haacked.com/archive/2010/08/10/versioning-issues-with-optional-arguments.aspx/

itaibh:
All done!

itaibh:
Done

ryangribble:
I'll take a look in the next day or two thanks

We still need to add unit tests and integration tests though 🤣

itaibh:
OK, then.
Tell me what tests are still missing and I'll add them to my backlog.

itaibh:
OK, then I'll get to it when I have the time, which currently I don't have enough of 😃 
I think I'll start with fixing according to your review notes, implement the missing parts that are straight-forward.
I'll read what you wrote about tests, check out the demo project.

Regarding tests, I'll just say that I currently use this code in a project here where I work, and for what we need, it is working and tested, so I think it can be merged, just so it won't get too big a merge. May I suggest that'll once I push the fixes you requested, you'll merge it and open another issue for tests?

ryangribble:
At a minimum we need unit tests before I'm comfortable to merge new features, and then before cutting a release (which I plan to do as soon as this feature is in) I also need to run integration tests to ensure things actually work! 

I'm happy to help with the tests if you like, also not too worried about the review size, as I prefer to deliver a feature and it's tests in one merge to master :+1:

itaibh:
In that case I'll be glad to have some help writing the tests. Especially integration tests that require connecting to GitHub. I didn't look yet at the demo projects and I currently work on completing the reviewed code.

itaibh:
I added some unit tests, but I don't know how to run the integration tests.

itaibh:
Now I figured out how to run the Integration tests, but I don't know how to set up a test project, and for some reason I can't debug it (it constantly fails to attach the debugger).

@ryangribble Can you please help?

ryangribble:
What are you running the integration tests via?  Most of the contributors to the project are using Visual Studio 2015 or 2017 and to be honest, that's probably the only way to effectively run the integration tests are there are way too many of them to run "everything" at once from command line or CAKE build task etc, so you really want the ability to select which tests to run

itaibh:
I am using Visual Studio 2017, that was the only version I managed to open the solution in.
But I can't get it to debug the integration tests.

itaibh:
How can I print to the log (output pane, show output from tests) in an integration test? Perhaps I can debug "the old way"...

ryangribble:
Ah sorry, I thought you had previously made a comment about doing this on linux!

In VS2017 in the test explorer you should be able to just select "Debug Selected Tests" rather than "Run Selected Tests" and it should then hit breakpoints that you have setup.

Unfortunately with automated tests (and perhaps specifically XUnit) you cant actually get console output to print as it gets swallowed.  There seems to be a long and sordid internet debate/commenting on this issue if you google it Im sure you can read lots about why this is so 😁   If you really have to see something you can write text to a file which is horrible 😱 but generally debugging with breakpoints should "just work"...

ryangribble:
Back to the topic of this PR I was just checking out the [API docs](https://developer.github.com/v3/apps) for the GitHub Apps implementation and it looks like we have a bit of reworking to do of the structure and naming of things.  Apologies for not noticing this sooner, and I appreciate you had picked up another PR that had already set this direciton... but it's important that we have consistency with the official docs and the upstream API we are wrapping, as well as consistency across the code base in terms of naming things and structure.  Feel free to let me know if you aren't keen to make these changes, and I can take over instead :+1:

We want to match the octokit structure to the API doc structure, so given the docs are like this:

image

We shouldn't really have `GitHubClient.Application` and `GitHubClient.Installation` both at that top level (the `Installation` should be a child of `Application`).   Also instead of `Application` it should be named `GitHubApp` (as a property) and `GitHubAppsClient` (class name) so we match the terminology of the API docs as well.

Infact looking at it more, that whole "Installations" area is actually a different set of API calls that are used once you are authenticated as a GitHubApp installation (same as the "Marketplace" area)...

The calls being implemented in this pull request are 
[get the authenticated github app](https://developer.github.com/v3/apps/#get-the-authenticated-github-app)
[create new installation token](https://developer.github.com/v3/apps/#create-a-new-installation-token)
[list installations for user](https://developer.github.com/v3/apps/#list-installations-for-user) 
all of which actually live in the "GitHubApps" section (not in the "Installations" section or in an "Access Tokens" section).  This is all rather confusing but appears to be the way it is in the official upstream API
 🤔 

So what I actually think we need to do is
- delete the `InstallationsClient` and `AccessTokensClient` classes
- rename `ApplicationsClient` to `GitHubAppsClient` 
- implement all 3 methods in `GitHubAppsClient` using consistent octokit method naming which would be called 
  - `GetCurrent()`
  - `CreateInstallationToken()`
  - `GetAllInstallationsForCurrent()`

itaibh:
@ryangribble You mean something like this?

itaibh:
What about adding methods specifically inside `Installation` and such, making it more OOP like? I mean, why write this:

``` c#
long installationId = myInstallation.Id;
AccessToken accessToken = await client.GitHubApps.CreateInstallationToken(installationId)
```

when you can write this:

``` c#
AccessToken accessToken = await myInstallation.CreateToken();
```

itaibh:
@ryangribble Can you please take another look at this PR? Thanks.

ryangribble:
Sorry @itaibh I had somehow missed that you had pushed the requested changes a while ago... Will get stuck into reviewing them tonight

Cheers

itaibh:
Hi there, @ryangribble!
Did you have a change to take a look on the changes? Do you have any comments?

ryangribble:
@itaibh it's on the list to review, have been quite busy sorry!

ryangribble:
This is tough to review sorry as it's a brand new area.  A few things I noticed

- `Application` response model already exists from years ago and is being re-used here, but doesnt have all of the fields of the actual response object and also has a `Url` field which doesnt map.  We may have to create a new `GitHubApp` response model for this stuff

- We probably need to implement more of the endpoints such as `Get` for an app, `GetInstallation` for an installation and `GetAllForUser` in order to support GitHubApp stuff enough to be useful to end users?

- XmlDocument tags on the clients and response model fields will be necessary to provide intellisense guidance to users

- Integration tests - we need to figure out how/if we can flesh out our integration test framework to handle doing these calls as a GitHubApp

- Some of the endpoints (such as list installations or generate installation token) require being authenticated as the GitHubApp which means we have to do the FWT token stuff and so on, we might need to think about whether we provide helper methods around this or just expect the consumers to have created the JWT from their app's certificate etc already

ryangribble:
Oh, forgot to mention that the initial round of requested changed regarding naming things and re-arranging the methods etc looks pretty good, thanks :+1:

ryangribble:
ive pushed up some of the mentioned stuff (implenting the `Get(string slug)` method to get an app, and also implementing the `GitHubApp` response model instead of re-using `Application`.

itaibh:
Thanks, @ryangribble!

I fixed the build (it seems you forgot to update a test) and now all tests pass.
Can you please merge it to the master now? I think the rest of the review notes we can fix in a later time.

ryangribble:
Before it goes to master we need xmldoc comments, unit tests and integration tests at a minimum. 

But as I mentioned I think we also need a more complete implementation and some solution around the JWT tokens as what we have so far probably isn't enough for end users to make good use of at the moment... So you disagree? 

If you want to test what we have so far, you can get nuget packages from this PR build from our appveyor feed :+1:

itaibh:
Hi @ryangribble,

I already added all the required unit test (I used other unit tests as reference to what is needed, so if I missed something please tell me).

I don't know how to write integration tests nor how to run them locally. I tried once (even twice 😃), and I get connection errors of various types. I started adding the `GitHubAppsClientTests` integration tests, but since I can't debug it locally I can't continue. I know for sure it doesn't throw exception, as the main function is getting called.

Regarding missing XML comments - what comments are missing exactly? I'm only asking because when I build the project I already get tons of "`missing xml comment for publicly visible type or member`" warnings, most of them for data fields, so I didn't bother adding XML comments for the new data fields. I did try to add XML comments to all new methods, so please tell me if I missed anything.

Regarding the more complete implementation and some solution around JWT tokens - up until I started writing this feature it was not implemented at all (no JWT token, I don't know what FWT token is), so I think users get at least something which allows them to start writing extensions, don't you agree?
I already use it in code that is going to be integrated into production soon, and I don't think there'll be too big a change after the fixes you request, if any.

Please tell me what you think.

haidelber:
I've been following this PR for 2 weeks now as I need this feature in one product I'm currently developing. I'd really appreciate the integration.
The security handling is sufficient enough in the current implementation via `ICredentialStore`. I could contribute with doc and some samples on how to implement these for:

* Access as GitHubApp
* Access as GitHubApp installation

ryangribble:
From my perspective we currently have no evidence that these calls even work because we don't have working integration tests. Unit tests make sure we call the correct url or correctly assert nulls etc but we need integration test to prove we actually "work" and handle the response from the API 

We also need xmldoc comments on the client methods and member attributes of the response classes. 

And given this is a new/complex feature we also need a sample/doc page about how to do this GitHubApp stuff  

So do you already have a working example where you create a signed JWT token using the private cert of an authorised app?

adriangodong:
Chiming a little bit here. I've managed to use my initial commits, created GitHub JWT, and use PEM from filesystem or env var. You can check this PR https://github.com/adriangodong/todo-bot/pull/6 for details.

It does require .NET Standard 2.0 to make the JWT part working.

adriangodong:
Also, would it be prudent to split this PR into several parts:
* Foundational work (new authentication option + access token client)
* GitHub JWT -- I'd argue this can be a separate package. Octokit can take a token string, not caring how to create it.
* Other clients (installation, etc.)

ryangribble:
Im happy with the approach to split it, and yes I do agree the JWT stuff doesnt necessarily belong in Octokit (however we need to be able to do it for integration tests, and we need to be able to tell users HOW to do it, so maybe your `GitHubJWT` stuff in todo-bot could be pulled out into it's own nuget package?).  

Even if we split the PR up in order to deliver the foundational Credential and token creation parts first, Im still keen to abide by our standards of each piece needing to tick certain boxes to merge to master (namely consistency in naming/code base hygiene, unit test coverage, integration test that excercises the function, xml doc comments for intellisense guidance and any updates to samples/documentation which particularly for GitHubApps being an entire new thing, is important)

adriangodong:
[GitHubJwt is now a Nuget package](https://www.nuget.org/packages/GitHubJwt/).

ryangribble:
That's awesome, although if it's targeting netstandard 2.0 we won't be able to use it since we are targeting netstandard 1.1... Does it require anything that isn't available for 1.1?

adriangodong:
.NET Core 1.0 has limited support for PEM files. I did not try .NET Core 1.1 since AWS Lambda has never supported the version. AWS Lambda recently added support for 2.0 and that cleared all the issues.

The test project itself can target netcoreapp2.0 and use both netstandard1.1 and netstandard2.0 libraries without issue. I'd expect Octokit stays at netstandard1.1 but any consumer that wants to use this specific library to generate JWT will need to run on netcoreapp2.0.

If I have some time, I can try out .NET Core 1.1 and see if there's any issue.

itaibh:
I've added and fixed some XML documentation tags.

I also want you to notice the name of this PR - it states this is an "initial support ...", not "complete support ...". Just saying.

ryangribble:
> .NET Core 1.0 has limited support for PEM files. I did not try .NET Core 1.1 since AWS Lambda has never supported the version. AWS Lambda recently added support for 2.0 and that cleared all the issues.
> 
> The test project itself can target netcoreapp2.0 and use both netstandard1.1 and netstandard2.0 libraries without issue. I'd expect Octokit stays at netstandard1.1 but any consumer that wants to use this specific library to generate JWT will need to run on netcoreapp2.0.
> 
> If I have some time, I can try out .NET Core 1.1 and see if there's any issue.

Yes this is true, we can make the test projects `netcoreapp2.0` however I guess the thing to think about is more just a design principle that in general a _library_ should target the *lowest* `netstandard` version it can get away with, in order to be able to be used by the largest possible number of applications, whilst an _application_ should target the *highest* possible `netstandard` version that supports the platforms it should be run on.  So being a library I would suggest the JWT should target a lower `netstandard` if possible.  Also don't be confused between `netstandard` and `netcoreapp` (`netstandard1.1`  is still supported by `netcoreapp1.0` so can still run on AWS Lambda (prior to them implementing `netcoreapp2.0` runtime (as long as you use the explicit metapackage reference 1.6.0 like we had to do in #1713)

ryangribble:
I've pushed up some changes... let me know what you think @itaibh . 

I agree this can just be "initial support" but that doesn't mean we get to skip the "standard" things that get a PR mergeable - unit tests, integration tests and documentation 😉 

I also reverted some unrelated commits that were fixing up stuff in other files unrelated to GitHubApps - let's keep this PR focused on GitHubApps implementation so it is easier to review and look at new PR's for those other fixups (incidentally I had also already made some of the same XmlDoc corrections for missing attributes, which ive pushed up on #1779)

From my perspective the remaining things to do on this PR are to implement Unit Tests for the Reactive client, and to hopefully get at least a basic integration test working (eg authenticate as a test GitHubApp with JWT token, create an installation token, then access some other endpoint).  Now that @adriangodong has provided the GitHub.JWT package I think this should be doable

itaibh:
sorry for the merges and reverts, my git client got me confused a little bit. :flushed:

Cyberboss:
Been playing around with this some and it seems to work well. I know it's
WIP but the lack of `/installation/repositories` is a bit of a downer
seeing as you need it to associate webhook payloads with installation
tokens afaik.

ryangribble:
> sorry for the merges and reverts, my git client got me confused a little bit. 😳

Hmm your last commit actually reverted all the changes I made 😮 

Ill push up a commit to re-instate them!

ryangribble:
image

Korporal:
Looking forward to leveraging this, I'm just at the earliest stages of planning a GitHub App so this PR is very timely.

Octokit/Models/Response/AccessToken.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// <summary>
+        /// The expiration date
+        /// </summary>
+        public DateTimeOffset ExpiresAt { get; protected set; }
+
+        internal string DebuggerDisplay
+        {
+            get
+            {
+                return string.Format(CultureInfo.InvariantCulture, "Token: {0}, ExpiresAt: {1}", Token, ExpiresAt);
```
Korporal: @itaibh - Just a minor simplification, this Format could be removed I think and use instead:

`   
   return $"{nameof(Token)}: {Token}, {nameof(ExpiresAt)}: {ExpiresAt}";
`

itaibh: This is a brand new feature of C# and won't compile on older versions of the compiler. Are you sure you want to switch to this?

ryangribble: We have previuosly discussed this and dont have a problem in using the string interpolation `$"blah {var} blah"` however would prefer this to be done across the code base in a consistent way via a Milestone and one/many PR's only making that change... rather than randomly using it in a few places while 100s of existing cases are still doing it "the old way").

ryangribble:
@itaibh im not sure what's going on but you seem to be doing merges and reverting changes we already have in master back to previously (bad) versions.

EG:

image

itaibh:
@ryangribble Hmmm... seems like the last merge messed things up. I'm on it.

itaibh:
@ryangribble Done (until further requests.)

ryangribble:
OK ive updated the integration tests to `netstandard2.0` and using your library @adriangodong (thanks! 💎 )

Ive also had a first crack at writing up some doco if everyone could please [have a look](https://github.com/itaibh/octokit.net/blob/7caef2427a6de188b1777a52e2e021a4df085db2/docs/github-apps.md)

Last issue to look at is the builds being broken because we need the 2.0 runtime on our appveyor and travis images now...

itaibh:
@ryangribble I read your docs. Looks good!

I just added a new feature allowing to get the Installation ID (wrapped in an empty Installation object, perhaps we should do something about it) from any event's payload. I think you should mention that in the docs to help developers get the access token for an installation in case they need it for tasks like creating status notifications.

ryangribble:
> wrapped in an empty Installation object, perhaps we should do something about it

Yeah it's nit so nice using the whole response class when only the Id property is returned. Could you create a new response class for this? Maybe it could be called`InstallationSummary` or `InstallationPayload`...

> I think you should mention that in the docs to help developers get the access token for an installation in case they need it for tasks like creating status notifications

How do you envisage people using it (via receiving the event as a webhook or by querying the activity events API?  Supporting webhooks better is something I'd like to add because at the moment users would have to use our deserializer directly. 
 
Can you provide a code sample or add to the doc file I started to show how this could be used currently?

ryangribble:
Thanks for the updates but for consistency with the rest if the octokit.net approach I don't want ti have "comfort" methods in response models that call client api's etc so could this please be removed? It's just there are hundreds of places this could be done, so it doesn't make sense to only do it here, and it's not something I'd want to start doing and proliferating across the code base. Our design is aimed at being a c# wrapper around the upstream API and mirroring the structure and behaviour of that

While it's a great idea it doesn't belong in the core library. 

Perhaps if it's something you want to undertake, you could develop an "extensions" repo with its own nuget package, that uses extension methods and so on, to provide the more object oriented comfort features you are wanting

Octokit/Models/Response/AccessToken.cs
Korporal: I agree with Ryan, I was not suggesting we do this piecemeal, I saw some earlier PR that used this string interpolation so I assumed it was established across the project, but if it isn't I agree 100% that this should be a dedicated focused PR that changes this across the board.

adriangodong:
@ryangribble @itaibh I've created a [`netstandard1.4` targeted GitHubJwt PR](https://github.com/adriangodong/githubjwt/pull/2). This will not get merged until I get the tests pass. With small modification, 1 out of 2 of the test passes on `netcoreapp1.1`.

Add the custom feed `https://ci.appveyor.com/nuget/githubjwt` and install the version you want to try out. I'd recommend trying out `0.0.1-2-g457db9a`.

ryangribble:
Today is the day folks!  Sorry for the delays, I was trying to chase down sourcelink issues on TravisCI which I thought may have been related to moving to the SDK2.0 tooling but probably weren't in the end.  Since the GitHubJWT library doesnt require netcoreapp2 anyomre anyway, I wont worry about tooling updates in this PR.

Appreciate the patience and many thanks to everyone who helped with this PR, especially @adriangodong and @itaibh.  Integration tests have been passing as I've been running them every now and then for the last few weeks so I'm happy to merge this and start putting together the next release 😁 

LGTM

ryangribble:
release_notes: Added initial support for GitHub Apps, see [the documentation](http://octokitnet.readthedocs.io/en/latest/github-apps/) for further information

PR #1739, opened 2018-01-11
Add a new column named PullRequestReviewId to PullRequestReviewComment

mirsaeedi:
I added a new column to PullRequestReviewComment model in order to capture a field called pull_request_review_id which is returned by [Github API](https://developer.github.com/v3/pulls/comments/) .

### DIFF
--- a/Octokit/Models/Response/PullRequestReviewComment.cs
+++ b/Octokit/Models/Response/PullRequestReviewComment.cs
@@ -15,8 +15,9 @@ namespace Octokit
             Id = id;
         }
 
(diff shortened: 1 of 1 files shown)


grokys:
👍 This is something we need over at [`github/VisualStudio`](https://github.com/github/VisualStudio) as well.

ryangribble:
Many thanks @mirsaeedi !  I just wanted to check that there aren't any other missing fields for this response class while we're here, but it looks like this was the only missing one :+1:

ryangribble:
release_notes: Add `PullRequestReviewId ` property to `PullRequestReviewComment` response, to indicate which `PullRequestReview` the comment is related to

ryangribble:
LGTM

Issue #1737, opened 2018-01-10
Get Rate Limit By Calling it's API


mirsaeedi:
Currently, we need to call GetLastApiInfo() to get the status of our usage. This API uses the information from the last request it has sent. The problem with this approach is that in some cases we need to fetch this data from Github and the information which we have from the previous call is stale and not sufficient. What's the reasoning behind your implementation? If my argument is valid, can I send a pull request?

ryangribble:
Hi @mirsaeedi the reasoning is simply that the remaining API rate limit is returned in the response headers of all calls, so we may as well provide a mechanism to access that information :)

It would be awesome if you wanted to send a PR to implement the actual [Rate Limit API](https://developer.github.com/v3/rate_limit/) though, then users can choose the method that is more appropriate for them :+1:

mirsaeedi:
Actually, I found that [Rate Limit API](https://developer.github.com/v3/rate_limit/) has been implemented before. I will update the documentation.

`gitHubeclient.Miscellaneous.GetRateLimits();`

ryangribble:
Fixed by #1742

PR #1742, opened 2018-01-20
Improve Rate Limits' Documentation

mirsaeedi:
I added an example and a bit of explanation about `GitHubClient.Miscellaneous.GetRateLimits()`.  #1737

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -71,9 +71,9 @@ var user = await client.User.Current();
 ```
 
 ### Too Much of a Good Thing: Dealing with API Rate Limits
-Like any popular API, Github needs to throttle some requests. The OctoKit.NET client allows you to get some insight into how many requests you have left and when you can start making requests again. In fact, there are two ways to get the Rate Limits via OctoKit.NET. Calling 'GitHubClient.GetLastApiInfo()' returns the Rate Limit status which has been returned with the last api call. So, calling 'GitHubClient.GetLastApiInfo()' will not send any extra HTTP requests to GitHub's servers. However, if in some cases you need to get the Rate Limit directly from Github, you should call 'GitHubClient.Miscellaneous.GetRateLimits()'.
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -90,6 +92,23 @@ var howManyRequestsDoIHaveLeft = rateLimit?.Remaining;
...
 
+However, if in some cases you need to get the Rate Limit directly from Github, you should call `GitHubClient.Miscellaneous.GetRateLimits()`.
+
+Example usage:
+
+```csharp
+GithubClient client; 
+//Create & initialize the client here
+
+var miscellaneousRateLimit = client.Miscellaneous.GetRateLimits();
```
ryangribble: Can we use,`await` syntax here to avoid the `.Result` below?

docs/getting-started.md
```diff
@@ -90,6 +92,23 @@ var howManyRequestsDoIHaveLeft = rateLimit?.Remaining;
...
+
+var rate = miscellaneousRateLimit.Result.Rate;
```
ryangribble: According to the docs the `Rate` property is deprecated so our docs should actually steer people towards the proper field (`.Resource.Core`)

ryangribble: Could probably also mention the search API rate limit in the code sample too

ryangribble:
Closes #1737

mirsaeedi:
@ryangribble  I tried to satisfy your points in my new commit.

docs/getting-started.md
mirsaeedi: I replaced the `Rate` property with the `.Resource.Core`. Also, I added a sample for the search API.

ryangribble:
Thanks so much @mirsaeedi for tidying up this doco!

LGTM

ryangribble:
release_notes: Updated Rate Limits documentation and samples

PR #1744, opened 2018-01-21
add NET Core

ShalokShalom:


ryangribble:
Fyi we actually support net core 1.0 onwards...

Though I think it would be better to state netstandard 1.1 rather than a net core version, and possibly hyperlink it to Microsoft page showing the netstandard versions and what each one aligns with

ShalokShalom:
Yep

ryangribble:
I've updated the PR to show what I mean... do you think this is sufficient?

ryangribble:
Yes, .NET Standard versions are cumulative, so by supporting .NET Standard 1.1 any framework that implements 1.1 or later, will be able to consume Octokit

ryangribble:
Update supported platforms in README.md to include `.NET Standard 1.1`

ShalokShalom:
So mean that higher versions are also supported? Might be confusing otherwise.

PR #1757, opened 2018-02-08
Asapferg/user email visibility

### LINKED ISSUE #1754: Email "visibility" Field Missing
```
var client = GetGitHubClient(accessToken);
var emailAddresses = await client.User.Email.GetAll();
```
dbg
curl

asapferg:
Fixes #1754

### DIFF
--- a/Octokit/Models/Response/EmailAddress.cs
+++ b/Octokit/Models/Response/EmailAddress.cs
@@ -36,7 +36,7 @@ namespace Octokit
         public bool Primary { get; protected set; }
 
         /// <summary>
-        /// private if the email is private; otherwise null
+        /// "private" if the email is private; otherwise null
         /// <summary>
         public string Visibility { get; protected set; }
 
--- a/Octokit/Models/Response/EmailAddress.cs
+++ b/Octokit/Models/Response/EmailAddress.cs
@@ -12,11 +12,12 @@ namespace Octokit
     {
         public EmailAddress() { }
 
-        public EmailAddress(string email, bool verified, bool primary)
+        public EmailAddress(string email, bool verified, bool primary, string visibility)
         {
             Email = email;
             Verified = verified;
             Primary = primary;
+            Visibility = visibility;
         }
 
         /// <summary>
@@ -34,6 +35,11 @@ namespace Octokit
         /// </summary>
         public bool Primary { get; protected set; }
 
+        /// <summary>
+        /// private if the email is private; otherwise null
+        /// <summary>
+        public string Visibility { get; protected set; }
+
         [SuppressMessage("Microsoft.Performance", "CA1811:AvoidUncalledPrivateCode",
             Justification = "Used by DebuggerDisplayAttribute")]
         internal string DebuggerDisplay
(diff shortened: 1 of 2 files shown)


ryangribble:
Hi @asapferg thanks for the PR!  Having a look at the doco, and my own API responses it seems like the visibility property can have a value of `private` `public` or `null` so it would be best to represent this as an `enum` rather than just a `string`.  We have a special `StringEnum<enum>` wrapper class to use for these cases so we can have the benefit of an `enum` but also be guarded against possible new values being introduced upstream.

Would you like to have a crack at changing it to a `StringEnum<>` type?  Otherwise let me know and I can push a change onto your branch that does this :+1:

asapferg:
Hey @ryangribble, thanks for the input. I'll take a crack at this!

Octokit/Models/Response/EmailAddress.cs
```diff
@@ -25,24 +27,54 @@ public EmailAddress(string email, bool verified, bool primary)
...
-        /// true if this is the users primary email; otherwise false
+        /// True if this is the users primary email; otherwise false
         /// </summary>
         public bool Primary { get; protected set; }
 
+        /// <summary>
+        /// "private" or "public" if the email address is the primary;
+        /// otherwise null
+        /// <summary>
+        public StringEnum<EmailVisibility> Visibility { get; protected set; }
```
ryangribble: The API response actually has a `null` value rather than the string "null" so what we need to do is make this property nullable (eg `StringEnum<EmailVisibility>? Visibility`)

Octokit/Models/Response/EmailAddress.cs
```diff
@@ -25,24 +27,54 @@ public EmailAddress(string email, bool verified, bool primary)
...
+        /// <summary>
+        /// Primary email address and is private
+        /// </summary> 
+        [Parameter(Value = "private")]
+        Private,
+
+        /// <summary>
+        /// Email is not the primary address
+        /// </summary>
+        [Parameter(Value = "null")]
```
ryangribble: Need to remove this `All` option from the enum as it wont deserialize correctly anyway, plus we can make the property above nullable, to handle this case

Octokit/Models/Response/EmailAddress.cs
ryangribble: (and note that this will not deserialize currently, but once #1760 is in, this should then work!)

Octokit/Models/Response/EmailAddress.cs
asapferg: Done, let me know how these changes look. Thanks for the continued help on this!

ryangribble:
#1760 is merged now, if you wanted to merge or rebase master into this branch and see if that clears up the failed tests

ryangribble:
Something a bit weird with your merge caused lots of commits to show up. I fixed it up for you and also ran integration tests to make sure things work.  Found an old skipped test that works now so was able to unskip it 😀

ryangribble:
Many thanks for the contribution @asapferg !

LGTM

ryangribble:
release_notes: Add `Visibility` property to `EmailAddress` response model, to indicate whether a primary email address is `Public` or `Private`

asapferg:
Thank you for all the help @ryangribble ! Couldn't have done it without you 😄

PR #1758, opened 2018-02-11
Add TLS1.2 to enabled security protocols (.NET Framework 4.5 only)

### LINKED ISSUE #1756: The request was aborted: Could not create SSL/TLS secure channel.
I'm running a simple .Net console app that I'm using to evaluate some logic that uses Octokit.Net.

It has been fine since I first wrote it a week ago, until a few hours ago when I began to get AggregateExceptions thrown during the call to

`var repo = client.Repository.Get("myorg","myrepo).Result;`

The exception has an inner exception message:

`An error occurred while sending the request. ---> System.Net.WebException: The request was aborted: Could not create SSL/TLS secure channel.`

What can account for this just starting to happen (consistently) all of a sudden?

In addition I use Smartgit and that too is failing to work because it is also seeing AggregateException (for example if I do a pull). I can login to Github fine but using API/tools to interact with it is suddenly failing (yes I rebooted!).

Thx

ryangribble:
Fixes #1756 

TLS1.2 will be required to connect to the GitHub API when [GitHub deprecates weak cryptographic algorithms on 22nd February](https://developer.github.com/changes/2018-02-01-weak-crypto-removal-notice/)

.NET Frameworks less than 4.6 do not have TLS1.2 enabled by default, so Octokit will now enable this protocol (on the `net45` target framework build only).

Notes
- `ServicePointManager.SecurityProtocol` is the mechanism used to enable the protocol... which isn't actually available on .NET Core until 2.0 anyway (but we don't need it since TLS1.2 is already enabled in .NET Core anyway)
- Added a `net452` target to the integration test project (bringing it into line with the other multi targeted test projects) in order to manually test that this value was set on .NET 4.5x

### DIFF
--- a/Octokit/Http/HttpClientAdapter.cs
+++ b/Octokit/Http/HttpClientAdapter.cs
@@ -27,6 +27,10 @@ namespace Octokit.Internal
         {
             Ensure.ArgumentNotNull(getHandler, "getHandler");
 
+#if HAS_SERVICEPOINTMANAGER
+            ServicePointManager.SecurityProtocol = ServicePointManager.SecurityProtocol | SecurityProtocolType.Tls12;
+#endif
+
             _http = new HttpClient(new RedirectHandler { InnerHandler = getHandler() });
         }
 
--- a/Octokit/Http/HttpClientAdapter.cs
+++ b/Octokit/Http/HttpClientAdapter.cs
@@ -28,7 +28,7 @@ namespace Octokit.Internal
             Ensure.ArgumentNotNull(getHandler, "getHandler");
 
 #if HAS_SERVICEPOINTMANAGER
-            ServicePointManager.SecurityProtocol = ServicePointManager.SecurityProtocol | SecurityProtocolType.Tls12;
+            ServicePointManager.SecurityProtocol |= SecurityProtocolType.Tls12;
 #endif
 
             _http = new HttpClient(new RedirectHandler { InnerHandler = getHandler() });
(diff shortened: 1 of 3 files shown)


ryangribble:
@Korporal @Haacked @shiftkey

Octokit/Http/HttpClientAdapter.cs
```diff
@@ -27,6 +27,10 @@ public HttpClientAdapter(Func<HttpMessageHandler> getHandler)
         {
             Ensure.ArgumentNotNull(getHandler, "getHandler");
 
+#if HAS_SERVICEPOINTMANAGER
+            ServicePointManager.SecurityProtocol = ServicePointManager.SecurityProtocol | SecurityProtocolType.Tls12;
```
Haacked: Alternatively:

```csharp
ServicePointManager.SecurityProtocol |= SecurityProtocolType.Tls12;
```

😛

ryangribble:
advisories: On February 22, 2018 19:00 UTC, GitHub will [disable permanently the use of weak cryptogrpahic standards](https://developer.github.com/changes/2018-02-01-weak-crypto-removal-notice/).  Applications targeting .NET Framework 4.5.x will be affected, as that framework does not enable the now required protocol (TLS1.2) by default.  This octokit.net release will automatically enable this protocol, when the `GitHubClient` is constructed.  Note that applications targeting .NET Framework 4.6+ or .NET Core should already have TLS1.2 enabled by default, and this release does nothing with enabled protocols on those platforms.

ryangribble:
advisories:  Affected clients that are unable to update octokit.net, can include their own [code change](https://github.com/octokit/octokit.net/blob/41b4059c110a60aee86a912dda2987fecc5a3fcb/Octokit/Http/HttpClientAdapter.cs#L31) to enable TLS1.2

ryangribble:
release_notes: Add TLS1.2 to enabled security protocols (.NET Framework 4.5 only) to avoid SSL connectivity errors when GitHub deprecates weak algorithms on February 22

PR #1759, opened 2018-02-11
Fix AccountType enum Parameter values to match the sentence case used in API responses

ryangribble:
While looking into another issue I found that the API responses for account type are `Bot` `User` `Organization` but we had these attributes in lower case on our enum.  This caused some cache misses in the deserializer enum field/attribute dictionary


ryangribble:
release_notes: Fix `AccountType` enum `Parameter` values to match the sentence case used in API responses

PR #1760, opened 2018-02-11
Deserializer should handle nullable StringEnum<T>

ryangribble:
When a nullable `StringEnum<T>?` member was added to a class, the deserializer failed to deserialize "good" values into the field as it tries to do a type conversion to the `Nullable<T>` wrapper.  This change ensures that when deserializing a non empty string value, we use the underlying type of any `Nullable` type.  We already were doing this for regular `enum`s but not `StringEnum<>`s

Added tests for enum, nullable enum, StringEnum and nullable StringEnum to capture the failing state, then proved the fix :+1:

### DIFF
--- a/Octokit/Models/Response/StringEnum.cs
+++ b/Octokit/Models/Response/StringEnum.cs
@@ -29,7 +29,7 @@ namespace Octokit
 
         public StringEnum(string stringValue)
         {
-            Ensure.ArgumentNotNull(stringValue, nameof(stringValue));
+            Ensure.ArgumentNotNullOrEmptyString(stringValue, nameof(stringValue));
 
             _stringValue = stringValue;
             _parsedValue = null;
--- a/Octokit/Models/Response/StringEnum.cs
+++ b/Octokit/Models/Response/StringEnum.cs
@@ -144,6 +144,11 @@ namespace Octokit
 
         private TEnum ParseValue()
         {
+            if (string.IsNullOrEmpty(_stringValue))
+            {
+                return default(TEnum);
+            }
+
             TEnum value;
             if (TryParse(out value))
             {
--- a/Octokit/Models/Response/StringEnum.cs
+++ b/Octokit/Models/Response/StringEnum.cs
@@ -68,12 +68,6 @@ namespace Octokit
                 return true;
             }
 
-            if (StringValue == null)
-            {
-                value = default(TEnum);
-                return false;
-            }
-
             try
             {
                 // Use the SimpleJsonSerializer to parse the string to Enum according to the GitHub Api strategy
@@ -146,11 +140,6 @@ namespace Octokit
 
         private TEnum ParseValue()
         {
-            if (_stringValue == null)
-            {
-                return default(TEnum);
-            }
-
             TEnum value;
             if (TryParse(out value))
             {
--- a/Octokit/Models/Response/StringEnum.cs
+++ b/Octokit/Models/Response/StringEnum.cs
@@ -29,7 +29,9 @@ namespace Octokit
 
         public StringEnum(string stringValue)
         {
-            _stringValue = stringValue ?? string.Empty;
+            Ensure.ArgumentNotNull(stringValue, nameof(stringValue));
+
+            _stringValue = stringValue;
             _parsedValue = null;
         }
 
@@ -66,7 +68,7 @@ namespace Octokit
                 return true;
             }
 
-            if (string.IsNullOrEmpty(StringValue))
+            if (StringValue == null)
             {
                 value = default(TEnum);
                 return false;
@@ -144,7 +146,7 @@ namespace Octokit
 
         private TEnum ParseValue()
         {
-            if (string.IsNullOrEmpty(_stringValue))
+            if (_stringValue == null)
             {
                 return default(TEnum);
             }
--- a/Octokit.Tests/Models/StringEnumTests.cs
+++ b/Octokit.Tests/Models/StringEnumTests.cs
@@ -8,9 +8,11 @@ namespace Octokit.Tests.Models
         public class TheCtor
         {
             [Fact]
-            public void ShouldThrowForNullStringValue()
+            public void ShouldThrowForNullOrEmptyStringValues()
             {
                 Assert.Throws<ArgumentNullException>(() => new StringEnum<AccountType>(null));
+
+                Assert.Throws<ArgumentException>(() => new StringEnum<AccountType>(""));
             }
 
             [Fact]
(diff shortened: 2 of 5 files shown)


khellang:
Looks like a legit failure:

```
Failed   Octokit.Tests.SimpleJsonSerializerTests+TheSerializeMethod.HandlesEnum
Error Message:
 System.ArgumentException : Value '' is not a valid 'SomeEnum' enum value.
```

khellang:
Well, the internal represenation will be `string.Empty` if `null` is passed, then it'll default to `default(T)` (where `T` is the enum type) if you try to parse it.

ryangribble:
The serualizer doesn't know about TryParse though, it just calls `Value` which throws an exception because empty string isn't a valid enum member. 

I've tweaked `.Value` to return `default(T)` rather than throw an exception but I'm also wondering whether it should even be "valid" to pass a `null` into the `ctor`. 

My reasoning is, if the response may contain nulls or omit a field, then the response class should just declare that member as a nullable type, just like we would do for a regular enum field. Only drama is any current response classes that are relying on the ctor allowing nulls to be passed in will need to be found and changed to declare as nullable. We've got pretty decent integration test coverage though so I'm considering doing this

khellang:
> I've tweaked `.Value` to return `default(T)` rather than throw an exception but I'm also wondering whether it should even be "valid" to pass a `null` into the `ctor`.

But doesn't that default the purpose of `.Value`? Isn't the point that it should throw an exception so you have to handle "unknown" values explicitly. Silently falling back to `default(T)` would make the application continue to run, but with possibly incorrect values.

khellang:
> My reasoning is, if the response may contain nulls or omit a field, then the response class should just declare that member as a nullable type, just like we would do for a regular enum field. Only drama is any current response classes that are relying on the ctor allowing nulls to be passed in will need to be found and changed to declare as nullable. We've got pretty decent integration test coverage though so I'm considering doing this.

Agreed. I think the current behavior (`null` throwing) should be expected (it *is* invalid after all), and the solution should be to make the response property nullable instead of silently falling back to `default(T)`.

Octokit/Models/Response/StringEnum.cs
```diff
@@ -66,7 +68,7 @@ public bool TryParse(out TEnum value)
                 return true;
             }
 
-            if (string.IsNullOrEmpty(StringValue))
+            if (StringValue == null)
```
khellang: How can `StringValue` be null at this point?

ryangribble:
@khellang could you take a look at the latest stuff?  It contains the commits that change the `StringEnum<T>` behaviour such that a `null` passed into the `ctor` now throws an exception, and an initialised member of type `StringEnum<T>` now returnss `default(T)` for `Value` property.  Thus, it behaves in line with a regular enum.


I've then gone down the rabbit hole in fixing the serialization side of things.  Ultimately i found the serializer was saving out the entire StringEnum struct as json (ie enumerating all properties and including `string_value` and `value` etc) so Ive added special handling for types of `StringEnum<T>` to get the enum value and then serialize it like a normal enum (taking into account `[Parameter()]` attributes and the like.

Octokit/Models/Response/StringEnum.cs
ryangribble: It seems to be `null` in the case where you have a class member of this type that is unitialized.

Octokit/Models/Response/StringEnum.cs
```diff
@@ -144,6 +146,11 @@ public override string ToString()
 
         private TEnum ParseValue()
         {
+            if (_stringValue == null)
```
khellang: You've already verified in the ctor that `_stringValue` can't be null?

Octokit/Models/Response/StringEnum.cs
ryangribble: I added a test to repro this here: https://github.com/octokit/octokit.net/pull/1760/files#diff-728143ef4b5a14ceed6d74d06f9c9df7R62

this is certainly doing my head in!

khellang: Couldn't this potentially lead to incorrect behavior?

If you have an uninitialized property, it means that the payload is missing the property, which means it's optional, but Octokit has modeled it as required (by it not being nullable).

By just silently falling back to the default value, you could potentially give the user the incorrect default (who decides what the "default value" is?)

ryangribble: right... I was losing sight of the fact we only use this `StringEnum<T>` in response models, because the unit tests are using dummy classes where it's possible to declare classes with uninitialised members and so on.

But from a technical point of view - the same would be true of an `enum` property on a response class now wouldn't it?  If it was not marked nullable, and the API response didnt specify it, then we will be providing the "incorrect" default enum value on the response model instead of a null.  

I guess the question is, should an incorrectly not nullable `StringEnum<T>` behave the same as an incorrectly not nullable regular `enum`?

khellang: > But from a technical point of view - the same would be true of an enum property on a response class now wouldn't it?

Yes, but that doesn't make it any less wrong 😉 There's a reason you should always declare all value-type properties as nullable when binding to external input. It's the only way to check whether that input actually existed, or if it's just the default (uninitialized) value.

> I guess the question is, should an incorrectly not nullable StringEnum<T> behave the same as an incorrectly not nullable regular enum?

I vote "no", simply because it's a nice way to "flush out" mistakes where we've assumed something was required, but was actually optional.

khellang: Let's say you have a model:

khellang: Let's say you have a model:

```csharp
public class Account
{
    public Guid Id { get; set; }
    public decimal Amount { get; set; }
}
```

And someone posts

```json
{ "id": "3d216536-0869-4c9d-b2b7-2a735498c634" }

**Comment by ryangribble** [comment]
let's see how this looks.  I've removed that "fallback `default(T)` behaviour.  But still asserting that `null` can't be passed into the `ctor`.

The only slightly hinky part is that the serializer blows up on a class with an uninitialized (not nullable) `StringEnum<T>` member.  Now as you say, given we use this on response classes and not on request classes AND the fact that anywhere it is used it should be declared nullable if it is infact optional... I guess this is OK that the test has that peculiarity (could always make the test class have a ctor that mandates the members be supplied/initialized)

File: Octokit/Models/Response/StringEnum.cs [hunk 3]
```diff
@@ -29,7 +29,9 @@ public StringEnum(TEnum parsedValue)
 
         public StringEnum(string stringValue)
         {
-            _stringValue = stringValue ?? string.Empty;
+            Ensure.ArgumentNotNull(stringValue, nameof(stringValue));
```
ryangribble: should we enforce no empty strings here too?

khellang: Yeah, I guess it wouldn't hurt.

ryangribble:
Is this looking OK to you @khellang  ?

Just running through all the integration tests now

khellang:
Yeah, I can't see anything wrong with it 😉 LGTM!

ryangribble:
Integration tests are 💎

ryangribble:
release_notes: Deserializer now handles nullable `StringEnum<T>` members

PR #1770, opened 2018-02-23
Add PreviousFileName to the PullRequestFile model

### LINKED ISSUE #1768: Add PreviousFileName property for PullRequestFile
When a file is renamed the github api returns a previous_filename property on [pull request files](https://developer.github.com/v3/pulls/#list-pull-requests-files). Could we add the property to PullRequestFile similar to how it's returned on [GithubCommitFile](https://github.com/octokit/octokit.net/blob/c9b2c1260bc87b7782d4fd9645d43a8885203923/Octokit/Models/Response/GitHubCommitFile.cs#L87)?

Kaneraz:
Fixes #1768 

Pretty straight forward change, and gets my consuming project rolling again. Unit test and convention tests pass, but it looks like I'm having unrelated issues with the integration tests.

### DIFF
--- a/Octokit/Models/Response/PullRequestFile.cs
+++ b/Octokit/Models/Response/PullRequestFile.cs
@@ -9,7 +9,7 @@ namespace Octokit
     {
         public PullRequestFile() { }
 
-        public PullRequestFile(string sha, string fileName, string status, int additions, int deletions, int changes, string blobUrl, string rawUrl, string contentsUrl, string patch)
+        public PullRequestFile(string sha, string fileName, string status, int additions, int deletions, int changes, string blobUrl, string rawUrl, string contentsUrl, string patch, string previousFileName)
         {
             Sha = sha;
             FileName = fileName;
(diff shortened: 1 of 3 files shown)


Octokit/Models/Response/PullRequestFile.cs
```diff
@@ -34,8 +35,10 @@ public PullRequestFile(string sha, string fileName, string status, int additions
         public string RawUrl { get; protected set; }
         public string ContentsUrl { get; protected set; }
         public string Patch { get; protected set; }
+        [Parameter(Key = "previous_filename")]
+        public string PreviousFileName { get; protected set; }
 
-        internal string DebuggerDisplay
+    internal string DebuggerDisplay
```
ryangribble: Something happened to whitespace/alignment here?

Kaneraz: Ah, sorry about that.

ryangribble:
Many thanks @Kaneraz

ryangribble:
release_notes: Add `PreviousFileName` field to `PullRequestFile` response

PR #1771, opened 2018-02-25
Adds MaintainerCanModify to PullRequest

### LINKED ISSUE #1769: Add MaintainerCanModify for PullRequest

Cyberboss:
Closes #1769


Octokit/Models/Response/PullRequest.cs
```diff
@@ -219,6 +220,11 @@ public bool Merged
         /// </summary>
         public bool Locked { get; protected set; }
 
+        /// <summary>
+        /// If users with write access to the base repository can push to the HEAD branch
+        /// </summary>
+        public bool MaintainerCanModify { get; protected set; }
```
ryangribble: I think we should make this nullable `bool? MaintainerCanModify` because responses that return lists of pull requests will not return this field, so rather than indicate it as `false` in that case, we should indicate it as `null`

Octokit/Models/Response/PullRequest.cs
```diff
@@ -219,6 +220,11 @@ public bool Merged
         /// </summary>
         public bool Locked { get; protected set; }
 
+        /// <summary>
+        /// If users with write access to the base repository can push to the HEAD branch
```
ryangribble: Could we reword this slightly?

"Whether maintainers of the base repository can push to the HEAD branch"

Cyberboss:
Yeah I'll add that too

Octokit/Models/Request/NewPullRequest.cs
```diff
@@ -15,7 +15,8 @@ public class NewPullRequest
         /// <param name="title">The title of the pull request.</param>
         /// <param name="head">The branch (or git ref where your changes are implemented. In other words, the source branch/ref</param>
         /// <param name="baseRef">The base (or git ref) reference you want your changes pulled into. In other words, the target branch/ref</param>
-        public NewPullRequest(string title, string head, string baseRef)
+        /// <param name="maintainerCanModify">Whether maintainers of the base repository can push to <paramref name="head"/> if it is a branch</param>
+        public NewPullRequest(string title, string head, string baseRef, bool maintainerCanModify)
```
ryangribble: we only want *mandatory* parameters to be in the constructor, so this can be removed sorry!

Octokit/Models/Request/NewPullRequest.cs
```diff
@@ -41,6 +43,11 @@ public NewPullRequest(string title, string head, string baseRef)
         /// </summary>
         public string Head { get; private set; }
 
+        /// <summary>
+        /// Whether maintainers of the base repository can push to <see cref="Head"/> if it is a branch
+        /// </summary>
+        public bool MaintainerCanModify { get; private set; }
```
ryangribble: This should be nullable `bool?` since it's an optional parameter.  That way when the user doesnt set it, we dont send anything in the request, and thus we default to whatever the API default is (which is important in this case since the API default is true, and if we leave this non nullable we would default to sending false)

Octokit/Models/Request/PullRequestUpdate.cs
```diff
@@ -29,6 +29,11 @@ public class PullRequestUpdate
         /// </summary>
         public string Base { get; set; }
 
+        /// <summary>
+        /// Whether maintainers of the base repository can push to the pull request's HEAD if it is a branch
+        /// </summary>
+        public bool MaintainerCanModify { get; set; }
```
ryangribble: This should also be nullable `bool?` so that the user is able to not specify a value and have it not change this property on the PR they are updating

ryangribble:
@Cyberboss just letting you know I havent fogotten about this!  I've been out of town but will get to this this week.  I wanted to add the new field to some integration tests just to prove that it is being exercised correctly

ryangribble:
release_notes: Added `MaintainerCanModify` field on `PullRequest` response and `NewPullRequest` and `UpdatePullRequest` requests

PR #1779, opened 2018-03-06
Tidy up compiler warnings

ryangribble:
- Suppress "missing XmlComment for publicly visible member" warning as it floods the build logs and we don't actually need XmlComments on _every_ public member
- Fix warnings regarding incorrect xml param comments
- Fix other compiler warnings


ryangribble:
release_notes: Correct missing/incorrect XmlDoc entries for parameters on some methods

PR #1784, opened 2018-03-17
Build/Test projects to use .NET Core SDK 2.0.x 

ryangribble:
split out of #1738 

Originally this was required to support using some nuget packages in integration tests that required `netcoreapp2`, but we ran into sourcelink errors when using SDK 2.0.  This turned out to be unrelated TravisCI errors around network calls timing out etc anyway... 

In the meantime, that dependent package was changed to only require an earlier `netstandard` again, so #1738 went back to 1.x SDK tooling.

We may as well move to SDK 2 tooling though, so that we aren't falling too far behind.  Note that the libraries themselves are still built targeting `netstandard1.1`, this PR only updates the SDK used for build steps, and moves test projects to `netcoreapp2.0` as well


ryangribble:
release_notes: Update Octokit build tooling to use .NET SDK 2.x (note that this is only an SDK tooling update - `Octokit` and `Octokit.Reactive` libraries are still targeting `netstandard1.1`)

PR #1788, opened 2018-03-24
Enable support for milestone-based issues queries

### LINKED ISSUE #1787: Allow querying by milestones
It seems it's not possible to query by `milestone` using `SearchIssueRequest`. Would be great to add support for it or at least some mechanism to specify custom set of parameters. I've tried to use the `Parameters` dictionary directly, but it is being overwritten every time it's accessed, based on the rest of the search request parameters.

mkArtak:
Fixes #1787

### DIFF
--- a/Octokit/Models/Request/SearchIssuesRequest.cs
+++ b/Octokit/Models/Request/SearchIssuesRequest.cs
@@ -258,6 +258,15 @@ namespace Octokit
         /// </remarks>
         public string User { get; set; }
 
+        /// <summary>
+        /// Gets or sets the milestone to filter issues based on
+        /// </summary>
+        public string Milestone
+        {
+            get;
+            set;
+        }
+
         [SuppressMessage("Microsoft.Usage", "CA2227:CollectionPropertiesShouldBeReadOnly")]
         public RepositoryCollection Repos { get; set; }
 
@@ -389,6 +398,11 @@ namespace Octokit
                 }
 
                 parameters.AddRange(Repos.Select(x => string.Format(CultureInfo.InvariantCulture, "repo:{0}", x)));
+            }
+
+            if (Milestone.IsNotBlank())
+            {
+                parameters.Add(string.Format(CultureInfo.InvariantCulture, "milestone:{0}", Milestone));
             }
 
             // Add any exclusion parameters
--- a/Octokit/Models/Request/SearchIssuesRequest.cs
+++ b/Octokit/Models/Request/SearchIssuesRequest.cs
@@ -1,11 +1,11 @@
-﻿using System;
+﻿using Octokit.Internal;
+using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.Linq;
-using Octokit.Internal;
 
 namespace Octokit
 {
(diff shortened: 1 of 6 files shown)


Octokit/Models/Request/SearchIssuesRequest.cs
```diff
@@ -398,7 +398,7 @@ public override IReadOnlyList<string> MergedQualifiers()
 
             if (Milestone.IsNotBlank())
             {
-                parameters.Add(string.Format(CultureInfo.InvariantCulture, "milestone:{0}", Milestone));
+                parameters.Add(string.Format(CultureInfo.InvariantCulture, "milestone:\"{0}\"", Milestone));
```
ryangribble: Im not so keen on the automatic quoting here.  Firstly, because other search criteria (eg Labels) don't do this, so this would mean behavior isnt consistent...  Secondly because if a user already specified quotes themselves, this would then end up double quoting their input.  

Thoughts?

mkArtak: I was trying to keep the code simple, so decided to add quotes for all the cases, rather than add them only when there are spaces in the value. Sure, from consistency perspective I see your point.
That will then leave the responsibility of specifying quotes on the user.
Let's play this out, and assume that the `Milestone` property is already available. From the library user's perspective, I would just see a property to set the milestone filter, so I'll apply whatever my milestone value is. I shouldn't really add a layer on top of the library I use which should be aware of the internals of how the interaction with GitHub works. And that "quoting" is actually an internal behavior.
The bottom line is, if we won't add quotes we force the library users to be aware of the internals of GitHub search API. So I still believe that adding quotes is correct. Even more, I would prefer to get another issue filed to add the quotes across all the other properties too (where applicable of course).

Thoughts?

ryangribble: I do agree with you from a usability point of view and yeah as mentioned my concern is about overall library consistency, so for me I would be happy to keep the auto-quoting in place here, provided:
- We handle the case if the user DOES provide quotes - either by stripping theirs to add yours, or by only adding quotes if there arent already some there etc.  
- We consider adding quotes is now something we should be doing in any similar cases in the future, and ideally also look at any existing places where this could be fixed up (on a new issue/PR)

mkArtak: Cool, thanks @ryangribble. I'll update the code to make sure quotes are added only if there aren't already.
As for the second point, I'll file an issue to address that in the future.

mkArtak: After some experimentation it looks like a milestone can have `"` as part of its name. That means, that we can't assume that the quotes the user will provide in the milestone property are to wrap it. Rather, those are just part of the milestone name. Here is a sample search output to confirm this:
image

When searching with just additional double quotes, or without - the search returns nothing:
image

So searching the way I did in the first image is correct, which is - to escape all existing quotes and to always wrap the milestone with double quotes. I'm going to implement that way.

ryangribble: :+1:

Oh one other thing I remembered while poking around into this side of things... we will probably also want to add milestone support to the `SearchIssuesRequestExclusions` request, assuming it's possible to search for items that DONT have a given milestone

mkArtak:
@ryangribble, do you have any other comments?

mkArtak:
Done, @ryangribble. Good suggestion!

ryangribble:
release_notes: Added the ability to search issues by milestones, using `SearchIssuesRequest.Milestone`

ryangribble:
Thanks @mkArtak for the PR and congrats on your first contribution to Octokit!

PR #1796, opened 2018-04-13
Added a pre-receive environments client

tasadar2:
This exposes the endpoints for working with pre-receive environments, [https://developer.github.com/v3/enterprise-admin/pre_receive_environments](https://developer.github.com/v3/enterprise-admin/pre_receive_environments)

Notes:
I did not find a good overload for a null payload with accept headers. And adding the obvious overload for this scenario would cause ambiguous call issues. So I ended up passing `new object()` as the payload for the `Delete` and `TriggerDownload` methods.
Integration tests are passing using our enterprise instance.


ryangribble:
Hey @tasadar2 , thanks for this PR.  I'll need a bit of time to go through it

In the meantime, I've merged some other PR's so this one needs a rebase or merge from master.  Thanks!

Octokit/Models/Request/Enterprise/NewPreReceiveEnvironment.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        /// <summary>
+        /// The name of the environment as displayed in the UI.
+        /// </summary>
+        public string Name { get; set; }
+
+        /// <summary>
+        /// URL to the tarball that will be downloaded and extracted.
+        /// </summary>
+        [Parameter(Key = "image_url")]
```
ryangribble: FYI, the octokit serializer/deserializer will automatically handle converting from ruby snake case (`image_url`) to c# camel case `ImageUrl` based on the capitalization converting to underscores... so you don't need to specify `[Parameter()]` attributes unless there is a mismatch.  So most of the attributes added in this PR can actually be removed.  (note, the exception is on `Enum` members, where the attribute is still required!)

Octokit/Models/Request/Enterprise/UpdatePreReceiveEnvironment.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.Diagnostics;
+
+namespace Octokit
+{
+    /// <summary>
+    /// Describes an update to an existing pre-receive environment.
+    /// </summary>
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
+    public class UpdatePreReceiveEnvironment : NewPreReceiveEnvironment
```
ryangribble: I think for consistency with the rest of the code base, it would be good to avoid an `Update` request inheriting from the `New` request, and instead implement the required fields directly in this class

Octokit/Models/Response/Enterprise/PreReceiveEnvironment.cs
```diff
@@ -0,0 +1,78 @@
...
+    /// <summary>
+    /// Describes a pre-receive environment.
+    /// </summary>
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
+    public class PreReceiveEnvironment
+    {
+        /// <summary>
+        /// Identifier for the pre-receive environment.
+        /// </summary>
+        public int Id { get; protected set; }
```
ryangribble: Can we use `long` for the Id field, just in case we ever reach max int?  Probably wouldnt happen on pre receive environments, but as a general policy all Id fields should be `long` (we got bitten by PullRequest Status Ids in the past!)

ryangribble:
> I did not find a good overload for a null payload with accept headers. And adding the obvious overload for this scenario would cause ambiguous call issues. So I ended up passing new object() as the payload for the Delete and TriggerDownload methods.
That's all good, we've done the same in a few other places as well, when `Post` or `Delete` methods don't have a payload.  The overloads on the `IConnection` calss are a bit out of hand unfortunately, but because they are `public` it will be a breaking change to clean things up and we try to avoid breaking changes where possible.

> Integration tests are passing using our enterprise instance.
That's good to know.  We have an enterprise instance too, but I've never looked into pre-receive environments as yet

Octokit/Models/Response/Enterprise/PreReceiveEnvironment.cs
tasadar2: Good call, i'll update that

Octokit/Models/Request/Enterprise/UpdatePreReceiveEnvironment.cs
tasadar2: I split these up, also updated the tests to account for optional values

Octokit/Models/Request/Enterprise/NewPreReceiveEnvironment.cs
tasadar2: Nice to find a serializer that accounts for that, I automatically assumed I would need an attribute;)

Octokit/Models/Request/Enterprise/UpdatePreReceiveEnvironment.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System.Diagnostics;
+using System.Globalization;
+
+namespace Octokit
+{
+    /// <summary>
+    /// Describes an update to an existing pre-receive environment.
+    /// </summary>
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
+    public class UpdatePreReceiveEnvironment
```
ryangribble: do you know if the `Name` or `ImageUrl` are required parameters for the update call?  Wondering if there should be a constructor that takes these

Octokit/Models/Response/Enterprise/PreReceiveEnvironment.cs
```diff
@@ -0,0 +1,76 @@
...
+using System.Diagnostics;
+using System.Globalization;
+
+namespace Octokit
+{
+    /// <summary>
+    /// Describes a pre-receive environment.
+    /// </summary>
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
+    public class PreReceiveEnvironment
```
ryangribble: response class needs parameterless ctor and ctor taking/setting all properties

Octokit/Models/Response/Enterprise/PreReceiveEnvironmentDownload.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Diagnostics;
+using System.Globalization;
+
+namespace Octokit
+{
+    /// <summary>
+    /// Describes the current download state of a pre-receive environment image.
+    /// </summary>
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
+    public class PreReceiveEnvironmentDownload
```
ryangribble: response class needs the 2 ctors mentioned previously

tasadar2:
The response classes are using implied parameterless constructors, so that is probably why the convention tests passed, and is for sure why the integration tests pass. I will get some constructors with all parameters added, which will require the explicit parameterless one as well.

Octokit/Models/Request/Enterprise/UpdatePreReceiveEnvironment.cs
tasadar2: Both are optional, and I have integration tests for omitting one or the other.

The call also works when no parameters are passed, although that doesn't really serve a purpose.
So, we could
1. Add a unit tests for the update nothing scenario
2. Create a constructor that ensures that at least one value is set. Can't really do separate constructors as both values are strings.
3. Leave it as is

Octokit/Models/Response/Enterprise/PreReceiveEnvironment.cs
tasadar2: Updated

Octokit/Models/Response/Enterprise/PreReceiveEnvironmentDownload.cs
tasadar2: Updated

ryangribble:
Hey @tasadar2 sorry for the delay!  I finally got a chance to run the integration tests against a GHE instance... made a couple of tweaks as you can see in the commits :+1:

Things are good to go from my perspective if you can please review the changes I just pushed up and let me know if you're good too?

tasadar2:
Changes look good to me.

ryangribble:
awesome stuff @tasadar2 I'd love us to get even more GitHub Enterprise features implemented :+1:

It reminds me I have some stale PR's that I should rebase and get merged (#1269, #1415 )

ryangribble:
LGTM

PR #1798, opened 2018-04-21
Adding a convention test to detect whether a model has a constructor exposing all properties

tasadar2:
This is a quick stab at the test, and it currently validates against the name.

Currently this would act more as a reminder/standard vs actually verifying if the ctor works as expected. It does not evaluate if the type is accurate(which nullable would probably need to be accounted for on both sides). It does not detect if the actual property is set with the corresponding parameter value.

- [X] Detect models without ctor exposing all properties
- [X] Display actionable members from closest ctor match
~~Instead of ignoring properties from base classes, possibly traverse the ctor tree to determine if all properties are exposed top down~~
- [x] Expose missing properties for the now offending models

There are a couple dozen legitimate offenders. And a couple with parameter names slightly different from the property name.


ryangribble:
Hehe I was actually digging into this and wrote a similar test last night myself :grinning:  yours is better though :+1: 

Mine did include base class properties but we had even more legit fails in that case... Eg [ApplicationAuthorization](https://github.com/octokit/octokit.net/blob/master/Octokit/Models/Response/ApplicationAuthorization.Cs) which inherits 
[Authorization](https://github.com/octokit/octokit.net/blob/master/Octokit/Models/Response/Authorization.cs) but doesn't allow setting the base class properties via the ctor.

ryangribble:
I should also mention if we encounter any cases where it doesn't make sense to abide by the convention we can introduce an attribute (decorated at the response class declaration level) to suppress the test for that class, and filter them out in the test

ryangribble:
release_notes;  Added convention tests to ensure Response classes have a `ctor` that exposes all properties (to support mocking)

tasadar2:
Thanks for going through the ctors.

I didn't see a good way through reflection to determine what base ctor was being called. We could open the rule up to all members by removing the `DeclaringType` condition. The scenario I was trying to solve for was for the ctors that pass explicit values to their base ctor. As mentioned, these could be ignored with attributes if it makes sense.
Ex. `AccountType` in [https://github.com/octokit/octokit.net/blob/c9b2c1260bc87b7782d4fd9645d43a8885203923/Octokit/Models/Response/Organization.cs#L12-L13](https://github.com/octokit/octokit.net/blob/c9b2c1260bc87b7782d4fd9645d43a8885203923/Octokit/Models/Response/Organization.cs#L12-L13)

ryangribble:
OK so if we include the base class stuff, there are 4 offences.  `User` and `Organization` have issues because they pass through the hardcoded `AccountType` enum value.

Then we have `ApplicationAuthorization` which actually doesn't call it's base `ctor` at all (quite odd!)
https://github.com/octokit/octokit.net/blob/c9b2c1260bc87b7782d4fd9645d43a8885203923/Octokit/Models/Response/ApplicationAuthorization.cs#L16

and `Merge` which again doesn't pass through to it's base `ctor` at all
https://github.com/octokit/octokit.net/blob/c9b2c1260bc87b7782d4fd9645d43a8885203923/Octokit/Models/Response/Merge.cs#L13

ryangribble:
In terms of the latter 2 I think it will be fine to just add the base ctor to the existing one, and all of the existing parameters that go with it. After all, these ctor really are only useful for creating mock responses, our deserializer uses the parameter less one, and the protected setter properties so the blast radius is pretty small on these and I like the idea of enforcing the convention that all fields could be mocked

I'm terms of the first 2 I'm thinking the best approach is to add an attribute but rather than suppressing the whole test, the attribute could take a parameter of field names to be exempt. That way we can still enforce the other fields are present, but exempt the Account Type field since it's not valid to create an Organization with any other type

Sound good?

tasadar2:
Sounds good.

I was thinking the same thing with excluding individual properties rather than the whole class. Pushed an update which should account for those last scenarios.

ryangribble:
Thanks @tasadar2 👏 

LGTM

PR #1801, opened 2018-05-07
[WIP] Implement the preview checks API

Cyberboss:
Very WIP at the moment, but has most of the checks runs portion (aside from response contructors). Advice and contributions welcome


Octokit/Clients/CheckRunsClient.cs
```diff
@@ -0,0 +1,133 @@
...
+
+            throw new NotImplementedException();
+        }
+
+        public Task<IReadOnlyList<CheckRun>> List(long repositoryId, long checkSuiteId, CheckRunRequest checkRunRequest, ApiOptions options)
+        {
+            Ensure.ArgumentNotNull(checkRunRequest, nameof(checkRunRequest));
+            Ensure.ArgumentNotNull(options, nameof(options));
+
+            throw new NotImplementedException();
```
Cyberboss: Listing check runs for a reference doesn't follow the standard pagination rules for most other APIs. Advice?

shiftkey: If it doesn't follow the conventions, you can exclude it here:

https://github.com/octokit/octokit.net/blob/3345f76fc9a499aca6f49defb345a7926ceb4c87/Octokit.Tests.Conventions/PaginationTests.cs#L96-L105

ryangribble: why do you say it doesnt follow the standard rules?  From the response in the docs it indicates it should support "normal" pagination:

image

ryangribble: that said, if there is a case where it doesnt follow normal conventions, another option rather than excluding the whole client interface as @shiftkey mentioned, is that we also have an attributes which can suppress pagination conventions for naming and/or ApiOptions overloads, on a per method basis :+1:
`[ExcludeFromPaginationNamingConventionTest]` and `[ExcludeFromPaginationApiOptionsConventionTest]`

ryangribble:
Thanks @Cyberboss for jumping on this so quickly... this is an awesome feature and we are very keen to get it implemented in octokit.net 😀 

I just had a quick 👀at this stage, I should have some more time over the weekend, but here is what jumped out immediately:

- "List" methods should be named `GetAllForReference` and `GetAllForCheckSuite`

- The annotations shouldnt be in a separate client as we structure our clients based on the API doc structure, so in this case there would be a top level `IChecksClient` with sub clients of `ICheckRunsClient` and `ICheckSuitesClient`.   Annotations would just be methods such as `ICheckRunsClient.GetAllAnnotations`

- We will need to expand our `GitHubApp` permission class to contain the new `check` permission

- A lot of `sealed` and `internal` classes which should just be `public`

- You forgot to include the `AcceptHeaders.ChecksApiPreview` in all the `ApiConnection` calls 😀 

- Watch out for the URI's as the owner/repo ones are `repos/:owner/:repo/etc` whilst the repository ID ones are `repositories/:id/etc` (you currently have `/repositories/` for all)

ryangribble:
there's alot to review here, would be better to focus on one client at a time, and ensure unit tests and integration tests are passing 😉 

A couple of other consistency things:

- to maintain consistency with other clients, can we list the `owner`/`repo` version of a method first, then the `repository_id` version?  This is because we are POC'ing some code generation tools so consistent ordering is going to be important

- similary, the naming convention for clients is that the class name is plural (which you've done, eg `IGitHubCheckRunsClient`) but the accessor method is singular.  so in terms of usage, we should be `github.Check.Run.GetAllForReference()` rather than `github.Checks.Runs.GetAllForReference()`

ryangribble:
yeah ideally in interfaces and implementation classes.  Hopefully the code base is reasonably consistent with this already

Octokit/Clients/CheckRunsClient.cs
```diff
@@ -0,0 +1,190 @@
...
+        public async Task<IReadOnlyList<CheckRun>> GetAllForReference(string owner, string name, string reference, CheckRunRequest checkRunRequest, ApiOptions options)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(owner, nameof(owner));
+            Ensure.ArgumentNotNullOrEmptyString(name, nameof(name));
+            Ensure.ArgumentNotNullOrEmptyString(reference, nameof(reference));
+            Ensure.ArgumentNotNull(checkRunRequest, nameof(checkRunRequest));
+            Ensure.ArgumentNotNull(options, nameof(options));
+
+            var results = await ApiConnection.GetAll<CheckRunList>(ApiUrls.ReferenceCheckRuns(owner, name, reference), checkRunRequest.ToParametersDictionary(), AcceptHeaders.ChecksApiPreview, options).ConfigureAwait(false);
+            return results.SelectMany(x => x.CheckRuns).ToList();
```
Cyberboss: This is how I ended up implementing the GetAll methods with object responses. Pagination on it's own doesn't handle the model

Cyberboss: Also, how would this translate to the reactive methods?

ryangribble:
Ive split this PR up, to make it easier to get it done.  Please see #1846 for the Check Suites side of things.  Then we can get the Check Runs side done...

PR #1802, opened 2018-05-12
Implement improved labels API

### LINKED ISSUE #1786: Label descriptions are not retrievable
`Octokit.Label` does not expose Description and I can't see an exposed method to fetch it.

jozefizso:
* [x] Add new fields to Label response class
* [x] New API: NewLabel and LabelUpdate contructors with description argument
* [x] write unit tests
* [x] write integration tests

Fixes #1786

### DIFF
--- a/Octokit/Models/Request/LabelUpdate.cs
+++ b/Octokit/Models/Request/LabelUpdate.cs
@@ -18,13 +18,15 @@ namespace Octokit
         /// </summary>
         /// <param name="name">The name of the label.</param>
         /// <param name="color">The color of the label.</param>
-        public LabelUpdate(string name, string color)
+        /// <param name="description">The description of the label.</param>
+        public LabelUpdate(string name, string color, string description = null)
         {
             Ensure.ArgumentNotNullOrEmptyString(name, nameof(name));
             Ensure.ArgumentNotNullOrEmptyString(color, nameof(color));
 
             Name = name;
             Color = color;
+            Description = description;
         }
 
         /// <summary>
@@ -48,6 +50,11 @@ namespace Octokit
                 _color = value;
             }
         }
+
+        /// <summary>
+        /// A short description of the label (optional).
+        /// </summary>
+        public string Description { get; set; }
 
         internal string DebuggerDisplay
         {
--- a/Octokit/Models/Request/LabelUpdate.cs
+++ b/Octokit/Models/Request/LabelUpdate.cs
@@ -18,15 +18,13 @@ namespace Octokit
         /// </summary>
         /// <param name="name">The name of the label.</param>
         /// <param name="color">The color of the label.</param>
-        /// <param name="description">The description of the label.</param>
-        public LabelUpdate(string name, string color, string description = null)
+        public LabelUpdate(string name, string color)
         {
             Ensure.ArgumentNotNullOrEmptyString(name, nameof(name));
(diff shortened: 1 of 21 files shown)


Octokit/Models/Response/Label.cs
```diff
@@ -30,6 +32,16 @@ public Label(string url, string name, string color)
...
 
+        /// <summary>
+        /// Description of the label
+        /// </summary>
+        public string Description { get; protected set; }
+
+        /// <summary>
+        /// Is default label
+        /// </summary>
+        public bool IsDefault { get; protected set; }
```
ryangribble: I think this will need to be named Default to match the json response field

Octokit/Models/Request/LabelUpdate.cs
```diff
@@ -18,13 +18,15 @@ public class LabelUpdate
         /// </summary>
         /// <param name="name">The name of the label.</param>
         /// <param name="color">The color of the label.</param>
-        public LabelUpdate(string name, string color)
+        /// <param name="description">The description of the label.</param>
+        public LabelUpdate(string name, string color, string description = null)
```
ryangribble: To be consistent with the rest of the library, the request model constructor should only take required parameters. Optional parameters can be set via object initializer pattern

Octokit/Models/Request/NewLabel.cs
```diff
@@ -18,23 +18,32 @@ public class NewLabel
         /// </summary>
         /// <param name="name">The name of the label.</param>
         /// <param name="color">The color of the label.</param>
-        public NewLabel(string name, string color)
+        /// <param name="description">The description of the label.</param>
+        public NewLabel(string name, string color, string description = null)
```
ryangribble: Optional field shouldn't be in the ctor

ryangribble:
Looking pretty 💎 @jozefizso 

I realised we also need to add preview header into the various `GetAll` methods in `ObservableIssuesLabelsClient` as these don't call through to the normal client.

eg: 
https://github.com/octokit/octokit.net/blob/master/Octokit.Reactive/Clients/ObservableIssuesLabelsClient.cs#L74

jozefizso:
Added the header to observable client. I hope I didn't miss come calls.

ryangribble:
Looks like you got them all 🤗

I'll do a final review tonight...  things are looking pretty good!

ryangribble:
breaking_change: Note that the `IssuesLabelsClient.RemoveFromIssue()` methods which previously had no return value, will now return an `IReadonlyList<Label>`.  This change is source compatible but not binary compatible.

ryangribble:
release_notes: Implemented support for [Label API Improvements](https://developer.github.com/changes/2018-02-22-label-description-search-preview/), including additional fields (`Description` and `Default`), emoji support and searching for labels (`SearchClient.SearchLabels()`)

ryangribble:
@jozefizso I found a few more methods that needed the preview header, and also decided to implement the final part of the "enhanced labels API changes", being the "search for labels" call.

Are you able to review what I've pushed up?  🙏

jozefizso:
It looks good. I've fixed the documentation a bit.

ryangribble:
release_notes: Implement [Label API Improvements (Preview)](https://developer.github.com/changes/2018-02-22-label-description-search-preview/)

ryangribble:
Thanks @jozefizso !!!

LGTM

PR #1805, opened 2018-06-01
Correct upload asset code

mungojam:
1. Remove incorrect usage of await on synchronous method
2. Dispose of stream after use


ryangribble:
release_notes: Improved the "Upload Release Asset" doc sample

ryangribble:
Thanks @mungojam

PR #1813, opened 2018-06-12
[WIP] Fixes #1718 Implement Repository Transfer functionality

### LINKED ISSUE #1718: Implement "Repository Transfer" functionality
As per [this recent blog post](https://developer.github.com/changes/2017-11-09-repository-transfer-api-preview/) there is now the ability to transfer ownership of a repository via the API (in preview mode)

Documentation: https://developer.github.com/v3/repos/#transfer-a-repository

- This should be implemented as a `Transfer` method on the `IRepositoriesClient` and `IObservableRepositoriesClient` clients
- Accept header `application/vnd.github.nightshade-preview+json` is required, to enable the preview functionality
- The `TeamId` field on the request object only applies to organisation owned repositories, so the `ctor` of the request object class should reflect this by oferring an overload that doesn't require this input, and the xmldoc should also make mention of the fact the team_id only applies to org repositories

wqferr:
Pull request addressing #1718 
- [x] (From issue) Repository transfer should be implemented as a ```Transfer``` method on the ```IRepositoriesClient``` and ```IObservableRepositoriesClient``` clients
- [x] (From issue) Accept header ```application/vnd.github.nightshade-preview+json``` is required, to enable the preview functionality
- [x] (From issue) The ```TeamId``` field on the request object only applies to organisation owned repositories, so the ```ctor``` of the request object class should reflect this by oferring an overload that doesn't require this input, and the xmldoc should also make mention of the fact the ```team_id``` only applies to org repositories
- [x] (From ```CONTRIBUTING.md```) XML documentation on the new methods
- [x] (From ```CONTRIBUTING.md```) Run convention tests successfully
- [x] (From ```CONTRIBUTING.md```) Create new unit tests for request object constructor
- [x] (From ```CONTRIBUTING.md```) Integration tests
- [x] Add unit tests for the ```Transfer``` client methods in [RepositoriesClientTest](https://github.com/octokit/octokit.net/blob/master/Octokit.Tests/Clients/RepositoriesClientTests.cs)
- [x] Implement 2nd endpoint (if it exists (it probably does)) that takes the repository Id ```/repositories/:id/transfer```
- [x] Use IReadOnlyList<int> rather than a normal int[] array for TeamId
- [x] Do something more succinct for ```DebuggerDisplay``` property in ```RepositoryTransfer```
- [x] Make Ensure.NotNullOrEmptyList into Ensure.NotNullOrEmptyEnumerable
- [x] XmlDoc comments should be added to RepositoryTransfer
- [x] Add XmlDoc for second endpoint
- [x] Implement integration tests
- [x] Debug integration tests for transferring with team ids
---
Points of concern:
- I'm not sure if I'm the one supposed to implement and/or run the integration tests, but if it is me then I have no idea of what to do. Could someone please point me in the right direction?
- Unit tests were all successful, but I'm not sure they've covered everything they were supposed to.

---
Edit: Updated "to do" list in replies
Edit 2: Update main "to do" list

### DIFF
--- a/Octokit.Reactive/Clients/IObservableRepositoriesClient.cs
+++ b/Octokit.Reactive/Clients/IObservableRepositoriesClient.cs
@@ -49,6 +49,8 @@ namespace Octokit.Reactive
         /// <param name="repositoryTransfer">Repository transfer information</param>
         /// <returns>A <see cref="Repository"/></returns>
         IObservable<Repository> Transfer(string currentOwner, string name, RepositoryTransfer repositoryTransfer);
+        
+        IObservable<Repository> Transfer(long repositoryId, RepositoryTransfer repositoryTransfer);
 
         /// <summary>
         /// Retrieves the <see cref="Repository"/> for the specified owner and name.
--- a/Octokit.Reactive/Clients/IObservableRepositoriesClient.cs
+++ b/Octokit.Reactive/Clients/IObservableRepositoriesClient.cs
@@ -39,6 +39,18 @@ namespace Octokit.Reactive
         IObservable<Unit> Delete(long repositoryId);
 
         /// <summary>
+        /// Transfers the ownership of the specified repository.
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="https://developer.github.com/v3/repos/#transfer-a-repository">API documentation</a> for more information.
+        /// </remarks>
+        /// <param name="currentOwner">The current owner of the repository</param>
+        /// <param name="name">The name of the repository</param>
+        /// <param name="repositoryTransfer">Repository transfer information</param>
+        /// <returns>A <see cref="Repository"/></returns>
+        IObservable<Repository> Transfer(string currentOwner, string name, RepositoryTransfer repositoryTransfer);
+
+        /// <summary>
         /// Retrieves the <see cref="Repository"/> for the specified owner and name.
         /// </summary>
         /// <param name="owner">The owner of the repository</param>
--- a/Octokit.Reactive/Clients/IObservableRepositoriesClient.cs
+++ b/Octokit.Reactive/Clients/IObservableRepositoriesClient.cs
@@ -44,11 +44,11 @@ namespace Octokit.Reactive
         /// <remarks>
         /// See the <a href="https://developer.github.com/v3/repos/#transfer-a-repository">API documentation</a> for more information.
         /// </remarks>
-        /// <param name="currentOwner">The current owner of the repository</param>
+        /// <param name="owner">The current owner of the repository</param>
         /// <param name="name">The name of the repository</param>
         /// <param name="repositoryTransfer">Repository transfer information</param>
         /// <returns>A <see cref="Repository"/></returns>
-        IObservable<Repository> Transfer(string currentOwner, string name, RepositoryTransfer repositoryTransfer);
+        IObservable<Repository> Transfer(string owner, string name, RepositoryTransfer repositoryTransfer);
 
         /// <summary>
         /// Transfers the ownership of the specified repository.
--- a/Octokit.Reactive/Clients/IObservableRepositoriesClient.cs
+++ b/Octokit.Reactive/Clients/IObservableRepositoriesClient.cs
@@ -49,7 +49,16 @@ namespace Octokit.Reactive
         /// <param name="repositoryTransfer">Repository transfer information</param>
         /// <returns>A <see cref="Repository"/></returns>
         IObservable<Repository> Transfer(string currentOwner, string name, RepositoryTransfer repositoryTransfer);
-        
+
+        /// <summary>
+        /// Transfers the ownership of the specified repository.
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="https://developer.github.com/v3/repos/#transfer-a-repository">API documentation</a> for more information.
+        /// </remarks>
+        /// <param name="repositoryId">The id of the repository</param>
+        /// <param name="repositoryTransfer">Repository transfer information</param>
+        /// <returns>A <see cref="Repository"/></returns>
         IObservable<Repository> Transfer(long repositoryId, RepositoryTransfer repositoryTransfer);
 
         /// <summary>
--- a/Octokit.Reactive/Clients/ObservableRepositoriesClient.cs
+++ b/Octokit.Reactive/Clients/ObservableRepositoriesClient.cs
@@ -111,6 +111,15 @@ namespace Octokit.Reactive
             return _client.Transfer(currentOwner, name, repositoryTransfer).ToObservable();
         }
 
+        /// <summary>
+        /// Transfers the ownership of the specified repository.
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="https://developer.github.com/v3/repos/#transfer-a-repository">API documentation</a> for more information.
+        /// </remarks>
+        /// <param name="repositoryId">The id of the repository</param>
+        /// <param name="repositoryTransfer">Repository transfer information</param>
+        /// <returns>A <see cref="Repository"/></returns>
         public IObservable<Repository> Transfer(long repositoryId, RepositoryTransfer repositoryTransfer)
         {
             return _client.Transfer(repositoryId, repositoryTransfer).ToObservable();
--- a/Octokit.Reactive/Clients/ObservableRepositoriesClient.cs
+++ b/Octokit.Reactive/Clients/ObservableRepositoriesClient.cs
@@ -102,13 +102,13 @@ namespace Octokit.Reactive
         /// <remarks>
         /// See the <a href="https://developer.github.com/v3/repos/#transfer-a-repository">API documentation</a> for more information.
         /// </remarks>
-        /// <param name="currentOwner">The current owner of the repository</param>
+        /// <param name="owner">The current owner of the repository</param>
         /// <param name="name">The name of the repository</param>
         /// <param name="repositoryTransfer">Repository transfer information</param>
         /// <returns>A <see cref="Repository"/></returns>
-        public IObservable<Repository> Transfer(string currentOwner, string name, RepositoryTransfer repositoryTransfer)
+        public IObservable<Repository> Transfer(string owner, string name, RepositoryTransfer repositoryTransfer)
         {
-            return _client.Transfer(currentOwner, name, repositoryTransfer).ToObservable();
+            return _client.Transfer(owner, name, repositoryTransfer).ToObservable();
         }
 
         /// <summary>
--- a/Octokit.Reactive/Clients/ObservableRepositoriesClient.cs
+++ b/Octokit.Reactive/Clients/ObservableRepositoriesClient.cs
@@ -108,6 +108,10 @@ namespace Octokit.Reactive
         /// <returns>A <see cref="Repository"/></returns>
         public IObservable<Repository> Transfer(string owner, string name, RepositoryTransfer repositoryTransfer)
         {
+            Ensure.ArgumentNotNullOrEmptyString(owner, nameof(owner));
+            Ensure.ArgumentNotNullOrEmptyString(name, nameof(name));
+            Ensure.ArgumentNotNull(repositoryTransfer, nameof(repositoryTransfer));
+
             return _client.Transfer(owner, name, repositoryTransfer).ToObservable();
         }
 
@@ -122,6 +126,8 @@ namespace Octokit.Reactive
         /// <returns>A <see cref="Repository"/></returns>
         public IObservable<Repository> Transfer(long repositoryId, RepositoryTransfer repositoryTransfer)
         {
+            Ensure.ArgumentNotNull(repositoryTransfer, nameof(repositoryTransfer));
+            
             return _client.Transfer(repositoryId, repositoryTransfer).ToObservable();
         }
 
--- a/Octokit.Reactive/Clients/ObservableRepositoriesClient.cs
+++ b/Octokit.Reactive/Clients/ObservableRepositoriesClient.cs
@@ -95,6 +95,21 @@ namespace Octokit.Reactive
         {
             return _client.Delete(repositoryId).ToObservable();
         }
+        
+        /// <summary>
+        /// Transfers the ownership of the specified repository.
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="https://developer.github.com/v3/repos/#transfer-a-repository">API documentation</a> for more information.
+        /// </remarks>
+        /// <param name="currentOwner">The current owner of the repository</param>
+        /// <param name="name">The name of the repository</param>
+        /// <param name="repositoryTransfer">Repository transfer information</param>
+        /// <returns>A <see cref="Repository"/></returns>
+        public IObservable<Repository> Transfer(string currentOwner, string name, RepositoryTransfer repositoryTransfer)
+        {
+            return _client.Transfer(currentOwner, name, repositoryTransfer).ToObservable();
+        }
 
         /// <summary>
         /// Retrieves the <see cref="Repository"/> for the specified owner and name.
--- a/Octokit.Reactive/Clients/ObservableRepositoriesClient.cs
+++ b/Octokit.Reactive/Clients/ObservableRepositoriesClient.cs
@@ -111,6 +111,11 @@ namespace Octokit.Reactive
             return _client.Transfer(currentOwner, name, repositoryTransfer).ToObservable();
         }
 
+        public IObservable<Repository> Transfer(long repositoryId, RepositoryTransfer repositoryTransfer)
+        {
+            return _client.Transfer(repositoryId, repositoryTransfer).ToObservable();
+        }
+
         /// <summary>
         /// Retrieves the <see cref="Repository"/> for the specified owner and name.
         /// </summary>
--- a/Octokit.Tests.Integration/Clients/RepositoriesClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/RepositoriesClientTests.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
@@ -1699,7 +1699,7 @@ public class RepositoriesClientTests
         }
         
         [IntegrationTest]
-        public async Task TransfersFromOrgToUserWithId()
+        public async Task TransfersFromOrgToUserById()
         {
             var github = Helper.GetAuthenticatedClient();
             var newRepo = new NewRepository(Helper.MakeNameWithTimestamp("transferred-repo"));
--- a/Octokit.Tests.Integration/Clients/RepositoriesClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/RepositoriesClientTests.cs
@@ -1682,10 +1682,10 @@ public class RepositoriesClientTests
         }
     }
 
-    public class TheTransferByNameMethod
+    public class TheTransferMethod
     {
         [IntegrationTest]
-        public async Task CanTransferOrgRepoToUser()
+        public async Task TransfersFromOrgToUser()
         {
             var github = Helper.GetAuthenticatedClient();
             var newRepo = new NewRepository(Helper.MakeNameWithTimestamp("transferred-repo"));
@@ -1697,9 +1697,23 @@ public class RepositoriesClientTests
                 var transferred = await github.Repository.Get(newOwner, context.RepositoryName);
             }
         }
+        
+        [IntegrationTest]
+        public async Task TransfersFromOrgToUserWithId()
+        {
+            var github = Helper.GetAuthenticatedClient();
+            var newRepo = new NewRepository(Helper.MakeNameWithTimestamp("transferred-repo"));
+            var newOwner = Helper.UserName;
+            using (var context = await github.CreateRepositoryContext(Helper.Organization, newRepo))
+            {
+                var transfer = new RepositoryTransfer(newOwner);
+                await github.Repository.Transfer(context.RepositoryId, transfer);
+                var transferred = await github.Repository.Get(context.RepositoryId);
+            }
+        }
 
         [IntegrationTest]
-        public async Task CanTransferUserRepoToOrg()
+        public async Task TransfersFromUserToOrg()
         {
             var github = Helper.GetAuthenticatedClient();
             var newRepo = new NewRepository(Helper.MakeNameWithTimestamp("transferred-repo"));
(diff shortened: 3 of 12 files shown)


Octokit/Models/Request/RepositoryTransfer.cs
```diff
@@ -0,0 +1,53 @@
+using System;
+using System.Text;
+using System.Diagnostics;
+using System.Globalization;
+
+namespace Octokit
+{
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
+    public class RepositoryTransfer
```
ryangribble: XmlDoc comments should be added to this class including it's ctor/s and the public properties

Octokit/Models/Request/RepositoryTransfer.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            Ensure.ArgumentNotNullOrEmptyArray(teamId, nameof(teamId));
+
+            TeamId = new int[teamId.Length];
+            Array.Copy(teamId, TeamId, teamId.Length);
+        }
+
+        public string NewOwner { get; set; }
+
+        public int[] TeamId { get; set; }
```
ryangribble: We would normally use `IReadOnlyList<int>` rather than a normal `int[]` array

take a look at other request models for examples (eg [here](https://github.com/octokit/octokit.net/blob/master/Octokit/Models/Request/PullRequestReviewRequest.cs))

Octokit/Models/Request/RepositoryTransfer.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        internal string DebuggerDisplay
+        {
+            get
+            {
+                StringBuilder strBuilder = new StringBuilder();
```
ryangribble: You could probably do something more succinctly like:

``` csharp
string.Format(CultureInfo.InvariantCulture, "NewOwner: {0}, TeamId: {1}", 
   NewOwner,  
   string.Join(",", TeamId ?? "null"));

**Comment by wqferr** [comment]
Thank you for such an in depth reply.
I'll do these things asap, although I'm a bit in a hurry this next week.

File: Octokit/Helpers/Ensure.cs [hunk 4]
```diff
@@ -49,16 +50,16 @@ public static void GreaterThanZero([ValidatedNotNull]TimeSpan value, string name
         }
 
         /// <summary>
-        /// Checks an array argument to ensure it isn't null or empty.
+        /// Checks a list argument to ensure it isn't null or empty.
         /// </summary>
         /// <param name = "value">The argument value to check</param>
         /// <param name = "name">The name of the argument</param>
-        public static void ArgumentNotNullOrEmptyArray<T>([ValidatedNotNull]T[] value, string name)
+        public static void ArgumentNotNullOrEmptyList<T>([ValidatedNotNull]IReadOnlyList<T> value, string name)
```
ryangribble: this method could probably just take an `IEnumerable<T>` rather than `IReadOnlyList<T>` and would then be more reusable for checking other potential lists in the future

ryangribble:
Coming along nicely 👍

wqferr:
Sorry to bug you with this but I'm new to this open source thing and I'm not sure where to ask, so I'll do it in this "thread".
My implementation of the API seems correct to me, I've checked the JSON serializer and (to me at least) it seems like it should be working, but the request sent never actually adds the teams it passes as ```team_id```. Other than that, the integration tests (org -> user and user -> org w/o teams) pass.

On another note, about the other end point: it's not listed in the [api](https://developer.github.com/v3/repos/#transfer-a-repository). It only mentions ```POST /repos/:owner/:repo/transfer```, not ```repositories/:id/transfer```. Is this a case of incomplete documentation? Or does it actually not exist. If it does exist but isn't listed, am I even supposed to try and implement it?

Octokit.Tests.Integration/Clients/RepositoriesClientTests.cs
```diff
@@ -1682,7 +1682,7 @@ public class TheGetLicenseContentsMethod
         }
     }
 
-    public class TheTransferMethod
+    public class TheTransferByNameMethod
```
ryangribble: These names should actually reflect the method name, which is `Transfer`.  So what we would do here is have one test class called `TheTransferMethod` and then have `[IntegrationTest]` methods called `TransfersFromOrgToUser` (for the "normal" endpoint) and `TransfersFromOrgToUserWithRepositoryId` (for the repositoryId endpoint)

Octokit/Clients/IRepositoriesClient.cs
```diff
@@ -112,6 +112,8 @@ public interface IRepositoriesClient
         /// <returns>A <see cref="Repository"/></returns>
         Task<Repository> Transfer(string currentOwner, string name, RepositoryTransfer repositoryTransfer);
 
+        Task<Repository> Transfer(long id, RepositoryTransfer repositoryTransfer);
```
ryangribble: by convention we always name the parameter `repositoryId`

Octokit.Tests.Integration/Clients/RepositoriesClientTests.cs
```diff
@@ -1697,57 +1697,42 @@ public class TheTransferByNameMethod
                 var transferred = await github.Repository.Get(newOwner, context.RepositoryName);
             }
         }
-
+        
         [IntegrationTest]
-        public async Task CanTransferUserRepoToOrg()
+        public async Task TransfersFromOrgToUserWithId()
```
ryangribble: to be consistent with your other method names, this should be `ById` rather than `WithId`

Octokit.Tests.Integration/Clients/RepositoriesClientTests.cs
```diff
@@ -1697,57 +1697,42 @@ public class TheTransferByNameMethod
...
-            var newOwner = Helper.Organization;
-            using (var context = await github.CreateRepositoryContext(newRepo))
+            var newOwner = Helper.UserName;
+            using (var context = await github.CreateRepositoryContext(Helper.Organization, newRepo))
             {
                 var transfer = new RepositoryTransfer(newOwner);
-                await github.Repository.Transfer(context.RepositoryOwner, context.RepositoryName, transfer);
-                var transferred = await github.Repository.Get(newOwner, context.RepositoryName);
+                await github.Repository.Transfer(context.RepositoryId, transfer);
+                var transferred = await github.Repository.Get(context.RepositoryId);
```
ryangribble: eventhough this would throw an exception if the repo didnt exist under the requested owner (thus failing the test), I think i'd still prefer to see some kind of `Assert` in each test, for consistency.

eg: `Assert.Equal(newOwner, transferred.Owner.Login);`

Octokit.Tests.Integration/Clients/RepositoriesClientTests.cs
```diff
@@ -1756,21 +1741,33 @@ public class TheTransferByIdMethod
...
         [IntegrationTest]
-        public async Task CanTransferUserRepoToOrg()
+        public async Task TransfersFromUserToOrgWithTeams()
         {
+            // FIXME API doesn't add teams when transferring to an organization
             var github = Helper.GetAuthenticatedClient();
             var newRepo = new NewRepository(Helper.MakeNameWithTimestamp("transferred-repo"));
             var newOwner = Helper.Organization;
+
+            Random rand = new Random();
```
ryangribble: we dont want to rely on the test org having to have existing teams.  We have a `TeamContext` that can handle creating a team and destroying it again when it goes out of scope.  So you could rewrite this test to something like this:

``` csharp
        [IntegrationTest]
        public async Task TransfersFromUserToOrgWithTeams()
        {
            // FIXME API doesn't add teams when transferring to an organization
            var newRepo = new NewRepository(Helper.MakeNameWithTimestamp("transferred-repo"));
            var newOwner = Helper.Organization;

            using (var teamContext = await github.CreateTeamContext(Helper.Organization, new NewTeam(Helper.MakeNameWithTimestamp("team"))))
            using (var context = await github.CreateRepositoryContext(newRepo))
            {
                var transfer = new RepositoryTransfer(newOwner, new int[] { teamContext.TeamId });
                await github.Repository.Transfer(context.RepositoryOwner, context.RepositoryName, transfer);

                var repoTeams = await github.Repository.GetAllTeams(context.RepositoryId);

                Assert.True(repoTeams.Any(x => x.Id == teamContext.TeamId));
            }
        }
```

wqferr: Fair enough, I didn't think that one through. I'll change this right now.

Octokit.Tests.Integration/Clients/RepositoriesClientTests.cs
wqferr: Oops.

ryangribble:
What I think we can do for now is [skip the failing tests](https://github.com/octokit/octokit.net/blob/3c818934b8b3b4cdc5472bf01e869c07343c3928/Octokit.Tests.Integration/Clients/RepositoryPagesClientTests.cs#L21) with a comment saying "The upstream API doesn't seem to be applying requested Teams at this time"

Octokit.Tests/Clients/RepositoriesClientTests.cs
```diff
@@ -211,6 +211,158 @@ public class TheCreateMethodForOrganization
...
+        public class TheTransferMethod
+        {
+            [Fact]
+            public void EnsuresNonNullArguments()
+            {
+                var connection = Substitute.For<IApiConnection>();
+                var client = new RepositoriesClient(connection);
+                var transfer = new RepositoryTransfer("newOwner");
+
+                Assert.ThrowsAsync<ArgumentNullException>(
```
ryangribble: All of these tests are actually not running properly because `Assert.ThrowsAsync` must be awaited

Octokit.Tests/Clients/RepositoriesClientTests.cs
```diff
@@ -211,6 +211,158 @@ public class TheCreateMethodForOrganization
...
+            
+            [Fact]
+            public void EnsuresNonNullArgumentsById()
+            {
+                var connection = Substitute.For<IApiConnection>();
+                var client = new RepositoriesClient(connection);
+                var transfer = new RepositoryTransfer("newOwner");
+                var repositoryId = 1;
+
+                Assert.ThrowsAsync<ArgumentNullException>(
```
ryangribble: needs to be `await`'ed

Octokit.Tests/Clients/RepositoriesClientTests.cs
```diff
@@ -211,6 +211,158 @@ public class TheCreateMethodForOrganization
...
+            }
+
+            [Fact]
+            public void EnsuresNonEmptyArguments()
+            {
+                var connection = Substitute.For<IApiConnection>();
+                var client = new RepositoriesClient(connection);
+                var transfer = new RepositoryTransfer("newOwner");
+
+                Assert.ThrowsAsync<ArgumentException>(
```
ryangribble: needs to be `await`'ed

Octokit/Clients/RepositoriesClient.cs
```diff
@@ -163,6 +163,35 @@ public Task Delete(long repositoryId)
...
+        /// <remarks>
+        /// See the <a href="https://developer.github.com/v3/repos/#transfer-a-repository">API documentation</a> for more information.
+        /// </remarks>
+        /// <param name="currentOwner">The current owner of the repository</param>
+        /// <param name="name">The name of the repository</param>
+        /// <param name="repositoryTransfer">Repository transfer information</param>
+        /// <returns>A <see cref="Repository"/></returns>
+        public Task<Repository> Transfer(string currentOwner, string name, RepositoryTransfer repositoryTransfer)
+        {
+            return ApiConnection.Post<Repository>(ApiUrls.RepositoryTransfer(currentOwner, name), repositoryTransfer, AcceptHeaders.RepositoryTransferPreview);
```
ryangribble: This is missing the `Ensure.ArgumentNotNullOrEmptyString()` checks for `currentOwner`, `name` and `Ensure.ArgumentNotNull()` check for `repositoryTransfer` (and the unit tests are not picking it up due to not being `await`'ed)

Octokit/Clients/RepositoriesClient.cs
```diff
@@ -163,6 +163,35 @@ public Task Delete(long repositoryId)
...
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="https://developer.github.com/v3/repos/#transfer-a-repository">API documentation</a> for more information.
+        /// </remarks>
+        /// <param name="repositoryId">The id of the repository</param>
+        /// <param name="repositoryTransfer">Repository transfer information</param>
+        /// <returns>A <see cref="Repository"/></returns>
+        public Task<Repository> Transfer(long repositoryId, RepositoryTransfer repositoryTransfer)
+        {
+            return ApiConnection.Post<Repository>(ApiUrls.RepositoryTransfer(repositoryId), repositoryTransfer, AcceptHeaders.RepositoryTransferPreview);
```
ryangribble: This is missing the `Ensure.ArgumentNotNull()` check for `repositoryTransfer` (and the unit tests are not picking it up due to not being `await`'ed)

Octokit/Helpers/ApiUrls.cs
```diff
@@ -1834,6 +1834,22 @@ public static Uri RepositoryTags(string owner, string name)
...
+        /// </summary>
+        /// <param name="currentOwner">The current owner of the repository</param>
+        /// <param name="name">The name of the repository</param>
+        /// <returns></returns>
+        public static Uri RepositoryTransfer(string currentOwner, string name)
+        {
+            return "repos/{0}/{1}/transfer".FormatUri(currentOwner, name);
+        }
+
+        public static Uri RepositoryTransfer(long repositoryId)
```
ryangribble: If the above owner/name method has XmlDoc comments provided then we should add them for this repositoryId one too

Octokit/Models/Request/RepositoryTransfer.cs
```diff
@@ -0,0 +1,70 @@
...
+    /// </summary>
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
+    public class RepositoryTransfer
+    {
+        /// <summary>
+        /// Creates a new repository transfer description with no team Ids.
+        /// </summary>
+        /// <param name="newOwner">The new owner of the repository after the transfer.</param>
+        /// <remarks>
+        /// This is the only valid constructor for repositories whose current owner is not an
```
ryangribble: I dont think we need this remark

Octokit/Models/Request/RepositoryTransfer.cs
```diff
@@ -0,0 +1,70 @@
...
+
+            NewOwner = newOwner;
+        }
+
+        /// <summary>
+        /// Creates a new repository transfer description.
+        /// </summary>
+        /// <param name="newOwner">The new owner of the repository after the transfer.</param>
+        /// <param name="teamId">A list of team Ids to add to the repository after the transfer.</param>
+        /// <remarks>
```
ryangribble: The wording on this remark is a bit odd (what do you mean by "invalid description"?).  Also <remarks> objects aren't visibile in intellisense AFAIK, so I would suggest removing this remark, and instead what I normally do is type a brief explanation in brackets of the description of the param element itself 

eg
`/// <param name="teamId">A list of team Ids to add to the repository after the transfer (only applies to Organization owned repositories).</param>`

Octokit/Models/Request/RepositoryTransfer.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        /// <summary>
+        /// The new owner of the repository after the transfer.
+        /// </summary>
+        public string NewOwner { get; set; }
+
+        /// <summary>
+        /// A list of team Ids to add to the repository after the transfer.
+        /// </summary>
+        /// <remarks>
```
ryangribble: ditto here.  I would remove this <remark> and put `(only applies to Organization owned repositories)` in the <summary>

ryangribble:
Good news, I heard back from the folks at github and it sounds like the docs have a typo - the parameter is actually called `team_ids` so let's rename it and see if your tests then work!

wqferr:
Integration tests work like a charm now, thanks for the help!

wqferr:
What is with this Travis CI test? It seems really inconsistent.
Twice the branch passed it in one commit, then I changed XmlDoc in the next one and it stops passing with some error associated with linksource.
Should I be doing something to fix it? I have no idea.

ryangribble:
Yeah the travis thing isn't anything you have done... We are unfortunately running into sporadic failures in the sourcelink test step which seem to be network issues. It's happening to others as well

Octokit.Reactive/Clients/ObservableRepositoriesClient.cs
```diff
@@ -95,6 +95,35 @@ public IObservable<Unit> Delete(long repositoryId)
...
+        /// <remarks>
+        /// See the <a href="https://developer.github.com/v3/repos/#transfer-a-repository">API documentation</a> for more information.
+        /// </remarks>
+        /// <param name="currentOwner">The current owner of the repository</param>
+        /// <param name="name">The name of the repository</param>
+        /// <param name="repositoryTransfer">Repository transfer information</param>
+        /// <returns>A <see cref="Repository"/></returns>
+        public IObservable<Repository> Transfer(string currentOwner, string name, RepositoryTransfer repositoryTransfer)
+        {
+            return _client.Transfer(currentOwner, name, repositoryTransfer).ToObservable();
```
ryangribble: Add `Ensure.xxx` guards

Octokit.Reactive/Clients/ObservableRepositoriesClient.cs
```diff
@@ -95,6 +95,35 @@ public IObservable<Unit> Delete(long repositoryId)
...
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="https://developer.github.com/v3/repos/#transfer-a-repository">API documentation</a> for more information.
+        /// </remarks>
+        /// <param name="repositoryId">The id of the repository</param>
+        /// <param name="repositoryTransfer">Repository transfer information</param>
+        /// <returns>A <see cref="Repository"/></returns>
+        public IObservable<Repository> Transfer(long repositoryId, RepositoryTransfer repositoryTransfer)
+        {
+            return _client.Transfer(repositoryId, repositoryTransfer).ToObservable();
```
ryangribble: Add `Ensure.xxx` guards

Octokit/Clients/IRepositoriesClient.cs
```diff
@@ -100,6 +100,29 @@ public interface IRepositoriesClient
...
+        /// Transfers the ownership of the specified repository.
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="https://developer.github.com/v3/repos/#transfer-a-repository">API documentation</a> for more information.
+        /// </remarks>
+        /// <param name="currentOwner">The current owner of the repository</param>
+        /// <param name="name">The name of the repository</param>
+        /// <param name="repositoryTransfer">Repository transfer information</param>
+        /// <returns>A <see cref="Repository"/></returns>
+        Task<Repository> Transfer(string currentOwner, string name, RepositoryTransfer repositoryTransfer);
```
ryangribble: rename to `owner`

Octokit/Clients/RepositoriesClient.cs
```diff
@@ -163,6 +163,41 @@ public Task Delete(long repositoryId)
...
+        /// Transfers the ownership of the specified repository.
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="https://developer.github.com/v3/repos/#transfer-a-repository">API documentation</a> for more information.
+        /// </remarks>
+        /// <param name="currentOwner">The current owner of the repository</param>
+        /// <param name="name">The name of the repository</param>
+        /// <param name="repositoryTransfer">Repository transfer information</param>
+        /// <returns>A <see cref="Repository"/></returns>
+        public Task<Repository> Transfer(string currentOwner, string name, RepositoryTransfer repositoryTransfer)
```
ryangribble: rename to `owner`

Octokit/Helpers/ApiUrls.cs
```diff
@@ -1834,6 +1834,27 @@ public static Uri RepositoryTags(string owner, string name)
             return "repos/{0}/{1}/tags".FormatUri(owner, name);
         }
 
+        /// <summary>
+        /// Returns the <see cref="Uri"/> for a repository transfer.
+        /// </summary>
+        /// <param name="currentOwner">The current owner of the repository</param>
+        /// <param name="name">The name of the repository</param>
+        /// <returns></returns>
+        public static Uri RepositoryTransfer(string currentOwner, string name)
```
ryangribble: rename to `owner`

Octokit.Reactive/Clients/IObservableRepositoriesClient.cs
```diff
@@ -38,6 +38,29 @@ public interface IObservableRepositoriesClient
...
+        /// Transfers the ownership of the specified repository.
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="https://developer.github.com/v3/repos/#transfer-a-repository">API documentation</a> for more information.
+        /// </remarks>
+        /// <param name="currentOwner">The current owner of the repository</param>
+        /// <param name="name">The name of the repository</param>
+        /// <param name="repositoryTransfer">Repository transfer information</param>
+        /// <returns>A <see cref="Repository"/></returns>
+        IObservable<Repository> Transfer(string currentOwner, string name, RepositoryTransfer repositoryTransfer);
```
ryangribble: rename to `owner`

Octokit.Reactive/Clients/ObservableRepositoriesClient.cs
```diff
@@ -95,6 +95,35 @@ public IObservable<Unit> Delete(long repositoryId)
...
+        /// Transfers the ownership of the specified repository.
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="https://developer.github.com/v3/repos/#transfer-a-repository">API documentation</a> for more information.
+        /// </remarks>
+        /// <param name="currentOwner">The current owner of the repository</param>
+        /// <param name="name">The name of the repository</param>
+        /// <param name="repositoryTransfer">Repository transfer information</param>
+        /// <returns>A <see cref="Repository"/></returns>
+        public IObservable<Repository> Transfer(string currentOwner, string name, RepositoryTransfer repositoryTransfer)
```
ryangribble: rename to `owner`

wqferr:
>since you've done such an amazing job so far

Aww, you're going to make me blush. Anyway, thanks!
Everything you asked for in this last batch is fixed, so I'll wait your next revision.

ryangribble:
Congrats on your first octokit.net contribution @wqferr - it is very much appreciated!  Fingers crossed you are keen to keep contributing!

PR #1818, opened 2018-06-18
A missing event state: comment_deleted

mirsaeedi:
Currently, Octokit does not support the comment_deleted event status. As an example, I was unable to get the Event.Value of one of the [VSCode issues](https://api.github.com/repos/Microsoft/vscode/issues/events/1246084209) and i recieved the following error:

> Requested value 'comment_deleted' was not found.

I'm using the latest version, which is 0.30.0

### DIFF
--- a/Octokit/Models/Response/EventInfo.cs
+++ b/Octokit/Models/Response/EventInfo.cs
@@ -284,6 +284,12 @@ namespace Octokit
         /// An issue that a user had previously marked as a duplicate of another issue is no longer considered a duplicate.
         /// </summary>
         [Parameter(Value = "unmarked_as_duplicate")]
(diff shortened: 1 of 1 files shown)


mirsaeedi:
I noticed there are [2 more Events](https://developer.github.com/v3/issues/events/#events-1) that are not defined in EventInfoState. So, I added them in my 2nd commit.

ryangribble:
Interesting... this `comment_deleted` event appears to be undocumented at this stage.

Thanks for finding this @mirsaeedi !  :+1:

ryangribble:
release_notes: Add missing `EventInfoState` values (`MarkedAsDuplicate`, `UnmarkedAsDuplicate` and `CommentDeleted`)

PR #1819, opened 2018-06-18
Mentioning the reset time is based on UTC standard

mirsaeedi:
At first sight, it not obvious that Reset shows time in UTC and I've seen a couple of times folks think it's based on their local timezone. I believe it will be a good addition to current document to mention that the time is UTC.


ryangribble:
Hi @mirsaeedi , thanks for the suggestion!

ryangribble:
release_notes: Clarify the rate limit `Reset` field is in UTC

PR #1822, opened 2018-06-20
Remove runtime dependency on SourceLink.Create.GitHub

ryangribble:
The last release inadvertently ended up with a nuget dependency on `SourceLink.Create.GitHub` which is only used as part of our build process...

Also bumped SourceLink packages to latest versions to keep them up to date

v0.30:
image

With this change:
image


ryangribble:
release_notes: Remove nuget dependency on `SourceLink.Create.GitHub` package

PR #1831, opened 2018-06-25
Fix exception in RepositoriesClient.GetAllLanguages() when no languages exist

### LINKED ISSUE #1820: GithubClient.Repository.GetAllLanguages throws Exception when the repo does not have a language
Example with this API endpoint: https://api.github.com/users/vinhhoa1/repos
Using the above  code will crash with this error for repository _RPG-ZombiesTCC_:

```
System.ArgumentNullException: Value cannot be null.

Parameter name: source

   at System.Linq.Enumerable.Select[TSource,TResult](IEnumerable`1 source, Func`2 selector)

   at Octokit.RepositoriesClient.<GetAllLanguages>d__84.MoveNext()
```

I think it would be better if _null_ value are marked as _""_ (empty) rather than throwing an exception here.

ryangribble:
Fixes #1820 

This endpoint is a bit "special" as it returns an object hash which we deserialize as a `Dictionary<string, long>` and then convert into a `ReadOnlyList`.  When no languages exist, the API returns `null` which then throws an exception in our conversion process.

This fix handles this situation by returning an empty list.

### DIFF
--- a/Octokit.Tests.Integration/Clients/RepositoriesClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/RepositoriesClientTests.cs
@@ -1343,6 +1343,30 @@ public class RepositoriesClientTests
             Assert.NotEmpty(languages);
             Assert.True(languages.Any(l => l.Name == "C#"));
         }
+
+        [IntegrationTest]
+        public async Task GetsEmptyLanguagesWhenNone()
+        {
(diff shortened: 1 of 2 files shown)


ryangribble:
cc @baonguyen96

ryangribble:
release_notes: Fix exception in `RepositoriesClient.GetAllLanguages()` when no languages exist

PR #1846, opened 2018-07-07
Implement Check Suites API

ryangribble:
Split out of #1801 to focus on Check Suites first

Remaining things to do:
- [x] XmlDoc for Response objects
- [x] XmlDoc for Observable client/interface
- [x] Tests for Observable client methods
- [x] Observable implementations

PR #1847, opened 2018-07-14
Implement Check Runs API

ryangribble:
Split out of #1801, focusing on Check Runs now Check Suites is done in #1846

Task List
- [x] Add CheckRunEvent activity/webhook payload
- [x] Create a check run
- [x] Update a check run
- [x] List check runs for a specific ref
- [x] List check runs in a check suite
- [x] Get a single check run
- [x] List annotations for a check run

### DIFF
--- a/Octokit/Models/Response/CheckRunAnnotation.cs
+++ b/Octokit/Models/Response/CheckRunAnnotation.cs
@@ -18,6 +18,7 @@ namespace Octokit
             EndLine = endLine;
             WarningLevel = warningLevel;
             Message = message;
+            Title = title;
             RawDetails = rawDetails;
         }
 
@@ -50,6 +51,11 @@ namespace Octokit
         /// A short description of the feedback for these lines of code
         /// </summary>
         public string Message { get; protected set; }
+
+        /// <summary>
+        /// The title that represents the annotation
+        /// </summary>
+        public string Title { get; set; }
 
         /// <summary>
         /// Details about this annotation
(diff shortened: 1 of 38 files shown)


ryangribble:
@Cyberboss @StanleyGoldman let me know what you think

Octokit/Models/Response/CheckRunAnnotation.cs
```diff
@@ -0,0 +1,61 @@
...
+namespace Octokit
+{
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
+    public class CheckRunAnnotation
+    {
+        public CheckRunAnnotation()
+        {
+        }
+
+        public CheckRunAnnotation(string filename, string blobHref, int startLine, int endLine, CheckWarningLevel warningLevel, string message, string title, string rawDetails)
```
StanleyGoldman: Missing a property for `Title`.

ryangribble: The response payload actually doesn't have that field, only the request seems to

StanleyGoldman: Odd, okay..

ryangribble: after checking internally, turns out this was an oversight and `title` will be added to this response, so I've added it to the mode :+1:  Also I realised I still had it in the `ctor` anyway, even though the class didn't have the property, which is probably what you were pointing out anyway!  🤣

ryangribble:
release_notes: Implemented Check Runs component of [New Checks Api (Public Beta)](https://developer.github.com/changes/2018-05-07-new-checks-api-public-beta/)

Cyberboss:
Read this over in email a few times, forgot to say, looks good

ryangribble:
Thanks for your help in kick-starting this @Cyberboss 👍

PR #1851, opened 2018-07-18
Release v0.31 - Check yo' self!

ryangribble:
- [x] generate release notes
- [x] run `.\build.ps1 -Target FormatCode` to tidy up whitespace/formatting
- [x] integration tests all pass (With a few known exceptions around collaborator/invite functionality having been locked down)
- [x] :thumbsup: from a contributor
- [ ] :shipit:

### DIFF
--- a/Octokit.Reactive/Clients/ObservableRepositoriesClient.cs
+++ b/Octokit.Reactive/Clients/ObservableRepositoriesClient.cs
@@ -95,7 +95,7 @@ namespace Octokit.Reactive
         {
             return _client.Delete(repositoryId).ToObservable();
         }
-        
+
         /// <summary>
         /// Transfers the ownership of the specified repository.
         /// </summary>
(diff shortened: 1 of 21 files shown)


Cyberboss:
Late to the party, but the CheckRunUpdate constructor requires a name field where it should be optional. https://developer.github.com/v3/checks/runs/#update-a-check-run

shiftkey:
@Cyberboss we might still be able to get that in - are you able to add in the right fix to `CheckRunUpdate`?

ryangribble:
Argh!  Yeah, looks like they [changed the documentation](https://github.com/octokit/routes/commit/bf289c0afba138d308af25267d7ac3b2861e3cc6#diff-5d76cdcdff3379ab7f53a945ceee5992) on the 18th July to say that field is no longer required!

I will quickly add another PR before I ship this release

ryangribble:
Actually I decided to just commit it here on the `release` branch.  Look ok now @Cyberboss  ?

ryangribble:
OK I think this is good to go, will :shipit: in a few hours

PR #1852, opened 2018-07-22
Fixes check runs taking over 2017 years to complete

Cyberboss:
Hahaha https://github.com/tgstation/tgstation/pull/39291/checks?check_run_id=8701837

Hopefully this doesn't break any tests

### DIFF
--- a/Octokit/Models/Request/CheckRunUpdate.cs
+++ b/Octokit/Models/Request/CheckRunUpdate.cs
@@ -38,7 +38,7 @@ namespace Octokit
         /// <summary>
         /// The time that the check run began
         /// </summary>
-        public DateTimeOffset StartedAt { get; set; }
+        public DateTimeOffset? StartedAt { get; set; }
 
         /// <summary>
         /// Required if you provide completed_at or a status of completed. The final conclusion of the check. Can be one of success, failure, neutral, cancelled, timed_out, or action_required. When the conclusion is action_required, additional details should be provided on the site specified by details_url.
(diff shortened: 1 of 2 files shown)


ryangribble:
Hmmm if this field allows nulls then we probably need to fix it in NewCheckRun request, and CheckRun response as well? 

https://github.com/Cyberboss/octokit.net/blob/2388138d33c0f401257c7fe31b6db417464fb60f/Octokit/Models/Request/NewCheckRun.cs#L50
https://github.com/Cyberboss/octokit.net/blob/2388138d33c0f401257c7fe31b6db417464fb60f/Octokit/Models/Response/CheckRun.cs#L71

Cyberboss:
It's not allowed to be null in those two iirc. Only when the check run is
updated.

ryangribble:
If it's not allowed to be null on creation then that should be marked as a required field for NewCheckRun then

Being null on the update does make sense though, so as not to overwrite the start time, however it wasn't evident whether an update "adds" to the initially created CheckRun or whether it "replaces" it with an updated one

I'll need to do some playing around before deciding exactly what this fix needs to cover

ryangribble:
release_notes: Adjust `StartedAt` and `Status` fields of `NewCheckRun` and `CheckRunUpdate` requests, to allow `null` values, avoiding resetting these to default values when not specified

ryangribble:
I tweaked the fields based on checking with GitHub folks on the detailed workings of these requests :+1:

Thanks @Cyberboss for spotting this

PR #1853, opened 2018-07-26
Docs: appClient - add GitHubApps subclass to method calls

d-a-s:
In implementing this, I found that `appClient.GetCurrent()` (etc.) don't work.
Appears that all the GitHub Apps methods were moved into a subclass called `GitHubApps`, so that `appClient.GitHubApps.GetCurrent()` etc does work. 

Could also be written by adding the `.GitHubApps` onto the end of the `appClient` constructor; not sure which would be better from a documentation standpoint.

Anyway, took me a while to figure it out, thought that updated docs might help others.

### DIFF
--- a/docs/github-apps.md
+++ b/docs/github-apps.md
@@ -54,13 +54,13 @@ Now we have an authenticated `GitHubApp` (`appClient`), we can query various top
 
 ``` csharp
 // Get the current authenticated GitHubApp
-var app = await appClient.GetCurrent();
+var app = await appClient.GitHubApps.GetCurrent();
 
 // Get a list of installations for the authenticated GitHubApp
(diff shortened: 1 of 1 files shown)


ryangribble:
Thanks @d-a-s and sorry for the confusion 🙏 

Your PR actually raises another /facepalm issue which is that the subclient is not named according to our conventions - it should be singular (`client.GitHubApp.XXX` instead of `client.GitHubApps.XXX`) - how did we miss that! 🤣

We will need to look at fixing this up although since there is already a release with `GitHubApps` we will need to deprecate the old property name over a couple of releases.

Thanks for fixing up the docs!

ryangribble:
LGTM

ryangribble:
release_notes: Fixed code samples in GitHub Apps sample docs

PR #1854, opened 2018-08-07
Additional GitHub Apps APIs

StanleyGoldman:
This API call comes in handy for integrators of GitHub Apps:

- https://developer.github.com/v3/apps/#list-installations-for-user
- https://developer.github.com/v3/apps/#find-repository-installation
- https://developer.github.com/v3/apps/#find-user-installation
- https://developer.github.com/v3/apps/#find-organization-installation
- https://developer.github.com/v3/apps/installations/#list-repositories
- https://developer.github.com/v3/apps/installations/#list-repositories-accessible-to-the-user-for-an-installation


StanleyGoldman:
There are a few more on that page that are not implemented. I'll attack them when I can.

StanleyGoldman:
I'm still dogfooding these.
The documentation on these APIs and which ones you would use then kind of leave me a bit confused.
So yea, I need to actually test them all in order to understand what data they are really getting.

StanleyGoldman:
I'm noticing a nuanced difference in the way some responses are delivered that is throwing this off..

https://developer.github.com/v3/apps/installations/#response

image

https://developer.github.com/v3/repos/#response

image

StanleyGoldman:
cc: @shiftkey ☝️

StanleyGoldman:
Nvm, I learned things today...

shiftkey:
@StanleyGoldman is this related to the `total_count` field (that feels a response from a search API, but `/apps/installations/` doesn't seem to behave like one) or was there some other questions/curiosities you had?

StanleyGoldman:
Yea, I was just mentally unprepared for things to be different between api calls like that.

Octokit.Reactive/Clients/IObservableGitHubAppsClient.cs
```diff
@@ -10,6 +10,14 @@ namespace Octokit.Reactive
     /// </remarks>
     public interface IObservableGitHubAppsClient
     {
+        /// <summary>
+        /// Access GitHub's Apps Installations API.
+        /// </summary>
+        /// <remarks>
+        /// Refer to the API documentation for more information: https://developer.github.com/v3/apps/installations/
+        /// </remarks>
+        IObservableGitHubAppsInstallationsClient Installations { get; }
```
ryangribble: The property name should be singular, eg `github.App.Installation.xxxx`

Octokit.Reactive/Clients/IObservableGitHubAppsClient.cs
```diff
@@ -53,5 +61,41 @@ public interface IObservableGitHubAppsClient
...
+        /// <remarks>https://developer.github.com/v3/apps/#list-installations-for-user</remarks>
+        IObservable<InstallationsResponse> GetAllInstallationsForUser(ApiOptions options);
+
+        /// <summary>
+        /// Enables an authenticated GitHub App to find the organizations's installation information.
+        /// </summary>
+        /// <remarks>https://developer.github.com/v3/apps/#find-repository-installation</remarks>
+        /// <param name="owner">The owner of the repo</param>
+        /// <param name="repo">The name of the repo</param>
+        IObservable<Installation> GetRepositoryInstallation(string owner, string repo);
```
ryangribble: I wonder if this should be named `GetRepositoryInstallationForCurrent()` since it uses the currently authenticated GitHub App... or otherwise `FindRepositoryInstallation()`?

ryangribble: also we need to support the `repositoryId` version of this call in addition to the `owner/repo` version

Octokit.Reactive/Clients/IObservableGitHubAppsClient.cs
```diff
@@ -53,5 +61,41 @@ public interface IObservableGitHubAppsClient
...
+
+        /// <summary>
+        /// Enables an authenticated GitHub App to find the repository's installation information.
```
ryangribble: comment says repository but should be organization

Octokit.Reactive/Clients/IObservableGitHubAppsClient.cs
```diff
@@ -53,5 +61,41 @@ public interface IObservableGitHubAppsClient
...
+        /// </summary>
+        /// <remarks>https://developer.github.com/v3/apps/#find-organization-installation</remarks>
+        /// <param name="owner">The owner of the repo</param>
```
ryangribble: incorrect params

Octokit.Reactive/Clients/IObservableGitHubAppsClient.cs
```diff
@@ -53,5 +61,41 @@ public interface IObservableGitHubAppsClient
...
+        /// <param name="repo">The name of the repo</param>
+        IObservable<Installation> GetOrganizationInstallation(string organization);
```
ryangribble: GetOrganizationInstallationForCurrent or FindOrganizationInstallation?

Octokit.Reactive/Clients/IObservableGitHubAppsClient.cs
```diff
@@ -53,5 +61,41 @@ public interface IObservableGitHubAppsClient
...
+
+        /// <summary>
+        /// Enables an authenticated GitHub App to find the users's installation information.
+        /// </summary>
+        /// <remarks>https://developer.github.com/v3/apps/#find-user-installation</remarks>
+        /// <param name="owner">The owner of the repo</param>
+        /// <param name="repo">The name of the repo</param>
+        IObservable<Installation> GetUserInstallation(string user);
```
ryangribble: GetUserInstallationForCurrent ?  Or FindUserInstallation?

Octokit.Reactive/Clients/IObservableInstallationsClient.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System;
+
+namespace Octokit.Reactive
+{
+    public interface IObservableGitHubAppsInstallationsClient
+    {
+        /// <summary>
+        /// List repositories of the authenticated GitHub App (requires GitHubApp JWT token auth).
```
ryangribble: The docs says this endpoint uses an installation token, not a JWT... and this would get repositories of an installation rather than a github app (which would have many installations each with potentially many repos)

So let's go with "List repositories of the authenticated installation (requires Installation Token auth)" or something similar

Octokit.Reactive/Clients/IObservableInstallationsClient.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// </summary>
+        /// <remarks>https://developer.github.com/v3/apps/installations/#list-repositories</remarks>
+        IObservable<RepositoriesResponse> GetAllRepositoriesForCurrent();
+
+        /// <summary>
+        /// List repositories of the authenticated GitHub App (requires GitHubApp JWT token auth).
```
ryangribble: same as above

Octokit.Reactive/Clients/IObservableInstallationsClient.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// </summary>
+        /// <param name="options">Options for changing the API response</param>
+        /// <remarks>https://developer.github.com/v3/apps/installations/#list-repositories</remarks>
+        IObservable<RepositoriesResponse> GetAllRepositoriesForCurrent(ApiOptions options);
+
+        /// <summary>
+        /// List repositories accessible to the user for an installation.
```
ryangribble: Comment probably needs to indicate this requires a user to server OAuth token

StanleyGoldman:
First off, sorry about my sloppiness, I was rushing so I can go and dogfood this. Thanks for reviewing.

1. I'm gonna start up a new pull request that targets this one, and I'll go through all of the GitHub Apps docs and make sure they are correct and link out correctly.
1. The `ForCurrent` moniker resonates better with me for some reason. I'll also wait for ☝️ to make sure they are consistent.
1. I'll make sure the docs I add have relevant links to the auth pages in the api.
1. Are there any other undocumented patters I should be looking for? 😼

StanleyGoldman:
I was gonna send the latest changes in a separate pull request, but then i got lazy.

StanleyGoldman:
Your changes so far get a 👍 from me.

ryangribble:
I think the last thing to do is unit tests for the observable client and ideally integration tests (at least for the things we can easily setup)

ryangribble:
OK I think im done with adding unit and integration tests for EVERYTHING.  Glad I did this too, as it found a few problems!  😀

ryangribble:
Hey @StanleyGoldman these latest changes I already had a separate PR for #1857 and I also did them in a way that keeps the old models working for GHE 2.14 since it won't get these changes

I think I did forget to add the start/end column members but we can just add that to the other PR and keep this one focused on the app/installation endpoints

StanleyGoldman:
Ah my bad, I was not aware of the nuance.

ryangribble:
release_notes: Implement additional endpoints for GitHub Apps to find installations for a given organization, repository or user

ryangribble:
release_notes: Implement GitHub Apps Installation API to allow listing all repositories a GitHub App Installation or GitHub App authenticated user has access to

ryangribble:
I think these are good to go!

LGTM

PR #1857, opened 2018-08-23
Implement changes to Checks API for Annotations models and re-request endpoint

### LINKED ISSUE #1856: API Changes Blog - Changes to the Checks REST API
As per [this blog post](https://developer.github.com/changes/2018-08-16-changes-to-the-checks-rest-api/) there are upcoming changes to Checks API

- Annotations has some field changes
- Deprecated "Request Check Run" endpoint
- New "Re-request Check Run" endpoint

The tricky thing is that GitHub Enterprise 2.14 will still have "the old" models and endpoints, whilst github.com will move to the new.  Ideally we can figure out a way to support both

ryangribble:
Fixes #1856 

Attempt to handle both old and new annotations models so we can support github.com and GHE 2.14

- [x] add `Path` and `AnnotationLevel` fields to `CheckRunAnnotation` response and `NewCheckRunAnnotation` request
- [x] flag `Filename` and `WarningLevel` as deprecated/obsolete but keep them around so we can use them for GHE 2.14
- [x] also flag `BlobHref` as deprecated on `NewCheckRunAnnotation`
- [x] Adjust ctors so we have a way to create the "old" and "new" model formats, ensuring the relevant fields are `null` so they don't get sent
- [x] Add new re-request endpoint
- [x] Flag old request endpoint for deprecation

### DIFF
--- a/Octokit.Reactive/Clients/IObservableCheckSuitesClient.cs
+++ b/Octokit.Reactive/Clients/IObservableCheckSuitesClient.cs
@@ -151,6 +151,7 @@ namespace Octokit.Reactive
         /// <param name="owner">The owner of the repository</param>
         /// <param name="name">The name of the repository</param>
         /// <param name="request">Details of the Check Suite request</param>
+        [Obsolete("This method has been deprecated in the GitHub Api, however can still be used on GitHub Enterprise 2.14")]
         IObservable<bool> Request(string owner, string name, CheckSuiteTriggerRequest request);
 
         /// <summary>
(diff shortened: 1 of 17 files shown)


ryangribble:
advisories: The changes made to `CheckRunAnnotation` response and `NewCheckRunAnnotation` request models in [PR #1857](https://github.com/octokit/octokit.net/pull/1857) will only apply to github.com (not GitHub Enterprise 2.14).  Backwards compatibility in this octokit.net version has been achieved by keeping the old request/response fields present (marked as deprecated - to be removed in the future when GHE2.14 is no longer supported).

ryangribble:
release_notes: Add new method `CheckSuitesClient.Rerequest()` and mark the old `CheckSuitesClient.Request()` method as deprecated (this will no longer function on github.com but will continue to be supported on GitHub Enterprise 2.14).

PR #1859, opened 2018-08-27
Implement renaming and deprecation of GitHub App Installation Access Token route

### LINKED ISSUE #1858: API Changes Blog - Renaming and deprecation of GitHub App Installation Access Token route
As per [this blog post](https://developer.github.com/changes/2018-08-16-renaming-and-deprecation-of-github-app-installation-access-token-route/) there are upcoming changes to GitHub App Installation Access Token route

Route is changing from
`/installations/:installation_id/access_tokens`
to
`/app/installations/:installation_id/access_tokens`

The tricky thing is that GitHub Enterprise 2.14 will still have "the old" route, whilst github.com will move to the new.  Ideally we can figure out a way to support both

ryangribble:
Fixes #1858

- Update `CreateInstallationToken()` method to handle the new `/app/installations/...` route
- Add an `[Obsolete]` flagged `CreateInstallationTokenPreview()` method still using the old `/installations/...` route, for GitHub Enterprise 2.14 compatibility

### DIFF
--- a/Octokit.Reactive/Clients/IObservableGitHubAppsClient.cs
+++ b/Octokit.Reactive/Clients/IObservableGitHubAppsClient.cs
@@ -53,5 +53,17 @@ namespace Octokit.Reactive
         /// </remarks>
         /// <param name="installationId">The Id of the GitHub App Installation</param>
         IObservable<AccessToken> CreateInstallationToken(long installationId);
+
+        /// <summary>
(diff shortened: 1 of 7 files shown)


ryangribble:
After some internal advice, it turns out the `/app/installations/...` route has been supported since June 2017 so even for GHE we don't need to keep the old one around.  I'll remove the `CreateInstallationTokenPreview()` method that was added in this PR

ryangribble:
Closing this, replaced with #1860

PR #1860, opened 2018-08-29
Implement renaming and deprecation of GitHub App Installation Access Token route 

### LINKED ISSUE #1858: API Changes Blog - Renaming and deprecation of GitHub App Installation Access Token route

ryangribble:
Replaces #1859 
Fixes #1858

- Update `CreateInstallationToken()` method to handle the new `/app/installations/...` route

This route has actually been supported since June 2017 so no special case is needed to maintain functionality for GitHub Apps Preview on GitHub Enterprise 2.14

### DIFF
--- a/Octokit.Reactive/Clients/IObservableGitHubAppsClient.cs
+++ b/Octokit.Reactive/Clients/IObservableGitHubAppsClient.cs
@@ -53,5 +53,17 @@ namespace Octokit.Reactive
         /// </remarks>
         /// <param name="installationId">The Id of the GitHub App Installation</param>
         IObservable<AccessToken> CreateInstallationToken(long installationId);
+
+        /// <summary>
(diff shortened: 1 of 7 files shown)


ryangribble:
release_notes: Adjust GitHub App Installation Access Token route in line with [announced API changes](https://developer.github.com/changes/2018-08-16-renaming-and-deprecation-of-github-app-installation-access-token-route/)

PR #1868, opened 2018-09-08
Fix `IIssueLabelsClient.RemoveFromIssue()`

### LINKED ISSUE #1865: IIssuesLabelClient.RemoveFromIssue started returning "Error parsing JSON"
NuGet package v0.31.0

I have code that uses IIssuesLabelClient.RemoveFromIssue(string owner, string name, int number, string labelName)  

Everything has been working fine until a couple days ago, when this method started throwing Bad Request 400: Error Parsing JSON API exceptions.  This happens on any issue when removing any existing label. Adding labels works fine.

ryangribble:
Fixes #1865 

`IIssueLabelsClient.RemoveFromIssue()` was incorrectly passing an accept header to the body parameter of the delete call.  Recent upstream api changes seem to have occured where this previously ignored situation is now being rejected as the body can't be parsed as valid json.  Also the preview functionality for label emojis/descriptions wouldnt have been working on this call due to the accept header not being passed correctly.

This PR fixes the issue by adding an overload for `ApiConnection.Delete<T>(Uri uri, string accepts)` that takes a Uri and accept header, but no body, which the existing code in `RemoveFromIssue()` will now use instead of `ApiConnection.Delete<T>(Uri uri, object data)`

### DIFF
--- a/Octokit.Tests/Reactive/ObservableIssuesLabelsClientTests.cs
+++ b/Octokit.Tests/Reactive/ObservableIssuesLabelsClientTests.cs
@@ -367,7 +367,7 @@ namespace Octokit.Tests.Reactive
 
                 client.RemoveFromIssue("fake", "repo", 42, "label");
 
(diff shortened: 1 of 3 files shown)


ryangribble:
release_notes: `IssueLabelsClient.RemoveFromIssue()` no longer fails with a HTTP 400 "Bad Request" error from the GitHub Api

PR #1877, opened 2018-09-27
Add Author Association for comments related models

mirsaeedi:
It seems that GitHub has recently added a new field called _[author_association](https://developer.github.com/v4/enum/commentauthorassociation/)_ to comment related response models such as[ Issue Comments](https://api.github.com/repos/dotnet/coreclr/issues/17920/comments), [Pull Request Review Comments](https://api.github.com/repos/dotnet/corefx/pulls/154/comments), and[ Pull Request Reviewers](https://api.github.com/repos/dotnet/corefx/pulls/11694/reviews
). This field shows how an author is related to a repo.

The proposed change adds a new enum named AuthorAssociate, changes constructors in affected models, and lastly adds a new attribute called AuthorAssociate to related models.

### DIFF
--- a/Octokit/Models/Common/AuthorAssociation.cs
+++ b/Octokit/Models/Common/AuthorAssociation.cs
@@ -1,0 +1,54 @@
+﻿using System;
+using Octokit.Internal;
+
+namespace Octokit
+{
+    /// <summary>
+    /// States of a Team/Organization Membership
+    /// </summary>
+    public enum AuthorAssociation
+    {
+
+        /// <summary>
+        /// Author has been invited to collaborate on the repository.
+        /// </summary>
+        [Parameter(Value = "COLLABORATOR")]
+        Collaborator,
+
+        /// <summary>
+        /// Author has previously committed to the repository.
+        /// </summary>
+        [Parameter(Value = "CONTRIBUTOR")]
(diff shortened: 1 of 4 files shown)


ryangribble:
Hey @mirsaeedi sorry I didn't have a chance to look at this yet... The documentation you linked to is for graphQL API, but this field is also being returned in REST API responses?

In the constructors can you remove the `=null` parts as we don't use default/optional parameters anywhere else so I'd like to stay consistent with that

ryangribble:
Oh yes sorry it only looked at the first link and not the others!

I understand the attempt to make it non breaking but these are only response classes which nobody would really be creating anyway (except in mocked tests) and just from a consistency point of view we don't do this anywhere else so I'd rather not do it now

mirsaeedi:
it's done. @ryangribble

ryangribble:
is this field provided in every API response that returns comments?  If not, it should be nullable (so it can be `null` when the API doesnt provide it)

ryangribble:
releaes_notes: Added `AuthorAssociation` field to `IssueComment`, `PullRequestReview` and `PullRequestReviewComment` response models

ryangribble:
Thanks @mirsaeedi !

LGTM

Issue #1904, opened 2018-11-01
DateRange does not include time


patricknolan:
Hi,

I'm trying to search for issues and setting the Closed property to a DateRange. However, the value produced by DateRange seems to ignore the Time.

Based on the following link, Time does appear to be supported by the GitHub API.
https://help.github.com/articles/understanding-the-search-syntax/

ryangribble:
Hi @patricknolan 

Yes it does look like we ignore the time component at the moment.  

I'm not sure if this is because the time support was added to GitHub after octokit had implemented request or whether it was just an oversight.

https://github.com/octokit/octokit.net/blob/8407369485541b1fce28f1c9c8d5bf5852579c57/Octokit/Models/Request/SearchRepositoriesRequest.cs#L307-L324

https://github.com/octokit/octokit.net/blob/8407369485541b1fce28f1c9c8d5bf5852579c57/Octokit/Models/Request/SearchRepositoriesRequest.cs#L331

What we could do in order to not affect existing behaviour is keep the existing `ctor`s that accept `DateTime` and mark them `[Obsolete]`, having them continue to only format the query string as YYYYMMDD, then add new `ctor`s that take `DateTimeOffset` types (giving us timezone info as well) and being formatted out to the full ISO8601 timestamp format including time zone info.  

Thoughts?  

If you are keen to send a PR that would be awesome :+1:

patricknolan:
Hey @ryangribble your suggestion solves the issue and is probably the best approach for backwards compatibility. I'll send a PR now.

catmorgan:
This bug still says it is open, but I saw the merged PR. I just updated to Octokit 0.32.0 and I am still experiencing this, but I am using DateRange on the Merged field

catmorgan:
I also noticed that the PR seems to address only  the bug with SearchRepositoryRequest, but I am experience it with SearchIssuesRequest

shiftkey:
Closing this out as we merged #1905, but this hasn't been published to NuGet so we have #1972 open to track that.

PR #1905, opened 2018-11-05
DateRange does not include time #1904

patricknolan:
Hi @ryangribble please let me know what you think.

### DIFF
--- a/Octokit.Tests/Clients/SearchClientTests.cs
+++ b/Octokit.Tests/Clients/SearchClientTests.cs
@@ -219,7 +219,7 @@ namespace Octokit.Tests.Clients
                 client.SearchUsers(request);
                 connection.Received().Get<SearchUsersResult>(
                     Arg.Is<Uri>(u => u.ToString() == "search/users"),
-                    Arg.Is<Dictionary<string, string>>(d => d["q"] == "github+created:>2014-01-01"));
+                    Arg.Is<Dictionary<string, string>>(d => d["q"] == $"github+created:{request.Created}"));
             }
 
             [Fact]
@@ -232,7 +232,7 @@ namespace Octokit.Tests.Clients
                 client.SearchUsers(request);
                 connection.Received().Get<SearchUsersResult>(
                     Arg.Is<Uri>(u => u.ToString() == "search/users"),
-                    Arg.Is<Dictionary<string, string>>(d => d["q"] == "github+created:>=2014-01-01"));
+                    Arg.Is<Dictionary<string, string>>(d => d["q"] == $"github+created:{request.Created}"));
             }
 
             [Fact]
@@ -245,7 +245,7 @@ namespace Octokit.Tests.Clients
                 client.SearchUsers(request);
                 connection.Received().Get<SearchUsersResult>(
                     Arg.Is<Uri>(u => u.ToString() == "search/users"),
-                    Arg.Is<Dictionary<string, string>>(d => d["q"] == "github+created:<=2014-01-01"));
+                    Arg.Is<Dictionary<string, string>>(d => d["q"] == $"github+created:{request.Created}"));
             }
 
             [Fact]
@@ -258,7 +258,7 @@ namespace Octokit.Tests.Clients
                 client.SearchUsers(request);
                 connection.Received().Get<SearchUsersResult>(
                     Arg.Is<Uri>(u => u.ToString() == "search/users"),
-                    Arg.Is<Dictionary<string, string>>(d => d["q"] == "github+created:<2014-01-01"));
+                    Arg.Is<Dictionary<string, string>>(d => d["q"] == $"github+created:{request.Created}"));
             }
 
             [Fact]
@@ -271,7 +271,7 @@ namespace Octokit.Tests.Clients
                 client.SearchUsers(request);
                 connection.Received().Get<SearchUsersResult>(
                     Arg.Is<Uri>(u => u.ToString() == "search/users"),
-                    Arg.Is<Dictionary<string, string>>(d => d["q"] == "github+created:2014-01-01..2014-02-01"));
+                    Arg.Is<Dictionary<string, string>>(d => d["q"] == $"github+created:{request.Created}"));
             }
 
             [Fact]
@@ -528,7 +528,7 @@ namespace Octokit.Tests.Clients
                 request.Created = DateRange.GreaterThan(new DateTime(2011, 1, 1));
                 client.SearchRepo(request);
                 connection.Received().Get<SearchRepositoryResult>(Arg.Is<Uri>(u => u.ToString() == "search/repositories"),
-                    Arg.Is<Dictionary<string, string>>(d => d["q"] == "github+created:>2011-01-01"));
+                    Arg.Is<Dictionary<string, string>>(d => d["q"] == $"github+created:{request.Created}"));
             }
 
             [Fact]
@@ -541,7 +541,7 @@ namespace Octokit.Tests.Clients
                 request.Created = DateRange.GreaterThanOrEquals(new DateTime(2011, 1, 1));
                 client.SearchRepo(request);
                 connection.Received().Get<SearchRepositoryResult>(Arg.Is<Uri>(u => u.ToString() == "search/repositories"),
-                    Arg.Is<Dictionary<string, string>>(d => d["q"] == "github+created:>=2011-01-01"));
+                    Arg.Is<Dictionary<string, string>>(d => d["q"] == $"github+created:{request.Created}"));
             }
 
             [Fact]
@@ -554,7 +554,7 @@ namespace Octokit.Tests.Clients
                 request.Created = DateRange.LessThan(new DateTime(2011, 1, 1));
                 client.SearchRepo(request);
                 connection.Received().Get<SearchRepositoryResult>(Arg.Is<Uri>(u => u.ToString() == "search/repositories"),
-                    Arg.Is<Dictionary<string, string>>(d => d["q"] == "github+created:<2011-01-01"));
+                    Arg.Is<Dictionary<string, string>>(d => d["q"] == $"github+created:{request.Created}"));
             }
 
 
@@ -568,7 +568,7 @@ namespace Octokit.Tests.Clients
                 request.Created = DateRange.LessThanOrEquals(new DateTime(2011, 1, 1));
                 client.SearchRepo(request);
                 connection.Received().Get<SearchRepositoryResult>(Arg.Is<Uri>(u => u.ToString() == "search/repositories"),
-                    Arg.Is<Dictionary<string, string>>(d => d["q"] == "github+created:<=2011-01-01"));
+                    Arg.Is<Dictionary<string, string>>(d => d["q"] == $"github+created:{request.Created}"));
             }
 
             [Fact]
@@ -580,7 +580,7 @@ namespace Octokit.Tests.Clients
                 request.Created = DateRange.Between(new DateTime(2011, 1, 1), new DateTime(2012, 11, 11));
                 client.SearchRepo(request);
                 connection.Received().Get<SearchRepositoryResult>(Arg.Is<Uri>(u => u.ToString() == "search/repositories"),
-                    Arg.Is<Dictionary<string, string>>(d => d["q"] == "github+created:2011-01-01..2012-11-11"));
+                    Arg.Is<Dictionary<string, string>>(d => d["q"] == $"github+created:{request.Created}"));
             }
 
             [Fact]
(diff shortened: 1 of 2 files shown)


patricknolan:
Hi @ryangribble build failed but cause does'nt seem to be related to my changes.

shiftkey:
@patricknolan I restarted the problem build - let's see if it goes green

Octokit/Models/Request/SearchRepositoriesRequest.cs
```diff
@@ -304,33 +304,43 @@ public class DateRange
         /// Matches repositories with regards to the <param name="date"/>.
         /// We will use the <param name="op"/> to see what operator will be applied to the date qualifier
         /// </summary>
-        public DateRange(DateTime date, SearchQualifierOperator op)
+        [Obsolete("This ctor has been deprecated as GitHub API now supports date and time")]
+        public DateRange(DateTime date, SearchQualifierOperator op) : this(new DateTimeOffset(date), op)
```
khellang: By forwarding these ctors to the new `DateTimeOffset` overloads, the behavior has changed by taking the time into account. That could be a breaking change.

I think the safest thing to do is to keep the existing logic (as @ryangribble noted in #1904) for these ctors and use the new (and improved) format for the new `DateTimeOffset` ctors.

Octokit/Models/Request/SearchRepositoriesRequest.cs
```diff
@@ -304,33 +304,43 @@ public class DateRange
...
+        /// Matches repositories with regards to both the <param name="from"/> and <param name="to"/> dates.
+        /// </summary>
+        [Obsolete("This ctor has been deprecated as GitHub API now supports date and time")]
+        public DateRange(DateTime from, DateTime to) : this(new DateTimeOffset(from), new DateTimeOffset(to))
+        {
+        }
+
+        public DateRange(DateTimeOffset from, DateTimeOffset to)
+        {
+            query = string.Format(CultureInfo.InvariantCulture, "{0:yyyy-MM-dd'T'HH:mm:ss zzz}..{1:yyyy-MM-dd'T'HH:mm:ss zzz}", from.DateTime, to.DateTime);
```
khellang: What will the `zzz` component be here, given you're using the `DateTime` property and not the `DateTimeOffset` itself?

khellang: But at which offset? `DateTime` doesn't store a separate offset.

khellang: See https://docs.microsoft.com/en-us/dotnet/api/system.datetimeoffset.datetime?view=netframework-4.7.2#remarks

Octokit/Models/Request/SearchRepositoriesRequest.cs
```diff
@@ -304,33 +304,43 @@ public class DateRange
...
+        }
+
+        public DateRange(DateTimeOffset date, SearchQualifierOperator op)
         {
             switch (op)
             {
                 case SearchQualifierOperator.GreaterThan:
-                    query = string.Format(CultureInfo.InvariantCulture, ">{0:yyyy-MM-dd}", date);
+                    query = string.Format(CultureInfo.InvariantCulture, ">{0:yyyy-MM-dd'T'HH:mm:ss zzz}", date);
```
khellang: I'd probably store this pattern in a constant and reuse it throughout.

Octokit/Models/Request/SearchRepositoriesRequest.cs
patricknolan: Hi sorry you are right @khellang. I was trying to do to code this while also doing my job which was a mistake. I'll fix in my next PR.

Octokit/Models/Request/SearchRepositoriesRequest.cs
patricknolan: Hi @khellang I've included this suggestion aswell.

I've also updated the relevant unit tests to use a Theory which tests both the `DateTime` and `DateTimeOffset` strategies.

patricknolan:
Hi @ryangribble and @khellang when you get a chance let me know what you think!

Octokit/Models/Request/SearchRepositoriesRequest.cs
```diff
@@ -298,37 +298,66 @@ public override string ToString()
     [DebuggerDisplay("{DebuggerDisplay,nq}")]
     public class DateRange
     {
+        public const string DateTimePatten = "yyyy-MM-dd'T'HH:mm:ss zzz";
+        public const string DatePatten = "yyyy-MM-dd";
```
khellang: Nit: `Patten` -> `Pattern`

Octokit/Models/Request/SearchRepositoriesRequest.cs
```diff
@@ -298,37 +298,66 @@ public override string ToString()
...
+
         private readonly string query = string.Empty;
 
         /// <summary>
         /// Matches repositories with regards to the <param name="date"/>.
         /// We will use the <param name="op"/> to see what operator will be applied to the date qualifier
         /// </summary>
+        [Obsolete("This ctor has been deprecated as GitHub API now supports date and time")]
```
khellang: Is it worth pointing out the preferred API and why you'd use it? Something like

> This constructor doesn't use the time component of the specified `DateTime`. Please use the overload accepting a `DateTimeOffset`, which also supports time.

patricknolan:
Hi @khellang all done.

Octokit/Models/Request/SearchRepositoriesRequest.cs
```diff
@@ -298,37 +298,66 @@ public override string ToString()
...
         /// We will use the <param name="op"/> to see what operator will be applied to the date qualifier
         /// </summary>
+        [Obsolete("This constructor doesn't use the time component of the specified DateTime. Please use the overload accepting a DateTimeOffset, which also supports time.")]
         public DateRange(DateTime date, SearchQualifierOperator op)
         {
             switch (op)
             {
                 case SearchQualifierOperator.GreaterThan:
-                    query = string.Format(CultureInfo.InvariantCulture, ">{0:yyyy-MM-dd}", date);
+                    query = string.Format(CultureInfo.InvariantCulture, ">{0:DatePatten}", date);
```
khellang: Does this compile? 🤔

Octokit/Models/Request/SearchRepositoriesRequest.cs
```diff
@@ -342,6 +371,7 @@ internal string DebuggerDisplay
         /// </summary>
         /// <param name="date">date to be used for comparison (times are ignored)</param>
         /// <returns><see cref="DateRange"/></returns>
+        [Obsolete("This ctor has been deprecated as GitHub API now supports date and time")]
```
khellang: These are not ctors. Maybe also change these to reflect the recommended APIs?

Octokit/Models/Request/SearchRepositoriesRequest.cs
khellang: I'm thinking it might be better to just use `ToString(DatePattern, CultureInfo.InvariantCulture)` and string concatenation or interpolation than the current format strings...

patricknolan: Wow! I got All Checks Passed "continuous-integration/travis-ci/pr — The Travis CI build passed"

patricknolan: Hi @khellang you're thinking something like this? 

`query = $"{from.ToString(DateTimePattern, CultureInfo.InvariantCulture)}..{to.ToString(DateTimePattern, CultureInfo.InvariantCulture)}";`

Octokit/Models/Request/SearchRepositoriesRequest.cs
```diff
@@ -298,37 +298,66 @@ public override string ToString()
...
         }
 
         /// <summary>
         /// Matches repositories with regards to both the <param name="from"/> and <param name="to"/> dates.
         /// </summary>
-        public DateRange(DateTime from, DateTime to)
+        [Obsolete("This constructor doesn't use the time component of the specified DateTime. Please use the overload accepting a DateTimeOffset, which also supports time.")]
+        public DateRange(DateTime from, DateTime to) 
+        {
+            query = string.Format(CultureInfo.InvariantCulture, "{0:DateTimePattern}..{1:DateTimePattern}", from, to);
```
ryangribble: Same here, it needs to use Date only, not Date+Time

Octokit/Models/Request/SearchRepositoriesRequest.cs
```diff
@@ -298,37 +298,66 @@ public override string ToString()
...
+        }
+
+        public DateRange(DateTimeOffset from, DateTimeOffset to)
         {
-            query = string.Format(CultureInfo.InvariantCulture, "{0:yyyy-MM-dd}..{1:yyyy-MM-dd}", from, to);
+            query = string.Format(CultureInfo.InvariantCulture, "{0:DateTimePatten}..{1:DateTimePatten}", from, to);
```
ryangribble: This still has the old spelling.  Im not sure you even can parameterize the format string like this anyway?!

Octokit.Tests/Clients/SearchClientTests.cs
```diff
@@ -209,69 +215,89 @@ public void TestingTheLanguageQualifier()
...
-                request.Created = DateRange.GreaterThan(new DateTime(2014, 1, 1));
-                client.SearchUsers(request);
+                request.Created = strategy == DateTimeStrategies.DateTimeStrategy 
+                    ? DateRange.GreaterThan(new DateTime(2014, 1, 1)) 
+                    : DateRange.GreaterThan(new DateTimeOffset(2014, 1, 1, 2, 4, 6, new TimeSpan(10, 0, 0)));
+                client.SearchUsers(request);        
                 connection.Received().Get<SearchUsersResult>(
                     Arg.Is<Uri>(u => u.ToString() == "search/users"),
-                    Arg.Is<Dictionary<string, string>>(d => d["q"] == "github+created:>2014-01-01"));
+                    Arg.Is<Dictionary<string, string>>(d => d["q"] == $"github+created:{request.Created}"));
```
ryangribble: I'd prefer the unit test actually tests the formatted string eg '2014-1-1'T'020406 1000' (or whatever it should be)

Otherwise these tests are susceptible to whatever formatting bug may exist inside `DateRange` class.  

EG Im pretty sure the way you've implemented it at the moment isn't correctly formatting the string, yet all these tests pass (because you are using the `DateRange` class to test the `DateRange` class! 🤣

Octokit/Models/Request/SearchRepositoriesRequest.cs
khellang: I'm thinking something like `query = FormattableString.Invariant($"{from:DateTimePattern}..{to:DateTimePattern}");` 😄

Octokit.Tests/Clients/SearchClientTests.cs
patricknolan: Hi @ryangribble 
Has taken me a while to make these changes due to other commitments, however, they're now there.

Let me know what you think.

patricknolan:
Hi @ryangribble and @khellang,,

Has taken me a while to make changes due to other commitments, however, I have now made the changes advised.

Let me know what you think.

ryangribble:
:wave: Hi @patricknolan 

Just trying to get back up to speed with this one!  The changes look pretty good now. 

Just on the `[Obsolete]` messages, I can see previously I suggested that this wording might be more clear... what do you think?

> In terms of the [Obsolete] messages, can we use these please?
>
> "This ctor does not support the time component or timezone and will be removed in a future release. Please use the DateTimeOffset overload instead"
>
> "This method does not support the time component or timezone and will be removed in a future release. Please use the DateTimeOffset overload instead"

patricknolan:
Hi @ryangribble changes for [Obsolete] messages have been done. The CI failed, however, it may be an intermittent error as I don't believe the changes could have caused this to fail. I'm not sure how to invoke the CI again without an additional check-in?

ryangribble:
As a user you can't rerun tyem, I've kicked them off again

patricknolan:
Thanks @ryangribble. You ok for this to be merged now?

ryangribble:
Yep I think we're good now!  Thanks @patricknolan :+1: 🚢 


LGTM

ryangribble:
release_notes: Enhanced `DateRange` class used in `SearchClient`  requests to support time and timezone components

patricknolan:
Hi @ryangribble just confirming if you do the release notes or do I? Sorry this is my first contribution no am not sure.

Am keen to get a new release available via Nuget.

ryangribble:
Hi @patricknolan sorry for the confusion, this is something I do... I'm really sorry I've just been tied up in lots of real life stuff and not cut a new release as yet.  If you add a custom nuget source of our appveyor feed `https://ci.appveyor.com/nuget/octokit-net`, you can get the latest "beta" packages from our master branch, but yes I really do need to make a release and push to nuget proper 🙏

patricknolan:
Not a problem @ryangribble completely understand

PR #1921, opened 2018-12-14
Update to Cake.Frosting 0.31.0

devlead:
* Uses Cake.Core & Cake.Common 0.31.0 ( read more about improvements at https://cakebuild.net/blog/2018/12/cake-v0.31.0-released )
* 0.31.0 is on NuGet.org so nuget.config no longer needed
* GetDirectories with predicate  is now obsolete, replaced by GlobberSettings

### DIFF
--- a/build/Build.csproj
+++ b/build/Build.csproj
@@ -6,7 +6,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Cake.Frosting" Version="0.30.0" />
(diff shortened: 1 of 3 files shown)


devlead:
Don't think the SourceLink eror is Cake related could be that I'm hitting ctaggart/SourceLink#322

Travis Mac & Linux seems happy.

LTS .NET SDK is 2.1.500, might be worth updating.

ryangribble:
yep, re-triggered appveyor and it's green now

Thanks @devlead

PR #1924, opened 2018-12-24
update docs to include the await keyword

### LINKED ISSUE #1923: Tutorial lists line of code that doesn't work
I was following along with the tutorial on:
https://octokitnet.readthedocs.io/en/latest/releases/

And it tells me to use:
var latest = releases[0]

To grab the most recent release, however, Visual Studio does not agree with this:
image

Please, advise?

shiftkey:
Fixes #1923

### DIFF
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -5,7 +5,7 @@
 To retrieve all releases for a repository:
 
 ```csharp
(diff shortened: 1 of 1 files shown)


ryangribble:
release_notes: Update `releases` sample code to await the async call

PR #1930, opened 2019-01-21
Enable building on linux without project file changes

### LINKED ISSUE #1745: Build on Linux
I'm trying to build this project on Ubuntu 16.04.
I installed Visual Studio Code with all relevant extensions, .Net Core and PowerShell.
I try building it, and this is what I get:

```powershell
> Executing task: dotnet build /home/itaibh/Projects/octokit.net/Octokit.Tests/Octokit.Tests.csproj <

Microsoft (R) Build Engine version 15.5.180.51428 for .NET Core
Copyright (C) Microsoft Corporation. All rights reserved.

  Restore completed in 16.53 ms for /home/itaibh/Projects/octokit.net/Octokit/Octokit.csproj.
  Restore completed in 16.53 ms for /home/itaibh/Projects/octokit.net/Octokit.Reactive/Octokit.Reactive.csproj.
/home/itaibh/Projects/octokit.net/Octokit.Tests/Octokit.Tests.csproj : warning NU1603: Castle.Core 4.0.0 depends on System.ComponentModel.TypeConverter (>= 4.0.1) but System.ComponentModel.TypeConverter 4.0.1 was not found. An approximate best match of System.ComponentModel.TypeConverter 4.1.0 was resolved.
  Restore completed in 32 ms for /home/itaibh/Projects/octokit.net/Octokit.Tests/Octokit.Tests.csproj.
  Restore completed in 16.8 ms for /home/itaibh/Projects/octokit.net/Octokit/Octokit.csproj.
  Restore completed in 22.02 ms for /home/itaibh/Projects/octokit.net/Octokit/Octokit.csproj.
/home/itaibh/Projects/octokit.net/Octokit.Tests/Octokit.Tests.csproj : warning NU1603: Castle.Core 4.0.0 depends on System.ComponentModel.TypeConverter (>= 4.0.1) but System.ComponentModel.TypeConverter 4.0.1 was not found. An approximate best match of System.ComponentModel.TypeConverter 4.1.0 was resolved.
/usr/share/dotnet/sdk/2.1.4/Microsoft.Common.CurrentVersion.targets(1124,5): error MSB3644: The reference assemblies for framework ".NETFramework,Version=v4.5.2" were not found. To resolve this, install the SDK or Targeting Pack for this framework version or retarget your application to a version of the framework for which you have the SDK or Targeting Pack installed. Note that assemblies will be resolved from the Global Assembly Cache (GAC) and will be used in place of reference assemblies. Therefore your assembly may not be correctly targeted for the framework you intend. [/home/itaibh/Projects/octokit.net/Octokit.Tests/Octokit.Tests.csproj]
  Octokit -> /home/itaibh/Projects/octokit.net/Octokit/bin/Debug/netstandard1.1/Octokit.dll
  Octokit.Reactive -> /home/itaibh/Projects/octokit.net/Octokit.Reactive/bin/Debug/netstandard1.1/Octokit.Reactive.dll
  Octokit.Tests -> /home/itaibh/Projects/octokit.net/Octokit.Tests/bin/Debug/netcoreapp1.0/Octokit.Tests.dll

Build FAILED.

/home/itaibh/Projects/octokit.net/Octokit.Tests/Octokit.Tests.csproj : warning NU1603: Castle.Core 4.0.0 depends on System.ComponentModel.TypeConverter (>= 4.0.1) but System.ComponentModel.TypeConverter 4.0.1 was not found. An approximate best match of System.ComponentModel.TypeConverter 4.1.0 was resolved.
/home/itaibh/Projects/octokit.net/Octokit.Tests/Octokit.Tests.csproj : warning NU1603: Castle.Core 4.0.0 depends on System.ComponentModel.TypeConverter (>= 4.0.1) but System.ComponentModel.TypeConverter 4.0.1 was not found. An approximate best match of System.ComponentModel.TypeConverter 4.1.0 was resolved.
/usr/share/dotnet/sdk/2.1.4/Microsoft.Common.CurrentVersion.targets(1124,5): error MSB3644: The reference assemblies for framework ".NETFramework,Version=v4.5.2" were not found. To resolve this, install the SDK or Targeting Pack for this framework version or retarget your application to a version of the framework for which you have the SDK or Targeting Pack installed. Note that assemblies will be resolved from the Global Assembly Cache (GAC) and will be used in place of reference assemblies. Therefore your assembly may not be correctly targeted for the framework you intend. [/home/itaibh/Projects/octokit.net/Octokit.Tests/Octokit.Tests.csproj]
    2 Warning(s)
    1 Error(s)

Time Elapsed 00:00:06.47
The terminal process terminated with exit code: 1

Terminal will be reused by tasks, press any key to close it.
```

I tried debugging it, but I'm new to this .Net Core thing, especially on Linux. Is this a problem with my configuration or is this a bug that need to be resolved?

Vogel612:
This enables building via cake on linux without requiring adjustments on
other platforms, fixes #1745

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,4 +19,5 @@ before_script:
 
 script:
   - git fetch --unshallow --tags
-  - ./build.sh --linksources=true --verbosity=verbose
+  # disengage core only switch because mono ships .NETFramework targets
+  - ./build.sh --coreonly=false --linksources=true --verbosity=verbose
--- a/Octokit.Reactive/Octokit.Reactive.csproj
+++ b/Octokit.Reactive/Octokit.Reactive.csproj
@@ -7,7 +7,7 @@
     <Version>0.0.0-dev</Version>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <TargetFrameworks>netstandard1.1</TargetFrameworks>
-    <TargetFrameworks Condition="$(NoFramework) != '' and !$(NoFramework)">$(TargetFrameworks);net45</TargetFrameworks>
+    <TargetFrameworks Condition="$(CoreOnly) != '' and !$(CoreOnly)">$(TargetFrameworks);net45</TargetFrameworks>
     <NetStandardImplicitPackageVersion>1.6.0</NetStandardImplicitPackageVersion>
     <AssemblyName>Octokit.Reactive</AssemblyName>
     <PackageId>Octokit.Reactive</PackageId>
--- a/Octokit.Reactive/Octokit.Reactive.csproj
+++ b/Octokit.Reactive/Octokit.Reactive.csproj
@@ -6,6 +6,7 @@
     <Authors>GitHub</Authors>
     <Version>0.0.0-dev</Version>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <CoreOnly Condition="$(CoreOnly) == '' and $([MSBuild]::IsOsPlatform('Windows'))">False</CoreOnly>
     <TargetFrameworks>netstandard1.1</TargetFrameworks>
     <TargetFrameworks Condition="$(CoreOnly) != '' and !$(CoreOnly)">$(TargetFrameworks);net45</TargetFrameworks>
     <NetStandardImplicitPackageVersion>1.6.0</NetStandardImplicitPackageVersion>
--- a/Octokit.Reactive/Octokit.Reactive.csproj
+++ b/Octokit.Reactive/Octokit.Reactive.csproj
@@ -6,7 +6,8 @@
     <Authors>GitHub</Authors>
     <Version>0.0.0-dev</Version>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
-    <TargetFrameworks>netstandard1.1;net45</TargetFrameworks>
+    <TargetFrameworks>netstandard1.1</TargetFrameworks>
+    <TargetFrameworks Condition="$(NoFramework) != '' and !$(NoFramework)">$(TargetFrameworks);net45</TargetFrameworks>
     <NetStandardImplicitPackageVersion>1.6.0</NetStandardImplicitPackageVersion>
     <AssemblyName>Octokit.Reactive</AssemblyName>
     <PackageId>Octokit.Reactive</PackageId>
--- a/Octokit.Tests.Conventions/Octokit.Tests.Conventions.csproj
+++ b/Octokit.Tests.Conventions/Octokit.Tests.Conventions.csproj
@@ -4,6 +4,7 @@
     <Description>Convention-based tests for Octokit</Description>
     <AssemblyTitle>Octokit.Tests.Conventions</AssemblyTitle>
     <Authors>GitHub</Authors>
+    <CoreOnly Condition="$(CoreOnly) == '' and $([MSBuild]::IsOsPlatform('Windows'))">False</CoreOnly>
     <TargetFrameworks>netcoreapp2.0</TargetFrameworks>
     <TargetFrameworks Condition="$(CoreOnly) != '' and !$(CoreOnly)">$(TargetFrameworks);net452</TargetFrameworks>
     <NoWarn>$(NoWarn);CS4014;CS1998</NoWarn>
--- a/Octokit.Tests.Conventions/Octokit.Tests.Conventions.csproj
+++ b/Octokit.Tests.Conventions/Octokit.Tests.Conventions.csproj
@@ -5,7 +5,7 @@
     <AssemblyTitle>Octokit.Tests.Conventions</AssemblyTitle>
     <Authors>GitHub</Authors>
     <TargetFrameworks>netcoreapp2.0</TargetFrameworks>
-    <TargetFrameworks Condition="$(NoFramework) != '' and !$(NoFramework)">$(TargetFrameworks);net452</TargetFrameworks>
+    <TargetFrameworks Condition="$(CoreOnly) != '' and !$(CoreOnly)">$(TargetFrameworks);net452</TargetFrameworks>
     <NoWarn>$(NoWarn);CS4014;CS1998</NoWarn>
     <AssemblyName>Octokit.Tests.Conventions</AssemblyName>
     <PackageId>Octokit.Tests.Conventions</PackageId>
--- a/Octokit.Tests.Conventions/Octokit.Tests.Conventions.csproj
+++ b/Octokit.Tests.Conventions/Octokit.Tests.Conventions.csproj
@@ -4,7 +4,8 @@
     <Description>Convention-based tests for Octokit</Description>
     <AssemblyTitle>Octokit.Tests.Conventions</AssemblyTitle>
     <Authors>GitHub</Authors>
-    <TargetFrameworks>netcoreapp2.0;net452</TargetFrameworks>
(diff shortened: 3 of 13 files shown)


ryangribble:
Hey @Vogel612, this is cool for building locally on a mac/Linux but the only thing is that on travisCI Mac and Linux builds we actually DO build the full framework as the build agent is setup with mono 

So maybe rather than hardcoding noframework=true in build.sh this should be passed in so we can set false in travis? 

Also I think the tool install step of the build might still be a problem on Mac/Linux if you don't have mono as it uses nuget (from mono) to pull down gitversion doesn't it ?

Definitely keen for us to have the noframework option but

Vogel612:
Yes, the tool install step requires nuget. GitVersion is then executed through mono (cf. tools/gitversion_wrapper.sh). I'm happy to remove the hardcoding there.

Currently I'm building locally by sidestepping the GitVersion invocation and returning a hardcoded JSON because neither mono nor dotnet invoke the downloaded binary without error. (mono segfaults and dotnet is missing a library).
I did install `nuget`, which depends on mono to fix the step in the first place. 

IIUC the tools are required for determining the version (which could possibly be implemented differently in the build project's `GitVersionRunner`?) and to format the code.  
If you want, I can try messing around with both these issues.  
Do give me a heads up what type of solution you had in mind :)

Vogel612:
Sidenote: I'm currently running into issues trying to replace `GitVersion.CommandLine (3.6.2)` as Tool-Dependency with `GitVersionCommon (4.0.0)` as NuGet package. The NativeBinaries from LibGit2Sharp are currently binding to curl & openssl with a version [apparently deprecated for security reasons](https://bugs.archlinux.org/task/28252).

LibGit2Sharp is looking at [replacing that native binding](https://github.com/libgit2/libgit2sharp.nativebinaries/issues/59#issuecomment-429467419), but unfortunately that has not happened yet. As such I am unable to replace the GitVersion related code right now.

ryangribble:
I'm thinking if we had "nofullframework" mode for local use only, it could skip running gitversion as well as not build the net45 targets you've already implemented

Vogel612:
@ryangribble were you thinking something along these lines? Happy to make some more adjustments :+1:

ryangribble:
Yep that's what i was thinking - love the warning you added by the way :+1:

A couple of other thoughts (apologies im on mobile at the moment so I dont have links to affected source code lines etc)
- We should skip the "ToolInstaller" of Gitversion and CodeFormatter tools when running in this mode (since it will fail if mono/nuget etc arent present)
- The code formatter also wouldn't work on linux/mac without mono (it may even not work WITH mono, i havent tried!) so perhaps that build task should log a warning if it is ever executed in NoFramework mode rather than trying to do stuff and blowing up (assuming it does blow up!)
- you could change default behaviour of build.sh to run in NoFramework mode by default (so things work for linux/mac local users "out of the box".  But then change travis.yml so that it still builds the full framework stuff on tarvis
- I wonder if it would be clearer to name the mode "CoreOnly" rather than "NoFramework" and invert the logic (`!$NoFramework` in the csproj conditions is a bit of a confusing double negative)... thoughts?

Vogel612:
> We should skip the "ToolInstaller" of Gitversion and CodeFormatter tools when running in this mode (since it will fail if mono/nuget etc arent present)

Done

> The code formatter also wouldn't work on linux/mac without mono (it may even not work WITH mono, i havent tried!) so perhaps that build task should log a warning if it is ever executed in NoFramework mode rather than trying to do stuff and blowing up (assuming it does blow up!)

The code formatter only runs on windows in the first place, but since we're not downloading the tools for CoreOnly builds, I added an explicit check to the `ShouldRun` implementation of the task.

> you could change default behaviour of build.sh to run in NoFramework mode by default (so things work for linux/mac local users "out of the box". But then change travis.yml so that it still builds the full framework stuff on tarvis

travis invokes the buildscript. Changing the buildscript's default would involve creating an explicit switch to turn off the default. My bash is somewhere between bad and gruesome, so I don't want to mess around there

> I wonder if it would be clearer to name the mode "CoreOnly" rather than "NoFramework" and invert the logic (!$NoFramework in the csproj conditions is a bit of a confusing double negative)... thoughts?

Agreed and done

ryangribble:
> . My bash is somewhere between bad and gruesome, so I don't want to mess around there

I year ya 🤣  

how about this though - the cake script could default CoreOnly to be true for Windows and false otherwise. Travis can pass in false which should override the default value.  The bash script should already pass through any parameter provided to it (you should see some already being passed in the travis yml)

ryangribble:
The linksources parameter is a good example to follow. Travis passes it in, nothing special for it in the build.sh (it passes all parameters through to cake) and the cake build uses the value provided or uses a default if none specified (here https://github.com/octokit/octokit.net/blob/master/build/Lifetime.cs#L14) 

We can follow similar approach except default for CoreOnly will be conditional on windows or not 

My aim here is for a contributor locally on Linux or OSX just runs build.sh and everything just works (without losing anything on the Windows or travis/appveyor side)

Vogel612:
@ryangribble could you kick travis for me? It seems there has been a timeout in the LinkSources check

ryangribble:
There's a section in the build that explicitly logs the value of the parameters, could we add the CoreOnly one to that?

This is looking good! I'll do a test on my MacBook today 😁

ryangribble:
Great stuff!  Tested on my macbook tonight, works perfectly :+1:

Is this ready for merging, from your perspective @Vogel612

Vogel612:
Anytime :)

Vogel612:
@ryangribble poke: what's the latest on this?

Vogel612:
@ryangribble poke: Do tell me when I start to seriously annoy you with this, otherwise I'll keep more or less regularly poking you here :wink:

ryangribble:
It's all good... I'm doing some serious juggling at the moment so poking is fine/necessary 😀 

So I was just doing some final testing and found an issue where on Windows in both VS2017 and `dotnet` cmdline, it now is only building the `netstandard1.1` targets instead of `netstandard1.1` and `net45`.  Makes sense I guess since only Cake currently has the logic to set this flag appropriately

If I were to propose the ideal result of this PR, it would be:

✅ Cake builds on CI (appveyor and travis, all OS'es), build both platforms
✅ Cake builds on linux/mac build only netstandard1.1
✅ Cake builds on windows build both platforms
✅ local "native" builds on linux/mac build only netstandard1.1
❌  local "native" builds on windows build both platforms

So far we have 4 out of 5.  Maybe we need to actually set a default value for `CoreOnly` in MSBuild land, by [detecting if on Windows](https://github.com/Microsoft/msbuild/issues/2468) or not?

Vogel612:
rough quick-patch: Add `<CoreOnly Condition="$(OS) == 'Windows NT'">False</CoreOnly>` to all property groups involved? Can you verify for me that this fixes the native windows build?

ryangribble:
yeah, sounds like the `IsOsPlatform()` is the recommended option.  I tested on windows adding your 2nd option to one of the csproj files and it appears to work :+1:

ryangribble:
Tested on windows and mac, plus reviewed the travis and appveyor builds and everything is looking great!  5 out of 5!

woohoo!  

LGTM


Thanks for working on this @Vogel612

ryangribble:
release_notes: Adjust Cake and native build configurations to allow building on OSX/Linux out of the box

PR #1931, opened 2019-01-21
Ensure test projects with with .NET 4.6

jcansdale:
This fork has been updated to compiled against .NET 4.6 and use an alternative public key. Unfortunately the test projects no longer compile and run.

### What this PR does

- Fix the `InternalsVisibleTo` attribute to use new public key
- Update the test projects to compile against .NET 4.6


jcansdale:
Sorry, I didn't mean to target upstream. 😊

Closing...

Issue #1646, opened 2017-08-03
Implement EventPayload models for the event types supported by GitHub API


ryangribble:
GitHub API has currently [37 event payloads](https://developer.github.com/v3/activity/events/types/) that the Activity API can return, and only 8 of these are implemented in Octokit.net!

To support these as first class objects, Octokit needs a response model class added [here](https://github.com/octokit/octokit.net/blob/master/Octokit/Models/Response/ActivityPayloads) for each one, and this [magic switch statement](https://github.com/octokit/octokit.net/blob/34d13743e5bdf70b7e2ab6b3e44531050f63ee26/Octokit/Http/SimpleJsonSerializer.cs#L159-L178) updated to suit.

We should also add a new [unit test](https://github.com/octokit/octokit.net/blob/master/Octokit.Tests/Clients/EventsClientTests.cs#L567-L833) and [integration test](https://github.com/octokit/octokit.net/blob/master/Octokit.Tests.Integration/Clients/EventsClientTests.cs) for each payload (although these may fail if they are uncommon events so maybe we can change to unit tests using hardcoded json content from the github api docs).

Related to #1635 (DeleteEvent not supported) and #1645 (proposed workaround to allow unsupported event payloads to be provided as a Json string)

List of [current event types](https://developer.github.com/v3/activity/events/types)
- [x] CommitCommentEvent
- [x] CreateEvent
- [x] DeleteEvent
- [ ] DeploymentEvent
- [ ] DeploymentStatusEvent
- [ ] DownloadEvent
- [ ] FollowEvent
- [x] ForkEvent
- [ ] ForkApplyEvent
- [ ] GistEvent
- [ ] GollumEvent
- [ ] InstallationEvent
- [ ] InstallationRepositoriesEvent
- [x] IssueCommentEvent
- [x] IssuesEvent
- [ ] LabelEvent
- [ ] MarketplacePurchaseEvent
- [ ] MemberEvent
- [ ] MembershipEvent
- [ ] MilestoneEvent
- [ ] OrganizationEvent
- [ ] OrgBlockEvent
- [ ] PageBuildEvent
- [ ] ProjectCardEvent
- [ ] ProjectColumnEvent
- [ ] ProjectEvent
- [ ] PublicEvent
- [x] PullRequestEvent
- [ ] PullRequestReviewEvent
- [x] PullRequestReviewCommentEvent
- [x] PushEvent
- [ ] ReleaseEvent
- [ ] RepositoryEvent
[...]

PR #1932, opened 2019-01-22
Add support for CreateEventPayload and DeleteEvent payload (#1646)

GMouron:
Hello, I tried to do the same as was done in #1732 
Please tell me if anything is missing.
Cheers

### DIFF
--- a/Octokit/Models/Response/RefType.cs
+++ b/Octokit/Models/Response/RefType.cs
@@ -1,0 +1,31 @@
+﻿using System;
+using System.Diagnostics;
+using System.Globalization;
+using Octokit.Internal;
+
+namespace Octokit
+{
+    /// <summary>
+    /// Represents the type of object created or deleted
+    /// </summary>
+    public enum RefType
+    {
+        /// <summary>
+        /// The object is of type repository
+        /// </summary>
+        [Parameter(Value = "repository")]
+        Repository,
+
+        /// <summary>
+        /// The object is of type branch
+        /// </summary>
+        [Parameter(Value = "branch")]
+        Branch,
+
+        /// <summary>
+        /// The object is of type tag
+        /// </summary>
+        [Parameter(Value = "tag")]
+        Tag
+    }
+}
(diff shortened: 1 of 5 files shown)


ryangribble:
Hi @GMouron, thanks for this!  It looks good although one thing I noticed when checking the docs is that the create event payload has  `master_branch` and `description` fields as well, so it would seem a good idea to add them to the response model in this PR.  What do you think?
https://developer.github.com/v3/activity/events/types/#createevent

Octokit/Models/Response/RefType.cs
```diff
@@ -0,0 +1,30 @@
...
+    {
+        /// <summary>
+        /// The object is of type repository
+        /// </summary>
+        [Parameter(Value = "repository")]
+        Repository,
+        /// <summary>
+        /// The object is of type branch
+        /// </summary>
+        [Parameter(Value = "branch")]
```
ryangribble: minor nit: missing a whitespace line

GMouron: fixed :)

ryangribble:
>> as well as the content of the events sent by github for the creation of a branch which does not contain them, I've decided to omit these here since, from my understanding, they will never be part of the payload we receive

Octokit.net uses these payload response classes for the results from "Activity Event" Api, which does include repository events and thus does include these fields.

The response classes can also be used to deserialise webhook payloads, although we dont natively have support for webhooks, so users doing this would be having to call the SipleJsonSerializer themselves currently.

So I think the extra fields should be included, since the current main supported use case in Octokit (the events API) would use them.  They would then just be `null` for anyone using them from a webhook perspective.

When we implement native webhook support we can consider whether to re-use these payload classes or have new ones specifically for the webhook payloads

GMouron:
OK, I'll be adding them then :)

GMouron:
And done :)

ryangribble:
release_notes: Support `CreateEvent` and `DeleteEvent` payloads, using new response models `CreateEventPayload` and `DeleteEventPayload`

ryangribble:
Thanks @GMouron!  (we have quite a few payload events that aren't implemented - it would be great if you wanted to add more!)

LGTM

GMouron:
> Thanks @GMouron! (we have quite a few payload events that aren't implemented - it would be great if you wanted to add more!)
> 
> LGTM

Hi, I might take a look into that a bit later but I really needed those ones and I have to first take care of the code that depends on them 😄

PR #1934, opened 2019-01-30
Allow fully qualified or short references to be used to generate "reference" Uri

### LINKED ISSUE #1933: Octokit says that a tag does not exist when it does, in fact
Using LINQpad and OctoKit, for my testing repository, I have utilized the GitHub website to add a tag called 'v0.7.3' for testing purposes.  The code
```
       /* Assuming the client variable is a GitHubClient that is properly authenticated and
          that repoOwner and repoName reference a repository to which I have full permissions */

	var refs = await client.Git.Reference.GetAll(repoOwner, repoName);
	refs.Select(r => r.Ref).Dump();
	return;
```
produces the result
image
Then, when I do
```
       /* Assuming the client variable is a GitHubClient that is properly authenticated and
          that repoOwner and repoName reference a repository to which I have full permissions */

        var refs = await client.Git.Reference.GetAll(repoOwner, repoName);
	
	foreach (var tagRef in refs)
	{
		if (!tagRef.Ref.Contains("v0.7.3"))
			continue;

		await client.Git.Reference.Delete(repoOwner,
			repoName, tagRef.Ref);
	}
```
Which should work, since the ```if``` branch does indeed let the loop move to the ```await client.Git.Reference.Delete``` line if I single-step, since tag matching the criteria does indeed exist (as has been shown above).  However, upon calling ```await client.Git.Reference.Delete(...)``` on a **tag that exists*** I get the ```ApiValidationException``` shown
image
This counter-intuitive.  Why is the API telling me my reference does not exist when it clearly does?

To reproduce:
1. Sign in to GitHub.com website in any browser.
2. Create a new Repo under your account with a README.md
3. Alter the README.md and commit it
4. Note the repo's name and your GitHub username
5. In your browser, create a tag with whichever name you want
6. Run the code in the second block above (with the appropriate authentication, repo owner, repo name, and tag search criteria.
7. Note that the exception occurs.

astrohart:
Fixes #1933 

30 Jan 2019 Updated the second overload of the ```ApiUrls.Reference``` to remove any occurrence of ```refs/``` from the string passed in for its ```referenceName``` parameter, so a reference from a tag retrievd with the API can be used natively with other reference APIs, such as Delete Reference.

### DIFF
--- a/Octokit/Models/Response/ActivityPayloads/PushWebhookCommit.cs
+++ b/Octokit/Models/Response/ActivityPayloads/PushWebhookCommit.cs
@@ -1,0 +1,41 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Globalization;
+
+namespace Octokit
+{
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
+    public class PushWebhookCommit
+    {
+        public string Id { get; protected set; }
+
+        public string TreeId { get; protected set; }
+
+        public bool Distinct { get; protected set; }
+
+        public string Message { get; protected set; }
+
+        public DateTimeOffset Timestamp { get; protected set; }
+
+        public Uri Url { get; protected set; }
+
+        public Committer Author { get; protected set; }
+
+        public Committer Committer { get; protected set; }
+
+        public IReadOnlyList<string> Added { get; protected set; }
+
+        public IReadOnlyList<string> Removed { get; protected set; }
+
+        public IReadOnlyList<string> Modified { get; protected set; }
+
+        internal string DebuggerDisplay
+        {
+            get
+            {
+                return string.Format(CultureInfo.InvariantCulture, "Sha: {0}", Id);
+            }
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


ryangribble:
Any reason why you've removed the `.dotsettings` resharper settings file @astrohart  ?

ryangribble:
I'm ok with this change - I agree it is robust and handles both the "existing" behaviour where `/refs` wasnt specified in the argument, plus works with "new" behaviour where you quite rightly point out, a ref received from another endpoint should be able to be passed straight into the delete method and work.

It actually stems from the GitHub Api documentation itself, where it indicates the URI for patch and put calls is `/refs/:ref` yet on the create call requires the "fully qualified ref" to be specified. 

Nevertheless I think this change makes things less surprising.

I would ask 2 things though
- if we no longer require the ref to be formatted in a particular way, could we remove that XmlComment in the code, that tells users not to pass in `/ref` ?
- Since the Update (patch) endpoint has the same behaviour, can we ensure this is also updated/tested? (i am on mobile so haven't checked if it uses the same ApiUrl overload or has the same comment guidance about not passing `/refs/` in)

shiftkey:
I accidentally closed this out because I could rewrite the branch to update the history and drop the tooling commit. I've opened #2104 to restart the process and see if we can land the fix.

Issue #1939, opened 2019-02-19
Retrieving events for issue throws System.OverflowException


matt-richardson:
I'm using Octokit version `0.29.0`, and I've just starting hitting an exception retrieving events for an issue:

```
System.OverflowException: Value was either too large or too small for an Int32.
   at System.Convert.ToInt32(Int64 value)
   at System.Int64.System.IConvertible.ToInt32(IFormatProvider provider)
   at System.Convert.ChangeType(Object value, Type conversionType, IFormatProvider provider)
   at Octokit.PocoJsonSerializerStrategy.DeserializeObject(Object value, Type type)
   at Octokit.Internal.SimpleJsonSerializer.GitHubSerializerStrategy.DeserializeObject(Object value, Type type)
   at Octokit.PocoJsonSerializerStrategy.DeserializeObject(Object value, Type type)
   at Octokit.Internal.SimpleJsonSerializer.GitHubSerializerStrategy.DeserializeObject(Object value, Type type)
   at Octokit.PocoJsonSerializerStrategy.DeserializeObject(Object value, Type type)
   at Octokit.Internal.SimpleJsonSerializer.GitHubSerializerStrategy.DeserializeObject(Object value, Type type)
   at Octokit.SimpleJson.DeserializeObject(String json, Type type, IJsonSerializerStrategy jsonSerializerStrategy)
   at Octokit.SimpleJson.DeserializeObject[T](String json, IJsonSerializerStrategy jsonSerializerStrategy)
   at Octokit.Internal.SimpleJsonSerializer.Deserialize[T](String json)
   at Octokit.Internal.JsonHttpPipeline.DeserializeResponse[T](IResponse response)
   at Octokit.Connection.<Run>d__56`1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Octokit.ApiConnection.<GetPage>d__41`1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #1940, opened 2019-02-19
Handle event ids greater than int32.maxvalue

matt-richardson:
Fixes https://github.com/octokit/octokit.net/issues/1939

### DIFF
--- a/Octokit/Models/Response/EventInfo.cs
+++ b/Octokit/Models/Response/EventInfo.cs
@@ -11,7 +11,7 @@ namespace Octokit
     {
         public EventInfo() { }
 
-        public EventInfo(int id, string nodeId, string url, User actor, User assignee, Label label, EventInfoState @event, string commitId, DateTimeOffset createdAt)
+        public EventInfo(long id, string nodeId, string url, User actor, User assignee, Label label, EventInfoState @event, string commitId, DateTimeOffset createdAt)
         {
             Id = id;
             NodeId = nodeId;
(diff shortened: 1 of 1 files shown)


ryangribble:
Thanks @matt-richardson 

LGTM

matt-richardson:
Thanks for the quick response. Whats your thoughts around when this could make it into a release?

ryangribble:
I've been meaning to push a release already however there wasn't anything that was urgent/pressing. This PR changes that though so I'll try to do it in the next few days. 

In the meantime you can pull the package built from merging this PR from our appveyor nuget feed if you need to get back up and running 👍

ryangribble:
release_notes: Update `EventInfo.Id` field from `int` to `long` to prevent overflow exceptions

adamralph:
Any update on getting this released? We're currently blocked on this over at https://github.com/Particular

hnrkndrssn:
@adamralph you could use the beta package https://github.com/octokit/octokit.net/blob/master/README.md#beta-packages

That's how we unblocked ourselves while waiting for the next release.

adamralph:
@hnrkndrssn in our case, we're using LINQPad scripts, and unfortunately we can't bake a package source into the script definitions.

Eilon:
I just hit this as well with some events with IDs greater than int.MaxValue. Would love to see a public release that has this fix!

Thraka:
I just ran into this issue too.

```csharp
var results = await client.Issue.Events.GetAllForRepository("dotnet", "docs");
```

Using octokit 0.36.0

shiftkey:
@Thraka please open a new issue as I suspect that requires a different fix to what was addressed in this PR.

PR #1946, opened 2019-02-21
Add Id field to Label model

### LINKED ISSUE #1941: Add Id to Octokit.Label
According to the GitHub API documentation, there's seems to be an `Id` property that identifies a specific label but it's not exposed via the Octokit API.

Is this something you would accept a pull request for?

Potapy4:
### Summary
Small change - I added Id to the Label model property.

Resolve #1941

### DIFF
--- a/Octokit/Models/Response/Label.cs
+++ b/Octokit/Models/Response/Label.cs
@@ -8,8 +8,9 @@ namespace Octokit
     {
         public Label() { }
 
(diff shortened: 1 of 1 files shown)


ryangribble:
Thanks @Potapy4 :+1:

ryangribble:
release_notes: Add `Id` field to `Label` response model

PR #1947, opened 2019-02-21
Added ID to label

### LINKED ISSUE #1941: Add Id to Octokit.Label

patriksvensson:
Couldn't find any tests that covers label, but if I've missed something please point it out and I'll fix it.

I also added an `id` parameter to the constructor which is pure cargo culting since I see no reference to the constructor anywhere 😄 

Closes #1941

### DIFF
--- a/Octokit/Models/Response/Label.cs
+++ b/Octokit/Models/Response/Label.cs
@@ -1,4 +1,4 @@
-﻿using System.Diagnostics;
+using System.Diagnostics;
 using System.Globalization;
 
 namespace Octokit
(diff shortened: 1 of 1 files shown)


M-Zuber:
If I recall correctly the ctors are used during deserialization.

M-Zuber:
Here are the related tests that I found if it helps any:
https://github.com/octokit/octokit.net/blob/3147ddd6941bc45fbe5fb2a2a958eb61e10edf4a/Octokit.Tests.Integration/Clients/IssuesLabelsClientTests.cs

https://github.com/octokit/octokit.net/blob/3147ddd6941bc45fbe5fb2a2a958eb61e10edf4a/Octokit.Tests/Clients/IssuesLabelsClientTests.cs

https://github.com/octokit/octokit.net/blob/3147ddd6941bc45fbe5fb2a2a958eb61e10edf4a/Octokit.Tests/Reactive/ObservableIssuesLabelsClientTests.cs

patriksvensson:
I see that there already is a PR for this (#1946) so I'm closing this one.

PR #1951, opened 2019-03-04
Correctly support language filter in repo search

benmcmorran:
The language was being directly converted to a string through `ToString()` instead of using the value in the `ParameterAttribute`.

### DIFF
--- a/Octokit.Tests/Clients/SearchClientTests.cs
+++ b/Octokit.Tests/Clients/SearchClientTests.cs
@@ -463,7 +463,7 @@ namespace Octokit.Tests.Clients
                 request.Language = Language.Ruby;
                 client.SearchRepo(request);
                 connection.Received().Get<SearchRepositoryResult>(Arg.Is<Uri>(u => u.ToString() == "search/repositories"),
-                    Arg.Is<Dictionary<string, string>>(d => d["q"] == "github+language:Ruby"));
+                    Arg.Is<Dictionary<string, string>>(d => d["q"] == "github+language:ruby"));
             }
 
             [Fact]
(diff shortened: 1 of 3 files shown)


ryangribble:
Hi @benmcmorran 

Did you have any more background or context on this change?

It seems github docs have changed and I cant actually find the list of language codes that are supported. 
 And many of our tests are using values of Csharp or Ruby which have the same name and value haha.

But even looking at your C++ example Im find that when doing searches with `C++` it is actualy returning results for repo's with language of just C.  Meanwhile using `cpp` does only find C++ language type repos.  So perhaps our parameter value for the `CPlusPlus` entry should be `cpp` and not `C++` anyway???

I cant seem to find the list of languages in docs anywhere anymore...

https://api.github.com/search/repositories?q=test+language:CPlusPlus (clearly wrong, there are all sorts of repos returned)
https://api.github.com/search/repositories?q=test+language:C++ (this still seems incorrect as you get `C` results)
https://api.github.com/search/repositories?q=test+language:cpp (this is now returning `C++` results)

hnrkndrssn:
I think I found the list of languages https://github.com/github/linguist/blob/master/lib/linguist/languages.yml

benmcmorran:
I initially ran into this when I was attempting to write a tool to download a bunch of C/C++ repos, but most of my results ended up being Java instead. That's a good catch though about C++ not actually returning C++ results. I had only played with the [web search](https://github.com/search?q=language%3Ac%2B%2B&type=Repositories), which does correctly handle `c++` as a language. Looking back at my results, it looks like I only ended up with C repositories in my dataset.

benmcmorran:
I pushed another commit to fix the C++ case.

ryangribble:
we probably have other language codes that need aligning but i think in the context of this PR, the changes you've made plus fixing up the obvious c++ language code issue, should be good enough :+1:

ryangribble:
release_notes: Fix `SearchRepositoriesRequest` `Language` filter option to use the parameter value rather than enum member name

PR #1954, opened 2019-03-15
Add a warning to the statistics client about caching

### LINKED ISSUE #1814: add warning from docs about statistics client data potentially being cached
The documentation for the statistics client includes a warning that the data might be cached.
(https://developer.github.com/v3/repos/statistics/#a-word-about-caching)
It might be worthwhile to include that warning or a link to it, in the intellisene on the methods

hnrkndrssn:
I had to add the warning to the `Statistics` property as well on the `*RepositoriesClient` otherwise the warning won't show up in intellisense when you do `github.Repositories.Statistics`

Fixes #1814

### DIFF
--- a/Octokit.Reactive/Clients/IObservableRepositoriesClient.cs
+++ b/Octokit.Reactive/Clients/IObservableRepositoriesClient.cs
@@ -203,17 +203,13 @@ namespace Octokit.Reactive
         /// <summary>
         /// Client for GitHub's Repository Statistics API
         /// </summary>
+        /// <summary>
+        /// Note that the GitHub API uses caching on these endpoints,
+        /// see <a href="https://developer.github.com/v3/repos/statistics/#a-word-about-caching">a word about caching</a> for more details.
+        /// </summary>
         /// <remarks>
         /// See the <a href="http://developer.github.com/v3/repos/statistics/">Statistics API documentation</a> for more details
         ///</remarks>
-        /// <remarks>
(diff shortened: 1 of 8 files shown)


ryangribble:
Unless I'm totally wrong on this, I thought the `<remarks>` tag didnt appear in intellisense - and it was only `<summary>` that does?

I'm also wondering whether we should embed the full text from GitHub here (what if the text changes in the future?) or whether we could just have a simpler note like `"Note that the GitHub API uses caching on these endpoints, see https://developer.github.com/v3/repos/statistics/#a-word-about-caching for futther details"`

Octokit.Reactive/Clients/IObservableRepositoriesClient.cs
```diff
@@ -203,6 +203,10 @@ public interface IObservableRepositoriesClient
         /// <summary>
         /// Client for GitHub's Repository Statistics API
         /// </summary>
+        /// <summary>
```
ryangribble: So this has added a 2nd `<summary>` section to the one that was already there.  I wasnt sure if this actually works or not.  Some brief testing I just did seems to indicate that the 2nd one is ignored... 

is this working for you?  If not, I think we need to append this info into the existing `<summary>` tags and not add duplicate ones

hnrkndrssn: Works On My Machine :tm:
image

This is in Rider though, I tested it in VS2017 and it does not show the second `<summary>` section.

PR #1956, opened 2019-03-29
Failing tests for GetAllContentsByRef on a branch with a root path

SimonCropp:
the `"/"` seems to force it back to the master branch. just wondering if this is by design?

### DIFF
--- a/Octokit.Tests.Integration/Clients/RepositoryContentsClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/RepositoryContentsClientTests.cs
@@ -223,6 +223,18 @@ namespace Octokit.Tests.Integration.Clients
             }
 
             [IntegrationTest]
+            public async Task GetsContentNonMasterWithRootPath()
+            {
+                var github = Helper.GetAuthenticatedClient();
+
+                var contents = await github
+                    .Repository
+                    .Content
+                    .GetAllContentsByRef("octocat", "Spoon-Knife", "/", reference: "test-branch");
+                Assert.Equal(4, contents.Count);
+            }
+
+            [IntegrationTest]
             public async Task GetsDirectoryContentWithRepositoryId()
             {
                 var github = Helper.GetAuthenticatedClient();


Octokit.Tests.Integration/Clients/RepositoryContentsClientTests.cs
```diff
@@ -222,6 +222,18 @@ public async Task GetsDirectoryContent()
...
 
+            [IntegrationTest]
+            public async Task GetsContentNonMasterWithRootPath()
+            {
+                var github = Helper.GetAuthenticatedClient();
+
+                var contents = await github
+                    .Repository
+                    .Content
+                    .GetAllContentsByRef("octocat", "Spoon-Knife", "/", reference: "test-branch");
```
shiftkey: The URL that gets generated in here looks like this:

https://github.com/octokit/octokit.net/blob/d16ac98d00780f86883133eabad100fc13201e00/Octokit/Helpers/ApiUrls.cs#L2353-L2356

I wonder if the `"/"` here is causing the confusing URL. @SimonCropp what happens if you make this an empty string?

SimonCropp: null and empty string throw an arg exception. https://github.com/octokit/octokit.net/blob/master/Octokit/Clients/RepositoryContentsClient.cs#L103

happy to attempt a PR where either null or empty string mean "root dir"

ryangribble:
I'm on mobile at the moment but I think there is another overload that doesn't take a lath, which you can use when you want the root

SimonCropp:
@ryangribble yep and i am already using that workaround. although it makes for some ugly code for anyone hitting this

        static Task<IReadOnlyList<RepositoryContent>> GetAllContents(IRepositoryContentsClient content, string owner, string repo, string path, string branch)
        {
            if (branch == null)
            {
                if (path != null)
                {
                    return content.GetAllContents(owner, repo, path);
                }
                return content.GetAllContents(owner, repo);
            }

            if (path != null)
            {
                return content.GetAllContentsByRef(owner, repo, path, reference: branch);
            }
            return content.GetAllContentsByRef(owner, repo, reference: branch);
        }

but ignoring that workaround. i think it is still a bug that you query for a branch and get results from master

shiftkey:
Moving this to #2105 as the fork has been removed

PR #1964, opened 2019-05-22
Add SubmittedAt field to PullRequestReview

ryangribble:
Field isnt shown in responses on the developer docs site, however poking the API shows it IS present in the json response payload and is thus populated successfully with this change

image


ryangribble:
release_notes: Implement `SubmittedAt` field on `PullRequestReview` response model

shiftkey:
Nice spot - will see if I can raise this with the right team internally to update the docs

martinnormark:
Any chance this will be merged soon?

PR #1978, opened 2019-07-31
(build) Updated Cake.Frosting to version 0.34.1

devlead:
Update build to use latest released version of Cake:Frosting (0.34.1 which uses same version of Cake.Core/Common).

### DIFF
--- a/GitVersion.yml
+++ b/GitVersion.yml
@@ -1,11 +1,13 @@
+mode: ContinuousDeployment
 branches:
   master:
     mode: ContinuousDeployment
     increment: Minor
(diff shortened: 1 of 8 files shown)


devlead:
I'll push a new commit shortly to address the GitVersion issue.

devlead:
Fixed GitVersion issue by upgrading to version 5.0.0, also switched to using the GitVersion.Tool so version can be asserted in .NET Core only environments too.

shiftkey:
Thanks so much @devlead for this update (and maintenance work!)

PR #1981, opened 2019-08-13
Fix credentials property instructions in getting-started

shaikatz:


shiftkey:
@shaikatz not sure I understand this change. `client.Credentials` is there to simplify things, and you're correct that it's wrapping `Connection.Credentials`, but I don't understand the benefits to updating this sample:

https://github.com/octokit/octokit.net/blob/01d5c8721c3fb613a8798bd3fdd443b737104e41/Octokit/GitHubClient.cs#L138-L156

shaikatz:
Since I've used IGithubClient with the DI I didn't have access to `Credentials` and accidantly assumed it doesn't exist.

PR #1988, opened 2019-09-12
fix timezone offset handling

### LINKED ISSUE #1972: DateRange formatted incorrectly
#1904 

https://github.com/octokit/octokit.net/blob/9f005c438ee393f4f687edf0b9f22d0b38ae3a37/Octokit/Models/Request/SearchRepositoriesRequest.cs#L301

I believe that this may be a bug- I don't think there should be a space for the timezone, between "ss" and "zzz" 

This is the difference between filtering merged by with the formatted space:
https://github.com/octokit/octokit.net/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aclosed+merged%3A%3E2018-07-01T21%3A11%3A01+%2B00%3A00

without the space:
https://github.com/octokit/octokit.net/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aclosed+merged%3A%3E2018-07-01T21%3A11%3A01%2B00%3A00+

shiftkey:
Fixes #1972

I was cleaning up some tests and stumbled upon this issue, and I think the fix here is to simply drop the space, as @catmorgan pointed out.

 - [x] build and tests pass
 - [x] integration tests for search also pass

### DIFF
--- a/Octokit.Tests/Clients/SearchClientTests.cs
+++ b/Octokit.Tests/Clients/SearchClientTests.cs
@@ -995,8 +995,6 @@ namespace Octokit.Tests.Clients
                     Arg.Is<Dictionary<string, string>>(d => d["q"] == "something+language:CSharp"));
             }
 
-            #region The following tests do not support the time component or timezone and will be removed in a future release
-
             [Fact]
             public void TestingTheCreatedQualifier_GreaterThan()
             {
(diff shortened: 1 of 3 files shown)


shiftkey:
release_notes: fix bug where `DateTimeOffset` is used to filter search results by date

PR #2005, opened 2019-09-20
no need to build and run tests against net46

shiftkey:
Follow-up from #2001


shiftkey:
There's an ongoing Appveyor error here because one file (a different one each run) isn't able to be validated:

```
https://raw.githubusercontent.com/octokit/octokit.net/af8ced49da4595a3a707dda07283ab0e3cdfcf82/Octokit/Models/Request/NewCheckRunAction.cs
error: url failed ServiceUnavailable: Backend unavailable, connection timeout
sourcelink test failed
failed for lib/net45/Octokit.dll
sourcelink test passed: lib/net46/Octokit.dll
sourcelink test passed: lib/netstandard1.1/Octokit.dll
1 files did not pass in C:/projects/octokit-net/packaging/Octokit.0.35.0-PullRequest2005-0006.nupkg
```

Gonna keep an eye on this, but I've seen it occur outside this PR so I think I'm fine with it for now.

PR #2006, opened 2019-09-20
add netstandard2.0 output

shiftkey:


shiftkey:
release_notes: added .NET Standard 2.0 output for NuGet packages

Issue #1970, opened 2019-06-18
Milestone is missing ID


patriksvensson:
It looks like `Id` is missing from `Octokit.Milestone`.

API example can be seen here: https://developer.github.com/v3/pulls/#get-a-single-pull-request

shiftkey:
@patriksvensson what did you need `Id` for? I believe that's an internal identifier for it, and `Number` is the public-facing ID that you should use when querying, based on https://developer.github.com/v3/issues/milestones/#get-a-single-milestone

patriksvensson:
I use it as an external identifier in a GitHub client i'm building.

What is the practice when it comes to expose the id field? A lot of other entities in the Octokit API exposes it (label, issue, pullrequest) and it's part of the public API.

I would gladly submit a PR for this.

shiftkey:
> What is the practice when it comes to expose the id field?

Some entities can be queried for using the `id`, but the only entity I can think of where this is explicitly support is for repositories and users - because they can renamed or transferred, integrators can query for that as it's considered more "stable". You'll also see these URLs in the pagination response to indicate there are more results, but it's not used much these days.

But for repository-specific entities (issues, labels, milestones, etc) I don't believe the same lookup was ever implemented.

> I would gladly submit a PR for this.

I'm not opposed to this, but I just wanted to indicate that the field is not what is used in the API itself.

cc @ryangribble just in case.

ryangribble:
If it's in the public API response payload I don't mind having it implemented here for completeness

hnrkndrssn:
@patriksvensson would you still like to submit a PR for this? Otherwise, I'll be happy to tackle it over the weekend.

PR #2007, opened 2019-09-21
Add missing ID property to Milestone

hnrkndrssn:
Not sure if I should add/update any tests to check for the existence of the `ID` property?

Fixes octokit/octokit.net#1970


shiftkey:
> Not sure if I should add/update any tests to check for the existence of the `ID` property?

I'm not worried about it

shiftkey:
release_notes: added `Id` property to `Milestone` response

Issue #1960, opened 2019-04-12
Missing Issue Comments Since Request Parameter


andyesys:
`Issue.Comment.GetAllForIssue()` has no way to provide the `since` request parameter described here: https://developer.github.com/v3/issues/comments/#list-comments-on-an-issue.

ryangribble:
Thanks for flagging this @andyesys! Would be great if you wanted to send a PR to add it!

6simon1112:
.c

PR #2008, opened 2019-09-21
Add ability to specify 'since' request parameter to issue comments

hnrkndrssn:
I noticed that `IObservable<IssueComment> GetAllForIssue(long repositoryId, int number, ApiOptions otpions)` didn't specify `AcceptHeaders.ReactionsPreview` so I added this as part of my changes.

- [x] Tests ~? (I checked the `GetAllForRepository` method and noticed that there are no tests that use the `IssueCommentRequest` parameter, but I can add tests for the `GetAllForIssue` if desired)~ `Codecov` clearly wants me to write some tests 😁 

Fixes octokit/octokit.net#1960


shiftkey:
release_notes: added `since` parameter to `Issue.Comments.GetAllForIssue()`

PR #2010, opened 2019-09-22
Begin implementation of Enterprise ManagementConsole API, redux

shiftkey:
Supersedes #1269 which predates a lot of nice things:

 - the cleanup of csproj files
 - maintainers being able to push to PR branches (so I can rebase the work)

This felt close enough, but it's been a few years so I'm gonna need time to read it over again.

 - [x] builds and tests pass
 - [x] review changes and tests make sense
 - [x] add some docs
 - [x] tidy up diff


shiftkey:
release_notes: added Management Console API to view and edit maintenance mode

Issue #1959, opened 2019-04-11
Support for "Remove organization membership" action


MikhailTymchukDX:
I'm using  `client.Organization.Member.Delete(orgName, loginToRemove)` method to remove member or pending member from organization.

The [documentation says](https://developer.github.com/v3/orgs/members/#remove-organization-membership):
> If the specified user has been invited to the organization, this will cancel their invitation. 

However, when I use this method, nothing happens.

MikhailTymchukDX:
Ok, github support said, that I need to used another method which is not implemented in Octokit.net.
It has OrganizationMembersClient.Delete, which corresponds to [Remove a member](https://developer.github.com/v3/orgs/members/#remove-a-member) API medhod.
I need [Remove organization membership](https://developer.github.com/v3/orgs/members/#remove-organization-membership) instead.

hnrkndrssn:
I'll grab this one, and at the same time add support for:
- [Get organization membership](https://developer.github.com/v3/orgs/members/#get-organization-membership) (as I couldn't find we support this endpoint)
- [Add or update organization membership](https://developer.github.com/v3/orgs/members/#add-or-update-organization-membership) (there's an open issue for tracking this [here](https://github.com/octokit/octokit.net/issues/1533))

PR #2014, opened 2019-09-23
Add get/update/delete methods for organization memberships

hnrkndrssn:
- [x] Unit Tests
- [x] Integration Tests

Fixes octokit/octokit.net#1959

### DIFF
--- a/Octokit/Clients/OrganizationMembersClient.cs
+++ b/Octokit/Clients/OrganizationMembersClient.cs
@@ -37,6 +37,22 @@ namespace Octokit
         Member
     }
 
+    public enum OrganizationMembershipsState
+    {
+        [Parameter(Value = "active")]
+        Active,
+        [Parameter(Value = "pending")]
+        Pending
+    }
+    
+    public enum OrganizationMembershipsRole
+    {
+        [Parameter(Value = "admin")]
+        Admin,
+        [Parameter(Value = "member")]
+        Member
+    }
+
     /// <summary>
     /// A client for GitHub's Organization Members API.
     /// </summary>
--- a/Octokit/Clients/OrganizationMembersClient.cs
+++ b/Octokit/Clients/OrganizationMembersClient.cs
@@ -460,6 +460,69 @@ namespace Octokit
         }
 
         /// <summary>
+        /// Get a user's membership with an organization.
+        /// </summary>
+        /// <remarks>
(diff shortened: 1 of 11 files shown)


Octokit/Clients/OrganizationMembersClient.cs
```diff
@@ -37,6 +37,22 @@ public enum OrganizationMembersRole
...
 
+    public enum OrganizationMembershipsState
+    {
+        [Parameter(Value = "active")]
+        Active,
+        [Parameter(Value = "pending")]
+        Pending
+    }
+    
+    public enum OrganizationMembershipsRole
```
hnrkndrssn: I'm not 100% sold on the naming of this `enum`...we now have 3 quite similarly named `enum`s:
- This newly added one
- [OrganizationMembersRole](https://github.com/octokit/octokit.net/blob/9bb344cd19d0b12024c534c0a3accc0467f7abd4/Octokit/Clients/OrganizationMembersClient.cs#L28) - this has the same roles as the one I've added, but with an extra value `All`
- [OrganizationMembershipRole](https://github.com/octokit/octokit.net/blob/7c170213fd2b2a803af3ea38e2ebe610ab738173/Octokit/Models/Response/OrganizationMembershipRole.cs#L5) - this one has **very** different values

Octokit/Clients/OrganizationMembersClient.cs
```diff
@@ -37,6 +37,22 @@ public enum OrganizationMembersRole
         Member
     }
 
+    public enum OrganizationMembershipsState
```
hnrkndrssn: I'm not 100% sold on this name either, maybe it should be singular `Membership` rather than plural `Memberships`...

hnrkndrssn: So I just found [`MembershipState`](https://github.com/octokit/octokit.net/blob/master/Octokit/Models/Response/MembershipState.cs#L9) 🤦‍♂

Octokit/Clients/OrganizationMembersClient.cs
hnrkndrssn: Now that I found the `MembershipState` `enum`, I'm thinking that `MembershipRole` is a better name for this `enum`

shiftkey:
release_notes: add support for organization memberships API

PR #2019, opened 2019-09-30
add support for multiple extension filters

ecarlson94:
Modifies `SearchCodeRequest.Extension` to be a list for filtering by `x` number of extensions. Helps reduce the number of required API calls.

Question: Is there an open issue to support the negation of qualifiers (`-extension:{0}`)? I couldn't find one.

### DIFF
--- a/Octokit.Tests/Clients/SearchClientTests.cs
+++ b/Octokit.Tests/Clients/SearchClientTests.cs
@@ -1801,14 +1801,16 @@ namespace Octokit.Tests.Clients
             {
                 var connection = Substitute.For<IApiConnection>();
                 var client = new SearchClient(connection);
-                var request = new SearchCodeRequest("something");
-                request.Extensions.Add("cs");
+                var request = new SearchCodeRequest("something")
+                {
+                    Extensions = new[] { "txt" }
+                };
 
                 client.SearchCode(request);
 
                 connection.Received().Get<SearchCodeResult>(
                     Arg.Is<Uri>(u => u.ToString() == "search/code"),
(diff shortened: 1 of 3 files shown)


shiftkey:
> Question: Is there an open issue to support the negation of qualifiers (`-extension:{0}`)? I couldn't find one.

@ecarlson94 not that I'm aware of. Please open one if this is something you'd like to see supported.

shiftkey:
@ecarlson94 thanks for opening this up. I also know there's a docs example in `docs/search.md` that will need to be updated to support this change:

```
var request = new SearchCodeRequest("auth")
{
    // we can restrict search to the file, path or search both
    In = new[] { CodeInQualifier.File, CodeInQualifier.Path },
    
    // how about we find a file based on a certain language
    Language = Language.JavaScript,
    
    // do we want to search forks too?
    Forks = true,

    // find files that are above 1000 bytes
    Size = Range.GreaterThan(1000),

    // we may want to restrict the search to the path of a file
    Path = "app/assets",
    
    // we may want to restrict the file based on file extension
    Extension = "json",
    
    // restrict search to a specific file name
    FileName = "app.json",
    
    // search within a users or orgs repo
    User = "dhh"
};
```

ecarlson94:
@shiftkey, any idea what I should do about this [Travis CI error](https://travis-ci.org/octokit/octokit.net/jobs/592674625#L2559)

shiftkey:
@ecarlson94 it's unrelated to the PR, so I've just re-run it to see if it'll pass

shiftkey:
release_notes: `SearchCodeRequest.Extension` has been renamed to `SearchCodeRequest.Extensions` to support multiple file extensions

PR #2032, opened 2019-10-17
add JupyterNotebook to Repository Search

### LINKED ISSUE #2030: Newer Languages are missing
Acording to this list :
https://github.com/github/linguist/blob/master/lib/linguist/languages.yml

those are the languages in github.

On ocktokit serveral are missing such as "Jupyter Notebooks", how can more languages be added? since currently is only a enum but i don't know if it affects other parts.

Dagizmo:
Related to #2030 

If there is need to add more languages let me know and i could get working on that :)

### DIFF
--- a/Octokit/Models/Request/SearchRepositoriesRequest.cs
+++ b/Octokit/Models/Request/SearchRepositoriesRequest.cs
@@ -83,7 +83,7 @@ namespace Octokit
         public Range Size { get; set; }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


shiftkey:
> If there is need to add more languages let me know and i could get working on that :)

For now I'm going leave #2030 open so I can think about the bigger picture of how to solve this.

shiftkey:
release_notes: Added "Jupyter Notebook" to Search Repositories API

PR #2038, opened 2019-10-24
fix language with spaces giving wrong results [#1607]

### LINKED ISSUE #1607: Search API: languages with space(s) fetch wrong results
The following code:

```csharp
var client = new GitHubClient();
var searchRepositoriesRequest = new SearchRepositoriesRequest { Language = Language.GettextCatalog };

var result = await client.Search.SearchRepo(searchRepositoriesRequest);
```

hits the following URL: https://api.github.com/search/repositories?q=language:Gettext%20Catalog, which returns incorrect results - see the first repo has `language: "JavaScript"`.

When surrounding the language with double quotes, like https://api.github.com/search/repositories?q=language:"Gettext%20Catalog", then the fetched results are correct as all the repos have `language: "Gettext Catalog"`.

Fix should be easy as we could use double quotes around the `language` value no matter if it contains spaces or not.

Dagizmo:
Fixes #1607

### DIFF
--- a/Octokit/Models/Request/SearchUsersRequest.cs
+++ b/Octokit/Models/Request/SearchUsersRequest.cs
@@ -117,7 +117,7 @@ namespace Octokit
 
             if (Language != null)
             {
-                parameters.Add(string.Format(CultureInfo.InvariantCulture, "language:\"{0}\"", Language));
+                parameters.Add(string.Format(CultureInfo.InvariantCulture, "language:\"{0}\"", Language.ToParameter()));
             }
 
             if (Created != null)
--- a/Octokit/Models/Request/SearchUsersRequest.cs
+++ b/Octokit/Models/Request/SearchUsersRequest.cs
@@ -117,7 +117,7 @@ namespace Octokit
 
             if (Language != null)
             {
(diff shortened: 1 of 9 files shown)


shiftkey:
@Dagizmo I think the only thing that's missing from this PR is a test that verifies we send through the space in the correct way. Would you be able to add that in?

Dagizmo:
ok, will look into that.

Octokit/Models/Request/SearchUsersRequest.cs
```diff
@@ -117,7 +117,7 @@ public override IReadOnlyList<string> MergedQualifiers()
 
             if (Language != null)
             {
-                parameters.Add(string.Format(CultureInfo.InvariantCulture, "language:\"{0}\"", Language));
+                parameters.Add(string.Format(CultureInfo.InvariantCulture, "language:\"{0}\"", Language.ToParameter()));
```
Dagizmo: Actually found that this part didn't use to Parameter so named languages such as CPlusPlus was parsed as "CPlusPlus" instead of "cpp"

Dagizmo:
updated :D

shiftkey:
release_notes: address encoding of spaces when using Search API

PR #2046, opened 2019-11-18
WIP: Get team by org and name

SeanKilleen:
This allows someone to use `Teams.Get("organizationName", "teamName")` to retrieve a `Team` object in cases where someone may not know the team ID off-hand. 

NOTE: This uses `.First()` on the result of `GetAll()`, which I think is sufficient in this case as we'd want an error to occur if no matching team is found. However, let me know if you'd like this error to be more expressive.

### DIFF
--- a/Octokit/Clients/TeamsClient.cs
+++ b/Octokit/Clients/TeamsClient.cs
@@ -40,15 +40,15 @@ namespace Octokit
 
         /// <summary>
         /// Gets a single <see cref="Team"/> by name.
-        /// Note: this is more expensive than using the
-        /// team ID, as first we perform a lookup to determine
-        /// matching teams.
         /// </summary>
         /// <remarks>
-        /// https://developer.github.com/v3/orgs/teams/#get-team
+        /// this is more expensive than using the team ID, as first we
+        /// perform a lookup to determine matching teams.
         /// </remarks>
         /// <param name="orgName">The organization name.</param>
         /// <param name="teamName">The team name.</param>
+        /// <exception cref="ArgumentNullException">if the list of teams returned from GetAll is null.</exception>
+        /// <exception cref="InvalidOperationException">if no matching team name is found.</exception>
         /// <returns>The <see cref="Team"/> with the given identifier.</returns>
         public async Task<Team> Get(string orgName, string teamName)
         {
--- a/Octokit/Clients/TeamsClient.cs
+++ b/Octokit/Clients/TeamsClient.cs
@@ -2,7 +2,6 @@
 using System.Net;
 using System.Threading.Tasks;
 using System.Collections.Generic;
-using System.Linq;
 
 namespace Octokit
 {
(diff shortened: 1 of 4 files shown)


Octokit/Clients/TeamsClient.cs
```diff
@@ -37,6 +38,28 @@ public Task<Team> Get(int id)
...
+        /// <param name="teamName">The team name.</param>
+        /// <returns>The <see cref="Team"/> with the given identifier.</returns>
+        public async Task<Team> Get(string orgName, string teamName)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(orgName, nameof(orgName));
+            Ensure.ArgumentNotNullOrEmptyString(teamName, nameof(teamName));
+
+            var allTeams = await this.GetAll(orgName);
+
+            return allTeams.First(x => string.Equals(x.Name, teamName, StringComparison.CurrentCultureIgnoreCase));
```
SeanKilleen: ℹ️ I think `CurrentCultureIgnoreCase` is acceptable here but want to call attention to it just in case.

Octokit/Clients/TeamsClient.cs
```diff
@@ -37,6 +38,28 @@ public Task<Team> Get(int id)
...
+        /// <param name="teamName">The team name.</param>
+        /// <exception cref="ArgumentNullException">if the list of teams returned from GetAll is null.</exception>
+        /// <exception cref="InvalidOperationException">if no matching team name is found.</exception>
+        /// <returns>The <see cref="Team"/> with the given identifier.</returns>
+        public async Task<Team> Get(string orgName, string teamName)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(orgName, nameof(orgName));
+            Ensure.ArgumentNotNullOrEmptyString(teamName, nameof(teamName));
+
+            var allTeams = await this.GetAll(orgName);
```
shiftkey: By hiding `GetAll` here we don't let the caller control pagination, which means for orgs with a large # of teams it'll take several requests to fetch them all, before doing the filtering.

Octokit/Clients/TeamsClient.cs
```diff
@@ -37,6 +38,28 @@ public Task<Team> Get(int id)
...
+
+            return allTeams.First(x => string.Equals(x.Name, teamName, StringComparison.CurrentCultureIgnoreCase));
```
shiftkey: I'm not a fan of `.First` here because it's behaviour is different to how Octokit itself handles cases where an entity cannot be found. In these cases the API response returns a 404, which we convert into a `NotFoundException` and throw as part of the plumbing. I'd prefer we stick to this pattern to avoid callers having to handle multiple things.

SeanKilleen:
@shiftkey thanks for the quick feedback! I appreciate it. 

Totally agree with your assessment and missed the get team by name API. Happy to re-work this around that.

Thanks again for jumping in before I got too far. 👍

shiftkey:
@SeanKilleen no problem - let me know if there's any questions you have as you rework it!

shiftkey:
@SeanKilleen i'm looking at my backlog of PRs and I see this one has stalled - anything I can do to help unblock it?

SeanKilleen:
Hey @shiftkey, thanks for the memory jog! Totally on me. I'm actually going to close this out for now because it was in very early stages and some priorities shifted. I may have cause to revisit soon in which case I'll open up a new PR.

PR #2054, opened 2019-12-24
GitHubClient: HttpClient: RequestTimeout property added

arxange1:
The problem: 
1. The request timeout is hard-coded inside Request class constructor (TimeSpan.FromSeconds(100))
2. GitHubClient.SetRequestTimeout() method does not change the request timeout as expected: hard-coded Request.Timeout is applied instead.
3. GitHubClient does not provide an ability to get/read an effective request timeout

The solution:
- The method "SetRequestTimeout()" replaced by "RequestTimeout" property: addresses the problem 3
- Request.Timeout property is initialized with "RequestTimeout" property value: addresses the problem 1 and 2

### DIFF
--- a/Octokit.Reactive/IObservableGitHubClient.cs
+++ b/Octokit.Reactive/IObservableGitHubClient.cs
@@ -7,14 +7,13 @@ namespace Octokit.Reactive
         IConnection Connection { get; }
 
         /// <summary>
-        /// Set the GitHub Api request timeout.
+        /// GitHub Api request timeout.
         /// Useful to set a specific timeout for lengthy operations, such as uploading release assets
         /// </summary>
         /// <remarks>
         /// See more information here: https://technet.microsoft.com/library/system.net.http.httpclient.timeout(v=vs.110).aspx
         /// </remarks>
-        /// <param name="timeout">The Timeout value</param>
-        void SetRequestTimeout(TimeSpan timeout);
+        TimeSpan RequestTimeout { get; set; }
 
         IObservableAuthorizationsClient Authorization { get; }
         IObservableActivitiesClient Activity { get; }
--- a/Octokit.Reactive/ObservableGitHubClient.cs
+++ b/Octokit.Reactive/ObservableGitHubClient.cs
@@ -57,16 +57,22 @@ namespace Octokit.Reactive
         }
 
         /// <summary>
-        /// Set the GitHub Api request timeout.
+        /// GitHub Api request timeout.
         /// Useful to set a specific timeout for lengthy operations, such as uploading release assets
         /// </summary>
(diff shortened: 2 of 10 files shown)


elv1s42:
Hi @shiftkey !
Any updates on this?
Looks like I'm having the same issue.

shiftkey:
@arxange1 @elv1s42 I can't take this PR in in it's current form because it's not clear to me what we're trying to achieve here. 

> * The request timeout is hard-coded inside Request class constructor (TimeSpan.FromSeconds(100))

This was chosen as it matches the default value for [`HttpClient.Timeout`](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.timeout?view=netframework-4.8).

> * GitHubClient.SetRequestTimeout() method does not change the request timeout as expected: hard-coded Request.Timeout is applied instead.

This is the bit I'd like to understand more about. Inside Octokit we use the request timeout to create a cancellation token (or combine it with an existing cancellation token):

https://github.com/octokit/octokit.net/blob/fe6639f270aa20f83a3fa6d47c88315d2c340eef/Octokit/Http/HttpClientAdapter.cs#L79-L91

This is the recommended approach to differentiate between global and per-request timeouts mentioned in the documentation link above:

> You may also set different timeouts for individual requests using a `CancellationTokenSource` on a task. **Note that only the shorter of the two timeouts will apply**.

What are we trying to do here? I don't believe plumbing a new property all the way through the internals is the best approach here, and will potentially confuse callers.

> * GitHubClient does not provide an ability to get/read an effective request timeout

It's not clear to me the value of this, and as mentioned in the previous point "effective" here requires a better understanding about what the goals or issues are.

arxange1:
@shiftkey, The mentioned statement "Note that only the shorter of the two timeouts will apply." is true. In the sample below the shorter is hardcoded in the Request class constructor.

`	var client = new GitHubClient(new ProductHeaderValue("MyAmazingApp"));
	client.SetRequestTimeout(TimeSpan.FromMinutes(3));
`
The code looks like the timeout is 3 minutes. But the timeout is 100 seconds instead. It is a bug.

Plus, if the library provides "SetRequestTimeout" method, I would expect it to provide "GetRequestTimeout" as well. Currently there is no API to get an effective timeout, even if you want to get it it for diagnostic purposes.

Eliminating hardcoded timeout - is a bug-fix actually.
New property - is a feature request to make an API clean. We may avoid it for sure, just let me know. if you need a working sample to demonstrate the buggy behavior, let me know as well.

shiftkey:
Let's step back and figure out the problem we're facing and how things should work in Octokit, before diving into the code changes. I've opened #2112 to discuss this further - please comment there.

Issue #1979, opened 2019-08-06
Retrieving events throws Stack.Overflow exception.


bobbytreed:
In the lastest local build (0.33.0), I still get a StackOverflow Exception.  While the type has been updated, something in PocoSerializationStrategy is still trying to convert to Int32
System.AggregateException: One or more errors occurred. ---> System.OverflowException: Value was either too large or too small for an Int32.
                            at System.Convert.ToInt32(Int64 value)
                            at System.Convert.ChangeType(Object value, Type conversionType, IFormatProvider provider)
                            at Octokit.PocoJsonSerializerStrategy.DeserializeObject(Object value, Type type)
                            at Octokit.PocoJsonSerializerStrategy.DeserializeObject(Object value, Type type)
                            at Octokit.PocoJsonSerializerStrategy.DeserializeObject(Object value, Type type)
                            at Octokit.SimpleJson.DeserializeObject[T](String json, IJsonSerializerStrategy jsonSerializerStrategy)
                            at Octokit.Internal.JsonHttpPipeline.DeserializeResponse[T](IResponse response)
                            at Octokit.Connection.<Run>d__57`1.MoveNext()
                         --- End of stack trace from previous location where exception was thrown ---
                            at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
                            at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
                            at Octokit.ApiConnection.<GetPage>d__42`1.MoveNext()
                         --- End of stack trace from previous location where exception was thrown ---
                            at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
                            at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
                            at Octokit.ApiConnection.<>c__DisplayClass17_0`1.<<GetAll>b__0>d.MoveNext()
                         --- End of stack trace from previous location where exception was thrown ---
                            at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
                            at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
                            at Octokit.ApiPagination.<GetAllPages>d__0`1.MoveNext()
                            --- End of inner exception stack trace ---
                         ---> (Inner Exception #0) System.OverflowException: Value was either too large or too small for an Int32.
[...]

PR #2037, opened 2019-10-24
Fix for ThrowInt32OverflowException in Issue.Events.GetAllForRepository

terrance00:
Fixes #2031

 - Updated Model with long Id
 - Fixed tests
 - Updated IssuesClient and interface
 - Updated IssuesEventsClient and interface

- Double checked that the same tests still failing for the same reasons.

Issue #2031, opened 2019-10-17
ThrowInt32OverflowException on Issue.Events.GetAllForRepository


Thraka:
Using octokit 0.36.0

The following code causes an exception:

```csharp
var results = await client.Issue.Events.GetAllForRepository("dotnet", "docs");
```

```
This exception was originally thrown at this call stack:
	System.Convert.ThrowInt32OverflowException()
	System.Convert.ToInt32(long)
	long.System.IConvertible.ToInt32(System.IFormatProvider)
	System.Convert.ChangeType(object, System.Type, System.IFormatProvider)
	Octokit.PocoJsonSerializerStrategy.DeserializeObject(object, System.Type) in SimpleJson.cs
	Octokit.Internal.SimpleJsonSerializer.GitHubSerializerStrategy.DeserializeObject(object, System.Type) in SimpleJsonSerializer.cs
	Octokit.PocoJsonSerializerStrategy.DeserializeObject(object, System.Type) in SimpleJson.cs
	Octokit.Internal.SimpleJsonSerializer.GitHubSerializerStrategy.DeserializeObject(object, System.Type) in SimpleJsonSerializer.cs
	Octokit.PocoJsonSerializerStrategy.DeserializeObject(object, System.Type) in SimpleJson.cs
	Octokit.Internal.SimpleJsonSerializer.GitHubSerializerStrategy.DeserializeObject(object, System.Type) in SimpleJsonSerializer.cs
	...
	[Call Stack Truncated]
```

shiftkey:
This is the problem property:

https://github.com/octokit/octokit.net/blob/74dc51a6f567395d0c46d97f7270f959d671573e/Octokit/Models/Response/IssueEvent.cs#L26-L29

And much like we did in #1940, changing it to a `long` should fix the issue.

terrance00:
I'd like to contribute, forked.

terrance00:
Hi @shiftkey, I've got a question.

When we change that model from `int` to `long` for the `Id`, some tests break:

```
    //IssueEventsClientTests.cs
    var issueEvents = await _issuesEventsClient.GetAllForRepository(_context.Repository.Id);
    int issueEventId = issueEvents[0].Id;
```

This does then also break those interfaces it calls.

Shall I follow the chain down and change the interfaces and implementations to accept longs etc.?

Let me know - I'll create a PR by following the chain all the way for the longs.

shiftkey:
[...]

PR #2060, opened 2020-01-08
Update IssueEvent ID field from int to long

stevedesmond-ca:
This PR fixes issues #1979 and #2031 (stack overflow when attempting to retrieve events on issues due to large ID numbers) and replaces the seemingly-abandoned PR #2037.

In addition to changing the types of the relevant field and parameters, the names of the parameters have also been updated (from `number` to `eventId`) to better reflect what's expected to be passed.


stevedesmond-ca:
(the code coverage reduction is a little disconcerting, given that this was all 1:1 type/name changes)

shiftkey:
> (the code coverage reduction is a little disconcerting, given that this was all 1:1 type/name changes)

Yeah, I had a poke at the places it say were regressed but I can't spot anything I should be worried about 🤷‍♂

shiftkey:
release_notes: changed `IssueEvent.Id` from `int` to `long` to address deserialization issue

PR #2064, opened 2020-01-27
add reference assemblies

shiftkey:
Looking to review some of the infrastructure things because how I build and cut Octokit releases currently is very fiddly.

This PR starts us off with reducing our dependency on Mono for builds by adding the `Microsoft.NETFramework.ReferenceAssemblies` package. 

This shouldn't affect our CI builds but I've tested locally on Ubuntu without Mono and (combined with some other small PRs) should get us to a good place.


shiftkey:
release_notes: add `Microsoft.NETFramework.ReferenceAssemblies` to allow building projects properly on platforms that only have .NET Core installed

PR #2065, opened 2020-01-27
upgrade sourcelink tools and GitVersion

shiftkey:
Following on from #2064 

Not sure why some of these dependencies aren't being detected by Dependabot, but for building and linking on Linux without Mono these upgrades are useful.

The change from `SourceLink.Create.GitHub` to `SourceLink.Create.CommandLine` was recommended in https://github.com/ctaggart/SourceLink/issues/321#issuecomment-373508013 due to problems raised previously with this project (the switch didn't happen earlier because it seemed to be an intermittent issue).


shiftkey:
release_notes: upgraded SourceLink and GitVersion tools

PR #2072, opened 2020-01-28
strip CoreOnly usages from codebase as they are no longer needed

shiftkey:
This was a vestige of the days when we needed Mono to target the classic NetFX targets, and shouldn't be needed now.


shiftkey:
release_notes: cleanup `CoreOnly` usage in project files

PR #2074, opened 2020-01-28
switch to Microsoft.SourceLink.GitHub package to perform linking

shiftkey:
I want to do away with the use of `DotNetCliToolReference` inside here (there's a kludge that I'll explain when we get to it) and for now it looks like `Microsoft.SourceLink.*` are the current ways to achieve this (with some options that I might look into later).

For now, I just want to confirm this works and passes `sourcelink test` for each build.


shiftkey:
release_notes: Switch over to `Microsoft.SourceLink.GitHub` for embeddeding GitHub URLs into assemblies

PR #2080, opened 2020-02-04
Add `stale` check conclusion

BenEmdon:
Hey Dotnet SDK team 👋 

I work on the team which maintains checks at GitHub and I wanted to add this conclusion ahead of release. This check conclusion will be fully documented on release.

I would also like to note that we could add new check/status states at any time, since adding states would be an additive change. The client-side logic should reflect this in some way so that new states don't break integrators systems.

### DIFF
--- a/Octokit/Models/Common/CheckStatus.cs
+++ b/Octokit/Models/Common/CheckStatus.cs
@@ -37,6 +37,9 @@ namespace Octokit
 
         [Parameter(Value = "skipped")]
         Skipped,
+        
+        [Parameter(Value = "stale")]
+        Stale,
     }
 
     public enum CheckAnnotationLevel


shiftkey:
> The client-side logic should reflect this in some way so that new states don't break integrators systems.

Yep, for states that we don't know they'll fallback to `null` until we add them.

BenEmdon:
> @shiftkey Yep, for states that we don't know they'll fallback to `null` until we add them.

Good to know! Apologies for my wrong assumption, I am not accustomed to the codebase 😅

shiftkey:
@BenEmdon no problem - totally understandable!

shiftkey:
release_notes: Add `CheckConclusion.Stale` value to `Conclusion` property for `CheckSuite` and `CheckRun`

PR #2081, opened 2020-02-04
Revert "upgrade sourcelink tools (#2065)"

shiftkey:
While CI has been fine for a while, I've not been able to `sourcelink test` locally for a while, and it prevented me from publishing packages from my local machine. After trying a bunch of things that might have explained why my development environment was different to Appveyor (where this works fine) I've decided to try and back out the commit that I think is to blame.

I want to get to a point where I can release from GitHub - push tag, CI passes, packages published to NuGet - but for now I'd like to just be able to publish to NuGet without dirty workaround.

- [ ] verify AppVeyor passes
- [ ] verify local Windows build passes
- [ ] verify GitHub Actions passes (or pause for now)
- [ ] ???

### DIFF
--- a/Octokit.Reactive/Octokit.Reactive.csproj
+++ b/Octokit.Reactive/Octokit.Reactive.csproj
@@ -46,9 +46,10 @@
 
   <ItemGroup>
     <PackageReference Include="System.Reactive" Version="3.1.0" />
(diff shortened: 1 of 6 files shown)


shiftkey:
I know what this is, and it's not something related to this commit.

Need to investigate, repro, and put it in front of the right contributors.

Issue #2078, opened 2020-01-30
Support for the newly announced Actions API?


JeffreyPalmer:
A kind soul at GitHub pointed me at this blog post announcing the launch of the Actions API beta: https://developer.github.com/changes/2020-01-28-actions-api/. This is super exciting, and I was wondering how this project would go about adding support for this API. I can see the general implementation patterns in the code for various other endpoints, and I would be happy to help with this effort. That said, I don't develop on Windows and am not sure if that makes my efforts more challenging than someone developing on Windows.

I guess I'm basically looking for guidance on some questions:

1) Is anyone already working on adding this API?
2) Am I misguided in thinking that developing this on a mac would be straightforward? My main concern would be that I would get 90% of the way through the implementation and then not be able to debug some test failure that I couldn't run, and the effort would stall out.
3) Would anyone be able to provide feedback during the implementation process if I have questions, etc.?

Thanks!

shiftkey:
>  anyone already working on adding this API?

Not that I'm aware of.

> Am I misguided in thinking that developing this on a mac would be straightforward? My main concern would be that I would get 90% of the way through the implementation and then not be able to debug some test failure that I couldn't run, and the effort would stall out.

We're on .NET Core these days and I've been able to to work with the solution in VSCode on Linux (I'll drop to the command line mostly to run tests) but there's a bunch of patterns and conventions with how things are organized that means it should be easy to follow/borrow/steal from existing components in the library.

> Would anyone be able to provide feedback during the implementation process if I have questions, etc.?

I'm more than happy to help with reviewing.

JeffreyPalmer:
I'll take a pass at it and see how it goes. Thanks!

JeffreyPalmer:
Okay I have a first pass at the entire API. I am *definitely* going to need a lot of feedback, both because I am an F# programmer and this is my first *real* C# development, and also because I don't really know the preferred idioms for this project (although I have copied a ton!).

I'm going to try to clean things up a little and submit a WIP PR to provide an opportunity for directional feedback. Gonna try for tomorrow. 🤞🏻

kentahikaru:
Hello,
is it still possible to implement this ? Or is anybody already working on this?
If i understand correctly , it is necessary to implement all of those 6 endpoints ?
https://developer.github.com/changes/2020-01-28-actions-api/

martincostello:
I'm going to take a look at contributing the client bits for Actions, as I've been using it in some automation I've been building.

[...]

PR #2082, opened 2020-02-05
WIP: Adding support for Actions API

JeffreyPalmer:
See #2078 - Feedback Requested!

Hello! I'm putting this up in order to get some feedback on this initial implementation before I take the time to start polishing and getting ready for a real mergeable PR. Any feedback would be great!

Things still to do:
- I haven't spent any time whatsoever on documentation - I'll clean that type of stuff up once I get some confirmation that we're past the "major changes required" stage.
- There are some tests missing on the Observable side; I'll add those later.
- I haven't really tested this implementation live against the API in anger yet. I see that there are integration tests, but I'm not sure how you'd suggest we approach those.
- There are now 3 different types of Token (AccessToken, RunnerRegistrationToken, and RunnerRemoveToken). This doesn't seem great, since they're all the same. I would have created a generic Token class, but I wasn't sure how the project handles changes to existing APIs (AccessToken already existed). Thoughts?
- I added some missing fields to CheckSuite and updated the tests.

Please let me know your thoughts - I'm happy to make changes as required. As I mentioned, I'm not really a C# developer, so any thoughts on where things can be made more idiomatic, etc., would be appreciated. Thanks!


shiftkey:
@JeffreyPalmer thanks for opening this up, and apologies for the delay in looking it over. Here's some quick thoughts:

> I haven't spent any time whatsoever on documentation - I'll clean that type of stuff up once I get some confirmation that we're past the "major changes required" stage.

:+1:  - if you're feeling like this is too much to tackle in one PR, feel free to focus on a subset of this so we can review, land and publish it sooner. Otherwise happy to proceed as-is.

> There are some tests missing on the Observable side; I'll add those later.

:+1: 

>  I haven't really tested this implementation live against the API in anger yet. I see that there are integration tests, but I'm not sure how you'd suggest we approach those.

I think initially focusing on tests against the read API (so we're not creating things and potentially triggering abuse mechanisms) would be a good start - this will help confirm we're getting information from the API and it's being deserialized correctly. Then we can expand things.

>    There are now 3 different types of Token (AccessToken, RunnerRegistrationToken, and RunnerRemoveToken). This doesn't seem great, since they're all the same. I would have created a generic Token class, but I wasn't sure how the project handles changes to existing APIs (AccessToken already existed). Thoughts?


I'd be hesitant to merge these unless we're very confident they'll stay the same. Looking at `AccessToken` for example, it's actually much more than just those two values which aren't deserialized from the API:

https://developer.github.com/v3/apps/#response-5

I'd prefer to enhance that with the additional fields in a follow-up PR. With these new `RunnerRegistrationToken` and `RunnerRemoveToken` types, I think it's fine to leave them as-is - who knows how the API might evolve? 

> I added some missing fields to CheckSuite and updated the tests.

:+1: 


I've skimmed the current code and it all looks good from a code organization perspective (CI is also happy, which is a great sign). Let me know if there's any other questions, otherwise I'm :+1: to proceeding!

JeffreyPalmer:
Hello!

Unfortunately, a series of events in my life have made it unlikely that I will be able to work on this for the foreseeable future. I hate to dump unfinished code in the project's lap like this, but I can't really change the situation. Sorry about that.

I'm hoping that this starting point will be enough for someone to get this over the finish line. I can probably respond to some questions here and there, if that is of any use at all. I'm hoping that I'll be able to contribute to the project again in the future, once my situation again allows for it.

shiftkey:
@JeffreyPalmer no problem, and totally understandable.

Gonna close this out as we have the original issue #2078 to discuss it further if anyone else is interested.

PR #2084, opened 2020-02-06
Double quote issue labels in search API

### LINKED ISSUE #2044: Search API: tags are not quoted
```csharp
var request = new SearchIssuesRequest()
{
  Labels = new[] {"With Space"},
};
SearchIssuesResult issues = await client.Search.SearchIssues(request);
```

This results in a search that looks like `label:With Space` which doesn't resolve, because the label needs to be quoted in the search string.

zHaytam:
Fixes #2044 

Hello,

I went with the easy way, double quote all labels. It will save us the check (whether the label has spaces or not) since having quotes on normal labels doesn't change anything.

I changed the test to include a test case `label 2` and tested it separately on this repository.


shiftkey:
release_notes: Search Issues endpoint to does not handles labels containing spaces correctly

PR #2085, opened 2020-02-06
move formatting code to optional flag

shiftkey:
I'm still digging into why `FormatCode` decides to insert some stray CR characters on Windows, but for now I'm making this optional rather than always running for local builds.


shiftkey:
release_notes: formatting code is now opt-in when running build scripts locally

PR #2089, opened 2020-02-09
upgrade project tooling to .NET Core 3.1

shiftkey:
- [x] Appveyor passes
 - [x] Actions passes
 - [x] tidy up about global tools
 - [x] review diff


shiftkey:
release_notes: upgraded project tooling to .NET Core 3.1

Directory.Build.targets
```diff
@@ -0,0 +1,9 @@
+<Project>
+  <PropertyGroup>
+    <EmbedUntrackedSources>true</EmbedUntrackedSources>
+    <TargetFrameworkMonikerAssemblyAttributesPath>$([System.IO.Path]::Combine('$(IntermediateOutputPath)','$(TargetFrameworkMoniker).AssemblyAttributes$(DefaultLanguageSourceExtension)'))</TargetFrameworkMonikerAssemblyAttributesPath>
```
casz: Instead of this, you can use https://www.nuget.org/packages/Microsoft.NETFramework.ReferenceAssemblies/
for compiling with mono.
https://stu.dev/dotnet-framework-reference-assemblies-via-nuget/

shiftkey: I'm already using `Microsoft.NETFramework.ReferenceAssemblies` in this project, and this workaround was due to an upgrade issue that I uncovered related to .NET Core 3.1 and Sourcelink. 

The details are over here, if you wanted to follow along: https://github.com/shiftkey/dotnetcore-sourcelink-test-bug/issues/1

casz: Neat 👍 
Thanks for sharing.

PR #2091, opened 2020-02-10
ensure datetime offset is encoded correctly for search APIs

### LINKED ISSUE #2056: Search requests need to correctly encode offset for date values
Test version 0.36.0

The repro snippet
```
            var endDate = new DateTimeOffset(2012, 12, 20, 16, 0, 0, TimeSpan.FromMinutes(8 * 60)); // offset +08:00
            var startDate = endDate.AddDays(-30);
            var request = new SearchRepositoriesRequest()
            {
                Updated = DateRange.Between(startDate, endDate)
            };
            var githubClient = new GitHubClient(new ProductHeaderValue("MyAmazingApp"));
            var repos = githubClient.Search.SearchRepo(request).Result;
            Console.WriteLine($"ReproCase: Find {repos.TotalCount} repos");
```

That is becoming to a http request using this url (from Fiddler)
`https://api.github.com/search/repositories?page=1&per_page=100&order=desc&q=pushed:2012-11-20T16:00:00+08:00..2012-12-20T16:00:00+08:00`

Note that "+" isn't escaped in this case and it only finds zero repositories.

Escaping "+" to "%2B" in a browser gets me total_count 98777

`https://api.github.com/search/repositories?page=1&per_page=100&order=desc&q=pushed:2012-11-20T16:00:00%2B08:00..2012-12-20T16:00:00%2B08:00`

I attach a .NET core project to demonstrate it.
[OctokitRepro.zip](https://github.com/octokit/octokit.net/files/4003519/OctokitRepro.zip)

shiftkey:
Fixes #2056

 - [x] CI passes
 - [x] diff makes sense

### DIFF
--- a/Octokit.Tests.Integration/Clients/SearchClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/SearchClientTests.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using Octokit;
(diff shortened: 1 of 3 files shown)


shiftkey:
release_notes: encodes offset part of `DateTimeOffset` when used in Search APIs

PR #2101, opened 2020-02-17
Add new events to the EventInfoState enum

maxim-lobanov:
Hello!
Recently, I have started to use this library and I have faced with issue that loading of events for some issues fail because of unrecognized events.
Unfortunately, I didn't find these events in the list of events in [GitHub API](https://developer.github.com/v3/issues/events/) page (it looks like it is pretty outdated) so I can't guarantee that my list is full. But I just added events that I have found in my repo so if you have the full list, please let me know

### DIFF
--- a/Octokit/Models/Response/EventInfo.cs
+++ b/Octokit/Models/Response/EventInfo.cs
@@ -193,6 +193,12 @@ namespace Octokit
         HeadRefForcePushed,
 
         /// <summary>
+        /// The pull request is ready for review
+        /// </summary>
+        [Parameter(Value = "ready_for_review")]
+        ReadyForReview,
+
+        /// <summary>
         /// The actor dismissed a review from the pull request.
         /// </summary>
         [Parameter(Value = "review_dismissed")]
@@ -308,6 +314,24 @@ namespace Octokit
         /// An issue was transferred.
         /// </summary>
         [Parameter(Value = "transferred")]
-        Transferred
+        Transferred,
+
+        /// <summary>
+        /// An issue was connected.
+        /// </summary>
+        [Parameter(Value = "connected")]
+        Connected,
+
+        /// <summary>
+        /// An issue was pinned.
+        /// </summary>
+        [Parameter(Value = "pinned")]
+        Pinned,
+
+        /// <summary>
+        /// An issue was unpinned.
+        /// </summary>
+        [Parameter(Value = "unpinned")]
+        Unpinned,
     }
 }


shiftkey:
release_notes:  Added new states to the `EventInfoState` enum

khellang:
> loading of events for some issues fail because of unrecognized events

Is there something that needs to be done to make sure this doesn't happen in the future, or does [`StringEnum<T>`](https://github.com/octokit/octokit.net/blob/master/docs/working-with-enums.md) already solve this?

shiftkey:
> or does [`StringEnum<T>`](https://github.com/octokit/octokit.net/blob/9b3cf309cf268916d54d73583869e65f95b6e627/docs/working-with-enums.md) already solve this, but wasn't used properly?

@khellang I suspect it was this, but @maxim-lobanov can provide more context

khellang:
Yeah, since the doc even [calls out `EventInfoState`](https://github.com/octokit/octokit.net/blob/9b3cf309cf268916d54d73583869e65f95b6e627/docs/working-with-enums.md#activitytimeline-apis-and-eventinfostate) explicitly as a good candidate for using `TryParse`, I think so as well.

It's great to keep enum values up-to-date though 😁👍

maxim-lobanov:
@khellang , @shiftkey 
The API response is parsed correctly to `StringEnum<EventInfoState>` even for unknown states and I believe that TryParse is not needed here.
But StringEnum has the following properties:
1) `StringValue` if we use this field, we can work even with unknown states and we will just compare them as strings (that is not very convenient)
2) `Value` that returns `TEnum` value. So if we try to get value from `StringEnum.Value` and `StringEnum` contains unknown value, exception will be raised.

image

Ideally, we should keep `EventInfoState` enum up-to-date with all new values but, actually, GitHub API docs are very outdated and it doesn't contain description of those fields that I have added in this PR. I have found these values by practice way - just run parsing events for all issues in my repo but, of course, it could be not full list.

khellang:
I know the properties, [I designed the type](https://github.com/octokit/octokit.net/pull/1595) 😉 

> I believe that TryParse is not needed here.

You basically have three options for consuming a `StringEnum<TEnum>`:

1. Use `StringValue` which is always safe, but as you noted, it's inconvenient.
1. Use `Value` which is unsafe (throws) for unknown values.
1. Use `TryParse`, which gives you back an enum value for known values, but still allows you to handle unknown values gracefully.

Personally, and especially for `EventInfoState`, I would always use `TryParse` to make sure my code doesn't start blowing up just because GitHub added a new event type 😄 

> GitHub API docs are very outdated and it doesn't contain description of those fields that I have added in this PR. I have found these values by practice way - just run parsing events for all issues in my repo but, of course, it could be not full list.

Exactly, that's what `StringEnum<T>.TryParse` is for. From [the docs](https://github.com/octokit/octokit.net/blob/9b3cf309cf268916d54d73583869e65f95b6e627/docs/working-with-enums.md):

> Since new functionality introduced on GitHub.com frequently leads to additional event types being received, Octokit was constantly playing catchup with these Enum values. Now, with `StringEnum<TEnum>` we have a solution to keep your code functioning, without needing to wait for a new Octokit release!

PR #2102, opened 2020-02-17
Add ProjectCard attribute to events model

maxim-lobanov:
Hello!
I have faced with issue that Event model doesn't contain `project_card` property that is important for all events that are related to GitHub boards (Projects).
For example, for event `added_to_project`, it contains the name of column where issue is added.
For event `moved_columns_in_project`, it contains the previous and new column names.
This PR includes the following changes:
1) Change `EventInfo` model to `IssueEvent` (please see note below for details)
2) Add model `IssueEventProjectCard` accordingly to [GitHub API description](https://developer.github.com/v3/issues/events/#attributes)
3) Add header `IssueEventsApiPreview = "application/vnd.github.starfox-preview"` that is needed to get `project_card` value
4) Update tests
5) Update `Octokit.Reactive` with the same changes

**Note:** Why `EventInfo` model was replaced by `IssueEvent` in `IssueClient`:
These models are pretty similar but `IssueEvent` contains additional fields like `Issue` and `ProjectCard`.
I am not sure why some methods in `IssueClient` used `IssueEvent` but some methods used `EventInfo`. Based on API, all methods in `IssueClient` should use `IssueEvent`.

### DIFF
--- a/Octokit.Tests/Clients/IssuesEventsClientTests.cs
+++ b/Octokit.Tests/Clients/IssuesEventsClientTests.cs
@@ -26,8 +26,7 @@ namespace Octokit.Tests.Clients
                 var client = new IssuesEventsClient(connection);
 
                 await client.GetAllForIssue("fake", "repo", 42);
-
-                connection.Received().GetAll<EventInfo>(Arg.Is<Uri>(u => u.ToString() == "repos/fake/repo/issues/42/events"), Args.ApiOptions);
+                connection.Received().GetAll<IssueEvent>(Arg.Is<Uri>(u => u.ToString() == "repos/fake/repo/issues/42/events"), null, Args.AnyAcceptHeaders, Args.ApiOptions);
             }
 
             [Fact]
@@ -38,7 +37,7 @@ namespace Octokit.Tests.Clients
 
                 await client.GetAllForIssue(1, 42);
 
-                connection.Received().GetAll<EventInfo>(Arg.Is<Uri>(u => u.ToString() == "repositories/1/issues/42/events"), Args.ApiOptions);
+                connection.Received().GetAll<IssueEvent>(Arg.Is<Uri>(u => u.ToString() == "repositories/1/issues/42/events"), null, Args.AnyAcceptHeaders, Args.ApiOptions);
             }
 
             [Fact]
@@ -56,7 +55,7 @@ namespace Octokit.Tests.Clients
 
                 await client.GetAllForIssue("fake", "repo", 42, options);
 
-                connection.Received().GetAll<EventInfo>(Arg.Is<Uri>(u => u.ToString() == "repos/fake/repo/issues/42/events"), options);
+                connection.Received().GetAll<IssueEvent>(Arg.Is<Uri>(u => u.ToString() == "repos/fake/repo/issues/42/events"), null, Args.AnyAcceptHeaders, options);
             }
 
             [Fact]
@@ -74,7 +73,7 @@ namespace Octokit.Tests.Clients
 
                 await client.GetAllForIssue(1, 42, options);
 
-                connection.Received().GetAll<EventInfo>(Arg.Is<Uri>(u => u.ToString() == "repositories/1/issues/42/events"), options);
+                connection.Received().GetAll<IssueEvent>(Arg.Is<Uri>(u => u.ToString() == "repositories/1/issues/42/events"), null, Args.AnyAcceptHeaders, options);
             }
 
             [Fact]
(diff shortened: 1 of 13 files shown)


maxim-lobanov:
@shiftkey , could you please take a look at this PR?
I have noticed that some test coverage checks failed due to new model that I have introduced for the project cards. Since it is just API models I am not sure what tests we need here so we should be okay with these changes.
If there is anything that should be improved, please let me know

shiftkey:
> I have noticed that some test coverage checks failed due to new model that I have introduced for the project cards. Since it is just API models I am not sure what tests we need here so we should be okay with these changes.

This is fine. I think I need to tweak the Codecov settings to focus on certain namespaces over others.

> I am not sure why some methods in `IssueClient` used `IssueEvent` but some methods used `EventInfo`. Based on API, all methods in `IssueClient` should use `IssueEvent`.

I am not sure about this either, and need to check the history to understand why these are different.

maxim-lobanov:
> I am not sure about this either, and need to check the history to understand why these are different.

@shiftkey ,
Unfortunately, nothing is in history. It was [introduced](https://github.com/octokit/octokit.net/commit/5ef086d44e8fde3e4fbd884e05a9f39059752ef2#diff-c50af1b3fa842efbd69894e011f2787b) with the first commit when `IssueEventClient` was implemented.

I believe that this change is safe enough because:
1) `EventInfo` model:
```
public long Id { get; protected set; }
public string NodeId { get; protected set; }
public string Url { get; protected set; }
public User Actor { get; protected set; }
public User Assignee { get; protected set; }
public Label Label { get; protected set; }
public StringEnum<EventInfoState> Event { get; protected set; }
public string CommitId { get; protected set; }
public DateTimeOffset CreatedAt { get; protected set; }
```
2) `IssueEvent` model:
```
public long Id { get; protected set; }
public string NodeId { get; protected set; }
public string Url { get; protected set; }
public User Actor { get; protected set; }
public User Assignee { get; protected set; }
public Label Label { get; protected set; }
public StringEnum<EventInfoState> Event { get; protected set; }
public string CommitId { get; protected set; }
public string CommitUrl { get; protected set; }
public DateTimeOffset CreatedAt { get; protected set; }
public Issue Issue { get; protected set; }
```
As you can see, `IssueEvent` contains all fields that `EventInfo` has.
But, also, it contains the following additional fields (which exist only for events that come with issues) :
1) `public string CommitUrl { get; protected set; }`
2) `public Issue Issue { get; protected set; }`
So we just extend model.

I have changed `EventInfo`-> `IssueEvent` for the method `GetAllForIssue`. [Based on GitHub API](https://developer.github.com/v3/issues/events/#list-events-for-an-issue), this method should return `Issue` field so we should use `IssueEvent` model here.

Octokit.Tests/Clients/IssuesEventsClientTests.cs
```diff
@@ -26,8 +26,7 @@ public async Task RequestsCorrectUrl()
                 var client = new IssuesEventsClient(connection);
 
                 await client.GetAllForIssue("fake", "repo", 42);
-
-                connection.Received().GetAll<EventInfo>(Arg.Is<Uri>(u => u.ToString() == "repos/fake/repo/issues/42/events"), Args.ApiOptions);
+                connection.Received().GetAll<IssueEvent>(Arg.Is<Uri>(u => u.ToString() == "repos/fake/repo/issues/42/events"), null, Args.AnyAcceptHeaders, Args.ApiOptions);
```
shiftkey: Can we assert that the preview URL is received here, rather than asserting that any header was found? Here's an example from elsewhere in the test suite:

https://github.com/octokit/octokit.net/blob/0074e76a8fdb00be22739bf65a1c4f48c43996e9/Octokit.Tests/Clients/DeploymentsClientTests.cs#L97-L101

We also include the preview header in the test rather than couple it to `AcceptHeader` to make preview headers more visible when reviewing.

Octokit/Models/Response/TimelineEventInfo.cs
```diff
@@ -40,9 +41,25 @@ public TimelineEventInfo(long id, string nodeId, string url, User actor, string
         public Label Label { get; protected set; }
         public User Assignee { get; protected set; }
         public Milestone Milestone { get; protected set; }
+
+        /// <summary>
```
shiftkey: :art: this is great - thanks for documenting both these current parts as well as the new property!

Octokit.Tests/Clients/IssuesEventsClientTests.cs
maxim-lobanov: Done! Thanks for the feedback

shiftkey:
release_notes: added `ProjectCard` property to `IssueEvent` response model

PR #2109, opened 2020-02-24
move dotnet-format global tool out of Lifetime

shiftkey:


shiftkey:
release_notes: Moved `dotnet-format` out to tool manifest file

PR #2114, opened 2020-02-26
Add Rocket and Eyes emoji reactions

ridhoq:
As per the [Reactions docs](https://developer.github.com/v3/reactions/#reaction-types), there's 2 reactions that are not available to ocktokit.net clients. They are added in this PR.

Let me know if there are any tests that need to be updated or added for this.


shiftkey:
@ridhoq thanks for opening this up - we've got decent coverage here already but I added a commit to update a test to use one of the new enums, just to be sure. All working good here!

shiftkey:
release_notes: Added :rocket: and :eyes: emoji to `Reaction` type

PR #2116, opened 2020-03-02
Update OAuth Token operations to new APIs

MGudgin:
Per ['Deprecating OAuth Application API'](https://developer.github.com/changes/2020-02-14-deprecating-oauth-app-endpoint/)
the HTTP API endpoints called by CheckApplicationAuthentication,
ResetApplicationAuthentication and RevokeApplicationAuthentication are
being deprecated.

This PR updates those APIs to call the new HTTP API endpoints as
documented at the above link.

# Details

Amend CheckApplicationAuthentication, ResetApplicationAuthentication and
RevokeApplicationAuthentication to create an object containing the OAuth
access token and to call the single arg version of
ApiUrls.ApplicationAuthorization. The object is used as the request
body.

Amend CheckApplicationAuthentication to use POST.

Amend ResetApplicationAuthentication to use PATCH.

Remove the two arg version of ApiUrls.ApplicationAuthorization as it is
no longer called. Amend the single arg version to use the new API path.

Amend unit tests to account for the above changes.

### DIFF
--- a/Octokit.Tests/Clients/AuthorizationsClientTests.cs
+++ b/Octokit.Tests/Clients/AuthorizationsClientTests.cs
@@ -292,7 +292,7 @@ namespace Octokit.Tests.Clients
 
                 client.Received().Post<ApplicationAuthorization>(
                     Arg.Is<Uri>(u => u.ToString() == "applications/clientId/token"),
-                    Args.Object);
+                    Arg.Is<Object>(o => o.GetType().GetProperty("access_token").GetValue(o).ToString() == "accessToken"));
             }
 
             [Fact]
@@ -320,7 +320,7 @@ namespace Octokit.Tests.Clients
 
                 client.Received().Patch<ApplicationAuthorization>(
                     Arg.Is<Uri>(u => u.ToString() == "applications/clientId/token"),
-                    Args.Object);
+                    Arg.Is<Object>(o => o.GetType().GetProperty("access_token").GetValue(o).ToString() == "accessToken"));
             }
 
             [Fact]
@@ -348,7 +348,7 @@ namespace Octokit.Tests.Clients
 
                 client.Received().Delete(
                     Arg.Is<Uri>(u => u.ToString() == "applications/clientId/token"),
-                    Args.Object);
+                    Arg.Is<Object>(o => o.GetType().GetProperty("access_token").GetValue(o).ToString() == "accessToken"));
             }
 
             [Fact]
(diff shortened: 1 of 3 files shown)


Octokit.Tests/Clients/AuthorizationsClientTests.cs
```diff
@@ -290,9 +290,9 @@ public async Task ChecksApplicationAuthenticateAtCorrectUrl()
 
                 authEndpoint.CheckApplicationAuthentication("clientId", "accessToken");
 
-                client.Received().Get<ApplicationAuthorization>(
-                    Arg.Is<Uri>(u => u.ToString() == "applications/clientId/tokens/accessToken"),
-                    null);
+                client.Received().Post<ApplicationAuthorization>(
+                    Arg.Is<Uri>(u => u.ToString() == "applications/clientId/token"),
+                    Args.Object);
```
shiftkey: Can we assert that the object has the expected `access_token` parameter here? I think we have some other examples of this floating around (I'm not sure whether it can be `AccessToken` and is serialized later to `access_token`, or must be `access_token`) but it'd be great to verify the behaviour...

shiftkey: Refreshed my memory a bit and it turns out that we don't do any conversion, so you're right to use `access_token` here:

https://github.com/octokit/octokit.net/blob/9b3cf309cf268916d54d73583869e65f95b6e627/Octokit/Clients/AuthorizationsClient.cs#L99-L105

The use of anonymous types however makes testing a bit of a pain. I was able to put together this one-liner to verify it, but I think we can do better on a follow-up PR. Thoughts?

```suggestion
                    Arg.Is<Object>(u => u.GetType().GetProperty("access_token").GetValue(u).ToString() == "accessToken"));
```

MGudgin: Yeah, was just going down that rabbit hole. I think to do better we would need a concrete type for the structure that has the access_token field. Happy to add that in a follow up PR. If we do add a concreate type, we should probably make a similar change for other cases where we just use an unadorned Args.Object in the test case. (e.g. TheGetOrCreateApplicationAuthenticationMethod.GetsOrCreatesAuthenticationAtCorrectUrl)

shiftkey: Because these are internal types (rather than ones the user creates in code) I'm fine with leaving them as-is. We don't have great test coverage over the parameters sent in `Post`/`Patch`/`Put` but I'm open to improvements here.

shiftkey:
Thanks for this @MGudgin!

Feel free to open other issues or PRs if you wanted to improve this area.

shiftkey:
release_notes: In preparation for the [Deprecating OAuth Application API](https://developer.github.com/changes/2020-02-14-deprecating-oauth-app-endpoint/) brownout and eventual removal on July 1 2020 we have updated the client internals to ensure we are calling the supported endpoint well in advance

PR #2121, opened 2020-03-03
Obsolete UnixTimestampExtensions methods

0xced:
* The equivalent of `public static DateTimeOffset FromUnixTime(this long unixTime)` exists in the framework since .NET Framework 4.6: https://docs.microsoft.com/en-us/dotnet/api/system.datetimeoffset.fromunixtimeseconds
* The equivalent of `public static long ToUnixTime(this DateTimeOffset date)` exists in the framework since .NET Framework 4.6: https://docs.microsoft.com/en-us/dotnet/api/system.datetimeoffset.tounixtimeseconds


shiftkey:
advisories: The `UnixTimestampExtensions.ToUnixTime` methods are marked obsolete because there are equivalent APIs available in .NET Framework 4.6. As these are also intended for internal usage in .NET and should not be imported into the library code, these will be removed in a later update.

shiftkey:
release_notes: Replaced old `UnixTimestampExtensions.ToUnixTime` extension methods with inbuild equivalents now that we target .NET Framework 4.6

Issue #2122, opened 2020-03-04
Review request of teams is not supported


ch1seL:
~~**Pull request** entity of GitHub API contains **requested_teams** property with an array of team slugs that will be requested but it missed in **Octokit.PullRequest** entity~~
Also, **Review Requests** entity contains similar **teams** property but **Octokit.IPullRequestReviewRequestsClient** gets the list of requested users only without teams

nrukavkov:
+1

EugeneMurS:
+

IvanQL:
+1

lady-guga:
+1

nkirsanov:
+1

AlexandraKim:
+1

irina-litvinova:
+1

beylkhanovdamir:
+1

shiftkey:
> Also, Review Requests entity contains similar teams property but Octokit.IPullRequestReviewRequestsClient gets the list of requested users only without teams

I think this the APIs as currently implemented are not designed to match the response returned from the GitHub API:

```
GET /repos/:owner/:repo/pulls/:pull_number/requested_reviewers

{
  "users": [
    {
      "login": "octocat",
      "id": 1,
      "node_id": "MDQ6VXNlcjE=",
      "avatar_url": "https://github.com/images/error/octocat_happy.gif",
      "gravatar_id": "",
      "url": "https://api.github.com/users/octocat",
      "html_url": "https://github.com/octocat",
      "followers_url": "https://api.github.com/users/octocat/followers",
      "following_url": "https://api.github.com/users/octocat/following{/other_user}",
      "gists_url": "https://api.github.com/users/octocat/gists{/gist_id}",
      "starred_url": "https://api.github.com/users/octocat/starred{/owner}{/repo}",
      "subscriptions_url": "https://api.github.com/users/octocat/subscriptions",
      "organizations_url": "https://api.github.com/users/octocat/orgs",
      "repos_url": "https://api.github.com/users/octocat/repos",
      "events_url": "https://api.github.com/users/octocat/events{/privacy}",
      "received_events_url": "https://api.github.com/users/octocat/received_events",
      "type": "User",
      "site_admin": false
    }
  ],
  "teams": [
    {
      "id": 1,
      "node_id": "MDQ6VGVhbTE=",
      "url": "https://api.github.com/teams/1",
      "html_url": "https://api.github.com/teams/justice-league",
      "name": "Justice League",
[...]

PR #2123, opened 2020-03-04
added Octokit.PullRequest.RequestedTeams property

ch1seL:
Related to #2122

### DIFF
--- a/Octokit/Models/Response/PullRequest.cs
+++ b/Octokit/Models/Response/PullRequest.cs
@@ -16,7 +16,7 @@ namespace Octokit
             Number = number;
         }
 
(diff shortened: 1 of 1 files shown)


Octokit/Models/Response/PullRequest.cs
```diff
@@ -242,6 +243,11 @@ public bool Merged
         /// Users requested for review
         /// </summary>
         public IReadOnlyList<User> RequestedReviewers { get; protected set; }
+        
+        /// <summary>
+        /// Teams requested for review
+        /// </summary>
+        public IReadOnlyList<Team> RequestedTeams { get; protected set; }
```
shiftkey: Given we have `RequestedReviewers` above that is configured in the same way, I think this is all we need to do.

Is there any other feedback you're looking for @ch1seL?

ch1seL: This is half the success of issue #2122
I think we can merge it but the issue will not be resolved totally

shiftkey: Now I understand. I'll merge this and add some more context to #2122 about the second item on the list once I know what needs to change.

shiftkey:
release_notes: added `RequestedTeams` property to `PullRequest` response model

PR #2124, opened 2020-03-05
add metadata to each client action

shiftkey:
This PR adds attributes to each client action to provide context about where each client method is sourced from. 

This comes with two main benefits:

 - by including the verb and path for each route, this will help me audit what parts of the API are not implemented currently
 - using `ManualRouteAttribute` will help me to track which code was hand-crafted, and hopefully which ones are generated by tooling in the future

I also noticed some places where things are likely broken, and some long-obsolete APIs, and I might need to make changes to this metadata down the track, but I can follow-up with other PRs later once I've given this another look.

  - [x] review diff
  - [x] extract issues for follow-up - #2125 #2126

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,5 +4,8 @@
         "**/obj": true,
         "**/packaging": true,
         "**/coverage-results": true
-    }
+    },
+    "files.trimTrailingWhitespace": true,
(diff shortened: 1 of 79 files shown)


shiftkey:
release_notes: add metadata to each client to assist with auditing API coverage

PR #2128, opened 2020-03-05
add attribute for tracking preview usage in the codebase

shiftkey:
Review the codebase for each of the previews from https://developer.github.com/v3/previews/ :

 - add attribute usage for each place in the codebase
 - review usage matches the documentation
 
OR

 - open an issue to track supporting the preview API

Current preview support:

 - [x] Migrations
 - [x] Enhanced deployments
 - [x] Reactions
 - [x] Timeline
 - [x] Integrations
 - [x] Projects
 - ~[ ] Commit search~ #2132
 - ~[ ] Community profile metrics~ #2133
 - ~[ ] User blocking~ #2134
 - ~[ ] Repository topics~ #1707 #1720
 - ~[ ] Codes of conduct~ #2135
 - ~[ ] Add lock reason~ #2136
 - ~[ ] Require signed commits~ #2137
 - [x] Require multiple approving reviews
 - [x] Check runs and check suites API
 - [x] Project card details
 - ~[ ] GitHub App Manifests~ #2138
 - [x] Deployment statuses
 - ~[ ] Repository creation permissions~ #2139
 - ~[ ] Content attachments~ #2140
 - ~[ ] Interaction restrictions for repositories and organizations~ #2141
 - [x] Draft pull requests
 - ~[ ] Enable and disable Pages~ #2142
 - ~[ ] List branches or pull requests for a commit~ #2143
 - ~[ ] Uninstall a GitHub App~ #2144
 - ~[ ] Enable or disable vulnerability alerts for a repository~ #2145
 - ~[ ] Update a pull request branch~ #2146
 - ~[ ] Enable or disable automated security fixes~ #2147
 - ~[ ] Create and use repository templates~ #2148
 - [x] New OAuth Applications API endpoints
 - ~[ ] New visibility parameter for the Repositories API~ #2149

PR #2131, opened 2020-03-06
move GitHubAppInstallationsClient into root namespace

shiftkey:
A follow-up from #2128 for a case that wasn't caught by our conventions, because it was created in the `Octokit.Clients` namespace rather than the root `Octokit` namespace.

This updates the namespace and adds the necessary route attributes to get the build passing again.


shiftkey:
release_notes: move `GitHubAppInstallationsClient` into root namespace

PR #2150, opened 2020-03-08
add support for enumerating users

### LINKED ISSUE #1915: User.GetAll endpoint
Hi,

I am working on a small project that does some statistics from GitHub data. I am so glad to find this library, which really helps me a lot. However I encountered some problems.

One thing I am trying to achieve is to sample some random users from GitHub, so I need an API to get a list of users. I know there are millions of users on GitHub and getting a full list in one API call is impossible, but I thing even a direct wrapper for the [original all users API](https://developer.github.com/v3/users/#get-all-users) would help.

From the implemented API and issue list I couldn't find such feature. Please let me know if I missed something. There is already a public repositories API so I believe there's no reason we can't add it. If you think it's OK to add it, I am very happy to spend some time to try to work on it and send a PR. 

Thanks,
Zhenwei

kentahikaru:
Resolves #1915

 - [x] Implemented User.GetAll()
 - [ ] Tests -- will make in next commit

### DIFF
--- a/Octokit.Reactive/Clients/IObservableUsersClient.cs
+++ b/Octokit.Reactive/Clients/IObservableUsersClient.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Diagnostics.CodeAnalysis;
-using System.Collections.Generic;
 
 namespace Octokit.Reactive
 {
@@ -34,7 +33,7 @@ namespace Octokit.Reactive
         /// </summary>
         /// <param name="since">ID after which the returned list begin</param>
         /// <returns>A list of <see cref="User"/> after specified ID</returns>
-        IObservable<List<User>> GetAll(string since);
+        IObservable<IReadOnlyPagedCollection<User>> GetAll(string since);
 
         /// <summary>
         /// A client for GitHub's User Followers API
--- a/Octokit.Reactive/Clients/IObservableUsersClient.cs
+++ b/Octokit.Reactive/Clients/IObservableUsersClient.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Diagnostics.CodeAnalysis;
+using System.Collections.Generic;
 
 namespace Octokit.Reactive
 {
@@ -27,6 +28,13 @@ namespace Octokit.Reactive
         /// <exception cref="AuthorizationException">Thrown if the client is not authenticated.</exception>
         /// <returns>A <see cref="User"/></returns>
         IObservable<User> Update(UserUpdate user);
+
+        /// <summary>
+        /// Returns list of all <see cref="User"/>, in the order that they signed up on GitHub after specified ID.
+        /// </summary>
+        /// <param name="since">ID after which the returned list begin</param>
+        /// <returns>A list of <see cref="User"/> after specified ID</returns>
+        IObservable<List<User>> GetAll(string since);
 
         /// <summary>
         /// A client for GitHub's User Followers API
--- a/Octokit.Reactive/Clients/IObservableUsersClient.cs
+++ b/Octokit.Reactive/Clients/IObservableUsersClient.cs
@@ -33,7 +33,7 @@ namespace Octokit.Reactive
         /// </summary>
         /// <param name="since">ID after which the returned list begin</param>
         /// <returns>A list of <see cref="User"/> after specified ID</returns>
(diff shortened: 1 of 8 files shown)


kentahikaru:
Where are tests for ObservableUsersClient ? Cant' seem to find them :(

shiftkey:
> Where are tests for ObservableUsersClient ? Cant' seem to find them :(

@kentahikaru it may be that this file doesn't exist - feel free to take a shot at creating it using the others as a reference.

kentahikaru:
I created ObservableUsersClientTests.cs but Test Explorer or CI Build doesn't seem to "see" it.
Please advise. What am i missing?

kentahikaru:
> I created ObservableUsersClientTests.cs but Test Explorer or CI Build doesn't seem to "see" it.
> Please advise. What am i missing?

Hmm, maybe the reference is missing in project file. I will check in the evening.

kentahikaru:
As none of project files has file list, this doesn't seems to be the case.
So i have no clue. Please help.

kentahikaru:
Now what happens ? :)

shiftkey:
> Now what happens ? :)

I need to find time to review, which I'm trying to carve out this long weekend...

Octokit.Reactive/Clients/IObservableUsersClient.cs
```diff
@@ -28,6 +28,13 @@ public interface IObservableUsersClient
         /// <returns>A <see cref="User"/></returns>
         IObservable<User> Update(UserUpdate user);
 
+        /// <summary>
+        /// Returns list of all <see cref="User"/>, in the order that they signed up on GitHub after specified ID.
+        /// </summary>
+        /// <param name="since">ID after which the returned list begin</param>
+        /// <returns>A list of <see cref="User"/> after specified ID</returns>
+        IObservable<IReadOnlyPagedCollection<User>> GetAll(string since);
```
shiftkey: This is not really important for the discussion but we generally "flatten" observables to not return lists, so I pushed a commit to change this to return `IObservable<User>` as that will be more familiar to callers of the API.

Octokit/Clients/IUsersClient.cs
```diff
@@ -54,6 +54,13 @@ public interface IUsersClient
         /// <returns>A <see cref="User"/></returns>
         Task<User> Update(UserUpdate user);
 
+        /// <summary>
+        /// Returns list of all <see cref="User"/>, in the order that they signed up on GitHub after specified ID.
+        /// </summary>
+        /// <param name="since">ID after which the returned list begin</param>
+        /// <returns>A list of <see cref="User"/> after specified ID</returns>
+        Task<IReadOnlyPagedCollection<User>> GetAll(string since);
```
shiftkey: This is almost correct - we have `IReadOnlyList<T>` here which is just what we need for this situation. I've updated this and pushed a commit.

Octokit.Tests.Integration/Clients/UsersClientTests.cs
```diff
@@ -98,4 +98,17 @@ public async Task FailsWhenAuthenticatedWithBadCredentials()
...
+    public class TheGetAllMethod
+    {
+        [IntegrationTest]
+        public async Task ReturnsListOfdUsers()
+        {
+            var github = Helper.GetAuthenticatedClient();
+
+            var users = await github.User.GetAll("1");
+
+            Assert.Equal(30, users.Count);
```
shiftkey, line 111: So while the other tests are looking good this one is the test I was waiting to dig into.

Because of how it's currently written, this API is only going to return the first page, with no options available for following the pagination. This doesn't really feel like a `GetAll` API, right? But we want to avoid this running forever, enumerating the millions of GitHub accounts, so how can we find this middle ground?

We have our existing pagination support so we could write the test like this:

```C#
[IntegrationTest]
public async Task ReturnsListOfUsers()
{
    var options = new ApiOptions
    {
        PageSize = 10,
        PageCount = 2,
    };

    var users = await github.User.GetAll("1", options);

    Assert.Equal(20, users.Count);
}
```


However this API doesn't provide a `page=` parameter so if we used that it would just continue forever without returning any records to the user. We'd need to provide custom pagination logic for this case, as well as making more strict what the user can pass in with `ApiOptions`.

PR #2152, opened 2020-03-13
Mark ReleasesClientTests.TheCreateReleasesMethod tests as integration

0xced:
They were using [Fact] instead of [IntegrationTest] so they would run and fail if no credentials are supplied.


shiftkey:
release_notes: mark `ReleasesClientTests.TheCreateReleasesMethod` tests as integration

PR #2154, opened 2020-03-15
add convention test for preview APIs

shiftkey:
This PR reviews and marks the preview headers that are no longer currently documented on the [GitHub Developer site](https://developer.github.com/v3/).

 - [x] add test
 - [x] mark headers that are now out of scope
 - [x] cleanup APIs that have graduated


shiftkey:
release_notes: Clean up preview accept headers which are no longer documented or have graduated to production.

shiftkey:
advisory: This release stops sending preview accept headers which are no longer mentioned in the developer documentation. These features should be working as expected if the service no longer looks for the preview header, but if you spot a situation where the code is not returning the expected response as it was in previous releases please open an issue so we can investigate.

azboyzranch (commented):
>.

PR #2157, opened 2020-03-18
first pass at cleaning up attributes

shiftkey:
This builds upon #2124 now that I've given it a first pass at auditing using the `octokit/routes` schema

 - [x] review diff


shiftkey:
release_notes: Clean up preview accept headers which are no longer documented or have graduated to production.

PR #2158, opened 2020-03-19
more cleanup of route metadata

shiftkey:


shiftkey:
release_notes: Tidy up route metadata attributes based on schema information

PR #2164, opened 2020-03-31
remove all usages of TaskFactory in test suite

shiftkey:
`Task.Factory.StartNew` is from a time before we had the ability to wrap a result in a `Task`, and as part of simplifying tests to be able to refactor the internals I decided to drop these usages.

You may also note the `TestSetup.GetApiResponse` helper function - that helps eliminate repetitive code and make the tests more readable.


shiftkey:
release_notes: replace `Task.Factory.StartNew` usage in tests with `Task.FromResult` to simplify test setup

PR #2169, opened 2020-04-05
Replace emojis with Unicode characters

NextTurn:
They don't render on https://octokitnet.readthedocs.io/en/latest/github-apps/


shiftkey:
release_notes: Updated GitHub Apps documentation to render emoji correctly

PR #2170, opened 2020-04-05
Add support for ReleaseEvent payload

NextTurn:
Contributes to #1646


shiftkey:
release_notes: Added `ReleaseEvent` payload

Issue #2132, opened 2020-03-06
add support for Commit Search preview


shiftkey:
https://developer.github.com/v3/search/#search-commits

leavelbw:
Didn't realize that @kentahikaru was working on this.  I just completed this task.

PR #2155, opened 2020-03-16
[WIP] search/commit , issue #2132

kentahikaru:
Initial commit, first baby steps

PR #2171, opened 2020-04-05
[WIP]Adds the ability to search commits

leavelbw:
Addresses #2132

Please let me know if you agree with these changes.  If you agree with the changes, I will add the unit tests.  Thanks.


shiftkey:
@leavelbw thanks for opening this PR. We also have #2155 so I need to figure out how to reconcile both PRs.

Octokit/Helpers/StringExtensions.cs
```diff
@@ -19,6 +19,16 @@ public static bool IsNotBlank(this string value)
         {
             return !string.IsNullOrWhiteSpace(value);
         }
+        
+        public static bool IsNotDefault(this DateTime value)
```
shiftkey, line 23: Can we move these new methods out to a separate file? They're not exactly string-based, but I'm fine with a single file for these "other" types...

Octokit/Clients/SearchClient.cs
```diff
@@ -83,5 +83,18 @@ public Task<SearchLabelsResult> SearchLabels(SearchLabelsRequest search)
...
+        /// search commits
+        /// https://developer.github.com/v3/search/#search-commits
+        /// </summary>
+        /// <param name="search"></param>
+        /// <returns>List of commits</returns>
+        [ManualRoute("GET", "/search/commits")]
+        public Task<SearchCommitsResult> SearchCommits(SearchCommitsRequest search)
+        {
+            Ensure.ArgumentNotNull(search, nameof(search));
+            return ApiConnection.Get<SearchCommitsResult>(ApiUrls.SearchCommits(), search.Parameters, @"application/vnd.github.cloak-preview");
```
shiftkey, line 97: The `application/vnd.github.cloak-preview` header should be in `Octokit/Helpers/AcceptHeaders.cs`.

Are you able to do these two things to be consistent with the rest of the codebase:

 - add a const for this header in `Octokit/Helpers/AcceptHeaders.cs`
 - add this as a valid preview header in the convention test so the build doesn't fail

https://github.com/octokit/octokit.net/blob/e41b2d19d74d84e5b5222ebce58d64fc308b1c2e/Octokit.Tests.Conventions/PreviewsTests.cs#L11-L21

Octokit/Models/Request/SearchCommitsRequest.cs
```diff
@@ -0,0 +1,291 @@
...
+        /// </remarks>
+        public CommitsSearchSort? SortField { get; set; }
+        public override string Sort
+        {
+            get { return SortField.ToParameter(); }
+        }
+
+        private string QueryDateTime(DateTime input)
+        {
+            return input.ToString("yyyy-MM-DD");
```
shiftkey, line 65: 🤔 I need to have another look at places where we need to handle date inputs, and see if this is a chance for us to standardize on date input.

Octokit/Helpers/StringExtensions.cs
leavelbw: I created 2 new files (DateTimeExtensions, and BooleanExtensions.)

Octokit/Clients/SearchClient.cs
leavelbw: I created the constant.  I also created a new array for non standard preview headers in the test since this header does not have "+json" at the end.

Octokit/Models/Request/SearchCommitsRequest.cs
leavelbw: @shiftkey have you decided on this?

leavelbw: @shiftkey Any update?

Issue #2161, opened 2020-03-26
GetLastApiInfo() returns only zero's after a call?


StefH:
Using Octokit version `0.32.0` to `0.40.0`, the `GetLastApiInfo()` returns 0 values ?

For an example project : see https://github.com/StefH/OctokitExample/tree/master/src

See also https://github.com/octokit/octokit.net/issues/1711

shiftkey:
@StefH thanks for the report. I've isolated this to how headers are returned when an unauthenticated request is made.

This tests shows how the headers for rate-limiting information differ:


```
$ curl -s -o /dev/null -D -  https://api.github.com/rate_limit | grep "X-Rate"
X-Ratelimit-Limit: 60
X-Ratelimit-Remaining: 60
X-Ratelimit-Reset: 1585235264
$ curl -s -o /dev/null -H "Authorization: token $OCTOKIT_OAUTHTOKEN" -D - https://api.github.com/rate_limit | grep "X-Rate"
X-RateLimit-Limit: 5000
X-RateLimit-Remaining: 5000
X-RateLimit-Reset: 1585235267
Access-Control-Expose-Headers: ETag, Link, Location, Retry-After, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval, X-GitHub-Media-Type, Deprecation, Sunset
```

Note the `Ratelimit` returned from the first command - the lower-case `l` here is why we are unable to parse the numbers from the header.

As a workaround, I recommend setting the `Credentials` property on your `GitHubClient` to perform an authenticated request. This will also give you a larger number of API calls to work with - unauthenticated requests are limited to 60 requests per hour.

I need to investigate the root cause of this on our end, and will close this out when it's been resolved.

StefH:
Ok. This seems like a bug in the API from GitHub...

I currently use octokit library in 
https://github.com/StefH/GitHubReleaseNotes

This is currently just using the free unauthenticated API.

Is it a good idea to include an authenticated call in an public application which is used by multiple people at the same time?
How is the rate limit calculated? Is this 5000 per system/user or 5000 in total?


An another request: if possible retarget .net framework 4.5.2 in octokit because this is missing from the latest?

shiftkey:
> How is the rate limit calculated? Is this 5000 per system/user or 5000 in total?

[...]

PR #2175, opened 2020-04-09
handle case insensitive headers when parsing for API rate limiting

shiftkey:
Related to #2161 but probably not the full fix, given we pass around a dictionary and lookup by key, rather than following the HTTP spec and supporting case insensitive headers.

I chatted with some others inside GitHub and they reminded me about [RFC 2616, section 4.2](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2):

> Each header field consists of a name followed by a colon (":") and the field value. **Field names are case-insensitive.**

I'd love to get to a place where we can leverage `HttpResponseHeaders` but we are very tightly coupled to `IResponse` in the internals, and it's this component that manages the headers as a dictionary:

https://github.com/octokit/octokit.net/blob/a25863f2289050fdfb091ba1ff2808cd1218d7e1/Octokit/Http/IResponse.cs#L32-L35

The setup is here, which converts the `HttpResponseHeaders` into a dictionary:

https://github.com/octokit/octokit.net/blob/a25863f2289050fdfb091ba1ff2808cd1218d7e1/Octokit/Http/HttpClientAdapter.cs#L125-L129

You'll also note that headers can have multiple values, so the eventual refactoring would need to support that. 

 - [x] identify other places where we `TryGetValue` without handling case-insensitive headers

### DIFF
--- a/Octokit.Tests/Clients/Enterprise/EnterpriseProbeTests.cs
+++ b/Octokit.Tests/Clients/Enterprise/EnterpriseProbeTests.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Net;
 using System.Net.Http;
 using System.Reactive.Linq;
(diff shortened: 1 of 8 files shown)


shiftkey:
release_notes: fixed parsing of response headers in `GitHubClient.GetLastApiInfo()`

PR #2176, opened 2020-04-09
Add support for querying the Code of Conduct API

neoncitylights:
While I've contributed to other repositories in the past, this is my first open-source contribution to another C# repository! I'm currently working on adding integration tests.

Design question: Should there be an enum for the Code of Conduct key? Currently there are two officially registered conducts with GitHub, which is the Citizen Conduct and Contributor Covenant.


neoncitylights:
Thanks for the feedback! I've adjusted accordingly. I also named it `CodesOfConductClient` instead of `CodeOfConductClient` (Code is singular instead of plural), since that's how GitHub officially refers it to: (https://developer.github.com/v3/codes_of_conduct/)

Octokit/Clients/CodesOfConductClient.cs
```diff
@@ -0,0 +1,64 @@
...
+        { }
+
+        /// <summary>
+        /// Gets all code of conducts on GitHub.
+        /// </summary>
+        /// <remarks>See the <a href="https://developer.github.com/v3/codes_of_conduct/#list-all-codes-of-conduct">API documentation</a> for more information.</remarks>
+        /// <returns>A <see cref="IReadOnlyList{CodeOfConduct}"/> on GitHub.</returns>
+        [Preview("scarlet-witch")]
+        [ManualRoute("GET", "/codes_of_conduct")]
+        public Task<IReadOnlyList<CodeOfConduct>> GetAll()
```
shiftkey, line 29: Because this API doesn't support pagination details, you can add this attribute to get one of the convention tests passing:

```suggestion
        [ExcludeFromPaginationApiOptionsConventionTest("Pagination not supported by GitHub API (tested 14/04/2020)")]
        public Task<IReadOnlyList<CodeOfConduct>> GetAll()
```

Octokit/Clients/CodesOfConductClient.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        /// <summary>
+        /// Gets an individual code of conduct.
+        /// </summary>
+        /// <remarks>See the <a href="https://developer.github.com/v3/codes_of_conduct/#get-an-individual-code-of-conduct">API documentation</a> for more information.</remarks>
+        /// <param name="key">The unique key for the Code of Conduct</param>
+        /// <returns>A <see cref="CodeOfConduct"/> that includes the code of conduct key, name, and API/HTML URL.</returns>
+        [Preview("scarlet-witch")]
+        [ManualRoute("GET", "/codes_of_conduct/{key}")]
+        public Task<CodeOfConduct> GetCodeOfConduct(CodeOfConductType key)
```
shiftkey, line 42: This could just be `Get` to keep the API as concise as possible:

```suggestion
        public Task<CodeOfConduct> Get(CodeOfConductType key)
```

Octokit/Clients/ICodesOfConductClient.cs
```diff
@@ -0,0 +1,58 @@
...
+
+namespace Octokit
+{
+    /// <summary>
+    /// GitHub client for the Codes of Conduct API.
+    /// </summary>
+    /// <remarks>
+    /// See the <a href="https://developer.github.com/v3/codes_of_conduct/">API documentation</a> for more information.
+    /// </remarks>
+    public interface ICodesOfConductClient
```
shiftkey, line 14: The other build error is the missing corresponding client for our reactive API. It's failing because it expects`Octokit.Reactive/IObservableGitHubClient.cs` to have a similar API, which internally just calls your original API. 

Have a look at `Octokit.Reactive/Clients/ObservableMiscellaneousClient.cs` for an example of how this works, and it looks like you've found the right bits to add that it should look familiar.

neoncitylights:
I'm currently working on adding tests now, I'll submit once I have them all passing locally. 👍

PR #2182, opened 2020-04-26
Update Repository to include WatchersCount

### LINKED ISSUE #2181: Subscribers Count not working
Seems like SubscribersCount from Repository is not working correctly?

KarolGrzesiak:
Resolves #2181


shiftkey:
@KarolGrzesiak thanks for opening this up!

Can we mark `SubscribersCount` as obsolete while we're making this change?

```c#
        [Obsolete("Update your code to use WatchersCount as this field will stop containing data in the future")]
        public int SubscribersCount { get; protected set; }
```

shiftkey:
release_notes: Add WatchersCount to Repository response model

PR #2187, opened 2020-05-04
add support for repository templates API

### LINKED ISSUE #2148: preview support for repository templates
- https://developer.github.com/v3/repos/#create-a-repository-using-a-template
 - https://developer.github.com/v3/repos/#create-a-repository-for-the-authenticated-user
 - https://developer.github.com/v3/repos/#get-a-repository

This preview flag was not added in #2090 so repositories returned from the API do not have the `is_template` field set.

FrediKats:
Fix #2148 

Add IsTemplate, add header, add method from https://developer.github.com/v3/repos/#create-a-repository-using-a-template

### DIFF
--- a/Octokit/Clients/RepositoriesClient.cs
+++ b/Octokit/Clients/RepositoriesClient.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.ObjectModel;
 using System.Linq;
 using System.Threading.Tasks;
@@ -50,6 +50,7 @@ namespace Octokit
         /// <param name="newRepository">A <see cref="NewRepository"/> instance describing the new repository to create</param>
         /// <exception cref="ApiException">Thrown when a general API error occurs.</exception>
         /// <returns>A <see cref="Repository"/> instance for the created repository.</returns>
+        [Preview(AcceptHeaders.TemplateRepositoryPreviewCodeName)]
         [ManualRoute("POST", "/user/repos")]
         public Task<Repository> Create(NewRepository newRepository)
         {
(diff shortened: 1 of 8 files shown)


Octokit/Clients/RepositoriesClient.cs
```diff
@@ -246,6 +261,7 @@ public Task<Repository> Edit(long repositoryId, RepositoryUpdate update)
         /// <param name="name">The name of the repository</param>
         /// <exception cref="ApiException">Thrown when a general API error occurs.</exception>
         /// <returns>A <see cref="Repository"/></returns>
+        [Preview(AcceptHeaders.TemplateRepositoryPreviewCodeName)]
```
shiftkey, line 264: In addition to this link, you need to pass in the accepts header to ensure the `Repository` response returns the additional field:

```c#
[Preview(AcceptHeaders.TemplateRepositoryPreviewCodeName)]
[ManualRoute("GET", "/repositories/{id}")]
public Task<Repository> Get(long repositoryId)
{
     return ApiConnection.Get<Repository>(ApiUrls.Repository(repositoryId), null, AcceptHeaders.TemplateRepositoryPreview);
}
```

This will also require the corresponding tests to be updated.

Octokit/Clients/RepositoriesClient.cs
```diff
@@ -79,11 +81,24 @@ public Task<Repository> Create(string organizationLogin, NewRepository newReposi
...
+        [ManualRoute("POST", "/repos/{owner}/{repo}/generate")]
+        public Task<Repository> Generate(string owner, string repo, NewRepositoryFromTemplate newRepository)
+        {
+            Ensure.ArgumentNotNull(owner, nameof(owner));
+            Ensure.ArgumentNotNull(repo, nameof(repo));
+            Ensure.ArgumentNotNull(newRepository, nameof(newRepository));
+            if (string.IsNullOrEmpty(newRepository.Name))
+                throw new ArgumentException("The new repository's name must not be null.");
+
+            return ApiConnection.Post<Repository>(ApiUrls.Repositories(owner, repo), newRepository, AcceptHeaders.TemplateRepositoryPreview);
```
shiftkey, line 94: You have some failing tests in `Octokit.Tests/Clients/RepositoriesClientTests.cs` because the assertions are not set up to look for this new preview header. These need to be updated, like this:

```C#
[Fact]
public async Task TheNewRepositoryDescription()
{
    var client = new RepositoriesClient(connection);
    var newRepository = new NewRepository("aName");

    await client.Create("aLogin", newRepository);

    connection.Received().Post<Repository>(Args.Uri, newRepository, "application/vnd.github.baptiste-preview+json");
}
```

We avoid using the `AcceptsHeader` value here as it feels like a shortcut - we want to assert the header value received matches what the API should see.

PR #2194, opened 2020-05-30
ChecksClient constructor should take interface arguments

fredrikhr:
The `ChecksClient` constructor originally took a concrete `ApiConnection` object as an argument. This PR changes the constructor to match all other clients in the library that take an `IApiConnection` argument.

This PR also adds a unit test theory, that asserts that all clients nested under `GitHubClient` follow this behaviour.

The assertion ensures
* All client interfaces have a publically visible concrete implementation type
* The Implementation type has a publically visible constructor that takes either `IConnection` or `IApiConnection` as its single argument.

### DIFF
--- a/Octokit.Tests/GitHubClientTests.cs
+++ b/Octokit.Tests/GitHubClientTests.cs
@@ -4,6 +4,8 @@ using NSubstitute;
 using Octokit.Internal;
 using Xunit;
 using System.Collections.Generic;
+using System.Reflection;
+using System.Linq;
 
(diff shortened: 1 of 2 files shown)


shiftkey:
@fredrikhr this is a great spot - thanks for adding in that convention test. 

Gonna have a read over the changes and see if I have any suggestions to improve it.

shiftkey:
release_notes: `ChecksClient` constructor should take interface arguments

PR #2202, opened 2020-06-08
drop unused code related to PORTABLE flag

shiftkey:
This component was marked as obsolete in `v0.47.0`, and the next release will drop this code as it shouldn't have been part of the public API at the time.


shiftkey:
release_notes: Removed unused helper class `ConcurrentCache`

PR #2224, opened 2020-07-09
Fix property types in PullRequestCommit

mmv:
PullRequestCommit Author and Committer fields were mistakenly typed with
Committer when they should be of type User.

Previously the types of properties were:

* `PullRequestCommit.Author/Committer`: `Committer`
* `Commit.Author/Committer`: `Committer`

Correct types should be:

* `PullRequestCommit.Author/Committer`: `User`
* `Commit.Author/Committer`: `Committer`

These fields always fail to deserialize in the API calls but produce no
errors, only objects with default values.

### DIFF
--- a/Octokit.Tests/Models/PullRequestCommitTests.cs
+++ b/Octokit.Tests/Models/PullRequestCommitTests.cs
@@ -1,0 +1,102 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using Octokit;
+using Octokit.Internal;
+using Xunit;
+
+public class PullRequestCommitTests {
+    // the following JSON is taken from the documentation available at https://developer.github.com/v3/pulls/#list-commits-on-a-pull-request
+
+    const string jsonPayload = @"
+    [
+  {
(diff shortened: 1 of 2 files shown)


haacked:
Gee, what dummy got this wrong?!  [Oh.](https://github.com/octokit/octokit.net/commit/b2d6e15d48ba63f3cdc21270d01de5d0188bcb35)

haacked (approved):
Hey @shiftkey. It's potentially a breaking change for users, but probably should be taken. I reviewed the change and it's correct.

shiftkey (approved):
Thanks for the improvement @mmv!

shiftkey:
release_notes: `Author` and `Committer` properties on `PullRequestCommit` are now of type `User`, not `Committer`

shiftkey:
breaking_change: Type of `PullRequestCommit.Author` and `PullRequestCommit.Commiter` properties is changed from `Committer` to `User`, please ensure serializing or deserializing this object in your code is not impacted by this change.

PR #2239, opened 2020-08-16
feat: add missed props for organization

InRedikaWB:
Organizations do not have some properties from https://developer.github.com/v3/orgs/ like description. Add missed properties for correct result parsing.


shiftkey (approved):
Thanks for the contribution @InRedikaWB!

shiftkey:
release_notes: Additional properties added to `Organization` response

Issue #2100, opened 2020-02-17
repository_dispatch event is not covered


justinyoo:
Hi, Team.

I'm heavily using GitHub Actions and just found that Octokit.NET doesn't have repository_dispatch event related logic, while rest.js does.

Would you please share your timeline to add this feature?

shiftkey:
@justinyoo thanks for your interest!

Currently my bandwidth for addressing changes to the APIs is limited, so I can't commit to a specific timeline, but I'm also happy to make time for reviewing contributions to keep things moving.

Can you refresh my memory on the sample that you would like to work? That'll help me write up what changes need to be made to support this event.

justinyoo:
@shiftkey Thanks for your reply!

I forked the repo and am in the middle of adding some logic for this, by changing `IRepositoriesClient`, `RepositoriesClient`, `IApiConnection` and `ApiConnection`. Once I complete my first draft, I'll send a PR for review.

MindaugasLaganeckas:
Our company needs this feature as well.
@justinyoo: can I start working on this feature or do you plan to resume working on it?

justinyoo:
@MindaugasLaganeckas Oh, that would be great! I’ve been fully stretched since this conversation. Thanks for your offer!

MindaugasLaganeckas:
Unfortunately, our focus has changed for now. Everyone is welcome to take over.

krokofant:
This is still missing. Could this be reopened?

EvoTechMike:
@justinyoo Did you make the PR?  I am looking to trigger actions and cant find anything documented in library regarding it.

jakub-lesniak-ikea:
This seems to be still missing after 4 years..

PR #2253, opened 2020-09-21
Implementation of workflow dispatch

MindaugasLaganeckas:


MindaugasLaganeckas:
Implementation of https://github.com/octokit/octokit.net/issues/2100

MindaugasLaganeckas:
@shiftkey : I would like to ask for an early review. I will improve the code coverage, when I finalize the PR.

Octokit/Helpers/ApiUrls.cs
```diff
@@ -1153,6 +1153,18 @@ public static Uri CreateTag(string owner, string name)
...
+        /// <summary>
+        /// Returns the <see cref="Uri"/> for creating the event.
+        /// </summary>
+        /// <param name="owner">The owner of the repository</param>
+        /// <param name="repo">The name of the repository</param>
+        /// <param name="workflowId">The ID of the workflow</param>
+        /// <returns></returns>
+        public static Uri CreateWorkflowDispatchEvent(string owner, string repo, string workflowId)
+        {
+            return "repos/{owner}/{repo}/actions/workflows/{workflow_id}/dispatches".FormatUri(owner, repo, workflowId);
```
khellang, line 1165: This uses named holes (is that even supported?), while the other method uses indexed holes. Also, there are a bunch of inconsistencies in `name` vs `repo` across the board. I suggest using `repo` throughout 😊

MindaugasLaganeckas: Thank you for such quick feedback. :smile: I will fix `name` vs `repo`. I have not tested yet if those holes are supported. I got inspired from 2 different (and inconsistent Client implementations). I am not so familiar with React framework and I was hoping to get some feedback there. Furthermore I am a bit limited in testing workflow dispatch or `GetAllForRepository` workflows. :smile:

PR #2258, opened 2020-10-02
Add license to repository search

Dagizmo:
Search Repositories by license is supported by github api:
https://help.github.com/articles/searching-repositories#search-by-license

Currently only a list of licenses is supported as stated in : 
https://docs.github.com/en/free-pro-team@latest/github/creating-cloning-and-archiving-repositories/licensing-a-repository#searching-github-by-license-type.

But using GitHub search I was able to find more licenses types keys.
https://github.com/search/advanced

This PR, adds all licenses found right now on GitHub.

One of the issues found while developing is that Octokit.License class does exist, while Octokit.Langues exist inside SearchRepositoriesRequest.cs and is also used on different classes I think we should move it later, for this PR I named the enum that has the licenses as : Octokit.RepoSearchLicense, the name can change, also I took the "license" word of the description of the enum, but I can put it back if needed.

Also test were added :)

### DIFF
--- a/Octokit.Tests/Models/SearchRepositoryRequestTests.cs
+++ b/Octokit.Tests/Models/SearchRepositoryRequestTests.cs
@@ -33,5 +33,13 @@ public class SearchRepositoryRequestTests
             var result = request.MergedQualifiers();
             Assert.Contains(result, x => string.Equals(x, "language:\"cpp\""));
         }
+
+        [Fact]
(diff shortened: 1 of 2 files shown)


shiftkey (approved):
Thanks for the contribution @Dagizmo!

shiftkey:
release_notes: `License` property added to repository search API

PR #2266, opened 2020-10-27
Case insensitive X-GitHub-Request-Id

jcansdale:
Add point fix from:
https://github.com/octokit/octokit.net/pull/2175

..for https://github.com/github/VisualStudio/issues/2560


jcansdale:
Wrong repo. 😊

PR #2291, opened 2021-02-08
Add support for multi users and orgs in SearchCodeRequest

ecarlson94:


shiftkey (approved):
Thanks for the contribution @ecarlson94!

I think this is a reasonable change to include, just needs some release notes to indicate the change from `string` to `string[]` for `Organizations` and `Users` on `SearchCodeRequest`

shiftkey:
release_notes: Added support for specifying multiple users or organizations when searching for code

shiftkey:
`breaking_change`: `SearchCodeRequest.Users` and `SearchCodeRequest.Organizations` now expects an array of strings, as the underling API supports searching for multiple users or organizations.

PR #2289, opened 2021-02-07
(build) Updated Cake Frosting to version 1.0.0

devlead:
This PR updates Cake Frosting to 1.0 and addresses breaking changes introduced with that release.

**Additional**
* Pinned coverlet version as latest not compact with 3.1
* Updated GitVersion as previous not compatible with GitHub actions
* Updated GitHub actions setup .NET to supported version

PR #2293, opened 2021-02-08
Attempt to fix the build

haacked:
Shouldn't specify an exact version of `setup-dotnet`. It won't use the latest patches.

### DIFF
--- a/.github/workflows/netcore.yml
+++ b/.github/workflows/netcore.yml
@@ -20,7 +20,7 @@ jobs:
     steps:
     - uses: actions/checkout@v1
     - name: Setup .NET Core
-      uses: actions/setup-dotnet@v1.4.0
+      uses: actions/setup-dotnet@v1
       with:
         dotnet-version: 3.1.201
     - name: Build with dotnet
--- a/.github/workflows/netcore.yml
+++ b/.github/workflows/netcore.yml
@@ -22,7 +22,7 @@ jobs:
     - name: Setup .NET Core
       uses: actions/setup-dotnet@v1
       with:
-        dotnet-version: 3.1.201
+        dotnet-version: 3.1.*
     - name: Build with dotnet
       run: ./build.sh --linksources=true --verbosity=verbose
       shell: bash
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -13,7 +13,7 @@ jobs:
     steps:
     - uses: actions/checkout@v1
     - name: Setup .NET Core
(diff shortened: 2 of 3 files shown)


haacked:
Ok, this is a necessary, but not sufficient patch. There's a lot of errors like:

```
Error: Unable to process command '::set-env name=GitVersion_Major::0' successfully.
Error: The `set-env` command is disabled. Please upgrade to using Environment Files or opt into unsecure command execution by setting the `ACTIONS_ALLOW_UNSECURE_COMMANDS` environment variable to `true`. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
```

I'm not sure where those are coming from and this is all the time I had today to investigate. :cry:

haacked:
Ok, I believe this is related to https://github.com/GitTools/GitVersion/issues/2421

I noticed this the build output:

```
You can invoke the tool using the following command: dotnet-gitversion
Tool 'gitversion.tool' (version '5.1.3') was successfully installed.
Calculating semantic version...
```

The build is using an old version of gitversion.tool. I still don't know what's invoking that (I couldn't find gitversion.tool in a source search). But it seems like upgrading that to the latest `gitversion.tool` will fix things. Or at least make them better.

I'll dig a little more.

haacked:
Also, the AppVeyor build fails with this message:

```
dotnet tool install --global coverlet.console
error NU1202: Package coverlet.console 3.0.2 is not compatible with netcoreapp3.1 (.NETCoreApp,Version=v3.1) / any. Package coverlet.console 3.0.2 supports: net5.0 (.NETCoreApp,Version=v5.0) / any
The tool package could not be restored.
Tool 'coverlet.console' failed to install. This failure may have been caused by:
```

For now, I'm trying to fix with an `explicit coverlet.console` version.

As for `gitversion.tool`, I can't figure out how that's being set up. It just needs to be updated.

haacked:
Hey! Using the explicit version of `coverlet.console` fixed the AppVeyor build. Unfortunately it's a pretty old version of it.

I have another PR experimenting with moving the builds to .NET 5 so the latest coverlet can be used. https://github.com/octokit/octokit.net/pull/2294

devlead:
Some things probably fixed in
https://github.com/octokit/octokit.net/pull/2289

shiftkey:
I think this is fine after merging #2289 and testing on some docs PRs so I'm gonna close this out.

Thanks for taking a stab at it @haacked!

PR #2294, opened 2021-02-08
Move the build to .net 5

haacked:
.NET 5 is fine for building projects that target older versions of .NET. This also should fix the AppVeyor build which fails because the latest `coverlet.console` only supports .NET 5.


haacked:
The failures seem to be the same failures in #2293, so it looks like using .net 5 for the build is fine. This fixes the broken AppVeyor build.

shiftkey:
release_notes: Now using .NET 5 to build and publish

PR #2295, opened 2021-02-09
Fix parsing validation errors

jairbubbles:
The default parsing does not work with validation error, I couldn't find any documentation on the subject so I tried not to break the existing parsing to be retro compatible.


shiftkey:
@jairbubbles I'm not sure I understand the issue at play here. Are you able to elaborate?

jairbubbles:
@shiftkey Sure! The default error parsing is too strict as it considers an error has four firelds. As a result other fields like documentation url are not parsed either as it just throws an exception during the deserialization.

In some cases, the error message is just a plain string.

I modified the code a bit to make it work. I don't find the code very pretty so I kept is as a draft for now.

PR #2299, opened 2021-02-13
tidy up workflow to address new restrictions

shiftkey:
Similar changes to #2289 to support publish workflow


shiftkey:
release_notes: Fixed publish flow for Actions due to environment changes

PR #2303, opened 2021-02-19
Change Payload to a Dictionary<string, string>

### LINKED ISSUE #2250: Invalid cast on Deployment.Create
As per the documentation the `deployment.payload` is intended to accept json.
Octokit.net has `NewDeployment.Payload` as string.

When creating a new deployment with a json serialized string in `Payload` the request is accepted and the deployment is created. However, it fails with an `InvalidCastException` when attempting return the `Deployment` response model.

On closer inspection it appears to be an issue with the Deployment response model expecting to be able to cast to a `IReadOnlyDictionary<string,string>` yet the NewDeployment request model accepts a string.
See https://github.com/octokit/octokit.net/blob/main/Octokit/Models/Response/Deployment.cs#L60
and https://github.com/octokit/octokit.net/blob/main/Octokit/Models/Request/NewDeployment.cs#L62

----

Here is some of my sandbox code I used that raised the error.
This is using Octokit 0.48.0

```csharp
new NewDeployment("master")
{
	Environment = "TestEnvironment1",
	Description = "ShipIt",

	//Serialize a model in to the payload
	Payload = JsonSerializer.Serialize(new
	{
		Property1 = "prop1value",
		Property2 = "prop2value",
	})
})
```

I get the following error
```
System.InvalidCastException
Invalid cast from 'System.String' to 'System.Collections.Generic.IReadOnlyDictionary`2[[System.String, System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[System.String, System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]]'.
   at System.Convert.DefaultToType(IConvertible value, Type targetType, IFormatProvider provider)
   at System.String.System.IConvertible.ToType(Type type, IFormatProvider provider)
   at System.Convert.ChangeType(Object value, Type conversionType, IFormatProvider provider)
   at Octokit.PocoJsonSerializerStrategy.DeserializeObject(Object value, Type type) in /home/runner/work/octokit.net/octokit.net/Octokit/SimpleJson.cs:line 1407
   at Octokit.Internal.SimpleJsonSerializer.GitHubSerializerStrategy.DeserializeObject(Object value, Type type) in /home/runner/work/octokit.net/octokit.net/Octokit/Http/SimpleJsonSerializer.cs:line 205
   at Octokit.PocoJsonSerializerStrategy.DeserializeObject(Object value, Type type) in /home/runner/work/octokit.net/octokit.net/Octokit/SimpleJson.cs:line 1494
   at Octokit.Internal.SimpleJsonSerializer.GitHubSerializerStrategy.DeserializeObject(Object value, Type type) in /home/runner/work/octokit.net/octokit.net/Octokit/Http/SimpleJsonSerializer.cs:line 205
   at Octokit.PocoJsonSerializerStrategy.DeserializeObject(Object value, Type type) in /home/runner/work/octokit.net/octokit.net/Octokit/SimpleJson.cs:line 1521
   at Octokit.Internal.SimpleJsonSerializer.GitHubSerializerStrategy.DeserializeObject(Object value, Type type) in /home/runner/work/octokit.net/octokit.net/Octokit/Http/SimpleJsonSerializer.cs:line 205
   at Octokit.SimpleJson.DeserializeObject(String json, Type type, IJsonSerializerStrategy jsonSerializerStrategy) in /home/runner/work/octokit.net/octokit.net/Octokit/SimpleJson.cs:line 590
   at Octokit.SimpleJson.DeserializeObject[T](String json, IJsonSerializerStrategy jsonSerializerStrategy) in /home/runner/work/octokit.net/octokit.net/Octokit/SimpleJson.cs:line 602
   at Octokit.Internal.SimpleJsonSerializer.Deserialize[T](String json) in /home/runner/work/octokit.net/octokit.net/Octokit/Http/SimpleJsonSerializer.cs:line 22
   at Octokit.Internal.JsonHttpPipeline.DeserializeResponse[T](IResponse response) in /home/runner/work/octokit.net/octokit.net/Octokit/Http/JsonHttpPipeline.cs:line 62
   at Octokit.Connection.Run[T](IRequest request, CancellationToken cancellationToken) in /home/runner/work/octokit.net/octokit.net/Octokit/Http/Connection.cs:line 653
   at Octokit.ApiConnection.GetPage[TU](Uri uri, IDictionary`2 parameters, String accepts, ApiOptions options) in /home/runner/work/octokit.net/octokit.net/Octokit/Http/ApiConnection.cs:line 622
   at Octokit.ApiConnection.<>c__DisplayClass18_0`1.<<GetAll>b__0>d.MoveNext() in /home/runner/work/octokit.net/octokit.net/Octokit/Http/ApiConnection.cs:line 212
```

Of note, i also tried creating a NewDeployment serializing a `Dictionary<string,string>` to payload but got the same result. Deployment was created on github but Octokit failed with a cast exception.

```csharp
new NewDeployment("master")
{
	Environment = "TestEnvironment1",
	Description = "ShipIt",
	Payload = JsonSerializer.Serialize(new Dictionary<string, string>
	{
		{"prop1", "value1"},
		{"prop2", "value1"}
	})
})
```

When calling github directly via curl I can see my deployment object created as expected. Note payload is a string and not json.
```
...
"ref": "master",
"task": "deploy",
"payload": "{\"prop1\":\"value1\",\"prop2\":\"value1\"}",
"environment": "TestEnvironment1",
"description": "ShipIt",
...
```

Is this a bug or am i over thinking it?

haacked:
When serializing the `NewDeployment` type, the `Payload` is serialized as an escaped string because JSON.NET doesn't know it's meant to be JSON.

This causes a problem when you call the API because the Payload is supposed to be a JSON dictionary that's just part of the overall payload. It's not supposed to be an escaped string.

That's why the JSON deserializer fails on it. Not only that, any deployments created using the current Octokit.net will create an invalid payload.

This PR fixes it by changing the type of `Payload` to a dictionary. THIS IS A BREAKING CHANGE, but the old behavior was broken so it forces a new correct behavior.

Fixes #2250

### DIFF
--- a/Octokit/Models/Request/NewDeployment.cs
+++ b/Octokit/Models/Request/NewDeployment.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.ObjectModel;
+﻿using System.Collections.Generic;
+using System.Collections.ObjectModel;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
@@ -59,7 +60,7 @@ namespace Octokit
         /// <summary>
         /// JSON payload with extra information about the deployment.
         /// </summary>
-        public string Payload { get; set; }
+        public Dictionary<string, string> Payload { get; set; }
 
         /// <summary>
         /// Optional name for the target deployment environment (e.g., production, staging, qa). Default: "production"
--- a/Octokit.Tests/Models/DeploymentTests.cs
+++ b/Octokit.Tests/Models/DeploymentTests.cs
@@ -9,6 +9,18 @@ namespace Octokit.Tests.Models
     public class DeploymentTests
     {
         [Fact]
+        public void CanSerialize()
+        {
+            var deployment = new NewDeployment("ref")
+            {
+                Payload = new Dictionary<string, string> {{"environment", "production"}}
+            };
+            var deserialized = new SimpleJsonSerializer().Serialize(deployment);
+            
+            Assert.Equal(@"{""ref"":""ref"",""task"":""deploy"",""payload"":{""environment"":""production""}}", deserialized);
+        }
+
+        [Fact]
         public void CanDeserialize()
         {
             const string json = @"{


Octokit/Models/Request/NewDeployment.cs
```diff
@@ -59,7 +60,7 @@ public NewDeployment(string @ref)
         /// <summary>
         /// JSON payload with extra information about the deployment.
         /// </summary>
-        public string Payload { get; set; }
+        public Dictionary<string, string> Payload { get; set; }
```
shiftkey, line 63: I'm curious whether this can be any sort of JSON object, and if we're being too strict here with assuming values will always be `string`. I can't find any interesting examples in the docs beyond [this mention](https://docs.github.com/en/rest/reference/repos#create-a-deployment) but maybe you're more familiar with this API currently than I am:

> The payload parameter is available for any extra information that a deployment system might need. **It is a JSON text field that will be passed on when a deployment event is dispatched.**

shiftkey (approved):
Happy to take this in as a bugfix for the original issue, and if we need to refine this later then I gather I'll hear from users of this API

haacked:
Yeah, `Dictionary<string, object> is probably fine. But we'd probably have to make that change to `Deployment` as well.

shiftkey:
> Yeah, `Dictionary<string, object>` is probably fine. But we'd probably have to make that change to `Deployment` as well.

Let's get this out and see what happens

shiftkey:
release_notes: Fixed deserialization issue for `Deployment.Payload` by returning `Dictionary<string,string>`

PR #2321, opened 2021-04-20
drop Appveyor and Codecov

shiftkey:
Due to [the Codecov security incident](https://about.codecov.io/security-update/) and a nudge from a colleague on the remaining AppVeyor usage in this project, this PR removes both while I also review for any tokens I need to roll...


nickfloyd:
release_notes: Removes Appveyor and Codecov usage due to a [codecov security incident](https://about.codecov.io/security-update/)

PR #2322, opened 2021-04-22
add size to Readme response payload

### LINKED ISSUE #2233: Cannot get Readme size via octokit.net
Need to access the `size` of README.md file, only to find that it's not a public property of Readme object. 

The property itself is indeed available  in native [Github API](https://docs.github.com/en/rest/reference/repos#get-a-repository-readme). Better to have it in octokit.net.

littlemisslilycane:
Fixes #2233

### DIFF
--- a/Octokit.Tests/Clients/RepositoryContentsClientTests.cs
+++ b/Octokit.Tests/Clients/RepositoryContentsClientTests.cs
@@ -40,6 +40,7 @@ namespace Octokit.Tests.Clients
                 var readme = await contentsClient.GetReadme("fake", "repo");
 
                 Assert.Equal("README.md", readme.Name);
+                Assert.Equal(100, readme.Size);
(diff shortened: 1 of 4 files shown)


M-Zuber:
Could you update/add a test for the new field that was added?

littlemisslilycane:
I have updated the existing unit tests to check the Size field. However, I noticed there is an Integration test file, "RepositoryContentsClientTests.cs" which tests GetReadme method using the "octokit.net" repository. Should that be modified too? If so how do I test the readme size of a repo which could change anytime.

lastthyme (approved):

M-Zuber:
I would suggest to look at how [this helper](https://github.com/octokit/octokit.net/blob/9558604273544ad45c143aed7427e1279d77192f/Octokit.Tests.Integration/Helpers/RepositorySetupHelper.cs#L45) is used, and create a new integration test that uses a test repository.

PR #2345, opened 2021-07-19
fix(docs): master --> main text in README

SeanKilleen:
Just happened to notice while browsing. After the conversion to the `main` branch, some text and old links remained.


shiftkey (approved):

nickfloyd:
release_notes: Cleans up the docs after the move from using the `master` moniker to `main`

PR #2352, opened 2021-08-23
WIP: Change all applicable IDs in the project from int --> long

### LINKED ISSUE #2351: GitHub IDs have exceeded Int32.MaxValue; Octokit uses int
Spotted in this tweet from @patriksvensson: https://twitter.com/firstdrafthell/status/1429738272574558210?s=19

I'll look into this ASAP but don't mind anyone else doing the same.

SeanKilleen:
It's been a long time coming. 😆

Resolves #2351.

⚠I assume this would mandate a breaking/major release.

## My Approach

If I'm going to be honest, it's a little bit "search and destroy". I'm going for wholesale change in the project and ensuring that all IDs can be `long`s going forward. So I'm searching with a regex of `int [^=]*Id` and `int\? [^=]*Id` and going from there.

## Work

* [x] Find ID values
* [x] Change said ID values to longs
* [x] Do it everywhere
* [x] Make sure the tests pass
* [ ] Double check `int [^=]*Id` search
* [ ] Double check `int\? [^=]*Id`
* [ ] Do a search for `int [^=]*number` just in case
* [ ] Do a search for `int\? [^=]*number` just in case
* [ ] Ensure that that the `int` IDs I accidentally captured as `long` have been changed back.

## Items To Confirm / Change based on feedback

* [x] :x: `PullRequestReview.Number` ~~Seems like it should remain int, given it's the display value and I don't think we're in danger of running out there)~~ leaving alone for now per comments.
* [x] :x: `Account.DiskUsage` ~~should this remain an int? Need to double-check on how it's used.~~ leaving alone per comments
* [x] :x: `GpgKey.PrimaryKeyId` ~~this is an `int?`. Not sure enough of the usage to change it right away~~ Treating like an ID but ensuring it handles null. (Update: leaving this alone as API defines it a an int.)

### DIFF
--- a/Octokit.Reactive/Clients/Enterprise/IObservableEnterpriseLdapClient.cs
+++ b/Octokit.Reactive/Clients/Enterprise/IObservableEnterpriseLdapClient.cs
@@ -43,7 +43,7 @@ namespace Octokit.Reactive
         /// <param name="teamId">The teamId to update LDAP mapping</param>
         /// <param name="newLdapMapping">The <see cref="NewLdapMapping"/></param>
         /// <returns>The <see cref="Team"/> object.</returns>
-        IObservable<Team> UpdateTeamMapping(int teamId, NewLdapMapping newLdapMapping);
+        IObservable<Team> UpdateTeamMapping(long teamId, NewLdapMapping newLdapMapping);
 
         /// <summary>
         /// Queue an LDAP Sync job for a team on a GitHub Enterprise appliance (must be Site Admin user).
@@ -53,6 +53,6 @@ namespace Octokit.Reactive
         /// </remarks>
         /// <param name="teamId">The teamId to update LDAP mapping</param>
         /// <returns>The <see cref="LdapSyncResponse"/> of the queue request.</returns>
-        IObservable<LdapSyncResponse> QueueSyncTeamMapping(int teamId);
+        IObservable<LdapSyncResponse> QueueSyncTeamMapping(long teamId);
     }
 }
--- a/Octokit.Reactive/Clients/Enterprise/ObservableEnterpriseLdapClient.cs
+++ b/Octokit.Reactive/Clients/Enterprise/ObservableEnterpriseLdapClient.cs
@@ -58,7 +58,7 @@ namespace Octokit.Reactive
         /// <param name="teamId">The teamId to update LDAP mapping</param>
         /// <param name="newLdapMapping">The <see cref="NewLdapMapping"/></param>
         /// <returns>The <see cref="Team"/> object.</returns>
-        public IObservable<Team> UpdateTeamMapping(int teamId, NewLdapMapping newLdapMapping)
+        public IObservable<Team> UpdateTeamMapping(long teamId, NewLdapMapping newLdapMapping)
         {
             return _client.UpdateTeamMapping(teamId, newLdapMapping).ToObservable();
         }
@@ -71,7 +71,7 @@ namespace Octokit.Reactive
         /// </remarks>
         /// <param name="teamId">The teamId to update LDAP mapping</param>
         /// <returns>The <see cref="LdapSyncResponse"/> of the queue request.</returns>
-        public IObservable<LdapSyncResponse> QueueSyncTeamMapping(int teamId)
+        public IObservable<LdapSyncResponse> QueueSyncTeamMapping(long teamId)
         {
             return _client.QueueSyncTeamMapping(teamId).ToObservable();
         }
--- a/Octokit.Reactive/Clients/IObservableAuthorizationsClient.cs
+++ b/Octokit.Reactive/Clients/IObservableAuthorizationsClient.cs
@@ -42,7 +42,7 @@ namespace Octokit.Reactive
         /// <returns>An <see cref="Authorization"/></returns>
         [SuppressMessage("Microsoft.Naming", "CA1716:IdentifiersShouldNotMatchKeywords", MessageId = "Get",
             Justification = "It's fiiiine. It's fine. Trust us.")]
-        IObservable<Authorization> Get(int id);
+        IObservable<Authorization> Get(long id);
 
         /// <summary>
         /// Creates a new personal token for the authenticated user.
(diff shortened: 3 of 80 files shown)


shiftkey:
> ⚠I assume this would mandate a breaking/major release.

I worry about updating to 1.0 but am happy to do a big breaking changes note for this release.

SeanKilleen:
@shiftkey actually I think given the weight of 1.0 release and the current availability of maintainers, I'd keep it as the next 0.x and publicize the changes. We can make sure there's a good write-up in the release notes.

SeanKilleen:
OK, at this point I believe I've adjusted the necessary IDs in the main project besides a few that I have to confirm still. 

Next up -- observable client changes. After that we're on to fixing the tests

shiftkey:
> * `GpgKey.PrimaryKeyId` -- this is an `int?`. Not sure enough of the usage to change it right away

Based on [the OpenAPI description](https://github.com/github/rest-api-description/blob/main/descriptions/api.github.com/api.github.com.json) I see these fields:

```json
"gpg-key": {
  "title": "GPG Key",
  "description": "A unique encryption key",
  "type": "object",
  "properties": {
    "id": {
      "type": "integer",
      "example": 3
    },
    "primary_key_id": {
      "type": "integer",
      "nullable": true
    },
    ...
```

👍🏻 to treating it the same as `id`, but also ensuring it handles `null`

shiftkey:
> * `Account.DiskUsage` -- should this remain an int? Need to double-check on how it's used.

```
"disk_usage": {
  "type": "integer",
  "example": 1
},
```

Also a number, I'd leave it as `int` as it's not an identifier...

> * `PullRequestReview.Number` -- Seems like it should remain int, given it's the display value and I don't think we're in danger of running out there).

I don't actually see this listed in the OpenAPI description, which is a bit surprising.

```json
"pull-request-review": {
  "title": "Pull Request Review",
  "description": "Pull Request Reviews are reviews on pull requests.",
  "type": "object",
  "properties": {
    "id": {
      "description": "Unique identifier of the review",
      "example": 42,
      "type": "integer"
    },
    "node_id": {
      "type": "string",
      "example": "MDE3OlB1bGxSZXF1ZXN0UmV2aWV3ODA="
    },
    "user": {
      "nullable": true,
      "allOf": [
        {
          "$ref": "#/components/schemas/simple-user"
        }
      ]
    },
    "body": {
      "description": "The text of the review.",
      "example": "This looks great.",
      "type": "string"
    },
    "state": {
      "type": "string",
      "example": "CHANGES_REQUESTED"
    },
    "html_url": {
      "type": "string",
      "format": "uri",
      "example": "https://github.com/octocat/Hello-World/pull/12#pullrequestreview-80"
    },
    "pull_request_url": {
      "type": "string",
      "format": "uri",
      "example": "https://api.github.com/repos/octocat/Hello-World/pulls/12"
    },
    "_links": {
      "type": "object",
      "properties": {
        "html": {
          "type": "object",
          "properties": {
            "href": {
              "type": "string"
            }
          },
          "required": [
            "href"
          ]
        },
        "pull_request": {
          "type": "object",
          "properties": {
            "href": {
              "type": "string"
            }
          },
          "required": [
            "href"
          ]
        }
      },
      "required": [
        "html",
        "pull_request"
      ]
    },
    "submitted_at": {
      "type": "string",
      "format": "date-time"
    },
    "commit_id": {
      "description": "A commit SHA for the review.",
      "example": "54bb654c9e6025347f57900a4a5c2313a96b8035",
      "type": "string"
    },
    "body_html": {
      "type": "string"
    },
    "body_text": {
      "type": "string"
    },
    "author_association": {
      "$ref": "#/components/schemas/author_association"
    }
  },
  "required": [
    "id",
    "node_id",
    "user",
    "body",
    "state",
    "commit_id",
    "html_url",
    "pull_request_url",
    "_links",
    "author_association"
  ]
},
```

Let's leave it alone for now.

SeanKilleen:
Alright, next up I'm on to fixing the integration tests and getting that all passing.

SeanKilleen:
Converted this to a Draft so as not to cause so much noise hopefully (I think it will help stop CI/CD from running, etc.)

SeanKilleen:
@shiftkey I noticed in going through these APIs that in some cases `number` and `id` are used interchangeably where perhaps they should not be. E.g. `number` being used with an xmldoc comment that says `the [x] id`. 

I could likely address these in this PR as well on a case-by-case basis. Let me know if you want me to call these out for follow-up as we go through the first round of changes.

shiftkey:
> I noticed in going through these APIs that in some cases `number` and `id` are used interchangeably where perhaps they should not be.

Yeah, `number` has a specific meaning and should be separate to `id` in the API. Happy to summarize what you've found and tackle it in a follow-up PR, especially if it's more related to docs then the overflow issue you're trying to fix.

SeanKilleen:
@shiftkey in the case of entities, such as `comment`, where the number/id are the same and id is considered an `integer`, it seems to me it would be ok to update it to a long in Octokit in the anticipation of it being enlarged later. However, would you prefer it to be kept as an `int` in the codebase in these cases? 

(Realized that my confusion about `id` vs `number` is that I didn't realize the API considered these the same for some elements.)

SeanKilleen:
An update here -- it's clear at this point from the many "oops" commits that I misunderstood that many IDs in the API level are indeed intended to be `int`s even if I might expect them to exceed that boundary at some point. 

So, my next steps to get through this are:

* Make the branch work, tests and all, with the changes I've already made
* Go back through every change and ensure it is in line with the API v3 specifications, and correct where it's not
* Once again ensure everything is building and passing.

At that point we should be back on the right track.

SeanKilleen:
Sigh....this is on me. @shiftkey, the more I look at it, the more these IDs are correctly `int`s and my mapping them to `long`s looks incorrect. I think I should have done more research before jumping into this.

We have a few options:

* Revert any IDs that aren't explicitly longs in the API docs
* Convert IDs to `long`s regardless of API docs, for graceful upgrades in the future, given that since we're reading there's little chance of an issue cropping up. 

I'm fine to proceed either way, including if I should close this out and start over to get rid of the mess.

SeanKilleen:
@shiftkey what do you think? 

1. Match the return types of the API (e.g. leave `int` if API specifies `integer`)
2. Convert all IDs to `long` regardless of API type
3. Convert all IDs -- and numbers -- to `long` regardless of API type

Happy to see this through, but realized I should probably get that clarification first.

PR #2375, opened 2021-11-18
Added enterprise pre-receive hooks client

tasadar2:


nickfloyd (approved):
Thanks for taking the time on these contributions!  They all look good! ❤️

PR #2378, opened 2021-11-30
fix: links into github documentation

Stefan-J-Hanke:
I was just curious to know about that `ProductHeaderValue` type and found references in this file. However the links were not 404's but also not what was originally intended. I didn't check the whole file, just these two URLs.


nickfloyd (approved):
Thanks for doing this! ❤️ You're right in assuming that there is more to change.  I am going to make an up-for-grabs issue to see if we can get the others updated - hopefully, once we automate things like model generation then these types of things will not get so far out of date.

nickfloyd:
https://github.com/octokit/octokit.net/issues/2474

nickfloyd:
release_notes: Fixes links to github API documentation

Issue #2112, opened 2020-02-25
improve the story about controlling request timeouts


shiftkey:
Opening this to track making this easier to understand and eliminate confusion about `HttpClient` global and per-`Request` timeouts (the lower of the two is what will be used).

- [ ] share some code samples where timeouts matter
- [ ] ???

Issues in the wild that may be relevant: #2045 #1069 #984

shiftkey:
Currently there are three areas where timeouts can be changed:

 - `GitHubClient.SetRequestTimeout` to set the timeout globally

```c#
var client = new GitHubClient(new Connection(new ProductHeaderValue("OctokitTests")));
client.SetRequestTimeout(TimeSpan.FromSeconds(15));
``` 

 - `IRequest.Timeout` is set per-request, and the constructor for `Request` sets a default timeout of 100 seconds. 
 - `ReleaseAssetUpload.Timeout` has a nullable `Timespan?` property that can be used to set a higher time.

One of the known issues here is where a global and per-request timeout are specified, **the lower value is the winner**. This is likely a source of confusion for users who set a higher value globally but can't access the inner default 100s on `Request`.

arxange1:
Lets double check the purpose of every timeout field we have:

1) HttpClient.Timeout in the field HttpClientAdapter._http

- Scope: per-request timeout
- In use by: HttpClient internaly
- How can be changed by library user: via SetRequestTimeout() method
- How can be read for diagnostics: not implemented

2) IRequest.Timeout (100 seconds)

- Scope: global timeout
- In use by: HttpClientAdapter.GetCancellationTokenForRequest() method only
- How can be changed by library user: not implemented
- But: Implemented for a few methods like GetArchive(owner, name, archiveFormat, reference, **timeout**)
- How can be read for diagnostics: not implemented

shiftkey:
@arxange1 isn't 1) the global timeout and 2) the per-request timeout?

arxange1:
@shiftkey To get a right answer we have to define what the global timeout is. Is pagination included in global scope?

arxange1:
According to docs.microsoft.com both are per-request. The difference is that the first will be applied to all requests and the send one to individual request. But the Octokit passes individual 100s timeout to all requests, so the Timeout property one does not matter, still not less than 100s

shiftkey:
> To get a right answer we have to define what the global timeout is.

I don't quite understand, but I want to focus on what Octokit **should** be doing, rather than what it's currently doing.

[...]

PR #2379, opened 2021-12-03
Allow end-user to reconfigure timeout on IRequest.Timeout #2112

tapika:
Allow to set timeout for UploadAsset

### DIFF
--- a/Octokit.Tests/GitHubClientTests.cs
+++ b/Octokit.Tests/GitHubClientTests.cs
@@ -209,8 +209,12 @@ namespace Octokit.Tests
 
                 client.SetRequestTimeout(TimeSpan.FromSeconds(15));
 
-
                 httpClient.Received(1).SetRequestTimeout(TimeSpan.FromSeconds(15));
+                
+                var connection = Substitute.For<IConnection>();
+                var t = TimeSpan.FromSeconds(20);
+                connection.SendDataTimeout = t;
+                connection.Received(1).SendDataTimeout = t;
             }
         }
 
--- a/Octokit/Http/Connection.cs
+++ b/Octokit/Http/Connection.cs
@@ -412,6 +412,11 @@ namespace Octokit
                 Endpoint = uri
             };
 
+            if (SendDataTimeout.HasValue)
+            {
+                request.Timeout = SendDataTimeout.Value;
+            }
+
             return SendDataInternal<T>(body, accepts, contentType, cancellationToken, twoFactorAuthenticationCode, request);
         }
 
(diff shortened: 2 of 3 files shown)


nickfloyd:
Hey @tapika, thanks for the change here.  I was wondering what you're specifically trying to accomplish here.  The client has a pubic `SetRequestTimeout` on it that set's the timeout for the connection associated with the client. 

Let me know, if you get the chance, I'd like to better understand the use case for the changeset.  Thanks!

tapika:
> Hey @tapika, thanks for the change here. I was wondering what you're specifically trying to accomplish here. The client has a pubic `SetRequestTimeout` on it that set's the timeout for the connection associated with the client.
> 
> Let me know, if you get the chance, I'd like to better understand the use case for the changeset. Thanks!

As far as I remember - `SetRequestTimeout` did not work for me out of box. That's why I've wired this change.

nickfloyd:
> As far as I remember - `SetRequestTimeout` did not work for me out of box. That's why I've wired this change.

Got it... I'll see if I can verify what you were seeing, especially given the previous implementation appeared that it would do the trick.

nickfloyd:
Hey @tapika,

I just verified this to be working in v1.0.0 of this SDK using:

```
var github = new GitHubClient(new ProductHeaderValue("TimeoutTest"));
var t = TimeSpan.FromMilliseconds(1);
github.SetRequestTimeout(t);
var user = await github.User.Get("nickfloyd");
```

It returns:

```
System.TimeoutException: A task was canceled.
```

A thing to note here, the timeout specified in the connection itself is only for the client connecting to the server.  All this does is pass through the value you pass in to `System.Net.Http.HttpClient` and assigns it to the `Timeout` property - which just sets the timespan to wait before the request itself times out.  

Also, in the case where you might expect deadline propagation (meaning if the GitHub API makes other requests to other services the knowledge of the deadline is not propagated upsteam) to occur on the GitHub REST API side, it will not.  

We need to correct the code docs for this API on the GitHubClient object. The code comments ("Set the GitHub Api request timeout.") implies that this will set the timeout for the GitHub API itself ([which is capped at 10 seconds](https://docs.github.com/en/rest/overview/troubleshooting#timeouts)) and not the client timeout to the API.

nickfloyd:
I have added this as an up-for-grabs : https://github.com/octokit/octokit.net/issues/2506

I will close this PR based on the above - but if you feel like that was in error, please feel free to reopen or continue the discussion on the issue above. 

Thank you for the contributions here; while this isn't what you probably expected it still helped to make the SDK better!

PR #2381, opened 2021-12-10
Add Draft to MergeableState enum

### LINKED ISSUE #2339: "Draft" state missing from "MergeableState" enum
We're receiving a payload from a GitHub webhook with `"mergeable_state": "draft"` when a PR is marked as a draft. We're seeing an error when deserializing the JSON payload using the Octokit `MergeableState` enum:
`System.ArgumentException: Value 'draft' is not a valid 'MergeableState' enum value.`

I think the fix should just be to add `draft` as an enum value here: [MergeableState enum](https://github.com/octokit/octokit.net/blob/e9fbe84657976326c8caed7289c0150e38c8adea/Octokit/Models/Response/PullRequest.cs#L263)

adtakle:
Fixes #2339

### DIFF
--- a/Octokit/Models/Response/PullRequest.cs
+++ b/Octokit/Models/Response/PullRequest.cs
@@ -302,6 +302,12 @@ namespace Octokit
         /// No conflicts, everything good. Merging is allowed.
         /// </summary>
         [Parameter(Value = "clean")]
-        Clean
+        Clean,
+
+        /// <summary>
+        /// Not ready for review. Merging is blocked.
+        /// </summary>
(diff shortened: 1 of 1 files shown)


adtakle:
@nickfloyd could you please take a look at this small PR? It's been open for a few months now. Thanks!

nickfloyd (approved):
✅  Thank you for the change ❤️ .  LGTM!

adtakle:
@nickfloyd could you please merge the PR since you approved? I don't have write access

nickfloyd:
release_notes: Adds Draft to MergeableState enum

Issue #2388, opened 2022-01-05
Converting Issue into Pull Request


miloush:
It seems that the GitHub API supports converting issues into pull requests. https://github.com/refined-github/refined-github/issues/2614 suggests this can be done by supplying issue number to the create PR API.

I verified it is working, using
```C#
public class IssueToPullRequest : NewPullRequest
{
    public IssueToPullRequest(string title, string head, string baseRef) : base(title, head, baseRef) { }

    public int? Issue { get; set; }
    public new string? Title => Issue == null ? base.Title : null;
}
...
IApiConnection connection = new ApiConnection(_github.Connection);
await connection.Post<PullRequest>(
    ApiUrls.PullRequests(Owner, RepoName), 
    new IssueToPullRequest(title, head, baseRef) { Issue = issue?.Number }, 
    AcceptHeaders.DraftPullRequestApiPreview);
```

Therefore I suggest a new overload to `NewPullRequest` is added:

```C#
public NewPullRequest(int issue, string head, string baseRef)
```

that sets a new `Issue` property instead of the `Title` property.

JonruAlveus:
Resolved by PR.

PR #2389, opened 2022-01-06
New pull request from existing issue

miloush:
Implements #2388


nickfloyd (changes requested):
Hey, @miloush thanks for the contributions here ❤️ !  I just had a couple of suggestions based on patterns that are already in the models.  Let me know what you think!

nickfloyd:
release_notes: Adds new pull request from an existing issue API

PR #2391, opened 2022-01-12
Adds the slug field to the GitHubApp response object

### LINKED ISSUE #2335: GitHubApp should include slug
This type https://github.com/octokit/octokit.net/blob/main/Octokit/Models/Response/GitHubApp.cs does not include the "slug" field documented here: https://docs.github.com/en/rest/reference/apps

Unfortunately, the "slug" field appears to be the name that appears with the app interacts with the repository (makes a comment or creates an issue), so we need access to that value for several scenarios.

Whitestripe7773:
I have fixed the issue #2335.

### DIFF
--- a/Octokit/Models/Response/GitHubApp.cs
+++ b/Octokit/Models/Response/GitHubApp.cs
@@ -12,9 +12,10 @@ namespace Octokit
     {
         public GitHubApp() { }
 
-        public GitHubApp(long id, string name, User owner, string description, string externalUrl, string htmlUrl, DateTimeOffset createdAt, DateTimeOffset updatedAt)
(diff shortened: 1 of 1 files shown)


nickfloyd (approved):
Hey, @Whitestripe7773 thanks for the contributions here.  ❤️  Let me know if you wouldn't mind updating the tests [here](https://github.com/octokit/octokit.net/blob/main/Octokit.Tests.Integration/Reactive/ObservableGitHubAppsClientTests.cs#L10).  It's not a show-stopper given this is just a property addition to a response object  but it would be nice to have the coverage (contractwise for the API).

Thanks again!

PR #2410, opened 2022-02-11
Fix full width apostrophe

onein528:
image

Garbled characters.

The character literals in question are treated the same as half-width apostrophes when searching on GitHub, so I couldn't currently found garbled characters in other files. There may be others.


onein528:
If there is another way to find it, this still needs to be open.

PR #2413, opened 2022-02-15
Deployment task is arbitrary string value

### LINKED ISSUE #2412: DeployTask should be a string, not an enum
Per docs and actual usage, there is no way to enumerate all possible values of `task`. It's an arbitrary string that only has meaning given to it via the consuming deployment mechanism. The example values are truly examples, not the only accepted values.

> The task parameter is used by the deployment system to allow different execution paths. In the web world this might be `deploy:migrations` to run schema changes on the system. In the compiled world this could be a flag to compile an application with debugging enabled.

https://docs.github.com/en/rest/reference/deployments#create-a-deployment

https://github.com/octokit/octokit.net/blob/b4d7fb09784f5695fb1360ca40d88262e9bc519f/Octokit/Models/Request/NewDeployment.cs#L41

mryanmurphy:
- Changed `NewDeployment` Task property to `string`
- Added `string Task` property to `Deployment` response
- Updated tests

Changes here align the project to the intended usage based on the documentation linked below.

> The task parameter is used by the deployment system to allow different execution paths. In the web world this might be `deploy:migrations` to run schema changes on the system. In the compiled world this could be a flag to compile an application with debugging enabled.

https://docs.github.com/en/rest/reference/deployments#create-a-deployment

Fixes #2412

### DIFF
--- a/Octokit.Tests/Models/DeploymentTests.cs
+++ b/Octokit.Tests/Models/DeploymentTests.cs
@@ -17,7 +17,7 @@ namespace Octokit.Tests.Models
             };
             var deserialized = new SimpleJsonSerializer().Serialize(deployment);
             
-            Assert.Equal(@"{""ref"":""ref"",""task"":""deploy"",""payload"":{""environment"":""production""}}", deserialized);
+            Assert.Equal(@"{""ref"":""ref"",""payload"":{""environment"":""production""}}", deserialized);
         }
 
         [Fact]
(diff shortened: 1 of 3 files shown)


nickfloyd:
Hey @mryanmurphy thank you for posting this, detailing the change, and making the contribution here ❤️ !  I am marking this as a breaking change just so that we can call it out in the release notes - technically this change shouldn't be problematic but since the type/representation has changed on the field I'd like to call that out... just giving you a head's up.

PR #2441, opened 2022-05-10
Creates the config file for stalebot to help us clean up old/inactive issues

nickfloyd:
We are in need of some house cleaning to get us back to a good "supported" state.  This will activate the stalebot to help us parse through any issues that are either no longer relevant, need to be rethought, or are not able to be addressed.  

Note:  It's not that these issues are not important or useful.  We simply need a clear starting point to begin actively introducing improvements and changes into this SDK.  If anyone feels like the issues that are marked as stale from the bot still need to be opened they can label them with the `pinned` and `followup-needed` labels with some details on the current state of things so that it will not auto close. That way we can have a closer look at the issue and see what we need to do to get it resolved.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,40 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale (approx. 9 months)
+daysUntilStale: 270
+
+# Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
+# Set to false to disable. If disabled, issues still need to be closed manually but will remain marked as stale.
+daysUntilClose: 7
+
+# Only issues or pull requests with all of these labels are check if stale. Defaults to `[]` (disabled)
+onlyLabels: []
+
+# Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
+exemptLabels:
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,40 @@
...
+  Hey Friends, this issue has been automatically marked as `stale` because it has not had
+  recent activity. It will be closed if no further activity occurs. Please add the 
+  `pinned` label if you feel that this issue needs to remain open/active. 
+  Thank you for your contributions and help in keeping things tidy!
+
+# Limit the number of actions per hour, from 1-30. Default is 30
+limitPerRun: 30
+
+# Limit to only `issues` or `pulls`
+only: issues
```
imwiss, line 40: Would it be worth adding `pulls` here, too?

imwiss (approved):
Thanks for setting this up. I agree with the benefits and clarity that stalebot will provide so that we can better focus on the most important friction points for users. 

I left one question that's totally up to you to decide on 👍

.github/stale.yml
nickfloyd: I'd like to parse through the pulls a bit more given that folks put in the extra effort to make the changeset and I'd like to validate if the changes are still relevant before something beneficial gets closed.

PR #2442, opened 2022-05-10
Adds stale bot functionality via action and removes the stalebot config

nickfloyd:
Using an Action (cron) instead of the stale app for ease of visibility and configuration.


imwiss (approved):

nickfloyd:
release_notes: Adds stale bot functionality via action and removes the stalebot config

PR #2447, opened 2022-05-18
Tweak `User-Agent` to refer to `Octokit.net`, not just `Octokit`

timrogers:
This improves the `User-Agent` header sent by Octokit.net to explicitly refer to `Octokit.net`, rather than just `Octokit`. This will make it easier to identify requests from this SDK in GitHub logs and other reporting.


nickfloyd (approved):
Thanks for getting to this @timrogers!

nickfloyd:
release_notes: Tweak User-Agent to refer to Octokit.net, not just Octokit

PR #2450, opened 2022-05-18
Explicitly configure identation settings for VSCode

timrogers:
This adds explicit configuration to `.vscode/settings.json` so we use two spaces for indentation, rather than relying on the user's own settings or automatic detection. It also fixes the indentation of the settings file itself 😂 😓


nickfloyd (approved):
🎉 Looks great!  Thank you!

PR #2456, opened 2022-06-09
Dev container rework to support OmniSharp changes

nickfloyd:
These changes are needed so that when working on code spaces and on OmniSharp server will properly start - https://github.com/OmniSharp/omnisharp-vscode#requirements

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -8,12 +8,8 @@ FROM mcr.microsoft.com/vscode/devcontainers/dotnet:0-${VARIANT}
 COPY --from=mcr.microsoft.com/dotnet/sdk:3.1 /usr/share/dotnet/shared /usr/share/dotnet/shared
 
 # OmniSharp no longer ships with mono so it needs to be added separately
-RUN apt-get update && apt-get install -y --no-install-recommends mono-xsp
+RUN apt-get update && apt-get install -y --no-install-recommends mono-complete
 
 # Add mkdocs for doc generation
 RUN apt-get update && export DEBIAN_FRONTEND=noninteractive && apt-get -y install --no-install-recommends python3-pip
 RUN pip3 install mkdocs
-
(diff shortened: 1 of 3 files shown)


nickfloyd:
> Does Omnisharp start for you with these changes? It doesn't for me. If I try to run the 'Restart OmniSharp' command for the command palette, I also get an error like this.

Odd I've never seen that error.  I am assuming steps to repro are:
1. Spin up a code space using this branch
2. Restart OmniSharp.

Quick question, we're you prompted to restart OmniSharp or did you just run the command out of habit?

Validate my assumed steps to repo - perhaps we can pair on it as well some time.

timrogers:
> > Does Omnisharp start for you with these changes? It doesn't for me. If I try to run the 'Restart OmniSharp' command for the command palette, I also get an error like this.
> 
> Yes it starts right away. Odd I've never seen that error. I am assuming steps to repro are:
> 
> 1. Spin up a code space using this branch
> 2. Restart OmniSharp.
> 
> Validate my assumed steps to repo - perhaps we can pair on it as well some time.

Exactly. I did that, and that's what I'm seeing. Also worth noting that IntelliSense etc. doesn't work for me.

timrogers:
I just ran the command because I saw IntelliSense wasn't working and I thought it might help. It seems that all the OmniSharp related commands generated similar errors.

nickfloyd:
> I just ran the command because I saw IntelliSense wasn't working and I thought it might help. It seems that all the OmniSharp related commands generated similar errors.

Interesting... Would you mind sharing what you are seeing in the VS Code "Output" window after the code space is generated?

Generally you'll see something like:
Screen Shot 2022-06-10 at 9 17 29 AM

timrogers (approved):
I restarted my container and it worked 🤔  What do you think of merging this, and coming back later if we or others run into issues?

nickfloyd:
release_notes: Adds support to the dev container for OmniSharp

Issue #2041, opened 2019-10-28
Expose `triage` and `maintain` permissions


terrajobst:
It looks like there is currently no way to see whether a user has `triage` or `maintain` permissions because the API only exposes the three levels `pull`, `push`, and `admin`.

Affected APIs:

* `GitHubClient.Organization.Team.GetAllRepositories()`
* `GitHubClient.Repository.Collaborator.GetAll()`

shiftkey:
I gather we're talking about https://developer.github.com/v3/repos/collaborators/#list-collaborators ? Or are you referring to the `permission` hash returned on https://developer.github.com/v3/repos/#get ?

I thought there might have been a preview API support for this, but it doesn't seem to be documented.

GraphQL supports this fine, FWIW:

```gql
query () {
  repository(owner: "owner", name: "repo") {
    collaborators(first: 50, affiliation: ALL) {
      edges {
        node {
          login
        }
        permission
      }
    }
  }
}
```

terrajobst:
> I gather we're talking about https://developer.github.com/v3/repos/collaborators/#list-collaborators ?

Yes, and `GitHubClient.Organization.Team.GetAllRepositories()`. I've updated the description.

hnrkndrssn:
Just so I don't go down the wrong 🐇 🕳 

I'm assuming that it's this class [RepositoryPermissions](https://github.com/octokit/octokit.net/blob/master/Octokit/Models/Response/RepositoryPermissions.cs) where these additional permissions would be added?

terrajobst:
That's the only types I'm using, so I think this looks good.

hnrkndrssn:
Cool 👍 PR incoming sometime this weekend 😁

shiftkey:
Closing this out as it's not currently available as a preview API.

If this does make an appearance on https://developer.github.com/v3/previews/ then we can revisit it.

fgarcia-cnb:
for what its worth, it looks like the REST API correctly shows triage and maintain permissions now (if you use "Accept: application/vnd.github.v3.repository+json" in the header: 
[https://developer.github.com/v3/teams/#list-team-repos](url)

and although the documentation only mentions being able to set pull/push/admin as permissions, triage and maintain does indeed work.

shiftkey:
@fgarcia-cnb that sounds like its close, but I'm going to wait for this to appear as a documented preview API https://developer.github.com/v3/previews/

terrajobst:
[...]

PR #2467, opened 2022-06-28
[WIP] feat: add Triage/Maintain permission

janv8000:
Syncs with current state at https://docs.github.com/en/rest/teams/teams#check-team-permissions-for-a-repository

Should fix https://github.com/octokit/octokit.net/issues/2041

### DIFF
--- a/Octokit/Models/Response/RepositoryPermissions.cs
+++ b/Octokit/Models/Response/RepositoryPermissions.cs
@@ -1,5 +1,5 @@
 ﻿using System.Diagnostics;
-using System.Globalization;
+using System;
 
 namespace Octokit
 {
@@ -8,10 +8,12 @@ namespace Octokit
     {
         public RepositoryPermissions() { }
 
-        public RepositoryPermissions(bool admin, bool push, bool pull)
+        public RepositoryPermissions(bool admin, bool maintain, bool push, bool triage, bool pull)
         {
             Admin = admin;
+            Maintain = maintain;
             Push = push;
+            Triage = triage;
             Pull = pull;
         }
 
@@ -21,9 +23,19 @@ namespace Octokit
         public bool Admin { get; protected set; }
 
         /// <summary>
+        /// Whether the current user has maintain permissions
+        /// </summary>
+        public bool Maintain { get; protected set; }
+
+        /// <summary>
         /// Whether the current user has push permissions
         /// </summary>
         public bool Push { get; protected set; }
+
+        /// <summary>
+        /// Whether the current user has triage permissions
+        /// </summary>
+        public bool Triage { get; protected set; }
 
         /// <summary>
         /// Whether the current user has pull permissions
@@ -32,7 +44,10 @@ namespace Octokit
 
         internal string DebuggerDisplay
         {
-            get { return string.Format(CultureInfo.InvariantCulture, "Admin: {0}, Push: {1}, Pull: {2}", Admin, Push, Pull); }
+            get
+            {
+                return FormattableString.Invariant($"Admin: {Admin}, Maintain: {Maintain}, Push: {Push}, Triage: {Triage}, Pull: {Pull}");
+            }
         }
     }
 }


sbest-gcm (approved):

Octokit/Models/Response/RepositoryPermissions.cs
```diff
@@ -20,19 +22,32 @@ public RepositoryPermissions(bool admin, bool push, bool pull)
         /// </summary>
         public bool Admin { get; protected set; }
 
+        /// <summary>
+        /// Whether the current user has maintain permissions
+        /// </summary>
+        public bool Maintain { get; protected set; }
```
nickfloyd, line 28: **Non-Blocker**: 
Would you mind changing all of these property accessors (on the `set`) to the more appropriate (given the use case) `private`? Using `protected` is currently the dominant pattern in this SDK but given that most of our models now implement `ctors` in one form or another so that makes it where I think we could change most of the protected access modifiers to the more appropriate private ones.

Let me know your thoughts - I'll be making a push to change all of the model's properties to use the private modifier soon so I am good with whatever you decide. Once done I'll be adding a theory to our [model tests](https://github.com/octokit/octokit.net/blob/main/Octokit.Tests.Conventions/ModelTests.cs) to help enforce this.

Octokit/Models/Response/RepositoryPermissions.cs
```diff
@@ -20,19 +22,32 @@ public RepositoryPermissions(bool admin, bool push, bool pull)
...
         /// Whether the current user has pull permissions
         /// </summary>
         public bool Pull { get; protected set; }
 
         internal string DebuggerDisplay
         {
-            get { return string.Format(CultureInfo.InvariantCulture, "Admin: {0}, Push: {1}, Pull: {2}", Admin, Push, Pull); }
+            get
+            {
+                return FormattableString.Invariant($"Admin: {Admin}, Maintain: {Maintain}, Push: {Push}, Triage: {Triage}, Pull: {Pull}");
```
nickfloyd, line 49: +1 thanks for the simplification here! 🎉

nickfloyd (approved):
Hey @janv8000!  Thank you so much for your contributions here! I left a comment on the setter accessors - make the changes if you'd like but it's not a blocker for this changeset.

Octokit/Models/Response/RepositoryPermissions.cs
janv8000: Sure thing, I wasn't sure whether the JSON deserializer in use supported private props.
I made them get-only in 40bdd8535ed6ef8a9303b9bf40b1533f6781ba40 or is that pushing the limits :D

nickfloyd: ✅ I could go for that... given C# 6  and greater these two are essentially equivalent when it comes to constructors but deviate when it comes to methods internal to the model needing to mutate the properties. I really don’t think there will ever be methods that mutate props on response models; that feels like a code smell - but if there is ever a need we can introduce private again. 

I think I like your omitted setter idea better than private for future generated response models. 🎉 Thanks for the thought and effort you put in here!  We should be good to merge this.

janv8000: I'm afraid I've been a bit too zealous.
While debugging this branch in my production code the ctor with the 5 arguments is never hit?

nickfloyd:
@janv8000 I'd like to get this change in for a release tomorrow - are you good with me merging this in or do you want to iterate on it more?

janv8000:
As I said in https://github.com/octokit/octokit.net/pull/2467#discussion_r911128250  I’m afraid there is some reflection magic going on somewhere I’m not aware of. 
Are you sure `get;` suffices for the simplejson serializer or does it really need `get; private set; `?

nickfloyd:
😬 oof, I completely missed that. So sorry. I'll take a look at this first thing tomorrow.

Octokit/Models/Response/RepositoryPermissions.cs
nickfloyd: Hey @janv8000, so I dug into this a bit more.  And it looks like you're correct.  We need to at minimum implement a setter with the private modifier.

`SimpleJson` makes heavy use of reflection to:
1. Get the [constructor](https://github.com/octokit/octokit.net/blob/main/Octokit/SimpleJson.cs#L1945) and a [delegate](https://github.com/octokit/octokit.net/blob/main/Octokit/SimpleJson.cs#L1958) for the constructor
2. Iterate through a dictionary (created from the incoming payload) and [assign values](https://github.com/octokit/octokit.net/blob/main/Octokit/SimpleJson.cs#L1495) to each given setter

This is why you only see the default constructor getting hit and not the parameter-based one on incoming payloads.  

 I am making an assumption here, but I believe that the parameterized constructor is primarily used for testing/mocking - because the delegate uses the memoized object map to make the assignments.

Let me know if that makes sense, and apologies again for missing your question the first time around.

Octokit/Models/Response/RepositoryPermissions.cs
```diff
@@ -8,31 +8,46 @@ public class RepositoryPermissions
...
             Push = push;
+            Triage = triage;
             Pull = pull;
         }
 
         /// <summary>
         /// Whether the current user has administrative permissions
         /// </summary>
-        public bool Admin { get; protected set; }
+        public bool Admin { get; }
```
nickfloyd, line 23:
```suggestion
        public bool Admin { get; private set; }
```

Octokit/Models/Response/RepositoryPermissions.cs
```diff
@@ -8,31 +8,46 @@ public class RepositoryPermissions
...
+
+        /// <summary>
+        /// Whether the current user has maintain permissions
+        /// </summary>
+        public bool Maintain { get; }
```
nickfloyd, line 28:
```suggestion
        public bool Maintain { get; private set; }
```

Octokit/Models/Response/RepositoryPermissions.cs
```diff
@@ -8,31 +8,46 @@ public class RepositoryPermissions
...
 
         /// <summary>
         /// Whether the current user has push permissions
         /// </summary>
-        public bool Push { get; protected set; }
+        public bool Push { get; }
```
nickfloyd, line 33:
```suggestion
        public bool Push { get; private set; }
```

Octokit/Models/Response/RepositoryPermissions.cs
```diff
@@ -8,31 +8,46 @@ public class RepositoryPermissions
...
+
+        /// <summary>
+        /// Whether the current user has triage permissions
+        /// </summary>
+        public bool Triage { get; }
```
nickfloyd, line 38:
```suggestion
        public bool Triage { get; private set; }
```

Octokit/Models/Response/RepositoryPermissions.cs
```diff
@@ -8,31 +8,46 @@ public class RepositoryPermissions
...
 
         /// <summary>
         /// Whether the current user has pull permissions
         /// </summary>
-        public bool Pull { get; protected set; }
+        public bool Pull { get; }
```
nickfloyd, line 43:
```suggestion
        public bool Pull { get; private set; }
```

nickfloyd (changes requested):
Just added comments on the things that need to change to make this work with the serializer.  Let me know your thoughts - i'd be glad to do these as well if you're unable to get to it.

Thanks again for the contributions ❇️

janv8000:
I made the requested changes + added a test for this convention in 94f23913f8832952a2d08b06eb83712f0b67bc83

nickfloyd (approved):

nickfloyd:
release_notes: Adds Triage/Maintain permission properties

PR #2468, opened 2022-06-29
Adds mono-complete to the dev container 

nickfloyd:
This dependency is needed to run the tests successfully from code spaces.

Note: prebuilds are enabled on main, so that if you create a branch from main you'll not have to worry about the spin-up time from docker generating the environment.

I tried other versions of mono from the Linux [packages](https://www.mono-project.com/download/stable/#download-lin) listed on mono-project and this seemed to be the only one that would not result in tests throwing errors. I'll need to take a sidebar to determine why mono complete is needed here.


nickfloyd:
release_notes: Adds mono-complete to the dev container

PR #2473, opened 2022-07-05
Adding in handling for secondary rate limit exceptions

### LINKED ISSUE #2466: "You have exceeded a secondary rate limit" does not throw RateLimitExceededException
As seen [here](https://github.com/octokit/octokit.net/blob/356588288e2d07fb4844911f6e03ef129540b124/Octokit/Http/Connection.cs#L722) when determining to throw a RateLimitExceededException it does not handle the case where it prints `You have exceeded a secondary rate limit. Please wait a few minutes before you try again.` and instead you get a `Octokit.ForbiddenException`

JonruAlveus:
fixes #2466

### DIFF
--- a/Octokit.Tests/Http/ConnectionTests.cs
+++ b/Octokit.Tests/Http/ConnectionTests.cs
@@ -180,7 +180,7 @@ namespace Octokit.Tests.Http
             }
 
             [Fact]
-            public async Task ThrowsRateLimitExceededExceptionForForbidderResponse()
+            public async Task ThrowsRateLimitExceededExceptionForForbiddenResponse()
             {
                 var httpClient = Substitute.For<IHttpClient>();
                 var response = CreateResponse(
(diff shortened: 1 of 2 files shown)


Octokit/Http/Connection.cs
```diff
@@ -719,7 +719,7 @@ static Exception GetExceptionForForbidden(IResponse response)
         {
             string body = response.Body as string ?? "";
 
-            if (body.Contains("rate limit exceeded"))
+            if (body.Contains("rate limit exceeded") || body.Contains("secondary rate limit"))
```
timrogers, line 722: I wonder if it might be better throw a different kind of exception for a secondary rate limit. What do you think? I'm just thinking that it would make it easier to handle this specific case as a caller, without doing string matching.

timrogers (commented):
Thanks for putting this PR together! ❤️

Octokit/Http/Connection.cs
JonruAlveus: I get where you are coming from, makes sense to me :)

timrogers: I’d just define a new exception and raise that. It should be a quick change, and then we can get this approved and merged ✅

JonruAlveus: I've changed it to be it's own exception.

timrogers (approved):
This looks good to me. Let me just tag in @nickfloyd who is more of a C# expert than me, and then we can get this merged hopefully. Thanks for your contribution ❤️

Octokit.Tests/Http/ConnectionTests.cs
```diff
@@ -180,7 +180,7 @@ public async Task ThrowsApiValidationExceptionFor422Response()
             }
 
             [Fact]
-            public async Task ThrowsRateLimitExceededExceptionForForbidderResponse()
```
Cook299, line 183: []()

Cook299 (commented):
I just learned to answer you I'll take it and run it on server see what it does get back to you.

nickfloyd:
release_notes: Adds handling for secondary rate limit exceptions from the GitHub REST API

PR #2477, opened 2022-07-14
Adding App slug tests

### LINKED ISSUE #2335: GitHubApp should include slug

JonruAlveus:
Fixes #2335 

Adds tests to cover the recently added github app slug field.


nickfloyd (approved):
Thank you so much for the follow-up on this!  You rock man! ❤️

PR #2479, opened 2022-07-14
Adding DetailsUrl to CheckRun

### LINKED ISSUE #2356: `DetailsUrl` property seems to be missing from `CheckRun`
According to the REST API docs a check run should have a `details_url` property in the body. However, it seems like the [`CheckRun`](https://github.com/octokit/octokit.net/blob/main/Octokit/Models/Response/CheckRun.cs) model class in this repo doesn't expose this property.

Am I looking in the wrong place or is this just missing? I'm happy to open a PR to add it.

JonruAlveus:
fixes #2356


nickfloyd (approved):
Thanks for the updates! 🎉

nickfloyd:
release_notes: Adds DetailsUrl to CheckRun

PR #2486, opened 2022-07-15
Trimming trailing slashes from user passed paths

### LINKED ISSUE #2243: GetAllContentsByRef doesn't properly handle trailing slash
Hi! We're using Octokit.NET in our service, and we found a bug where it doesn't look like it's properly handling trailing slashes in `GetAllContentsByRef`. The repo, commitId, and tree we use for testing is [vs-threading](https://github.com/microsoft/vs-threading/tree/1e187c0a93afb43f08cc36517b44b653a8934b79/src/IsolatedTestHost).

In our first test call, we call:
```cs
client.Repository.Content.GetAllContentsByRef("Microsoft", "vs-threading", "src/IsolatedTestHost", "1e187c0a93afb43f08cc36517b44b653a8934b79")
``` 
and we get the expected response of 4 files in the tree. When we call it again though but with a trailing slash on our path (like so):
```cs
client.Repository.Content.GetAllContentsByRef("Microsoft", "vs-threading", "src/IsolatedTestHost/", "1e187c0a93afb43f08cc36517b44b653a8934b79")
```
we get an additional file `App.Config` which does not exist at the commit specified, but does at a later commit. 

Looks like somehow the response is coming from the default branch?

### LINKED ISSUE #2390: Requesting repository contents by ref does not trim input parameters
The issue was observed on method specified below but would be also produced on some similar methods.

Task<IReadOnlyList<RepositoryContent>> GetAllContentsByRef(string owner, string name, string path, string reference);

Let's have repo "Test_Repo" of "Programmer" owner with following structure:
/src
--/back-end
----code.cs

Let's have 2 branches of this repo: main and dev.

I want to get contents of "src/back-end" folder in "dev" branch.
I call the method with following arguments: "Programmer", "Test_Repo", "src/back-end", "dev". 
OK, it returns actual file from dev branch.
But if I pass "src/back-end/" (path ends with slash) then I will get the file from main branch.

As traced with Fiddler, in the second scenario the built URL is ...src/back-end/?ref=dev and the lib sees that it is strange and performs request to check if such resource exists. Obviously it is not found and then it simply throws out ref parameter value and executes request for main branch.

Probably would be better to trim path parameter value in such case.

JonruAlveus:
Fixes #2390 
Fixes #2243 

Trimming trailing slashes from user passed paths that appear before query string parameters

### DIFF
--- a/Octokit.Tests.Integration/Clients/RepositoryContentsClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/RepositoryContentsClientTests.cs
@@ -216,7 +216,21 @@ namespace Octokit.Tests.Integration.Clients
                 var contents = await github
                     .Repository
                     .Content
-                    .GetAllContentsByRef("octokit", "octokit.net", "Octokit", "master");
+                    .GetAllContentsByRef("octokit", "octokit.net", "Octokit", "main");
+
+                Assert.True(contents.Count > 2);
+                Assert.Equal(ContentType.Dir, contents.First().Type);
+            }
+
+            [IntegrationTest]
+            public async Task GetsDirectoryContentWithTrailingSlash()
+            {
+                var github = Helper.GetAuthenticatedClient();
+
+                var contents = await github
+                    .Repository
+                    .Content
+                    .GetAllContentsByRef("octokit", "octokit.net", "Octokit/", "main");
 
                 Assert.True(contents.Count > 2);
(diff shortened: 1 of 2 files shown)


nickfloyd (approved):
Looks good!  Thanks for the contribution!

nickfloyd:
release_notes: Trims trailing slashes from user passed paths that appear before query string parameters

PR #2487, opened 2022-07-15
Update organizations-related `GetAllForUser` method docs to make it clear that only public memberships are returned

### LINKED ISSUE #2470: Unable to fetch Organisations list if membership is private
i have an  C# wpf Desktop APP  trying to fetch the organizations list i have access to or i'm a member of

This app has been added to trusted apps to my account and by the organization ( i own) also
Applications ->Authorised OAUth APPs ->MYAPP, its trusted by the Organization too.

Scopes used are : `user` and `read:org`  while fetching access token

But still using this token im unable to fetch full organizations list.

**Organizations in which my membership is Public are being retrieved**.

but for **Organizations in which my membership is Private are not retrieved**.

is this expected?

timrogers:
Fixes #2470.

### DIFF
--- a/Octokit/Clients/IOrganizationsClient.cs
+++ b/Octokit/Clients/IOrganizationsClient.cs
@@ -65,19 +65,27 @@ namespace Octokit
         Task<IReadOnlyList<Organization>> GetAllForCurrent(ApiOptions options);
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


nickfloyd (approved):
💥

PR #2501, opened 2022-07-20
Fix null reference access when ContentType is null

diogorolo:
When we receive an ApiException and the ContentType is null, if we try to log the message (or otherwise call the ToString method), we're throwing an NullReferenceException because HttpResponseBodySafe is accessing ContentType without checking if it's null.


nickfloyd (approved):
LGTM! Thanks for the test as well ❤️

nickfloyd:
release_notes: Prevents a null reference exception when ContentType is null on ApiException

PR #2519, opened 2022-08-02
move deprecation warning from subscribers_count to watchers_count.

### LINKED ISSUE #2509: Repository WatchersCount actually refers to Starred
The response for getting a repository contains a watchers_count property. This doesn't actually refer to watchers, but to Stars (unless stars add watchers?). I've created a repository, added a watch via the Activity.Watch client and the property is 0 when retrieved. If I start it, the property becomes 1.

I wonder if there is something wrong in the background as watchers_count seems to follow stargazers.

This shows what I mean, it's watched by 1 and starred by 0, but watchers_count and stargazers are both 0.
image

Brend-Smits:
Subscribers_count shows the actual watchers_count. Many people think they are retrieving the watchers_count, but are actually retrieving the stargazers_count...
This is because the GitHub UI is still calling subscribers, "Watchers".
To prevent many others from falling into this trap, a warning should be added to the watchers_count property.

On top of this, subscribers_count is no longer deprecated as far as the GitHub documentation goes.

Fixes #2509

Signed-off-by: Brend Smits <brend.smits@philips.com>

### DIFF
--- a/Octokit/Models/Response/Repository.cs
+++ b/Octokit/Models/Response/Repository.cs
@@ -108,6 +108,7 @@ namespace Octokit
 
         public int StargazersCount { get; protected set; }
 
+        [Obsolete("WatchersCount returns the same data as StargazersCount. You are likely looking to use SubscribersCount. Update your code to use SubscribersCount, as this field will stop containing data in the future")]
         public int WatchersCount { get; protected set; }
 
         public string DefaultBranch { get; protected set; }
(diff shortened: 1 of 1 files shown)


nickfloyd (approved):
@Brend-Smits thank you for adding this ❤️

nickfloyd:
release_notes: Updates deprecation warning from subscribers_count to watchers_count

PR #2520, opened 2022-08-03
[WIP] update code docs for SetRequestTimeout

### LINKED ISSUE #2506: Correct SetRequestTimeout code does to be more descriptive of what it's actually doing
A few spots in the code exist where `SetRequestTimeout` is used and documented. 

The timeout specified in the connection itself is only for the client connecting to the server. All this does is pass through the value you pass into `System.Net.Http.HttpClient` and assigns it to the `Timeout` property - which just sets the timespan to wait before the request itself times out.

We need to correct the code docs for this API on the GitHubClient object. The code comments ("Set the GitHub Api request timeout.") implies that this will set the timeout for the GitHub API itself ([which is capped at 10 seconds](https://docs.github.com/en/rest/overview/troubleshooting#timeouts)) and not the client timeout to the API.

This documentation should better represent what calling this method does - it sets the timeout for the connection between the client and server and does not assign the timeout of the GitHub REST API (it's a client connection timeout and not a deadline propagation).

notauserx:
updating code docs to reflect what SetConnectionTimeout does. Fixes #2506

### DIFF
--- a/Octokit.Reactive/IObservableGitHubClient.cs
+++ b/Octokit.Reactive/IObservableGitHubClient.cs
@@ -7,7 +7,8 @@ namespace Octokit.Reactive
         IConnection Connection { get; }
 
         /// <summary>
-        /// Set the GitHub Api request timeout.
+        /// Sets the timeout for the connection between the client and the server.
+        /// Github will terminate the request if it takes more than 10 seconds to process the request
         /// Useful to set a specific timeout for lengthy operations, such as uploading release assets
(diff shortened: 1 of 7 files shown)


notauserx:
#2520

notauserx:
My pleasure, @nickfloyd.

nickfloyd:
release_notes: Updates code docs for SetRequestTimeout to more appropriately reflect what the method does.

Issue #2500, opened 2022-07-19
Drop support for .NET Framework 4.6


JamieMagee:
Currently `Octokit` targets `netstandard2.0` and `net46`:

https://github.com/octokit/octokit.net/blob/73b7a7a18ce33862ca48892982d69d952d755df4/Octokit/Octokit.csproj#L9

.NET Framework 4.6 (and 4.6.1) went out of support on April 26 2022[^1]. Continuing to publish `Octokit` for this target adds additional support burden.

.NET standard 2.0 covers .NET Framework 4.6.2 already[^2].

Supporting a single target framework, .NET standard 2.0, would simplify maintenance of this project while still supporting all target frameworks that are currently supported upstream.

[^1]: https://docs.microsoft.com/en-us/lifecycle/products/microsoft-net-framework
[^2]: https://docs.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-0

eriawan:
I agree. We should remove support for .NET Framework before v4.7, to be in sync with .NET Framework lifecycle and related Windows support due to tight relationship of .NET Framework and Windows release cycle.

nickfloyd:
I agree with this 💯.  I'd like to get this done sooner than later. I've also marked this as a breaking change - since technically older applications will not be able to use the new version.

My only hesitation (which might not be valid at all) is that we are moving toward SDK [generation](https://github.com/octokit/octokit.net/discussions/2495#discussioncomment-3173935) soon with this SDK. This also means abstracting the models into a separate package so that this type of thing won't matter in the future.

I am thinking about consumers who have pre .NET Standard 2.0 apps -  if we made this move after getting generation in place, they would benefit from the current API set before we move forward.

The real question is, when would we be able to get generation done, and does it make sense to wait for it? While generation is not impossible(especially now that GitHub is shipping OpenAPI descriptions), generation is now even more achievable than it ever has been but there are still fringe challenges that might get in the way.

The question is: Which provides more value, removing legacy .NET support first or building generation, updating the models and then shipping that before we remove legacy support.

I'd love to hear yall's thoughts on this - I could easily be overthinking this. 

cc: @timrogers @JonruAlveus

JonruAlveus:
I too believe we should move up to Standard2. It opens up a whole world of new things (and challenges). We can look into different serialisers, newer code styles etc. But, as you say, what do we do with our legacy application users. I feel like we have a couple of options:
[...]

PR #2521, opened 2022-08-03
Drop support for .NET 4.6

nickfloyd:
Fixes: https://github.com/octokit/octokit.net/issues/2500

Already covered by the explanation via @JamieMagee - this needs to be done so we can move forward with the framework and innovate using the new APIs and assemblies.

Highlights from the [issue](https://github.com/octokit/octokit.net/issues/2500) description:
> .NET Framework 4.6 (and 4.6.1) went out of support on April 26, 2022. Continuing to publish Octokit for this target adds a support burden.

> .NET standard 2.0 covers .NET Framework 4.6.2 already.

> Supporting a single target framework, .NET standard 2.0, would simplify maintenance of this project while supporting all target frameworks currently supported upstream.

For reference, these are the supported versions going forward

.NET implementation | Version support
-- | --
.NET and .NET Core | 2.0, 2.1, 2.2, 3.0, 3.1, 5.0, 6.0
.NET Framework  | 4.6.1 2, 4.6.2, 4.7, 4.7.1, 4.7.2, 4.8

### DIFF
--- a/Octokit.Reactive/Octokit.Reactive.csproj
+++ b/Octokit.Reactive/Octokit.Reactive.csproj
@@ -37,7 +37,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="System.Reactive" Version="4.4.1" />
+    <PackageReference Include="System.Reactive" Version="5.0.0" />
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.1" PrivateAssets="All"/>
   </ItemGroup>
 </Project>
--- a/Octokit.Reactive/Octokit.Reactive.csproj
+++ b/Octokit.Reactive/Octokit.Reactive.csproj
@@ -6,7 +6,7 @@
     <Authors>GitHub</Authors>
     <Version>0.0.0-dev</Version>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
-    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+    <TargetFramework>netstandard2.0</TargetFramework>
     <AssemblyName>Octokit.Reactive</AssemblyName>
     <PackageId>Octokit.Reactive</PackageId>
     <DebugType>embedded</DebugType>
--- a/Octokit.Reactive/Octokit.Reactive.csproj
+++ b/Octokit.Reactive/Octokit.Reactive.csproj
@@ -6,7 +6,7 @@
     <Authors>GitHub</Authors>
     <Version>0.0.0-dev</Version>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
-    <TargetFrameworks>netstandard2.0;net46</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
     <AssemblyName>Octokit.Reactive</AssemblyName>
     <PackageId>Octokit.Reactive</PackageId>
     <DebugType>embedded</DebugType>
@@ -19,11 +19,8 @@
     <Copyright>Copyright GitHub 2017</Copyright>
   </PropertyGroup>
 
-  <PropertyGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
-    <NetStandardImplicitPackageVersion>2.0.0</NetStandardImplicitPackageVersion>
-  </PropertyGroup>
-
   <PropertyGroup>
+    <NetStandardImplicitPackageVersion>2.0.0</NetStandardImplicitPackageVersion>
     <NoWarn>1591;1701;1702;1705</NoWarn>
   </PropertyGroup>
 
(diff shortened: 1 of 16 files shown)


Octokit.Tests.Integration/Clients/RepositoriesClientTests.cs
```diff
@@ -536,7 +536,7 @@ public async Task UpdatesNameObsolete()
             using (var repoContext = await _github.CreateUserRepositoryContext())
             {
                 var updatedName = Helper.MakeNameWithTimestamp("updated-repo");
-                var update = new RepositoryUpdate(updatedName);
+                var update = new RepositoryUpdate() { Name = updatedName };
```
nickfloyd, line 539: This addressed a compiler warning - I wanted to ensure the output was as clean as possible and, therefore, clean a little bit of the house regardless of whether it was a related change or not.

Octokit.Reactive/Octokit.Reactive.csproj
```diff
@@ -6,7 +6,7 @@
     <Authors>GitHub</Authors>
     <Version>0.0.0-dev</Version>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
-    <TargetFrameworks>netstandard2.0;net46</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
```
JamieMagee, line 9:
```suggestion
    <TargetFramework>netstandard2.0</TargetFramework>
```

Octokit.Reactive/Octokit.Reactive.csproj
```diff
@@ -19,11 +19,8 @@
     <Copyright>Copyright GitHub 2017</Copyright>
   </PropertyGroup>
 
-  <PropertyGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
-    <NetStandardImplicitPackageVersion>2.0.0</NetStandardImplicitPackageVersion>
-  </PropertyGroup>
-
   <PropertyGroup>
+    <NetStandardImplicitPackageVersion>2.0.0</NetStandardImplicitPackageVersion>
```
JamieMagee, line 23: I don't believe this property is necessary anymore.

nickfloyd:
Packages inspected from CI.

Before Octokit
after-octokit

After Octokit
before-octokit

Before Octokit.Reactive
before

After Octokit.Reactive
after

nickfloyd:
cc: @eriawan @JonruAlveus

Octokit.Tests.Integration/Clients/RepositoriesClientTests.cs
JonruAlveus: These two tests were there to test the obsolete method, which technically does still work. The previous 2 tests in the file check the newer way of doing it, so if we don't want the warnings it would be better to get rid of those two tests (ending Obsolete) entirely 👍

Octokit/Octokit.csproj
```diff
@@ -19,26 +19,11 @@
     <Copyright>Copyright GitHub 2017</Copyright>
   </PropertyGroup>
 
-  <PropertyGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
-    <DefineConstants>$(DefineConstants);HAS_TYPEINFO;SIMPLE_JSON_INTERNAL;SIMPLE_JSON_OBJARRAYINTERNAL;SIMPLE_JSON_READONLY_COLLECTIONS;SIMPLE_JSON_TYPEINFO</DefineConstants>
-    <NetStandardImplicitPackageVersion>2.0.0</NetStandardImplicitPackageVersion>
-  </PropertyGroup>
-
-  <PropertyGroup Condition=" '$(TargetFramework)' == 'net46' ">
-    <DefineConstants>$(DefineConstants);HAS_ENVIRONMENT;HAS_REGEX_COMPILED_OPTIONS;SIMPLE_JSON_INTERNAL;SIMPLE_JSON_OBJARRAYINTERNAL;SIMPLE_JSON_READONLY_COLLECTIONS;HAS_SERVICEPOINTMANAGER</DefineConstants>
```
JonruAlveus, line 28: I think there are some usages of these constants in the code, should these be removed/tidied as well as part of this?

For example HAS_ENVIRONMENT has been removed entirely but is present/used in Connection.cs.

nickfloyd: Good highlight.  I can clean these up as well while I am in there.

Octokit.Tests.Integration/Clients/RepositoriesClientTests.cs
nickfloyd: Great point.  I'm thinking we also opt to remove the constructor as well given we are making a breaking change with this change set - and I can note it in the release.

Octokit/Octokit.csproj
nickfloyd: Scratch that.  There are implementations in the other projects that I'd like to take a closer look at - this is feeling like a separate changeset. I'm going to hold off so that we can do this separately - https://github.com/octokit/octokit.net/issues/2526

JonruAlveus (approved):
👍  Looks good to me!

JamieMagee (approved):

eriawan:
@nickfloyd 

Sorry to join late! Thanks for this PR, now we can move to newer .NET Framework version! 👍

nickfloyd:
release_notes: Drops support for .NET 4.6 and updates all projects to .NET Standard 2.0

PR #2532, opened 2022-08-07
Update dotnet supported version in readme

FrediKats:
Ref to https://github.com/octokit/octokit.net/releases/tag/v2.0.0


nickfloyd (approved):
Thanks for catching this and getting it updated! ❤️

nickfloyd:
release_notes:  Updated the .NET supported version in readme

Issue #2496, opened 2022-07-18
Code warnings and obsoletes should be removed


JonruAlveus:
We’ve just moved to v1, it would be lovely to clean up the Repo a bit:

* Fix any obsolete warnings
* Fix any other warnings
* Remove or fix any skipped tests

JonruAlveus:
All obsoletes warnings and other warnings are now gone. All unit tests pass.

Now to work on skipped tests.

JonruAlveus:
Skipped tests are all skipped for a reason so this is done.

PR #2535, opened 2022-08-09
Fixing the use of the binary formatter in net461 so all unit tests now pass.

JonruAlveus:
Also removed the need to have binary formatter tests run only in framework as it also works on netcoreapp3.1.
#2496


nickfloyd (approved):
Nice!  Solid contrib here @JonruAlveus, thank you ❤️.  I love cleaning the campground kind of things!

nickfloyd:
release_notes: Fixed the use of the binary formatter in net461 so all unit tests now pass. Also removed the need to have it run only in framework as it also works on netcoreapp3.1

Issue #2534, opened 2022-08-08
Add generators project for current and future generative features


nickfloyd:
related: https://github.com/octokit/octokit.net/pull/2516

From the reference PR

Given that we are headed down to a more generative SDK  (see the discussions https://github.com/octokit/octokit.net/discussions/2499 and https://github.com/octokit/octokit.net/discussions/2495 and https://github.com/octokit/octokit.net/issues/2517 for more info on that)

IMO the preferred approach would be to have a generators project with namespaced scrips i.e.
`Octokit.Generators` and the generator above would be something like `Octokit.Generators.AsyncPaginationExtension` (as an example).

This way, we're establishing the future location for all generators to remove this question later. Also, this will allow us to use CI to build the generators, generate the code, and commit it before the other aspects of the SDK are built.

As a side note, in my perfect world brain, all models and related libs would be hoisted out of this SDK and referenced in as a package so that we could better support versioning and be able to ship SDK core changes separate from SDK model changes.

Let's create a generators project, add the [script](https://github.com/octokit/octokit.net/pull/2516#issue-1321664302) detailed in the PR, then document how to run and what to expect - we can do a follow-up for CI to implement this more automatedly.

PR #2537, opened 2022-08-10
Add generators project for source generation

nickfloyd:
Fixes: https://github.com/octokit/octokit.net/issues/2534

This PR intends to add an iteration 0 for code generation. The initial generator was sourced from the work @Saalvage did when adding the [AsyncPaginationExtensions](https://github.com/octokit/octokit.net/pull/2516).

This PR:
- [x] Adds the generator project
- [x] Adds the first generator - AsyncPaginationExtensions (initial approach - future iterations will dive into using things such as Rosyln and source templates
- [x] Adds a readme explaining how to use this project and why we are doing it
- [x] Regenerates the extensions - there are deltas due to white space, some template modification, and ordinality of the generated implementations

Note: The existing test coverage for the extensions is still intact, and I intentionally left testing off of the generations project because:
1. It was redundant to test the generated extensions
2. This project is effectively a prototype for code generation and is proven through compiling the source it has generated

### DIFF
--- a/Octokit.Generators/AsyncPaginationExtensionsGenerator.cs
+++ b/Octokit.Generators/AsyncPaginationExtensionsGenerator.cs
@@ -1,0 +1,119 @@
+﻿using System.Linq;
+using System.IO;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+using System.Data;
+using System.Text;
+
+namespace Octokit.Generators
+{
+
+  /// <summary>
+  /// AsyncPaginationExtensionsGenerator for generating pagination extensions for Octokit.net Clients that return collections.
+  /// </summary>
+  /// <remarks>
+  /// This generator originally appeared in https://github.com/octokit/octokit.net/pull/2516
+  /// The generator solves a small part of a larger effort that is being discussed:
+  ///  https://github.com/octokit/octokit.net/discussions/2499
+  ///  https://github.com/octokit/octokit.net/discussions/2495
+  ///  https://github.com/octokit/octokit.net/issues/2517
+  ///  In the future, we should be able to unify generation for
+  ///   * models (request and response)
+  ///   * clients
+  ///   * routing and related helpers
+  ///   TODO: Convert to use Rosyln source generators
+  /// </remarks>
+  class AsyncPaginationExtensionsGenerator
+  {
+
+  private const string HEADER = (
+@"using System;
+using System.Collections.Generic;
+
+namespace Octokit.AsyncPaginationExtension
+{
+  /// <summary>
+  /// Provides all extensions for pagination.
+  /// </summary>
+  /// <remarks>
+  /// The <code>pageSize</code> parameter at the end of all methods allows for specifying the amount of elements to be fetched per page.
+  /// Only useful to optimize the amount of API calls made.
+  /// </remarks>
+  public static class Extensions
+  {
+    private const int DEFAULT_PAGE_SIZE = 30;
+  ");
+
+  private const string FOOTER = (
+@"
+  }
+}");
+
+    /// <summary>
+    /// GenerateAsync static entry point for generating pagination extensions.
+    /// </summary>
+    /// <remarks>
+    /// This defaults the search path to the root of the project
+    /// This expects to generate the resulting code and put it in Octokit.AsyncPaginationExtension
+    /// This does a wholesale overwrite on ./Octokit.AsyncPaginationExtension/Extensions.cs
+    /// </remarks>
+    public static async Task GenerateAsync(string root = "./")
+    {
+      var sb = new StringBuilder(HEADER);
+      var enumOptions = new EnumerationOptions { RecurseSubdirectories = true };
+      var paginatedCallRegex = new Regex(@".*Task<IReadOnlyList<(?<returnType>\w+)>>\s*(?<name>\w+)(?<template><.*>)?\((?<arg>.*?)(, )?ApiOptions \w*\);");
+
+      foreach (var file in Directory.EnumerateFiles(root, "I*.cs", enumOptions)) {
+          var type = Path.GetFileNameWithoutExtension(file);
+
+          foreach (var line in File.ReadAllLines(file)) {
+              var match = paginatedCallRegex.Match(line);
+
+              if (!match.Success) { continue; }
+              sb.Append(BuildBodyFromTemplate(match, type));
+          }
+      }
+
+      sb.Append(FOOTER);
+
+      await File.WriteAllTextAsync("./Octokit.AsyncPaginationExtension/Extensions.cs", sb.ToString());
+    }
+
+    /// <summary>
+    /// BuildBodyFromTemplate uses the match from the regex search and parses values from the given source
+    /// to use to generate the paging implementations.
+    /// </summary>
+    /// <remarks>
+    /// TODO: This should be reworked to use source templates
+    /// </remarks>
+    private static string BuildBodyFromTemplate(Match match, string type)
+    {
+      var argSplitRegex = new Regex(@" (?![^<]*>)");
+      var returnType = match.Groups["returnType"].Value;
+      var name = match.Groups["name"].Value;
+      var arg = match.Groups["arg"].Value;
+      var template = match.Groups["template"];
+      var templateStr = template.Success ? template.Value : string.Empty;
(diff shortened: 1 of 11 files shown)


Octokit.Generators/README.md
```diff
@@ -0,0 +1,58 @@
+# Octokit.net Generators
+
+## About
+
+We've been discussing and thinking about code generators for a [while now](https://github.com/octokit/octokit.net/discussions/2527) and they are part of the [future vision](https://github.com/octokit/octokit.net/discussions/2495) for where we'd like to head with this SDK.
+
+Consider this to be iteration 0.  Meaning while we want to potentially move to source generators, templates, and using the features that Rosyln offers we need to do some functional experiments and solve our current needs while iterating on the future.  Acknoledging that code generation is a solved problem (menaing there is existing work out there) we should be intentional with the direction we take.
```
JonruAlveus, line 7: Acknoledging > Acknowledging
Menaing > meaning

Octokit.Generators/Octokit.Generators.csproj
```diff
@@ -0,0 +1,23 @@
...
+    <DebugType>embedded</DebugType>
+    <RepositoryUrl>https://github.com/octokit/octokit.net</RepositoryUrl>
+    <PackageProjectUrl>https://github.com/octokit/octokit.net</PackageProjectUrl>
+    <PackageIconUrl>https://f.cloud.github.com/assets/19977/1510987/64af2b26-4a9d-11e3-89fc-96a185171c75.png</PackageIconUrl>
+    <PackageIcon>octokit.png</PackageIcon>
+    <PackageLicenseExpression>MIT</PackageLicenseExpression>
+    <PackageTags>GitHub API Octokit dotnetcore dotnetstandard2.0</PackageTags>
+    <Copyright>Copyright GitHub 2022</Copyright>
+    <LangVersion>9</LangVersion>
+	  <Nullable>enable</Nullable>
```
JonruAlveus, line 21: Extra space. Is nullable applicable to net 5?

Octokit.Generators/Octokit.Generators.csproj
```diff
@@ -0,0 +1,23 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <Description>A set of code generators for Octokit.NET backed by the GitHub REST API Open API descriptions</Description>
+    <AssemblyTitle>Octokit.Generators</AssemblyTitle>
+    <Authors>GitHub</Authors>
+    <Version>0.0.0-dev</Version>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net5.0</TargetFramework>
```
JonruAlveus, line 9: Why net 5 and not 6? 5 isn’t LTS.

Octokit.Generators/Octokit.Generators.csproj
```diff
@@ -0,0 +1,23 @@
...
+    <AssemblyName>Octokit.Generators</AssemblyName>
+    <PackageId>Octokit.Generators</PackageId>
+    <DebugType>embedded</DebugType>
+    <RepositoryUrl>https://github.com/octokit/octokit.net</RepositoryUrl>
+    <PackageProjectUrl>https://github.com/octokit/octokit.net</PackageProjectUrl>
```
JonruAlveus, line 14: Does this project need package details as I assume it’s not getting packed up?

Octokit.Generators/Generator.cs
```diff
@@ -0,0 +1,29 @@
...
+  /// Provides an entry point for code generation of various types.
+  /// </summary>
+  /// <remarks>
+  /// The backing source for generation will either be the source files in this repo or
+  /// the OpenAPI Descriptions from the GitHub REST API: https://github.com/github/rest-api-description
+  /// </remarks>
+  class Generator
+    {
+
+        static void Main(string[] args)
```
JonruAlveus, line 15: Async Task Main?

Octokit.Generators/AsyncPaginationExtensionsGenerator.cs
```diff
@@ -0,0 +1,119 @@
...
+    /// This defaults the search path to the root of the project
+    /// This expects to generate the resulting code and put it in Octokit.AsyncPaginationExtension
+    /// This does a wholesale overwrite on ./Octokit.AsyncPaginationExtension/Extensions.cs
+    /// </remarks>
+    public static async Task GenerateAsync(string root = "./")
+    {
+      var sb = new StringBuilder(HEADER);
+      var enumOptions = new EnumerationOptions { RecurseSubdirectories = true };
+      var paginatedCallRegex = new Regex(@".*Task<IReadOnlyList<(?<returnType>\w+)>>\s*(?<name>\w+)(?<template><.*>)?\((?<arg>.*?)(, )?ApiOptions \w*\);");
+
```
JonruAlveus, line 65: What’s the rules on { on same line or new line?

Octokit.Generators/Octokit.Generators.csproj
nickfloyd: Oof, a tab got in there somehow... good catch.  Yeah, it is, for .NET 5 lang version 9.  I prefer the `?` preceding syntax - and since all of the other projects have this set I just decided to use it here.

Octokit.Generators/Octokit.Generators.csproj
nickfloyd: I did this primarily because I didn't want to add another SDK runtime to the container (see ci build [output](https://github.com/octokit/octokit.net/runs/7758616506?check_suite_focus=true)).  Looking at it now - that does not seem like a great reason given that I enabled prebuilds on this repo. 

I'll add the SDK and change it.

Octokit.Generators/Octokit.Generators.csproj
nickfloyd: So this is me thinking forward... I am still not sold as this being part of the .NET SDK itself, so I added that in there as a placeholder just in case we want to ship this as a package... I could remove it or leave it - I can see how it's a little extra cruft.

Octokit.Generators/AsyncPaginationExtensionsGenerator.cs
nickfloyd: We should define that and then add a linter.  I favor concise over verbose unless it makes good sense to be expressive - so if it can be a clean one like expression, why not? That's my opinion, though.

Octokit.Generators/Generator.cs
nickfloyd: I didn't see a good reason to make that jump right now... it wouldn't provide any value IMO. I could go either way.

Octokit.Generators/Octokit.Generators.csproj
nickfloyd: Yeah, I just verified.. adding it in really didn't impact the CI build time. I've added .NET 6.0 to the workflows and changed the project ref.

Octokit.Generators/Octokit.Generators.csproj
JonruAlveus: 👍

Octokit.Generators/Octokit.Generators.csproj
JonruAlveus: Ahh yep sorry, didn’t think of that! Definitely a good thing to have even if it’s just there to make life easier in the future. ❤️

Octokit.Generators/Generator.cs
JonruAlveus: 👍 I'm easy either way.

Octokit.Generators/AsyncPaginationExtensionsGenerator.cs
JonruAlveus: I’m happy with anything really as long as it’s the same throughout. I’ve found editorconfig files quite handy but we’d need to make sure it doesn’t fight with Resharper settings 😀

Octokit.Generators/Octokit.Generators.csproj
JonruAlveus: ❤️

JonruAlveus (approved):

nickfloyd:
release_notes: Adds generators project and AsyncPaginationExtension generator

PR #2538, opened 2022-08-10
Tidying up DefineConstants

### LINKED ISSUE #2526: Review/Cleanup DefineConstants in the projects
There seems to be a mismatch between what is defined in the `< DefineConstants>` nodes in the csproj files from what is being used in the code. We need to look at the following projects and determine what env constants are still being used:

- [ ] Octokit
- [ ] Octokit.Tests
- [ ] Octokit.Tests.Integration

JonruAlveus:
fixes #2526 
- [X] GITHUBJWT_HELPER_AVAILABLE - Removed as we now target netstandard2.0 and so does the github jwt package
- [X] NET_45 - Removed as nothing uses it
- [X] NO_SERIALIZABLE - Remove as it's for SimpleJson and it requires the Serializable attribute to work in netstandard.
- [X] HAS_DEFAULT_ENCODING - Removed as Encoding.GetEncoding(0) is the netstandard/netcore way of doing things
- [X] HAS_TYPEINFO - Removed as netstandard's System.Reflection supports both options. Retaining the version of the code which uses GetTypeInfo()
- [X] HAS_REGEX_COMPILED_OPTIONS - Keep as it doesn't appear to be to do with the .net version in use
- [X] HAS_ENVIRONMENT - Removed as netstandard supports both options. Changed to use the Environment version as it gives more info
- [X] HAS_SERVICEPOINTMANAGER - Removed as netstandard and net462 has TLS1.2 enabled by default (or system defaults). In the future, if TLS1.3 becomes the requirements, we will have to change code to handle this, which isn't desirable
- [X] NETFX_CORE - Don't want to touch simplejson!
- [X] SIMPLE_JSON_NO_LINQ_EXPRESSION - Don't want to touch simplejson!
- [X] SIMPLE_JSON_DYNAMIC - Don't want to touch simplejson!
- [X] SIMPLE_JSON_OBJARRAYINTERNAL - Don't want to touch simplejson!
- [X] SIMPLE_JSON_INTERNAL - Don't want to touch simplejson!
- [X] SIMPLE_JSON_READONLY_COLLECTIONS - Don't want to touch simplejson!
- [X] SIMPLE_JSON_REFLECTION_UTILS_PUBLIC - Don't want to touch simplejson!
- [X] SIMPLE_JSON_TYPEINFO - Don't want to touch simplejson!


nickfloyd (approved):
Nice! ❤️

nickfloyd:
release_notes: Cleans up unused or unnecessary DefineConstants

PR #2540, opened 2022-08-11
Adding the SHA that a CommitFile was committed against, parsed from RAW url

### LINKED ISSUE #2411: Question: compare branches and get files changed for each commit
Hello,

I'm trying to compare 2 branches - so far pretty easy with `/repos/{owner}/{repo}/compare/{basehead}`

but for some reason, the returned result comes with a list of commits and a list of files changed separated,
and i can't seem to be able to connect between them by any identifier. 
i would like for each commit to link the associated file changes. 
is it possible?

the only solution i found is for each commit to have a separate api call like so:
`/repos/{owner}/{repo}/commits/{ref}` but it seems very inefficient.

JonruAlveus:
fixes #2411

### DIFF
--- a/Octokit/Models/Response/GitHubCommitFile.cs
+++ b/Octokit/Models/Response/GitHubCommitFile.cs
@@ -86,6 +86,27 @@ namespace Octokit
         [Parameter(Key = "previous_filename")]
         public string PreviousFileName { get; protected set; }
 
+        /// <summary>
+        /// The SHA the file was commited in
+        /// </summary>
+        public string CommitSha
+        {
+            get
+            {
+                if (RawUrl != null)
+                {
+                    var splitItems = RawUrl.Split(new[] { '/' }, System.StringSplitOptions.RemoveEmptyEntries);
+                    
+                    if (splitItems.Length > 2)
+                    {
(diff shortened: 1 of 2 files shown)


Octokit/Models/Response/GitHubCommitFile.cs
```diff
@@ -86,6 +86,27 @@ public GitHubCommitFile(string filename, int additions, int deletions, int chang
         [Parameter(Key = "previous_filename")]
         public string PreviousFileName { get; protected set; }
 
+        /// <summary>
+        /// The SHA the file was commited in
+        /// </summary>
+        public string CommitSha
```
nickfloyd, line 92: This feels like it would be more appropriate as an extension or helper.  If possible, I'd like to keep composite and logic-filled properties out of the models.  Otherwise, we will have to develop a more complex / override-based generation method.

My preference would be that if we are going to provide this type of helper that it be hoisted out of the model itself or in the related [question](https://github.com/octokit/octokit.net/issues/2411) you offer it as a solution that could be implemented in the user's app itself.

nickfloyd (changes requested):
Just a comment/thought around keeping logic out of our model properties  - let me know what you think.

PR #2551, opened 2022-08-24
Adding Packages Api

### LINKED ISSUE #2360: Packages API
Hi, it looks like this library doesn't implement the [packages API](https://docs.github.com/en/rest/reference/packages)?

For example I wanted to delete some packages (which github doesn't support through `dotnet nuget delete`)

JonruAlveus:
fixes #2360

### DIFF
--- a/Octokit.Reactive/Clients/IObservablePackagesClient.cs
+++ b/Octokit.Reactive/Clients/IObservablePackagesClient.cs
@@ -7,12 +7,132 @@ namespace Octokit.Reactive
     {
         IObservablePackageVersionsClient PackageVersions { get; }
 
-        IObservable<Unit> Delete(string org, PackageType packageType, string packageName);
-        
-        IObservable<Package> Get(string org, PackageType packageType, string packageName);
-        
-        IObservable<Package> GetAll(string org, PackageType packageType, PackageVisibility? packageVisibility = null);
-        
-        IObservable<Unit> Restore(string org, PackageType packageType, string packageName);
+        /// <summary>
+        /// List all packages for an organisations, readable by the current user
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="https://docs.github.com/rest/packages#list-packages-for-an-organization">API documentation</a> for more details
+        /// </remarks>
+        /// <param name="org">Required: Organisation Name</param>
+        /// <param name="packageType">Required: The type of package</param>
+        /// <param name="packageVisibility">Optional: The visibility of the package</param>
+        IObservable<Package> GetAllForOrg(string org, PackageType packageType, PackageVisibility? packageVisibility = null);
+
+        /// <summary>
+        /// Get a specific package for an Organization.
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="https://docs.github.com/rest/packages#get-a-package-for-an-organization">API documentation</a> for more details
+        /// </remarks>
+        /// <param name="org">Required: Organisation Name</param>
+        /// <param name="packageType">Required: The type of package</param>
+        /// <param name="packageName">Required: The name of the package</param>
+        IObservable<Package> GetForOrg(string org, PackageType packageType, string packageName);
+
+        /// <summary>
+        /// Delete a specific package for an Organization.
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="https://docs.github.com/rest/packages#delete-a-package-for-an-organization">API documentation</a> for more details
+        /// </remarks>
+        /// <param name="org">Required: Organisation Name</param>
+        /// <param name="packageType">Required: The type of package</param>
+        /// <param name="packageName">Required: The name of the package</param>
+        IObservable<Unit> DeleteForOrg(string org, PackageType packageType, string packageName);
+
+        /// <summary>
+        /// Restore a specific package for an Organization.
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="https://docs.github.com/rest/packages#restore-a-package-for-an-organization">API documentation</a> for more details
+        /// </remarks>
+        /// <param name="org">Required: Organisation Name</param>
+        /// <param name="packageType">Required: The type of package</param>
+        /// <param name="packageName">Required: The name of the package</param>
+        IObservable<Unit> RestoreForOrg(string org, PackageType packageType, string packageName);
+
+        /// <summary>
+        /// Lists packages owned by the authenticated user within the user's namespace
+        /// </summary>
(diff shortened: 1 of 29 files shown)


Octokit/IGitHubClient.cs
```diff
@@ -87,6 +87,14 @@ public interface IGitHubClient : IApiInfoProvider
         /// </remarks>
         IOrganizationsClient Organization { get; }
 
+        /// <summary>
+        /// Access GitHub's Pacakges API.
```
timrogers, line 91: *Packages

Octokit/Helpers/ApiUrls.cs
```diff
@@ -4457,5 +4457,19 @@ public static Uri Meta()
         {
             return "meta".FormatUri();
         }
+
+        /// <summary>
+        /// Returns the <see cref="Uri"/> for the Pacakges GET request
```
timrogers: *Packages

Octokit/GitHubClient.cs
```diff
@@ -225,6 +226,14 @@ public Uri BaseAddress
         /// </remarks>
         public IOrganizationsClient Organization { get; private set; }
 
+        /// <summary>
+        /// Access GitHub's Pacakges API.
```
timrogers, line 230: *Packages

Octokit/Clients/IPackagesClient.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Octokit
+{
+    public interface IPackagesClient
+    {
+        Task<IReadOnlyList<Package>> List(string org, PackageType packageType, PackageVisibility? packageVisibility = null);
```
timrogers: Do you think `List` is the right name here? Should we use `ListForOrganization` or something like that, given that there are other list APIs?

Octokit.Reactive/Clients/ObservablePackagesClient.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// </summary>
+        /// <remarks>
+        /// See the <a href="https://docs.github.com/rest/packages#delete-a-package-for-an-organization">API documentation</a> for more details
+        /// </remarks>
+        /// <param name="org">Required: Organisation Name</param>
+        /// <param name="packageType">Required: The type of package</param>
+        /// <param name="packageName">Required: The name of the package</param>
+        public IObservable<Unit> Delete(string org, PackageType packageType, string packageName)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(org, nameof(org));
```
timrogers, line 72: Should we be enforcing that the `packageName` is not null or empty, as we do for the `org`?

Octokit.Reactive/Clients/ObservablePackagesClient.cs
```diff
@@ -32,8 +30,12 @@ public ObservablePackagesClient(IGitHubClient client)
         public IObservable<Package> GetAll(string org, PackageType packageType, PackageVisibility? packageVisibility = null)
         {
             Ensure.ArgumentNotNullOrEmptyString(org, nameof(org));
+            Ensure.ArgumentNotNull(packageType, nameof(packageType));
```
timrogers, line 35: C# newbie question: when the method is declared with argument `PackageType packageType`, is it implicitly possible to pass null?

timrogers (approved):
Thanks for taking the time to add this ❤️  I've reviewed commit-by-commit and it looks good to me from an API and basic implementation perspective, but I'm not a .NET expert, so let's wait for @nickfloyd to return from holiday before we merge (if that's okay!).

Octokit.Reactive/Clients/ObservablePackagesClient.cs
JonruAlveus: 👍

Octokit.Reactive/Clients/ObservablePackagesClient.cs
JonruAlveus: No probs! No, an enum is a value type (like an int) so it can only be null if it’s specified with a ?

I had to look that one up and I’ve been using c# for around a decade 😨

Octokit/IGitHubClient.cs
JonruAlveus: 👍

Octokit/Helpers/ApiUrls.cs
JonruAlveus: 👍

Octokit/GitHubClient.cs
JonruAlveus: 👍

Octokit/Clients/IPackagesClient.cs
JonruAlveus: Yeah it wasn't the best name. I believe I changed it in later commits to "GetAllForOrg" and so on for the various combinations.

Octokit.Reactive/Clients/IObservablePackagesClient.cs
```diff
@@ -0,0 +1,138 @@
+﻿using System;
+using System.Reactive;
+
+namespace Octokit.Reactive
+{
+    public interface IObservablePackagesClient
+    {
+        IObservablePackageVersionsClient PackageVersions { get; }
```
nickfloyd, line 8: So this is one of those weird API domains where package versions are a subset of packages, yet a package version is a package.

What you've done here feels like it works - I'm glad you approached the relationship this way to help reduce namespace confusion - I think leaving PackageVersions out of this interface would've made things a bit more clumsy.

```
github.Packages.PackageVersions.GetAllForUser
github.Packages.GetAllForOrg
```

feels more elegant than something like (though possible):

```
github.PackageVersions.GetAllForUser
github.Packages.GetAllForOrg
```

While, as you've implemented, it can lead to a bit more verbose API call, I think the expression feels more natural ❤️  solid work here!

nickfloyd (approved):
Yeah this all looks fantastic @JonruAlveus - other than the typos, this looks g2g! :heart:

Octokit.Reactive/Clients/IObservablePackagesClient.cs
JonruAlveus: Thanks! It is a little odd. I took my inspiration from the GitHub Api docs where it's all grouped together in one Packages section.

I've just pushed a final fix for the typos also.

nickfloyd:
release_notes: Added Packages and Package versions APIs

PR #2553, opened 2022-08-25
Add ability to pass ApiOptions to PullRequestsClient.Files

### LINKED ISSUE #2552: Ability to specify per_page for PullRequest.Files?
Is it somehow possible to specify the per_page parameter when calling PullRequest.Files?
By default Github only returns 30 files per page but could return up to 100.
This means when retrieving a list of modified files on a large PR a lot of unneeded API calls are being made.

Apparently other GetAll methods allow you to pass an ApiOptions parameter but it doesn't look like PullRequest.Files has an overload for that.

Alternatively would it be possible for PullRequest.Files to include per_page set to 100 by default?
I don't see any reason why one wouldn't want that method to use as little API calls as possible.

Gamer025:
This fixes #2552 by allowing you to pass an ApiOptions object to the Files method of PullRequestsClient.
Files can still be called without ApiOptions and will simply pass on ApiOptions.None then.

### DIFF
--- a/Octokit.Reactive/Clients/IObservablePullRequestsClient.cs
+++ b/Octokit.Reactive/Clients/IObservablePullRequestsClient.cs
@@ -238,7 +238,26 @@ namespace Octokit.Reactive
         /// <param name="owner">The owner of the repository</param>
         /// <param name="name">The name of the repository</param>
         /// <param name="number">The pull request number</param>
+        /// <param name="options">Options for changing the API response</param>
+        IObservable<PullRequestFile> Files(string owner, string name, int number, ApiOptions options);
+
+        /// <summary>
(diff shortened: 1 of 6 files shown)


timrogers (approved):
Thank you so much for taking the time to contribute to Octokit! I really appreciate it ✨ This looks good to me - but I'm not a .NET expert by any means. Before we merge, I'd love @nickfloyd (currently on vacation) or @JonruAlveus to review as well.

timrogers:
Fixes #2552.

JonruAlveus (approved):
Looks good to me! ⭐️

nickfloyd:
release_notes: Adds the ability to pass ApiOptions to PullRequestsClient.Files

Issue #2472, opened 2022-06-30
Standardize response model property accessors


nickfloyd:
As we start making our way toward generating models based on the OpenAPI definitions from the GitHub REST API we need to do some pre-work to help us know what the result of the generated code might look like.  This begins with standardization, establishing patterns, and approaches to the way we build our models.

One area where we need to do this is with our [response models](https://github.com/octokit/octokit.net/tree/main/Octokit/Models/Response).  Given that most of them have constructors now we should be able to move to a standardized way of writing these models.  Considering the following:

- Some models use `protected`, `private`, or no modifier on the setters
- Some models are missing constructors
- Some models implement `DebuggerDisplay` property that has been implemented in a few different ways.

I recommend we standardize on the following.

- We still use DebuggerDisplay
- There should be two public constructors one that is empty and the other setting all the properties from parameters being passed in; this allows the construction of an empty object as well as providing a construction contract.
- Setters will be completely removed favoring an omission pattern over property accessors - this clearly communicates the intent of the model itself - that it is only ever hydrated from a source once and should never mutate (that's what the request objects are for)
- The DebuggerDisplay property can be simplified using (assuming a non static model) - ` get { return JsonSerializer.Serialize(this);}`.  This eliminates the need for having to update it every time properties are introduced, will provide the same shape as the actual model, and provides a serialized version during debugging that can be used for various dev needs.

```
using System.Text.Json;
using System.Diagnostics;

namespace Octokit
{
    [DebuggerDisplay("{DebuggerDisplay,nq}")]
    public class SourceInfo
    {
        public SourceInfo() { }

        public SourceInfo(User actor, int id, Issue issue, string url)
        {
            Actor = actor;
            Id = id;
            Issue = issue;
            Url = url;
        }

        public User Actor { get; }
        public int Id { get; }
        public Issue Issue { get; }
        public string Url { get; }

        internal string DebuggerDisplay
        {
            get { return JsonSerializer.Serialize(this);}
        }
    }
}
```

JonruAlveus:
Which JsonSerializer are you referring to? I can't see any static serializers currently available in the repo (and system.text.json requires .net 4.6.1).

JonruAlveus:
Changing this test to throw when models have non private setters (or none) reveals... only a small amount of work!

![image](https://user-images.githubusercontent.com/15954362/177265887-f6401c72-62ec-4392-87a1-89763162df31.png)

[...]

PR #2565, opened 2022-09-08
Convert response model property accessors from protected to the more appropriate private for consistency

nickfloyd:
From the [issue](https://github.com/octokit/octokit.net/issues/2472):

> As we start making our way toward generating models based on the OpenAPI definitions from the GitHub REST API we need to do some pre-work to help us know what the result of the generated code might look like. This begins with standardization, establishing patterns, and approaches to building our models.

> One area where we need to do this is with our [response models](https://github.com/octokit/octokit.net/tree/main/Octokit/Models/Response). Given that most of them have constructors now we should be able to move to a standardized way of writing these models.

Notes about the changes in this PR:
- The existing serialization and hydration tests cover this changeset
- Common and request models will be modified in a separate PR if needed
- Some object's properties could not be changed due to inheritance (listed below)

The following models remain unchanged due to inheritance constraints (where the `protected` accessor is required).  We can evaluate these relationships either prior or post generation (depending on of we approach the generation byte for byte or if it's a complete rewrite):
- Account
- Author
- Authorization
- CommitContent
- GitReference
- InstallationId
- LicenseMetadata
- SerchResults

### DIFF
--- a/Octokit/Models/Response/AccessToken.cs
+++ b/Octokit/Models/Response/AccessToken.cs
@@ -18,12 +18,12 @@ namespace Octokit
         /// <summary>
         /// The access token
         /// </summary>
-        public string Token { get; protected set; }
+        public string Token { get; private set; }
 
         /// <summary>
         /// The expiration date
         /// </summary>
-        public DateTimeOffset ExpiresAt { get; protected set; }
+        public DateTimeOffset ExpiresAt { get; private set; }
 
         internal string DebuggerDisplay
         {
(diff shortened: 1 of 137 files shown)


timrogers:
Would these visibility changes be breaking?

nickfloyd:
> Would these visibility changes be breaking?

Not at all.  They are the setters that should be internal to the model, and the fields already become immutable after hydration.  So no breaking changes are happening here.

timrogers (approved):

PR #2571, opened 2022-09-15
Update some docs links

### LINKED ISSUE #2474: [DOCS] Update all links to developer.github.com in docs and code comments to the more accurate docs.github.com
It's been a while since we've updated the documentation for octokit.net (*.md). The hostname for the docs site on GitHub has since changed from `developer.github.com` to `docs.github.com` so all of the references need to change given the redirects might not always be there and might not currently be accurate.

Things to watch out for:
- This also includes the v3 URI segment - in most cases that can be removed for docs
- There are close to 4000 lines that have a string segment of `developer.github.com/v3` - most of them currently redirect property but we'll need to be more cautious around the more complex URLs
- Make sure to sanitize out localization - en, `https://docs.github.com/en/rest/issues` should be `https://docs.github.com/rest/issues` etc... the docs site should handle all of that

SeanKilleen:
Supports #2474.

This PR:

* Converts the links in markdown
* Removes the Beta packages section of the README, which had bad links
* Moves all current docs.github.com links to be `https`.
* Removes `/en` from existing routes.

## Remaining Work

* [x] Convert Markdown links
* [x] Decide what to do about the README beta packages section
* [x] Remove `/en` from all changed links (docs know how to handle language now)
* [x] Correct `http` links in changed links


SeanKilleen:
@nickfloyd as part of adding the GH Action to check links, I saw that the following text in `README` appears to 404:

```
### Beta packages ###

Unstable NuGet packages that track the `main` branch of this repository are available at
[https://ci.appveyor.com/nuget/octokit-net](https://ci.appveyor.com/nuget/octokit-net)
```

Is there a new beta packages link? Or should I remove that section?

SeanKilleen:
Hmm....tried configuring Lychee to fail when links are redirected, since most of the links we're seeing are 301 permanent redirects. However, setting `args: --max-redirects 0` in the github action YAML didn't do what I expected it to. Another time.

nickfloyd:
> @nickfloyd as part of adding the GH Action to check links, I saw that the following text in `README` appears to 404:
> 
> ```
> ### Beta packages ###
> 
> Unstable NuGet packages that track the `main` branch of this repository are available at
> [https://ci.appveyor.com/nuget/octokit-net](https://ci.appveyor.com/nuget/octokit-net)
> ```
> 
> Is there a new beta packages link? Or should I remove that section?

That should now be removed - good catch, thx! ❤️

SeanKilleen:
@nickfloyd while changing links in the `MiscellaneousClient`, I noticed that the misc docs now redirect to https://docs.github.com/en/rest/emojis

The `MiscellaneousClient` should likely be split up into MarkownClient, EmojiClient, GitignoreClient, etc. and then have the larger MiscellaneousClient retired in a future major release. Noting it here because you may want to add it to a backlog somewhere (happy to create an issue if you'd like).

For now, I will remove the top-level "misc" link as it no longer exists, and will add text to refer to the doc link at the method level.

nickfloyd:
> Noting it here because you may want to add it to a backlog somewhere (happy to create an issue if you'd like).

If you get the chance, please feel free to create an issue so that we can get what you're suggesting here implemented.  Thanks!

SeanKilleen:
Down to 3519 results. 😆  

ℹ️ ⚠️ One thing I've noticed: Many of the links refer to specific headings as anchors, and the text of many of those anchors has changed, so I need to re-evaluate it.

💡 One thought: if I can find an HTML proofer (lychee, html-proofer, or similar) that will validate the existence of anchors, I can focus on the base URLs en masse and then do anchor cleanups as a separate step.

SeanKilleen:
ℹ️ Realizing that lychee doesn't support anchor tags so its usefulness will be limited in this case. Will likely end up removing it, as the majority of the links are in C# files and we'd need to convert them to UTF-8 first and then I'm still not sure we'd get the benefit.

SeanKilleen:
@nickfloyd question for you -- how important is linking to the correct anchor tag on the page? 

If we wanted to go more toward the 80/20 rule, we could remove the anchor references in the links and link to the correct page. This would let us update the links quickly, but we'd lose some specificity, and the links would be more fragile as heading/anchor names change.

I'm fine to proceed either way, but if you don't care about anchor-level specificity, I can remove them.

SeanKilleen:
@nickfloyd I realized that this is going to be a larger effort than thought so I'll split it over multiple PRs. 

I'll probably update #2474 with a breakdown list so that we can slice it in more meaningful ways going forward. 

I'm going to do a review pass for basic fixes & comments and then it'll be ready for review.

SeanKilleen:
Actually as much as this pains me, I'm going to close this PR and start over because:

* I don't like the way it's structured overall -- changes aren't targeted, too hard to review.
* I didn't realize and I accidentally changed all the whitespace in C# files due to VS Code formatting on save. Again, very hard to review.


Look for smaller, targeted PRs toward this task going forward.

Issue #1604, opened 2017-05-09
Strongly named assembly issue


CortanaAnalyticsGallery:
I'm working at Azure Machine Learning Team and we found the kit very useful for the project we've been currently working on.  Unfortunately we hit a blocker when it turns out that the dlls distributed through the nuget are not signed. 
Because we can only consume strongly named assemblies in production environment I would like to ask if there are any plans to make official nuget distribution with the signed dlls? 

Here is the exception:
"An exception of type 'System.IO.FileLoadException' occurred in <theDll>.dll but was not handled in user codeAdditional information: Could not load file or assembly 'Octokit, Version=0.24.0.0, Culture=neutral, PublicKeyToken=null' or one of its dependencies. A strongly-named assembly is required. (Exception from HRESULT: 0x80131044)"
![image](https://cloud.githubusercontent.com/assets/26587387/25864165/482baa26-34a3-11e7-94ee-33fbe5b238fc.png)

khellang:
I think [the docs](https://github.com/octokit/octokit.net/blob/master/docs/strong-naming.md) explains it pretty clearly. If you're going to read the [epic thread](https://github.com/octokit/octokit.net/issues/405), I suggest you get some popcorn :wink:

Lately, there's been quite a few projects pointing to [StrongNamer](https://github.com/dsplaisted/strongnamer) for questions like these. The library lets you transparently and automatically sign the assemblies you reference as part of the build process.

shiftkey:
@AzureMLTest @khellang I'm mostly neutral on this due to the fact that it doesn't hurt me in my day-to-day work, but I'll defer to @ryangribble and @mderriey who have been doing a lot of work recently on getting .NET Standard support in on whether they want to support this at some stage.

And :thumbsup: to StrongNamer as an escape hatch for this.

mderriey:
I don't have a strong opinion on the subject as I've never used it/experienced any pain with it.

I'd lean towards StrongNamer as the process seems straightforward.

ryangribble:
I wasn't aware of the history (definitely enjoyed reading the epic thread! :popcorn:)

I am also from the "don't have a strong opinion" stance.  The fact it hasn't come up in the 15 months I've been contributing to Octokit would seem to indicate it isnt a hugely requested feature.  Given there are things like StrongNamer now to let others effectively make any of their dependencies strongly named when they need it, without needing the library to be strong named for all other users... is there an actual need to do anything in octokit.net here?

[...]

Issue #2382, opened 2021-12-13
Still not strong named at nuget.org ?


wangyoutian:
After I installed octokit from nuget.org, I exported the meta of the installed dll and found it's  still not strong-named. The dumped meta is :

"Octokit, Version=0.50.0.0, Culture=neutral, PublicKeyToken=null"

Being unstrong-named, it brings about some issue in my project, which is strong named. 

I saw at this repo that whether to strong-name has been discussed and the decision to strong-name is made. So where can I found the strong-name version?

I pulled this repo to my local computer. I found the source code is not strong named. I tried to strong-name it locally; it is done. But I think I would prefer the official strong name version to avoid some issues such as mismatched ".snk" file.

Any advice? 

Thanks.

nickfloyd:
Possble duplicate of: https://github.com/octokit/octokit.net/issues/1604

JonruAlveus:
We are now strong signed 👍

Issue #405, opened 2014-02-28
Signed Octokit Releases - Yea or Nay?


shiftkey:
So #404 landed, and I wanted to have a proper discussion about whether we should do this in an official capacity for Octokit releases. 

So let the fur fly (remember, bonus points for good reasoning and being gentlemanly) and I'll check back in a few days to see who is left standing...

![fight3](https://f.cloud.github.com/assets/359239/2291194/a8aa9d7a-a037-11e3-9621-474b59d9e8c8.gif)
![fight2](https://f.cloud.github.com/assets/359239/2291195/a8ade336-a037-11e3-96f0-4e2083f446cc.gif)
![fight4](https://f.cloud.github.com/assets/359239/2291198/b7b0eb62-a037-11e3-98c3-f9ecbc8c49db.gif)
![fight1](https://f.cloud.github.com/assets/359239/2291196/a8b172e4-a037-11e3-914c-1a6ce3c66fac.gif)

nickfloyd:
Hey Friends,

As a factor of closing the loop on this - we have just shipped the first signed Octokit .NET.  Regardless of where you might've been originally on this approach - after a few years, we have completed what [we set out to do](https://github.com/octokit/octokit.net/blob/main/docs/strong-naming.md) after this conversation.

❤️ Thank you to everyone who contributed to the above and to Octokit .NET itself - we could not do any of this without y'all!

Special hat-tip to @shiftkey for having the courage to begin this discussion and making the [decision](https://github.com/octokit/octokit.net/blob/master/docs/strong-naming.md) and to @AndreyTretyak for [implementing](https://github.com/octokit/octokit.net/pull/2575) it.

PR #2575, opened 2022-09-17
Add strong name signing

AndreyTretyak:
It's a follow-up after my comment here https://github.com/octokit/octokit.net/issues/1604#issuecomment-1209916385 that intended to resolve https://github.com/octokit/octokit.net/issues/1604, https://github.com/octokit/octokit.net/issues/2382, and similar issues. I would be grateful if you would consider this change since it would simplify my team's life significantly.

**Main justification points:**
* Decision on strong naming [here](https://github.com/octokit/octokit.net/blob/master/docs/strong-naming.md) was to sign `Octokit` before version `1.0`, and currently we are at version `3.0`, so I hope we can do this long overdue change.
* Other `Octokit` packages like `Octokit.GraphQL` are signed for 5 years already ([PR link](https://github.com/octokit/octokit.graphql.net/commit/f386a60f84acf210e46227b380e4921a99b6aa73)) and it would be nice to make them consistent.
* Using `StrongNamer` as a workaround is a constant source of build and debug problems, since it's essentially decompiling `Octokit.dll` and compiling it again with the strong name during each build. So customers running not the dll that you've shipped.
* Strong naming required for multiple use cases for example Visual Studio extensions.

**Implementations details:**
Signing logic is done in the same way as in `octokit.graphql.net` using the same key for consistency and also without versioning changes.

Since signing is enabled in the `Directory.Build.props` it would be applied by default to every new project created in the future in this repository without additional effort.

If for some reason there is a need of disabling strong naming for the specific project this could be done by setting `<SignAssembly>false</SignAssembly>` in the project. As currently done for `Build.csproj` and `Octokit.Tests.Integration.csproj` that reference unsigned dlls.

Another slight change is in applying `InternalsVisibleTo` attribute for signed dlls. It should be done in the project file directly and with `$(StrongNameSuffix)` after dll name (to include the public key without duplicating this horrible string).

_Side note:_ One of the options discussed in the epic thread [here](https://github.com/octokit/octokit.net/issues/405) was to change dll version only on major version changes of NuGet package (Newtonsoft.Json model) and if needed this could be easily achieved by using declaring `AssemblyFileVersion` MSBuild property with the required dll version. Although I'm not sure how useful it is now when SDK-based projects are able to automatically generate binding redirects for executables during the build.


AndreyTretyak:
> Hey @AndreyTretyak Thank you for the work here ❤️. I feel like a lot of folks have been waiting for this.
> 
> I think I'd like to push this after our next release just so that it is an encapsulated change - I'll most likely release it as a breaking change as well, given the type of change it is.

Thank you, Nick. I'm looking forward to this. Are there any plans for when the next and following releases might happen?

nickfloyd:
> Thank you, Nick. I'm looking forward to this. Are there any plans for when the next and following releases might happen?

Hey @AndreyTretyak, We'll try to drop what is in `main` out there tomorrow and most likely push your work here on Monday, tentatively.

nickfloyd:
Hey @AndreyTretyak, We just dropped our last [release](https://github.com/octokit/octokit.net/releases/tag/v3.0.1) before we get this merged in on Monday.  Just a quick question - was there any reason you didn't end up signing [Octokit.Reactive](https://github.com/octokit/octokit.net/tree/main/Octokit.Reactive)?

If not, let me know if you'd be able to get that one done or if you'd like me to do it in your branch.  I'd like to get both done, so we only have to release one breaking change.  Let me know what you think.

Thanks again for getting this hammered out!

AndreyTretyak:
> Hey @AndreyTretyak, We just dropped our last [release](https://github.com/octokit/octokit.net/releases/tag/v3.0.1) before we get this merged in on Monday. Just a quick question - was there any reason you didn't end up signing [Octokit.Reactive](https://github.com/octokit/octokit.net/tree/main/Octokit.Reactive)?
> 
> If not, let me know if you'd be able to get that one done or if you'd like me to do it in your branch. I'd like to get both done, so we only have to release one breaking change. Let me know what you think.
> 
> Thanks again for getting this hammered out!

Just checked and `Octokit.Reactive.dll` would be signed as well. 
Setting to enable signing is in the `Directory.Build.props` file in the root of the repo, so by default all current and future projects would be signed, unless we disable it explicitly by setting `<SignAssembly>false</SignAssembly>` in the project file the way I've done for `Octokit.Tests.Integration.csproj`.

nickfloyd:
> Just checked and `Octokit.Reactive.dll` would be signed as well.

Perfect.  Thanks for validating that... we should be g2g on Monday then. 💥

nickfloyd (approved):

nickfloyd:
@AndreyTretyak,
@kfcampbell has just shipped this 🎉 - https://github.com/octokit/octokit.net/releases/tag/v4.0.0

PR #2577, opened 2022-09-17
Fix deserializing of Emoji types

### LINKED ISSUE #2576: Miscellaneous Client - GetEmojis does not work
MiscellaneousClient.GetAllEmojis currently fails because it expects to receive a list of Emojis in an [ ] format. It actually returns as an object list. This needs to change to be handled properly without breaking the public api.

JonruAlveus:
Adding in assignable extensions to allow us to tell whether a type of IDictionary<T1,T2> which allows us to deserialise an object list to a dictionary. Currently if you use IDictionary<T1,T2> it is not recognised as a dictionary type and the body is wrapped as if it were an array of 1 object.

closes #2576


nickfloyd:
release_notes: Fixes deserializing of Emoji types

PR #2579, opened 2022-09-17
Fixes the meta dto and meta tests so they pass and are fully the same…

JonruAlveus:
… as from github. Closes #2578

PR #2574, opened 2022-09-16
Splitting out the misc client into separate clients as per current documentation

JonruAlveus:
Splitting out Miscellaneous client into new clients.

1. Split clients and interfaces
2. Redirect misc client to new interfaces
3. Obsolete methods with descriptions
4. Repeat for Observable
5. Duplicate unit tests
6. Duplicate integration tests

PR #2581, opened 2022-09-20
updates tests to include new fields

nickfloyd:
This came up as a result of a combo merge.  This PR simply adds the new fields to the tests that the related PRs added.

Currently, the mainline build is failing due to the test missing the additional fields.

Related:
https://github.com/octokit/octokit.net/pull/2579
https://github.com/octokit/octokit.net/pull/2574
https://github.com/octokit/octokit.net/pull/2577


nickfloyd:
📓 Bypassed approval given the build was broken in main.

PR #2585, opened 2022-10-01
Add missing properties to `IssueEvent`.

### LINKED ISSUE #2550: IssueEvent is missing payload for code review set of events
I am writing a review request notification bot and as part of that I need to get information about when somebody was requested to review a PR. As part of `review_requested` event [payload](https://docs.github.com/en/developers/webhooks-and-events/events/issue-event-types#review_requested) there should be `review_requester`, `requested_reviewer` and `requested_team` fields but they are not in the IssueEvent class so I can't get this information using OctoKit. Same problem for other similar events like `reviewed`, `review_request_removed`, etc.

bloslo:
Fixes #2550.


kfcampbell (approved):
Reminder to ourselves to bump the minor version of octokit.net when releasing this PR. Although the constructor changes to `IssueEvent` are technically breaking, the model is used internally in Octokit and this shouldn't cause practical issues for users.

kfcampbell:
The checks weren't showing until I merged in the latest main, which is interesting. I think this is good to go now!

PR #2592, opened 2022-10-07
Use correct methods and URL to generate release notes

### LINKED ISSUE #2589: ReleasesClient.GenerateReleaseNotes uses the wrong URL and HTTP method
Both overloads of `ReleasesClient.GenerateReleaseNotes` have incorrect routes.

**EDIT:** They also use the GET method instead of POST.

---

https://github.com/octokit/octokit.net/blob/8624d90e8070e2594f2b1634d4f196363bb013a0/Octokit/Clients/ReleasesClient.cs#L33-L34

Here, the route is missing `/generate-notes` at the end, as per [REST API docs](https://docs.github.com/en/rest/releases/releases#generate-release-notes-content-for-a-release).

What's funny is that all parameters to the `.../releases/generate-notes` entry point are _also_ valid parameters to the `.../releases` entry point, thus the HTTP request returns `200 OK`.

Furthermore, the `.../releases` response contains, among others, _all_ fields of the `.../releases/generate-notes` response and is thus successfully deserialized to `GeneratedReleaseNotes`.

The end result is that the method succeeds, even if the outcome is far from what one may expect: a new release is created and published, triggering validation errors in a later call to either `ReleasesClient.Create` or `ReleasesClient.Edit` with the same tag name.

By the way, the route is wrong in unit tests, too: https://github.com/octokit/octokit.net/blob/8624d90e8070e2594f2b1634d4f196363bb013a0/Octokit.Tests/Clients/ReleasesClientTests.cs#L34-L36

---

https://github.com/octokit/octokit.net/blob/8624d90e8070e2594f2b1634d4f196363bb013a0/Octokit/Clients/ReleasesClient.cs#L53-L54

This route is missing `/generate-notes` at the end too, plus it doesn't even match method parameters, as it has `/repos/{owner}/{repo}` instead of `/repositories/{id}`.

I haven't experimented with this one, but I strongly suspect it throws.

The unit test here tests the right route though (minus `/generate-notes`): https://github.com/octokit/octokit.net/blob/8624d90e8070e2594f2b1634d4f196363bb013a0/Octokit.Tests/Clients/ReleasesClientTests.cs#L48-L50

How can the test pass? It may be related to how method parameters are mapped to `ManualRoute` attributes, but I'm just guessing here.

**EDIT:** Turns out `ManualRoute` attributes are actually not used to generate URLs: static methods of class `ApiUrls` are called instead. `GenerateReleaseNotes` methods call `ApiUrls.Releases` methods, which construct URLs suitable for listing releases, not for generating release notes. What's missing are two `ApiUrls.ReleasesGenerateNotes` methods (one for owner / repo, the other for repository ID) that return URLs with `/generate-notes` appended.

---

Please forgive me if I came out as blunt: it is not in my intentions to offend anyone. I think @Cristaling did a great job with the `GenerateReleaseNotes` implementation, by the way. I was just trying to be concise, fighting against my over-verbose Italian nature, :smile: and I realized later I might have erred on the other side.

How about I open a PR for this?

rdeago:
Fixes #2589.

I also changed the documentation for `GenerateReleaseNotes`, adding links to relevant GitHub docs, an example of generated text, and clearer example code.

### DIFF
--- a/Octokit.Tests/Clients/ReleasesClientTests.cs
+++ b/Octokit.Tests/Clients/ReleasesClientTests.cs
@@ -31,7 +31,7 @@ namespace Octokit.Tests.Clients
 
                 await releasesClient.GenerateReleaseNotes("fake", "repo", data);
 
-                client.Received().Post<GeneratedReleaseNotes>(Arg.Is<Uri>(u => u.ToString() == "repos/fake/repo/releases"),
+                client.Received().Post<GeneratedReleaseNotes>(Arg.Is<Uri>(u => u.ToString() == "repos/fake/repo/releases/generate-notes"),
                     data,
                     "application/vnd.github.v3");
             }
@@ -45,7 +45,7 @@ namespace Octokit.Tests.Clients
 
                 await releasesClient.GenerateReleaseNotes(1, data);
 
-                client.Received().Post<GeneratedReleaseNotes>(Arg.Is<Uri>(u => u.ToString() == "repositories/1/releases"),
+                client.Received().Post<GeneratedReleaseNotes>(Arg.Is<Uri>(u => u.ToString() == "repositories/1/releases/generate-notes"),
                     data,
                     "application/vnd.github.v3");
             }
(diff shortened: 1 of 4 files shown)


nickfloyd (approved):
LGTM! ❤️

nickfloyd:
release_notes: Use correct methods and URL to generate release notes

rdeago:
Wow @nickfloyd that was fast! Thanks a lot!

PR #2600, opened 2022-10-25
Adding repo allowupdatebranch

michal-ciechan:
As per https://docs.github.com/en/rest/repos/repos#update-a-repository, there is the following setting:

> `allow_update_branch`  `boolean`
Either `true` to always allow a pull request head branch that is behind its base branch to be updated even if it is not required to be up to date before merging, or false otherwise.
> 
> Default: `false`

- [x] Code change
- [x] Locally tested
- [x] Fix existing Unit/Integration Tests
- [ ] Add new unit/integration tests


nickfloyd (approved):
This looks good!  Thanks for the changes! ❤️ Given that our serialization and other tests cover this response model, we should be good.

PR #2607, opened 2022-11-08
Remove Microsoft.Bcl.AsyncInterfaces dependency

kfcampbell:
I don't believe that this package is required anymore given [this comment](https://www.nuget.org/packages/Microsoft.Bcl.AsyncInterfaces/#readme-body-tab) on its Nuget page:

> This package is not required starting with .NET Standard 2.1 and .NET Core 3.0.

I'd like to test that assumption by running the Actions builds.


kfcampbell:
Looks like I was incorrect! It must have succeeded on my local machine because I have a somewhat up-to-date .NET install.

Issue #2587, opened 2022-10-04
If a folder or file name contains "#", the content cannot be retrieved.


minoura-a:
In the following folder structure of GitHub, I tried to get the contents under #aaa1 by specifying "AAA/#aaa1" in the `path` of `Client.GetAllContents()`.

```
GitHub
├AAA
│#aaa1
│└││aaa_xxx.xlsx
│#aaa2
└#aaa3
└BBB
  └#bbb1
    └│bbb_xxx.xlsx
```

I thought I could get #aaa_xxx.xlsx, but #aaa1, #aaa2, and #aaa3 were got.
This bug occurs only when the folder or file name contains #.

nickfloyd:
Hey @minoura-a,  Thanks for reporting this ❤️ .  We'll prioritize this to be worked on, but feel free to submit a PR for a fix if you have time!

minoura-a:
Hello @nickfloyd, thank you for your comments.
I would like to make a pull request.
I think I can solve the problem by URL encoding the #.
May I implement a method that specifies the path to a folder or file that encodes #?

nickfloyd:
Sounds good!  Thanks for taking the time to make the fix! ❤️ Go ahead and PR your change, and we'll get it reviewed.

nickfloyd:
Hey @minoura-a, happy Tuesday!

Just a friendly ping - we're you still planning on giving this fix a shot and submitting a PR?

minoura-a:
Hello @nickfloyd , thanks for the follow up.
Sorry for the delay.
I have not taken the time to work on it, but will create a PR.
I can create it by the end of next week. Is that OK with you?

nickfloyd:
Hey @minoura-a, for sure! I just wanted to follow up with you to see if you were still interested.  Thank you so much for helping to make our community and this SDK better! ❤️

brent85:
> In the following folder structure of GitHub, I tried to get the contents under #aaa1 by specifying "AAA/#aaa1" in the `path` of `Client.GetAllContents()`.
> 
> ```
> GitHub
> ├AAA
> │#aaa1
> │└││aaa_xxx.xlsx
> │#aaa2
> └#aaa3
> └BBB
>   └#bbb1
>     └│bbb_xxx.xlsx
> ```
> 
> I thought I could get #aaa_xxx.xlsx, but #aaa1, #aaa2, and #aaa3 were got. This bug occurs only when the folder or file name contains #.
[...]

PR #2608, opened 2022-11-08
Fixed a bug that content cannot be retrieved if "#" is included in a file or folder name. 

minoura-a:
I have confirmed that content can be retrieved even if "#" is included in the file or folder name.
I have confirmed that all unit tests passed.
#2587


kfcampbell:
@minoura-a nice work! Would you mind adding tests for the new helper you created to match those [here](https://github.com/octokit/octokit.net/blob/main/Octokit.Tests/Helpers/StringExtensionsTests.cs#L68)?

minoura-a:
@kfcampbell 
Thank you for confirming this.
I will create a test, please wait a couple of days.

minoura-a:
@kfcampbell 
Since the EncodeSharp method is only used within the StringExtensions class, I have changed the accessibility of the method to private.
[This](https://github.com/minoura-a/octokit.net/commit/b22bd13a51de02789226a7062b8fc248921d8998) is Comimt.
For this reason, I do not think testing is necessary.
Is this correct?

kfcampbell:
I think it's still worthwhile to create a test for this behavior, similar to the one for `EscapeDoubleQuotes`.

minoura-a:
@kfcampbell 
I have created a test. Could you confirm it?
[This](https://github.com/octokit/octokit.net/blob/8c2c199e058f193fa12ecfba85d3a1aff3c34476/Octokit.Tests/Helpers/StringExtensionsTests.cs#L93) is a commit.

minoura-a:
@kfcampbell 
I seem to be failing the Sourcelink test, is there anything I should do?

nickfloyd (approved):

Issue #2329, opened 2021-05-31
Get/Create/Update Environments


flcdrg:
I'd like to see support added for environments - https://docs.github.com/en/rest/reference/repos#environments

semyon-p:
Hey! I've added Environments list Get feature.
How can I send a pull request?

kfcampbell:
@semyon-p if you've forked the project and created the feature on your fork, you can find instructions on opening a PR [here](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request-from-a-fork). If you haven't forked the repo yet, that's a good first step!

semyon-p:
Thank you!
Created PR: https://github.com/octokit/octokit.net/pull/2613

nickfloyd:
This has been fixed and merged  in: https://github.com/octokit/octokit.net/pull/2613

danielshendfs:
Thanks a lot for the Get environment feature, is there any plan to add Create as well?

mikebollandajw:
also looking to default setup environments

PR #2613, opened 2022-11-13
Added Environments API - GetAll list only feature

semyon-p:
Let's add Environment API support.
https://github.com/octokit/octokit.net/issues/2329

As a 1st step I've implemented GetAll methods for Environments.

### DIFF
--- a/Octokit.Reactive/Clients/IObservableEnvironmentsClient.cs
+++ b/Octokit.Reactive/Clients/IObservableEnvironmentsClient.cs
@@ -1,0 +1,57 @@
+﻿using Octokit.Models.Response;
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Octokit.Reactive
+{
+    public interface IObservableEnvironmentsClient
+    {
+        /// <summary>
+        /// Gets all the environments for the specified repository. Any user with pull access
+        /// to a repository can view deployments.
+        /// </summary>
+        /// <remarks>
+        /// https://docs.github.com/en/rest/deployments/environments#list-environments
+        /// </remarks>
+        /// <param name="owner">The owner of the repository</param>
+        /// <param name="name">The name of the repository</param>
+        IObservable<DeploymentEnvironmentsResponse> GetAll(string owner, string name);
+
+        /// <summary>
+        /// Gets all the environments for the specified repository. Any user with pull access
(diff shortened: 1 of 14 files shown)


nickfloyd:
Hey @semyon-p Thank you so much for the contributions here.  If you get the chance, would you mind adding any test coverage that would be appropriate?  If needed, you can look at what others have done before you [here](https://github.com/octokit/octokit.net/tree/main/Octokit.Tests).

Thank you again for making things better! ❤️

nickfloyd (changes requested):
Overall, things look good here.  Just a comment about the tests and also you'll need to add an interface to the Reactive Project - `Octokit.Reactive.IObservableRepositoryDeployEnvironmentsClient` we keep these in parity with the sync SDK APIs and interfaces.  Let me know if you need any help with any of this!  Thanks again for the contributions!

semyon-p:
@nickfloyd hi!
I've added tests and IObservableRepositoryDeployEnvironmentsClient.
All tests are passed now. Please check.
I've made a small change in Octokit.Models.Request.ApiOptions class to become such kind of constructions work correctly:
            if (options != ApiOptions.None)
            {
               //do something
            }

semyon-p:
@nickfloyd also I wanted to let you know, that I was trying to implement CreateEnvironment method but figured out, that API call described in the documentation does not work for me.
https://docs.github.com/en/rest/deployments/environments#create-or-update-an-environment
I've tried to test on a public personal repository.

DeleteEnvironment call works fine, by the way.
https://docs.github.com/en/rest/deployments/deployments#delete-a-deployment

Maybe I've done something wrong, but if you'll have some time please check it too. Maybe there is a bug in Github API itself.

nickfloyd (approved):

mikebollandajw:
Value was either too large or too small for an Int32.    at System.Convert.ThrowInt32OverflowException()
         at System.Convert.ToInt32(Int64 value)
         at System.Convert.ChangeType(Object value, Type conversionType, IFormatProvider provider)
         at Octokit.PocoJsonSerializerStrategy.DeserializeObject(Object value, Type type) in /_/Octokit/SimpleJson.cs:line 1437
         at Octokit.PocoJsonSerializerStrategy.DeserializeObject(Object value, Type type) in /_/Octokit/SimpleJson.cs:line 1368
         at

PR #2625, opened 2022-11-30
Add missing properties for meta and app payloads

### LINKED ISSUE #2624: [FEAT]: Add missing properties for meta and app payloads
### Describe the need

I'm writing a GitHub App that is deployed to GHES and noticed several properties missing from some of the models:

* `InstalledVersion` is missing from the `Meta` model (only present on GHES)
* `Permissions` and `Events` are missing from the `GitHubApp` model
* Some permissions are missing from the `InstallationPermissions` model

### SDK Version

4.0.2

### API Version

_No response_

### Relevant log output

_No response_

### Code of Conduct

MatisseHack:
Resolves #2624

----

### Before the change?

* `InstalledVersion` was missing from the `Meta` model (only present on GHES)
* `Permissions` and `Events` were missing from the `GitHubApp` model
* Some permissions were missing from the `InstallationPermissions` model

### After the change?

* The missing properties have been added

### Other information

* I pulled the list of permissions from the [OpenAPI description](https://github.com/github/rest-api-description/blob/main/descriptions/api.github.com/api.github.com.json) using the following JSONPath query: `$.components.schemas.app-permissions.properties`

----

### Pull request checklist
- [ ] Added the appropriate label for the given change (I don't have permission to)

- [ ] No

If `Yes`, what's the impact:  

* Low - adding properties requires adding constructor arguments, but consumers are unlikely to be using model constructors. Would it make sense to start making the model constructors `internal` to prevent future breaking changes?

### Pull request type

 
 

- Bugfix: `Type: Bug`

----


Octokit.Tests/Clients/MetaClientTests.cs
```diff
@@ -23,7 +23,8 @@ public async Task RequestsTheMetadataEndpoint()
                      new[] { "1.1.6.1/24", "1.1.6.2/24" },
                      new[] { "1.1.7.1", "1.1.7.2" },
                      new[] { "1.1.8.1/24", "1.1.8.2/24" },
-                     new[] { "1.1.9.1", "1.1.9.2" }
+                     new[] { "1.1.9.1", "1.1.9.2" },
+                     null
```
kfcampbell, line 27: Do you mind explaining the addition of the `null` here? I'm not sure I follow.

MatisseHack: That was just to satisfy the required constructor arguments. I left it as `null` here because `installed_version` isn't present on calls to github.com. However, I'm happy to change it to `3.7.0` if you'd prefer to have all the values filled in. I could also make that constructor parameter optional if that's preferable.

kfcampbell: Hmm...I would probably prefer an explicit argument there, but I'm unable to articulate a good case for why. Matching other behavior is probably our best bet.

kfcampbell:
@nickfloyd has mentioned that in the past we typically don't treat constructor/model changes like this one as a breaking change. Is there another aspect that makes this change breaking I'm not seeing?

Octokit.Tests/Clients/MetaClientTests.cs
MatisseHack: Makes sense! I've updated it to an actual value

MatisseHack:
No, just the constructor change. I saw other PRs were approved with constructor changes so I assumed it would be okay, but still marked it as a breaking change since it seemed applicable.

kfcampbell (approved):
Got it, thank you! I'll get this merged shortly and released at some point in the future.

MatisseHack:
Thanks!

PR #2661, opened 2023-01-23
Update permissions

### LINKED ISSUE #2323: Team model uses wrong Permission enum
It looks like commit 598c7c31bf0b4cc0eb1f722c555600d3d01b3e84 broke deserialization of permissions on the `Team` model when getting teams with access to a repository (we're on GitHub Enterprise, but AFAICS the Team DTO is the same on GitHub.com).

Curling the [repository list teams endpoint](https://docs.github.com/en/rest/reference/repos#list-repository-teams) on our GitHub Enterprise installation gives responses like

```
[
  {
    ...
    "permission": "push",
    ...
  },
  {
    ...
    "permission": "pull",
    ...
  },
  {
    ...
    "permission": "admin",
    ...
  },
]
```

Which matches [`Permission`](https://github.com/octokit/octokit.net/blob/main/Octokit/Models/Request/Permission.cs) (which the field on `Team` used to be before that commit), but not [`PermissionLevel`](https://github.com/octokit/octokit.net/blob/main/Octokit/Models/Response/PermissionLevel.cs) (what it is now).

When I use a collaborator endpoint and try to examine the permissions on the response I get a stack trace like

```
System.ArgumentException: Value 'push' is not a valid 'PermissionLevel' enum value.
   at Octokit.StringEnum`1.ParseValue() in /home/runner/work/octokit.net/octokit.net/Octokit/Models/Response/StringEnum.cs:line 149
   at Octokit.StringEnum`1.get_Value() in /home/runner/work/octokit.net/octokit.net/Octokit/Models/Response/StringEnum.cs:line 45
   at Foo.ConvertPermission(StringEnum`1 permission) 
   at Foo.<GetRepositoryTeamsAsync>b__7_0(Team x) 
```

(This is running against GitHub Enterprise versions 2.22.7 and 3.0.5)

(Edited to correct a link to the GitHub docs which linked to the wrong API endpoint)

notauserx:
Resolves #2323 

----

### Before the change?

* App Installations and Collaborators endpoints does not contain the correct permissions on the request and response contracts.

### After the change?

* Permissions will be updated to match the github open api specs

----

### Pull request checklist

- [ ] No

If `Yes`, what's the impact:  

* N/A

### Pull request type

 
 

- Bugfix: `Type: Bug`

----


notauserx:
There are some integration tests failing for IRepoCollaboratorsClient's GetAll method. The tests are adding a collaborator and expecting the collaborator count to be updated. However, adding a collaborator will send an invitation and the collaborator has to accept the invitation before the count will be updated.

The integration tests for ReviewPermission also follow the same route, a collaborator is added with 'write' permission and the test expects a collaborator with 'write' permission, without waiting for the collaborator to accept the invitation. 

In these tests [octokitnet-test1](https://github.com/octokitnet-test1) and [m-zuber-octokit-integration-tests](https://github.com/m-zuber-octokit-integration-tests) accounts are used for adding a collaborator, and I don't have their creds to accept the invitation. 

This probably is out of scope for this issue, but we probably should create a new issue and think about possible workarounds.

For app installation permissions, there are some permissions that have only 'read' as a value, some have only 'write', most with 'read' and 'write', and some with 'read', 'write' and 'admin'. 

I have named the enums InstallationReadPermissionLevel,  InstallationWritePermissionLevel, InstallationReadWritePermissionLevel, and InstallationReadWriteAdminPermissionLevel respectively. 

I have also moved the Permission.cs to the common folder because it houses all the permissions from the API surface.

kfcampbell:
@notauserx is the `[wip]` description in the PR still valid?

notauserx:
@kfcampbell I have completed updating all references the old Permission enums and made them context-specific. I was also planning to update some endpoints that use the updated permission in request/response.

However, the integration tests are failing for some of these endpoints, as I've mentioned in my previous comments. I do need some feedback on how to proceed.

kfcampbell:
The integration testing question with the invites is a tough one. As far as I know, we don't have a convenient way to do that. I'll have to think about that more and ask my team.

nickfloyd:
Hey @notauserx, Thanks again for these contributions ❤️ .  As you've detailed above, we need to rethink some of the strategies in our integration tests.  I cannot be 100% certain, but the workflow for the ones you referenced has changed since they were originally implemented.

We have a few options here:

1. We can `skip` to unblock this by adding the skip attribute i.e., `[Fact(Skip = "Reference to this PR")]`.  That will give us more time to figure out what we want to do without looking at code context clues or outright deleting them.
2. We can delete the tests, looking at the ones [here](https://github.com/octokit/octokit.net/blob/00692397746d9afa20aaf7895a719939fca68dcd/Octokit.Tests/Clients/RepoCollaboratorsClientTests.cs) some of these seems a bit contrived and fragile (as you discovered). 
3. We can see if there is an alternate strategy and implement that - at its most basic, GetAll could check for something else or check that it pulled back the one contributor it already has.
4. We could rework things so that it is dependent on less fragile and more public fixtures - though I think we'd probably still run into what you are running into around account access.

I don't want to get into philosophical aspects of how useful functional tests are - especially since there is already pretty good unit test coverage on `RepoCollaboratorsClient`, for instance - so I'll leave the decision on how to proceed up to you. I'll trust your judgment. I see this as a valuable change and would like to get it in. 

I appreciate you highlighting this fragile part of this SDK's coverage.  We're working daily to try to improve this for everyone, and I'm glad you've decided to come alongside us and do the same ❤️ !

notauserx:
This PR is now ready for review. Please give me some feedback and I will update the PR accordingly.

nickfloyd (approved):
Thank you for getting this in ❤️

PR #2682, opened 2023-03-10
Correctly deserialize CollaboratorPermissionResponse

### LINKED ISSUE #2681: Octokit.net v5 doesn't deserialize user permissions response correctly
Based on https://github.com/octokit/octokit.net/blob/7d01bad2971b751f3b6736c57c9349d318a37293/Octokit/Clients/RepoCollaboratorsClient.cs#L242, a `CollaboratorPermissionResponse` needs to have a `Collaborator` since you're deserializing the response, but this, in practice, is `null because the [response schema](https://docs.github.com/en/rest/collaborators/collaborators?apiVersion=2022-11-28#get-repository-permissions-for-a-user) returns "user" and there's no obvious mapping between "user" and "collaborator".

/cc @JimSuplizio

heaths:
Fixes #2681

### DIFF
--- a/Octokit.Tests/Models/CollaboratorPermissionResponseTests.cs
+++ b/Octokit.Tests/Models/CollaboratorPermissionResponseTests.cs
@@ -1,0 +1,52 @@
+using Octokit;
+using Octokit.Internal;
+using Xunit;
+
+public class CollaboratorPermissionResponseTests
+{
+  [Fact]
+  public void CanDeserialize()
+  {
+    const string json = @"{
+      ""permission"": ""read"",
+      ""user"": {
+        ""login"": ""octocat"",
+        ""id"": 583231,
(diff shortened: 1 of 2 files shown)


nickfloyd:
Hey @heaths, I am taking a look at this now... thank you for jumping on this so quickly. I missed it in my review the other day; appologies for that oversight! ❤️

nickfloyd (approved):
Looks good.  Thanks for the going the extra mile with tests and a fixture ❤️

heaths:
Might we see a 5.1.1 in the near-ish future? For now we used the legacy `PCollaboratorPermissionResponse.Permission` string property but seems the new, more granular permission model gives us the better access we need than checking for "admin" or "write" - now we can just test if the nested "push" is true.

/cc @JimSuplizio who is working on incorporating this new version into our bot.

PR #2685, opened 2023-03-19
Add missing workflow run conclusions

### LINKED ISSUE #2684: [BUG]: Missing workflow run conclusions
### What happened?

Our company has a service that is grabbing GitHub workflow run data and then store it locally for further analysis. However the library was throwing exceptions for some of those workflow runs with the message `Value 'startup_failure' is not a valid 'WorkflowRunConclusion' enum value`. I've checked the source code of Octokit.NET and it wasn't there indeed, so I've checked GitHub API reference to make sure this is properly documented. However, all of the REST API documentation doesn't mention this conclusion anywhere ([Using the REST API to interact with checks](https://docs.github.com/en/rest/guides/using-the-rest-api-to-interact-with-checks), [Check Runs REST API](https://docs.github.com/en/rest/checks/runs), [Workflow runs REST API](https://docs.github.com/en/rest/actions/workflow-runs), [Creating CI tests with the Checks API](https://docs.github.com/en/apps/creating-github-apps/guides/creating-ci-tests-with-the-checks-api)) and the only place this conclusion is mentioned is in [GraphQL reference](https://docs.github.com/en/graphql/reference/enums#checkconclusionstate).

I have prepared a branch with the fix to the library (however the GitHub documentation is still broken), so the pull request should follow shortly after the issue.

also misc: the Code of Conduct link is broken in the issue template and actually is leading [to issues](https://github.com/octokit/octokit.net/issues/CODE_OF_CONDUCT.md) :p

### Versions

Octokit.Net 4.0.2, .NET 6.0.14

### Relevant log output

```shell
System.ArgumentException: Value 'startup_failure' is not a valid 'WorkflowRunConclusion' enum value.
  File "/home/runner/work/octokit.net/octokit.net/Octokit/Models/Response/StringEnum.cs", line 146, col 17, in ParseValue
  File "/home/runner/work/octokit.net/octokit.net/Octokit/Models/Response/StringEnum.cs", line 45, col 19, in get_Value
  ... application stacktrace ...
```

### Code of Conduct

ezhevita:
Resolves #2684

----

### Before the change?

* The property to get a workflow run conclusion throws an exception when trying to get value of the workflow run with the conclusion of `startup_failure` or `skipped`.

### After the change?

* The property returns the workflow run conclusion correctly and doesn't throw an exception.

### Other information

* I've mentioned in the original issue that GitHub documentation doesn't include these values as well, so it will be great if you  pass this information internally to the docs team 😄

----

### Pull request checklist

- [ ] Yes
- [x] No

### Pull request type

 
 

- Bugfix: `Type: Bug`

----

### DIFF
--- a/Octokit/Models/Response/WorkflowRunConclusion.cs
+++ b/Octokit/Models/Response/WorkflowRunConclusion.cs
@@ -18,5 +18,9 @@ namespace Octokit
         ActionRequired,
         [Parameter(Value = "stale")]
         Stale,
+        [Parameter(Value = "startup_failure")]
+        StartupFailure,
+        [Parameter(Value = "skipped")]
+        Skipped,
     }
 }


ezhevita:
also GitHub doesn't seem to allow me to add a label to the PR, so if someone adds it for me I'll be happy 😁
the appropriate label is `Type: Bug`

nickfloyd (approved):
Hey @ezhevita, Thanks for the work here ❤️ .  We're still chatting about the API docs and the need for the enum values to be surfaced in a more visible way but this change is a solid addition to the SDK, thanks again for doing it!

PR #2686, opened 2023-03-20
Variables API

Aragas:
This is a draft of the Variable API implementation.
It lacks the environment part, but it's not required for my case, so if anyone want a jumpstart, this draft could be used to complete the implementation!


nickfloyd:
Hey @Aragas, thanks for getting this started.  I labeled this with `up-for-grabs` to help with the community optics on this PR - so that others can see that you'd like some help seeing this through.,

Thanks again for the kickstart on this API set ❤️

PR #2690, opened 2023-03-28
[FEAT] Allow users to fetch all commits for two commits comparision

zolroman:
### Before the change?
Method Compare in RepositoryCommitsClient returns CompareResult with limit  in collections Commits and Files up to 250 items.

* 

### After the change?
Added a parameter `bool fetchAllCommits` to the method that allows users to fetch all Commits and Files. It uses a pagination inside.

* 

----

### Pull request checklist
- [x ] Tests for the changes have been added (for bug fixes / features)
- [ x] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [ x] Added the appropriate label for the given change

- [ x] No

### Pull request type

Type: Feature

----

### DIFF
--- a/Octokit/Clients/RepositoryCommitsClient.cs
+++ b/Octokit/Clients/RepositoryCommitsClient.cs
@@ -1,4 +1,6 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 
 namespace Octokit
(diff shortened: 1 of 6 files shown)


Octokit/Clients/RepositoryCommitsClient.cs
```diff
@@ -105,6 +107,74 @@ public Task<CompareResult> Compare(long repositoryId, string @base, string head)
...
+        private async Task<CompareResult> Compare(Uri uri, ApiOptions options)
+        {
+            var results = await ApiConnection.GetAll<CompareResult>(uri, options);
+            if (results.Count == 1) return results[0];
+
+            var firstCompareResult = results[0];
+            var commits = firstCompareResult.Commits.ToList();
+            var files = firstCompareResult.Files.ToList();
+
+            foreach (var compareResult in results.Skip(1))
```
kfcampbell, line 156: I'm sorry I'm blanking on this: do you mind explaining to me why we're skipping a result here?

Octokit/Clients/RepositoryCommitsClient.cs
```diff
@@ -105,6 +107,74 @@ public Task<CompareResult> Compare(long repositoryId, string @base, string head)
...
+            Ensure.ArgumentNotNullOrEmptyString(@base, nameof(@base));
+            Ensure.ArgumentNotNullOrEmptyString(head, nameof(head));
+            Ensure.ArgumentNotNull(options, nameof(options));
+
+            return Compare(ApiUrls.RepoCompare(repositoryId, @base, head), options);
+        }
+
+        private async Task<CompareResult> Compare(Uri uri, ApiOptions options)
+        {
+            var results = await ApiConnection.GetAll<CompareResult>(uri, options);
```
kfcampbell, line 149: Do you mind explaining the reason for this custom logic where it doesn't seem to exist for other comparisons? How is this tested?

kfcampbell (commented):
I had a couple of quick questions! Thank you for contributing.

Octokit/Clients/RepositoryCommitsClient.cs
zolroman: this method returns only one object (not a collection) but it contains properties with collections: `Commits` and `Files`. 
Without `ApiOptions` it returns up to 250 elements in this collections, but the API supports the pagination. I used an existing method `ApiConnection.GetAll` to work with pagination but a result of this method is a collection of the elements.
I've added a simple loop to aggregate data from collection to one object.

 in results we have a data structure like that 

```

[
	{
		<fields with the same values for all objects in list>,
		Commits:[0..100]
		Files: [0..100]
	},
	{
		<fields with the same values for all objects in list>,
		Commits:[101..200]
		Files: [101..200]
	},
	...
]

```

and we need to return an object like that
```
{
	<fields with the same values for all objects in list>,
	Commits:[0..end],
	Files: [0..end]
}
```

zolroman: I've added an integration test to check it: https://github.com/octokit/octokit.net/pull/2690/files#diff-0390f8750f37f9bd7484e089ecbab5421588cfd100243b64c6cab2bd0c6b20dcR302

kfcampbell: Understood, thanks!

Octokit/Clients/RepositoryCommitsClient.cs
kfcampbell: Oh of course, :facepalm:. Thanks.

kfcampbell (approved):

PR #2698, opened 2023-04-13
Removing empty object paramater in MigrationsClient.DeleteArchive

### LINKED ISSUE #2697: [BUG]: .NET 7.0 causes ArgumentException in Migrations.DeleteArchive
### What happened?

I encountered a similar error to #2659 for Migrations.DeleteArchive.  MigrationsClient.DeleteArchive throws an ArgumentException - The value cannot be null or empty. (Parameter 'mediaType') when invoked in an application compiled using .Net 7.0.

In this case it appears the issue is caused by creating an empty object parameter '**new object()**' in line: https://github.com/octokit/octokit.net/blob/aed70e335a59ef3f67fa6afdecc3aab10efc318e/Octokit/Clients/MigrationsClient.cs#L133 

**NOTE**: Removing the '**new object()**' Changing line 133 to '**return ApiConnection.Delete(endpoint);**' no longer generates the **ArgumentException - The value cannot be null or empty. (Parameter 'mediaType')** and the call succeeds.

This change avoids the exception which is generated when the **request.Body** is **"{}"** https://github.com/octokit/octokit.net/blob/aed70e335a59ef3f67fa6afdecc3aab10efc318e/Octokit/Http/HttpClientAdapter.cs#L144

### Versions

Octokit.Net (5.0.3)

### Relevant log output

_No response_

### Code of Conduct

gitasaurus:
Resolves #2697

----

### Before the change?

* MigrationsClient.DeleteArchive throws an ArgumentException - The value cannot be null or empty. (Parameter 'mediaType') when invoked in an application compiled using .Net 7.0

### After the change?

* MigrationsClient.DeleteArchive succeeds on .Net 7.0.  

### Other information

* Requires modifying Octokit.Tests/Clients/MigrationsClientTests.cs (any object was required by the test, hence the **new object()** parameter being added to pass the test)

----

### Pull request checklist

- [x] No

If `Yes`, what's the impact:  

* N/A

### Pull request type

 
 

- Bugfix: `Type: Bug`

----

### DIFF
--- a/Octokit.Tests/Clients/MigrationsClientTests.cs
+++ b/Octokit.Tests/Clients/MigrationsClientTests.cs
@@ -177,8 +177,7 @@ namespace Octokit.Tests.Clients
                 client.DeleteArchive("fake", 69);
 
                 connection.Received().Delete(
-                    Arg.Is<Uri>(u => u.ToString() == "orgs/fake/migrations/69/archive"),
-                    Arg.Any<object>());
+                    Arg.Is<Uri>(u => u.ToString() == "orgs/fake/migrations/69/archive"));
             }
 
             [Fact]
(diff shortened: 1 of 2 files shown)


kfcampbell (approved):
Thank you!

PR #2704, opened 2023-05-03
Fix runner application deserialization

### LINKED ISSUE #2703: [BUG]: Error deserializing runner application response
### What happened?

Octokit throws a deserialization error when calling `ListAllRunnerApplicationsForOrganization` in the `ActionsSelfHostedRunnersClient`

### Versions

Octokit.net 5.1.0

### Relevant log output

_No response_

### Code of Conduct

MatisseHack:
Resolves #2703

----

### Before the change?

* Octokit throws a deserialization error when calling `ListAllRunnerApplicationsForOrganization` in the `ActionsSelfHostedRunnersClient`

### After the change?

* No error is thrown

### Other information

I fixed the issue by removing the extra `IReadOnlyList` layer since `ApiConnection.GetAll` already returns an `IReadOnlyList`. I also made a few other updates to the`RunnerApplication` model:

* Added a missing property in the model
* Fixed a property that was always null because it had the wrong casing (didn't match the JSON payload).
* Created a new `RunnerLabel` model since the `Label` model is for issue/PR labels and didn't match the properties in the JSON payload

----

### Pull request checklist

- [ ] No

If `Yes`, what's the impact:  

* The casing for one property has been changed, which should be very low impact since the model was only added a few days ago.
* The model for one property has been changed, which should also be very low impact since most properties on the old model would be null anyway.

### Pull request type

 
 

Bugfix

- Bugfix: `Type: Bug`

----

### DIFF
--- a/Octokit.Tests/Clients/ActionsSelfHostedRunnersClientTests.cs
+++ b/Octokit.Tests/Clients/ActionsSelfHostedRunnersClientTests.cs
@@ -133,7 +133,7 @@ namespace Octokit.Tests.Clients
 
                 await client.ListAllRunnerApplicationsForEnterprise("fake");
 
-                connection.Received().GetAll<IReadOnlyList<RunnerApplication>>(
+                connection.Received().GetAll<RunnerApplication>(
                   Arg.Is<Uri>(u => u.ToString() == "enterprises/fake/actions/runners/downloads"), Args.ApiOptions);
 
             }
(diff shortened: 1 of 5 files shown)


nickfloyd (approved):
Hey @MatisseHack, this change looks good.  Thank you for doing it ❤️.  I will mark and ship this as a breaking change in light of the type change.  Even though it's low risk, bumping major does not cost anything and is a signal in case consumers need to know.

PR #2740, opened 2023-06-30
Fix ActionsSelfHostedRunnersClient Delete Calls

### LINKED ISSUE #2739: [BUG]: Fix ActionsSelfHostedRunnersClient Delete calls
### What happened?

Calling `ActionsSelfHostedRunnersClient.DeleteEnterpriseRunner`, `ActionsSelfHostedRunnersClient.DeleteOrganizationRunner`, `ActionsSelfHostedRunnersClient.DeleteRepositoryRunner` throw exceptions with the following message: `The value cannot be null or empty. (Parameter 'mediaType')`.

The source of this error is that `ApiOptions` are being passed in to the delete methods which don't expect this kind of object. They expect a request body type, but for these calls we can actually just omit this argument altogether. I have a fix for this that I've tested.

### Versions

Octokit.net 6.2.1

### Relevant log output

_No response_

### Code of Conduct

lineville:
* Resolves #2739 by resolving an exception being thrown by all of the `Delete___Runner` calls
* Passes `CancellationToken cancellationToken = default` to `Create___Token` calls to allow for an optional cancellation token.

----

### Before the change?

* Calls to all of the `Delete___Runner` methods would throw an exception caused by passing the wrong kind of `Object` to the underlying `Delete` call. We we're previously passing `ApiOptions` which are the wrong type and not relevant.

### After the change?

* Calls to all of the `Delete___Runner` methods succeed.
* Calls to the `Create__Token` methods allow for passing a `CancellationToken` and still succeed.

### Other information

* 

----

### Pull request checklist
- [x] Added the appropriate label for the given change

- [x] No

If `Yes`, what's the impact:  

* N/A

### Pull request type

 
 

- Bugfix: `Type: Bug`

----

### DIFF
--- a/Octokit.Reactive/Clients/IObservableActionsSelfHostedRunnersClient.cs
+++ b/Octokit.Reactive/Clients/IObservableActionsSelfHostedRunnersClient.cs
@@ -1,258 +1,197 @@
 ﻿using System;
-using System.Collections.Generic;
 using System.Reactive;
+using System.Threading;
 
 namespace Octokit.Reactive
 {
+  /// <summary>
+  /// A client for GitHub's Actions Self-hosted runners API.
+  /// </summary>
+  /// <remarks>
+  /// See the <a href="https://developer.github.com/v3/actions/self-hosted-runners/">Actions Self-hosted runners API documentation</a> for more information.
+  /// </remarks>
+  public interface IObservableActionsSelfHostedRunnersClient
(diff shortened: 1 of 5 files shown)


nickfloyd (approved):
This LGTM ❤️.  Thanks for the 🏕️ work on the comments and descriptions as well.

PR #2749, opened 2023-07-18
Support refreshtokens in OAuth flow

### LINKED ISSUE #2731: [FEAT]: Support refreshtokens in OAuth flow
### Describe the need

While the library currently supports getting access tokens from an OAuth callback, the resulting object doesn't include the refresh token, which allows an application to get a new access token for the user, without having to go through the entire OAuth flow again.

Additionally, the OAuthClient doesn't support a way to get an access token from a refresh token.

The refresh token is returned in the original request to get an access token.
https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/generating-a-user-access-token-for-a-github-app
https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/refreshing-user-access-tokens

Proposal:
Add RefreshToken as a property on [Octokit/Models/Response/OauthToken.cs](https://github.com/octokit/octokit.net/blob/afac3e0bcfc853a61f0decde6fe1da1b1124f53b/Octokit/Models/Response/OauthToken.cs)
Add RefreshToken as a property on [Octokit/Models/Request/OauthTokenRequest.cs](https://github.com/octokit/octokit.net/blob/afac3e0bcfc853a61f0decde6fe1da1b1124f53b/Octokit/Models/Request/OauthTokenRequest.cs)

If a OauthTokenRequest contains a RefreshToken but the AuthToken is null only the refresh_token property should be added to the request body due to the null check in [RequestParameters.cs::ToParametersDictionary()](https://github.com/octokit/octokit.net/blob/afac3e0bcfc853a61f0decde6fe1da1b1124f53b/Octokit/Models/Request/RequestParameters.cs#L24)

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

_No response_

### Code of Conduct

Kencdk:
Resolves #2731

----

### Before the change?

* OAuth Client doesn't support getting the refresh token, nor utilizing it to get a new access token.
* Additionally, it doesn't return information about when the access token expires.

### After the change?

* CreateAccessToken now returns all fields from the returned body, including token expiration, refresh token and refresh token expiration time.
* CreateAccessTokenFromRefreshToken accepts a clientId, clientSecret and refresh token, in order to create a new access token on behalf of the user, the refresh token was created for.

### Other information
While the endpoint is the same for creating an access token from a response code and for creating an access token based on the refresh token, there are key differences in the expected parameters. When creating based on a refresh token, the grant_type needs to be set.

As a result, I opted to create a new function instead of adding a mix of optional and required parameters (depending on input) to the existing CreateAccessToken function. Please advise if you'd prefer otherwise.

* https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/generating-a-user-access-token-for-a-github-app
*  https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/refreshing-user-access-tokens

----

## Additional info
Added &lt;inheritdoc /&gt; instead of having the xml doc in both interfaces and implementation.
The end-result is the same, it just avoids having to keep both up-to-date.

### Pull request checklist
- [x] Added the appropriate label for the given change

- [x] No

If `Yes`, what's the impact:  

* N/A

### Pull request type

Type: Feature

----

### DIFF
--- a/Octokit.Reactive/Clients/ObservableOauthClient.cs
+++ b/Octokit.Reactive/Clients/ObservableOauthClient.cs
@@ -3,6 +3,10 @@ using System.Reactive.Threading.Tasks;
 
 namespace Octokit.Reactive
 {
+    /// <summary>
+    /// Wrapper around <see cref="IOauthClient"/> for use with <see cref="IObservable{T}"/>
+    /// </summary>
+    /// <inheritdoc />
     public class ObservableOauthClient : IObservableOauthClient
     {
         readonly IGitHubClient _client;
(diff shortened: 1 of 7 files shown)


Octokit/Models/Response/OauthToken.cs
```diff
@@ -20,6 +20,19 @@ public OauthToken(string tokenType, string accessToken, IReadOnlyList<string> sc
             this.ErrorUri = errorUri;
         }
 
+        public OauthToken(string tokenType, string accessToken, int expiresIn, string refreshToken, int refreshTokenExpiresIn, IReadOnlyList<string> scope, string error, string errorDescription, string errorUri)
```
kfcampbell, line 23: Do you think it'd be worthwhile to add a comment here and above describing the situations when you'd want to use each constructor?

kfcampbell (commented):
This approach seems reasonable and I love the functionality!

Octokit.Reactive/Clients/ObservableOauthClient.cs
```diff
@@ -14,59 +18,30 @@ public ObservableOauthClient(IGitHubClient client)
             _client = client;
         }
 
-        /// <summary>
-        /// Gets the URL used in the first step of the web flow. The Web application should redirect to this URL.
-        /// </summary>
-        /// <param name="request">Parameters to the Oauth web flow login url</param>
-        /// <returns></returns>
```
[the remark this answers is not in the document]
Kencdk: I tried adding it to the PR message, but it's mistaken &lt;inheritdoc /&gt; for an HTML tag.

When inheritdoc is added on the class level, is automatically inherits the xml doc from the interface.
As such, there is no need to have it in both places and risk one getting out of date :)

Octokit/Models/Response/OauthToken.cs
Kencdk: I primarily kept the old one, to keep the change binary compatible - to avoid breaking changes :)
Added a summary for good measure.

Octokit.Reactive/Clients/ObservableOauthClient.cs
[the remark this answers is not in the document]
kfcampbell: Oh that's awesome, thank you! Apologies for missing that in the PR description.

Octokit/Models/Response/OauthToken.cs
kfcampbell: Got it, thank you! I think it's reasonable to mark the old one as deprecated to be removed in a future major release. After that's done, I can merge and release this!

kfcampbell:
@Kencdk are you okay with the updates I've made? If so, I can ship this!

Kencdk:
> @Kencdk are you okay with the updates I've made? If so, I can ship this!

Yup, looks good to me!

kfcampbell (approved):

Issue #2602, opened 2022-10-28
Pagination with explicitly set StartPage value doesn't work 


hrrrrustic:
```C#
using Octokit;

var github = new GitHubClient(new ProductHeaderValue("MyAmazingApp"));
var options = new ApiOptions()
{
    PageSize = 3,
    StartPage = 1,
    PageCount = 2
};

var a = await github.Issue.GetAllForRepository("dotnet", "runtime", options);
```

I'm expecting that this code should return 6 issues, but actually it just crashes with api limit exception in endless loop requesting only first page

The problem is that next Uri value passed correctly here (with page=2)
https://github.com/octokit/octokit.net/blob/5b28c7ee68ceb82cd9c64516546a8b892b8fce85/Octokit/Http/ApiConnection.cs#L629-L639

But it anyways will be replaced with page=1 here
https://github.com/octokit/octokit.net/blob/5b28c7ee68ceb82cd9c64516546a8b892b8fce85/Octokit/Helpers/UriExtensions.cs#L85-L95

Because parameters dictionary was calculated once before first request (and has page=1) and never updated later. 
https://github.com/octokit/octokit.net/blob/5b28c7ee68ceb82cd9c64516546a8b892b8fce85/Octokit/Http/ApiConnection.cs#L205-L213

Without `StartPage` option everything works as expected

Looks like parameters dictionary should be updated after each page request or manually passed every time same as nextPageUri

hrrrrustic:
Uhhh, found a https://github.com/octokit/octokit.net/issues/1955

PR #2778, opened 2023-09-19
[Bug]Fix pagination when ApiOptions.StartPage is explicitly set.

### LINKED ISSUE #1955: ApiOptions and PageCount
Hi,

I'm just after some clarification regarding the ApiOptions PageCount.

I have a Repository which contains 12 labels. Could you explain why "call 1" would fail (just seems to hangs) and yet "call 2" would succeed and return back 12 labels as expected?

It's a little confusing what PageCount is meant to do when calling GetAllForRepository as it returns a list of Labels and not a list of pages containing labels.

**_call 1_**
`            var result = await this.Client.Issue.Labels.GetAllForRepository(repositoryId, new ApiOptions()
            {
                PageCount = 2,
                PageSize = 11,
                StartPage = 1
            });`

**_call 2_**
`            var result = await this.Client.Issue.Labels.GetAllForRepository(repositoryId, new ApiOptions()
            {
                PageCount = 2,
                PageSize = 12,
                StartPage = 1
            });`

pkindruk:
Resolves #1955, #2602

----

### Before the change?

* Following code never returns a list of release and hits API rate limit:
```
var client = new GitHubClient(new ProductHeaderValue("Test"));
var releases = await client.Repository.Release.GetAll("octokit", "octokit.net",
        new ApiOptions { StartPage = 1, PageSize = 10, PageCount = 2 });
```
* Looks like any method prefixed with `GetAll*` is affected when called with `ApiOptions.StartPage != null` specified and one of the following is true:
  - `ApiOptions.PageCount > 1`
  - `ApiOptions.PageCount == null`

### After the change?

* Example code returns 20 releases as expected

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/Octokit.Tests/Helpers/UriExtensionsTests.cs
+++ b/Octokit.Tests/Helpers/UriExtensionsTests.cs
@@ -161,6 +161,64 @@ namespace Octokit.Tests.Helpers
                 Uri uri = null;
                 Assert.Throws<ArgumentNullException>(() => uri.ApplyParameters(new Dictionary<string, string>()));
             }
+
+            [Fact]
+            public void AppendsStartPageIfNone()
+            {
+                var uri = new Uri("https://api.github.com/repositories/1/labels");
+
+                var parameters = new Dictionary<string, string>();
+                Pagination.Setup(parameters, new ApiOptions { StartPage = 3, PageSize = 7 });
+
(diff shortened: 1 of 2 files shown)


nickfloyd (approved):
Thank you for knocking this out @pkindruk ❤️

PR #2782, opened 2023-09-23
[feat]: SDKs for ActionsArtifacts APIs

### LINKED ISSUE #2780: [Feature]: IActionsArtifactsClient is empty and has no methods available to call
### What happened?

I'm not sure to raise this as a bug or feature, as I'm not sure whether it's just not been developed yet, or it's been overlooked/forgotten about.

The `IActionsArtifactsClient` is empty - So it seems pointless to expose this to the user.

I'd like to be able to download/upload artifacts from Workflows using Octokit, but there's no methods available to call to do this.

```csharp
_gitHubClient.Actions.Artifacts. // No methods other than `object` methods available to call
```

```csharp
namespace Octokit
{
    /// <summary>
    /// A client for GitHub's Actions Artifacts API.
    /// </summary>
    /// <remarks>
    /// See the <a href="https://developer.github.com/v3/actions/artifacts/">Actions Artifacts API documentation</a> for more information.
    /// </remarks>
    public interface IActionsArtifactsClient
    {
    }
}

```

### Versions

Octokit (NuGet) 7.1.0

### Relevant log output

_No response_

### Code of Conduct

thomhurst:
Resolves #2780

----

### Before the change?

* The artifacts client isn't usable currently - There are no methods

### After the change?

Methods to interact with the APIs listed here: https://docs.github.com/en/rest/actions/artifacts

### Pull request checklist

- [ ] Yes
- [x] No

----

I couldn't add integration tests as there isn't any API to create artifacts or anything. But I've added a deserialization test. And then I've hooked into the existing HTTP functionality which should already be thoroughly tested.

I also slightly reworked the internal HTTP sending to receive binary file responses as Streams, rather than raw byte arrays. Artifacts could potentially be large file sizes, and loading these directly into a byte array may not be ideal for memory. So by utilising streams instead, we can reduce memory and write to the disk as we receive data.

### DIFF
--- a/Octokit/Http/HttpClientAdapter.cs
+++ b/Octokit/Http/HttpClientAdapter.cs
@@ -83,25 +83,24 @@ namespace Octokit.Internal
                 "application/x-gzip" ,
                 "application/octet-stream"};
 
-            using (var content = responseMessage.Content)
+            var content = responseMessage.Content;
+            if (content != null)
             {
-                if (content != null)
-                {
-                    contentType = GetContentMediaType(responseMessage.Content);
+                contentType = GetContentMediaType(content);
 
-                    if (contentType != null && (contentType.StartsWith("image/") || binaryContentTypes
+                if (contentType != null && (contentType.StartsWith("image/") || binaryContentTypes
                         .Any(item => item.Equals(contentType, StringComparison.OrdinalIgnoreCase))))
-                    {
-                        responseBody = await responseMessage.Content.ReadAsByteArrayAsync().ConfigureAwait(false);
-                    }
-                    else
-                    {
-                        responseBody = await responseMessage.Content.ReadAsStringAsync().ConfigureAwait(false);
-                    }
-
-                    if (!(preprocessResponseBody is null))
-                        responseBody = preprocessResponseBody(responseBody);
+                {
+                    responseBody = await content.ReadAsStreamAsync().ConfigureAwait(false);
                 }
+                else
+                {
+                    responseBody = await content.ReadAsStringAsync().ConfigureAwait(false);
+                    content.Dispose();
+                }
+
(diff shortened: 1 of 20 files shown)


Octokit/Http/HttpClientAdapter.cs
```diff
@@ -83,25 +83,24 @@ protected virtual async Task<IResponse> BuildResponse(HttpResponseMessage respon
...
-                if (content != null)
-                {
-                    contentType = GetContentMediaType(responseMessage.Content);
+                contentType = GetContentMediaType(content);
 
-                    if (contentType != null && (contentType.StartsWith("image/") || binaryContentTypes
+                if (contentType != null && (contentType.StartsWith("image/") || binaryContentTypes
                         .Any(item => item.Equals(contentType, StringComparison.OrdinalIgnoreCase))))
-                    {
-                        responseBody = await responseMessage.Content.ReadAsByteArrayAsync().ConfigureAwait(false);
```
nickfloyd, line 95: I agree with the preference of using Stream instead of ByteArray here. Given that this has changed would you mind updating the related code docs around the [Response](https://github.com/octokit/octokit.net/blob/03e7644f5ec2bbb9299d4531ca17a14c7569af60/Octokit/Http/Response.cs#L39C84-L39C104) implementation as well, just for clarity sake.

nickfloyd (approved):
This looks great @thomhurst, just a little nit about some code doc clarity... it's definitely not super important, but it would be some nice cleanup.

This looks fantastic otherwaise, thank you for the effort here! ❤️

Octokit/Http/HttpClientAdapter.cs
thomhurst: Updated :)

PR #2787, opened 2023-09-26
Rate Limit Handling

thomhurst:


Octokit/Http/HttpClientAdapter.cs
```diff
@@ -198,6 +198,28 @@ public async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, Can
...
+                var nowInEpochSecondsUtc = DateTimeOffset.UtcNow.ToUnixTimeSeconds();
+                var resetTimeInEpochSecondsUtc = long.Parse(response.Headers.GetValues(ResponseHeaders.RateLimitReset).First());
+
+                var timeToWait = resetTimeInEpochSecondsUtc > nowInEpochSecondsUtc 
+                    ? TimeSpan.FromSeconds(resetTimeInEpochSecondsUtc - nowInEpochSecondsUtc)
+                    : TimeSpan.Zero;
+
+                // Could rate limit refreshes be up to an hour in the future?
+                // Should we conditionally retry only if it's a short period in the future
+                if (timeToWait < TimeSpan.FromSeconds(30))
```
martincostello, line 215: This is _quite a long time_ to wait in the case of microservices. Places where I use Octokit would have a request timeout in the region of a few seconds _at most_, so this would be redundant in those cases as the `CancellationToken` would always fire and prevent another attempt ever happening while masking the HTTP request exception from the 403(429?) with a timeout-related exception instead.

thomhurst: Yeah I do agree. From the documentation, unless I'm wrong, it sounded like rate limits reset hourly. In which case this functionality might not be particularly useful as it would very rarely be hit.

martincostello: Yeah from my experience, once you hit a limit then you just have to wait, unless it's a secondary rate-limit - those reset much sooner (I don't know how often though, I think it's the order of minutes).

dirtygooback:
I have implemented a leaky bucket type API filter in my own code. Not sure how I would port it over to here but the idea is that you have x number of calls that can be used in y time. Only allow one call per z milliseconds. That way you never really run out and it seems to adhere to the rate AND abuse limits.

Reading
https://dzone.com/articles/detailed-explanation-of-guava-ratelimiters-throttl#:~:text=The%20leaky%20bucket%20limits%20the%20constant%20outflow%20rate%2C,is%20always%20stable%2C%20regardless%20of%20the%20inflow%20rate.

PR #2791, opened 2023-10-02
[fix] Fix handling for Streams in IConnection for raw content

### LINKED ISSUE #2789: [BUG]: NullReferenceException thrown using IConnection.GetRaw() for JSON content in Octokit 8.0.1
### What happened?

Calling `IConnection.GetRaw()` for an endpoint that responds with non-binary content (e.g. `application/json`) results in a `NullReferenceException` being thrown when updating to Octokit 8.0.1.

This appears to be [due to the changes](https://github.com/octokit/octokit.net/pull/2782/files#diff-2787e44423613b17ad12829a18645c782b89a5d4fa3cd945274e103c7167e77aR704-R713) in #2782 to handle large binary responses.

In the case where the response is not a `Stream` but instead is just a `string`, `response.Body as Stream` returns `null`, and then causes an exception here:

https://github.com/octokit/octokit.net/blob/0238092ca0370f2edc2ee27e31b7c6e8ba747280/Octokit/Http/Connection.cs#L724

Fix incoming shortly, but a workaround is to use `IConnection.GetRawStream()` and refactor calling code to deal with `Stream` instead of `byte[]`.

### Versions

8.0.1

### Relevant log output

```shell
System.NullReferenceException: Object reference not set to an instance of an object.
    at Octokit.Connection.StreamToByteArray(Stream stream) in /_/Octokit/Http/Connection.cs:line 724
    at Octokit.Connection.GetRaw(IRequest request) in /_/Octokit/Http/Connection.cs:line 704
    at MartinCostello.Costellobot.Pages.DeliveryModel.OnGet(Int64 id) in /_/src/Costellobot/Pages/Webhooks/Delivery.cshtml.cs:line 64
```

### Code of Conduct

martincostello:
Fix `NullReferenceException` if raw content was handled as a string rather than a stream by `HttpClientAdapter.BuildResponse()`.

I haven't added a test for this, but have validated that the apps that surfaced this issue are working when using a local build of Octokit including this change. If you can direct me to exactly where you'd like a test for this, I'll add one.

Resolves #2789

----

### Before the change?

* A `NullReferenceException` would be thrown.

### After the change?

* The call succeeds.

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/Octokit.Tests/Http/ConnectionTests.cs
+++ b/Octokit.Tests/Http/ConnectionTests.cs
@@ -416,6 +416,116 @@ namespace Octokit.Tests.Http
             }
         }
 
+        public class TheGetRawMethod
+        {
+            [Fact]
+            public async Task SendsProperlyFormattedRequestWithProperAcceptHeader()
+            {
+                var httpClient = Substitute.For<IHttpClient>();
+                var response = CreateResponse(HttpStatusCode.OK);
+                httpClient.Send(Args.Request, Args.CancellationToken).Returns(Task.FromResult(response));
+                var connection = new Connection(new ProductHeaderValue("OctokitTests"),
+                    _exampleUri,
+                    Substitute.For<ICredentialStore>(),
+                    httpClient,
+                    Substitute.For<IJsonSerializer>());
+
+                await connection.GetRaw(new Uri("endpoint", UriKind.Relative), new Dictionary<string, string>());
+
+                httpClient.Received(1).Send(Arg.Is<IRequest>(req =>
+                    req.BaseAddress == _exampleUri &&
+                    req.ContentType == null &&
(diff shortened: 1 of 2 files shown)


martincostello:
Looks like this got closed by accident because I put the wrong `#` in the commit message for the other PR, and I can't re-open it.

martincostello:
@nickfloyd Anything more you need from me to progress this fix?

kfcampbell:
I do think it'd be possible to write a test for this case in `Octokit.Tests/Http/ConnectionTests.cs`, since the modified `GetRaw` is called by two public functions. Why don't we try doing so?

kfcampbell (approved):
Looks good, thank you!

PR #2803, opened 2023-10-14
Getarchive

### LINKED ISSUE #2802: [BUG]: RepositoryContentsClient.GetArchive does not return the expected binary content
### What happened?

This [commit](https://github.com/octokit/octokit.net/commit/958bc5f1f87cd97cd8520ac74f958c2df12903b3?diff=unified) which was published in Octokit v8.1.1 introduced a change in `HttpClientAdapter.cs` to return binary content as a stream rather than an array of byte. However, the code in RepositoryContentsClient.GetArchive still expects `byte[]` and was not adjusted to handle a stream.

This problem was not caught in unit testing because none of the tests in `RepositoryContentsClientTests` validate that GetArchive returns the expected array of bytes. I wrote the following unit test to demonstrate the problem. This unit test fails because the response does not contain the expected bytes:
```csharp
[Fact]
public async Task ReturnsExpectedContent()
{
	var headers = new Dictionary<string, string>();
	var response = TestSetup.CreateResponse(HttpStatusCode.OK, new byte[] { 1, 2, 3, 4 }, headers);
	var responseTask = Task.FromResult<IApiResponse<byte[]>>(new ApiResponse<byte[]>(response));

	var connection = Substitute.For<IConnection>();
	connection.Get<byte[]>(Arg.Is<Uri>(u => u.ToString() == "repos/org/repo/tarball/"), null)
		.Returns(responseTask);

	var apiConnection = Substitute.For<IApiConnection>();
	apiConnection.Connection.Returns(connection);

	var client = new RepositoryContentsClient(apiConnection);

	var actual = await client.GetArchive("org", "repo");

	Assert.Equal(new byte[] { 1, 2, 3, 4 }, actual); // <-- this assertion fails, thereby demonstrating that GetArchive does not return the expected array of bytes
}
```

I will submit a PR to resolve this problem and ensure that the unit test presented above completes successfully.

### Versions

This problem can be observed in Octokit 8.1.1.

I verified 7.1.0, 7.2.0, 8.0.0 and 8.0.1 and they work fine. Meaning: they return the expected array of bytes.

### Relevant log output

_No response_

### Code of Conduct

Jericho:
Resolves #2802

----

### Before the change?
RepositoryContentsClient.GetArchive expects the result of the API call to contain a byte[] when in fact a stream is returned. This bug is due to a change introduced [here](https://github.com/octokit/octokit.net/commit/958bc5f1f87cd97cd8520ac74f958c2df12903b3?diff=unified#diff-a2515f648f91f996285bf547708f9e1e8450b04ac226e3f97114510e0cb476e7).

### After the change?
RepositoryContentsClient.GetArchive invokes Connection.GetRaw rather than Connection.Get<byte[]> in order to be the binary content of the response.

### Pull request checklist
- [X] Tests for the changes have been added (for bug fixes / features)
- [X] Docs have been reviewed and added / updated if needed (for bug fixes / features)

- [ ] Yes
- [x] No

To ensure that I was not introducing a breaking change, I created an overload of the existing `GetRaw` method to accept a timeout parameter. I could have added an additional parameter to the existing GetRaw method, but that would have been a breaking change. If you prefer the "additional-parameter-to-existing-method" rather than "overload-existing-method", I'll be happy to modifiy my PR.

----

### DIFF
--- a/Octokit.Tests/Clients/RepositoryContentsClientTests.cs
+++ b/Octokit.Tests/Clients/RepositoryContentsClientTests.cs
@@ -4,6 +4,8 @@ using System.Threading.Tasks;
 using NSubstitute;
 using Xunit;
 using System.Collections.Generic;
+using System.Net;
+using Octokit.Internal;
 
 namespace Octokit.Tests.Clients
 {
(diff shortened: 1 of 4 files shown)


kfcampbell (approved):
I'm comfortable with the decision to overload here.

PR #2821, opened 2023-11-24
Allow the constructor for `ProductHeaderValue` to accept a framework version of the object

Cyberboss:

- [ ] Yes
- [x] No

----

I think this is fairly straightforward. Currently, if you already have a `System.Net.Http.Headers.ProductHeaderValue`, you have to deconstruct it to pass it to the Octokit constructor. Then, it proceeds to just make an unnecessary copy of it.

Simplify things by allowing one to be passed in directly.


Octokit/Http/ProductHeaderValue.cs
```diff
@@ -38,9 +40,16 @@ public ProductHeaderValue(string name, string version)
...
         }
 
-        ProductHeaderValue(System.Net.Http.Headers.ProductHeaderValue productHeader)
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ProductHeaderValue"/> class.
+        /// </summary>
+        /// <remarks>
+        /// See more information regarding User-Agent requirements here: https://developer.github.com/v3/#user-agent-required
+        /// </remarks>
+        /// <param name="name">The <see cref="System.Net.Http.Headers.ProductHeaderValue"/>.</param>
```
kfcampbell, line 49: Can you fix these check warnings please?

Octokit/Http/ProductHeaderValue.cs
```diff
@@ -38,9 +40,16 @@ public ProductHeaderValue(string name, string version)
...
+        public ProductHeaderValue(System.Net.Http.Headers.ProductHeaderValue productHeader)
```
kfcampbell, line 50: Is this the meaningful part of the change? I'm having a hard time figuring out how the internal _productHeaderValue was set before the change.

Cyberboss: The changing of the constructor from private to public is the meaningful bit yeah. All other constructors call it.

PR #2862, opened 2024-01-25
Fix namespaces for Artifacts models

### LINKED ISSUE #2861: [BUG]: class Artifact is in global namespace
### What happened?

I am using Nuke. 
I have my custom code to define Build and have defined the Artifact class in my custom code. 
After updating Nuke to the new version 8.0.0 which depends on >= 9.0.0, the Artifact class from Octokit wins with my custom class which breaks usages of my library and I can't fix this without breaking change. 
This is because Artifact from Octokit is in the global namespace. 

Is there any chance that this type could be moved to some namespace? :) 

### Versions

9.0.0 for sure. 

### Relevant log output

```shell
Namespace '<global namespace>' contains a definition conflicting with alias 'Artifact'
```

### Code of Conduct

thomhurst:
Resolves #2861

Fixes namespace clashes

### DIFF
--- a/Octokit/Models/Response/Artifact.cs
+++ b/Octokit/Models/Response/Artifact.cs
@@ -2,82 +2,85 @@
 using System.Diagnostics;
 using System.Globalization;
 
-[DebuggerDisplay("{DebuggerDisplay,nq}")]
-public class Artifact
+namespace Octokit
 {
-    public Artifact()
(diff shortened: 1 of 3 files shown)


amis92 (approved):
Thanks a lot!

darthkurak (approved):

nickfloyd (approved):
Thanks for the follow up here, I completely missed this during the review.  ❤️

PR #2865, opened 2024-01-27
[Feat] Add Advanced Security Properties to Repository Get/Update

SlyckLizzie:
Resolves #ISSUE_NUMBER
N/A
----

### Before the change?

* Repository get/update did not implement Advanced Security properties

### After the change?

* Repository get/update implements Advanced Security properties

### Pull request checklist

- [x] Yes
- [ ] No

----

### DIFF
--- a/Octokit/Models/Common/Status.cs
+++ b/Octokit/Models/Common/Status.cs
@@ -1,0 +1,12 @@
+﻿using Octokit.Internal;
+
+namespace Octokit
+{
+    public enum Status
+    {
+        [Parameter(Value = "enabled")]
+        Enabled,
+        [Parameter(Value = "disabled")]
+        Disabled
+    }
+}
--- a/Octokit/Models/Request/RepositoryUpdate.cs
+++ b/Octokit/Models/Request/RepositoryUpdate.cs
@@ -1,4 +1,5 @@
 ﻿using Octokit.Internal;
+using Octokit.Models.Response;
 using System;
 using System.Diagnostics;
 
@@ -132,6 +133,8 @@ namespace Octokit
         /// Optional. Gets or sets whether to enable discussions for the repository. The default is null (do not update). The default when created is false.
         /// </summary>
         public bool? HasDiscussions { get; set; }
+
+        public SecurityAndAnalysisRequest SecurityAndAnalysis { get; set; }
 
         internal string DebuggerDisplay => new SimpleJsonSerializer().Serialize(this);
     }
--- a/Octokit/Models/Request/SecurityAndAnalysisRequest.cs
+++ b/Octokit/Models/Request/SecurityAndAnalysisRequest.cs
@@ -1,0 +1,184 @@
+﻿using Octokit.Internal;
+using System.Diagnostics;
+
+
+namespace Octokit
+{
+    /// <summary>
+    /// The security and analysis features that are enabled or disabled for the repository
+    /// </summary>
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
+    public class SecurityAndAnalysisRequest
+    {
+        public SecurityAndAnalysisRequest()
+        { }
+
+        public SecurityAndAnalysisRequest(AdvancedSecurityRequest advancedSecurity, DependabotSecurityUpdatesRequest dependabotSecurityUpdates, 
+            SecretScanningRequest secretScanning, SecretScanningPushProtectionRequest secretScanningPushProtection, 
+            SecretScanningValidityChecksRequest secretScanningValidityChecks)
+        {
+            this.AdvancedSecurity = advancedSecurity;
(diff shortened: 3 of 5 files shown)


SlyckLizzie:
Breaking change is because the current C# language is targeting 7.3 and in order to make a reference object nullable it needs to be targeting at least 8.0. The repository response and update objects will require implementers to add in the SecurityAndAnalysis objects.

kfcampbell:
@nickfloyd are you comfortable with this support bump? C# 7.0 came out in 2017, but I can't seem to find any information on the support windows. Some versions of .NET [still default to C# 7.x](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version#defaults), though I'm not sure if we should still support those.

nickfloyd:
> Breaking change is because the current C# language is targeting 7.3 and in order to make a reference object nullable it needs to be targeting at least 8.0. The repository response and update objects will require implementers to add in the SecurityAndAnalysis objects.

Hey @SlyckLizzie, for clarity reference types in c# have always been nullable.  I am guessing you are referring to the enum that was added - `status`, is that correct? 

Enums, which are value types, will allow the nullable operator (since it was introduced in c# 2.0) the only real difference is that we have to do some safety checks like `.Value`, `.HasValue`, or string comparison checks.  

If you get the chance, would you let me know where in your change set is the modification that will require .NET 8.0? I am sure I am missing it - perhaps with the way we serialize and deserialize things?  Let me know so that I can better understand the constraint.

Somewhat related we still have users on this SDK that are still on C# 7.x ([LTS](https://learn.microsoft.com/en-us/lifecycle/products/microsoft-net-and-net-core)) - which is still in maintenance until May 14, 2024 and .NET 6 which is active until November 12, 2024 - we tend to follow the language release cycle for this SDK.

FYI, our generative [.NET SDK](https://github.com/octokit/dotnet-sdk) is targeting .NET 8 as is. 

Thanks for this addition, I'm looking forward to getting it in!

SlyckLizzie:
@nickfloyd - It's not the framework version but the language version targeted by the project file. netstandard2.0 targets the c# language 7.3. In order for me to be able to make the  SecurityAndAnalysisRequest input parameter of the Respository Response model nullable, e.g. SecurityAndAnalysisRequest? the target framework of the project needs to be netstandard2.1 to target the c# language version 8.0

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version

SlyckLizzie:
@nickfloyd - Is there still an issue with this? I noticed a new breaking change version update to the package. Why not add this in with that?

nickfloyd:
> @nickfloyd - Is there still an issue with this? I noticed a new breaking change version update to the package. Why not add this in with that?

Because I completely dropped the ball on this.  I'm sorry about that... 

We can move forward with what you've suggested.  I'll need to roll it out in a new major release (that's on me) but it should be fine. Go ahead and change the target to 2.1 when you get the chance and we can get this out the door.

Again, apologies for the delay.

SlyckLizzie:
@nickfloyd  - It's all good. I'll make those changes by the end of the weekend. :)

SlyckLizzie:
@nickfloyd - Apparently that change isn't going to play nice with the .Net Framework. If it's not .net core or later then it has to reference net standard 2.0

https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-1
image

nickfloyd:
> @nickfloyd - Apparently that change isn't going to play nice with the .Net Framework. If it's not .net core or later then it has to reference net standard 2.0. The test and pagnationextension projects still have a reference to 4.62. Everything seems to compile fine without the 4.62 but I want to make sure that's the direction you want to go.
> 

I was really hoping that all of the projects were going to play well with this change.  Let me see if I can determine if 4.6.2 is actually a requirement in those projects or if they can be bumped up as well. 

I still believe moving this direction is the right way to go - but it might take a bit more work to get there.  Thanks again for the heads up... I'll have a look and let you know what I find / or if you beat me to it - either way.

ddaniels-andmore:
@nickfloyd any status update on this. It's really like to see it get included in the next release if possible

nickfloyd:
> @nickfloyd any status update on this. I'd really like to see it get included in the next release if possible

Hey @ddaniels-andmore I am planning on taking a look this Friday (06/07).  I'll comment and let you know what I find/come up with.

SlyckLizzie:
@nickfloyd - What's blocking this from being added as part of a breaking change release?

nickfloyd (approved):

nickfloyd:
Just a head's up: I'm going to make more changes from this [issue](https://github.com/octokit/octokit.net/issues/2893) before shipping this to avoid shipping another major release right after this one.

PR #2871, opened 2024-01-30
bugfix #2870

### LINKED ISSUE #2870: [BUG]: exception "mediatype is null or empty" when trying to delete file
### What happened?

When trying to delete a file using 
```csharp
var deleteRequest =   new DeleteFileRequest($"delete file", content.content.Sha);
await gitHubClient.Repository.Content.DeleteFile(RepositoryId,$"notes/{noteName}.md" , deleteRequest);
```

I get the following exception 
```
System.ArgumentException: The value cannot be null or empty. (Parameter 'mediaType')
   at System.Net.Http.Headers.MediaTypeHeaderValue.CheckMediaTypeFormat(String mediaType, String parameterName)
   at System.Net.Http.StringContent..ctor(String content, Encoding encoding, String mediaType)
   at Octokit.Internal.HttpClientAdapter.BuildRequestMessage(IRequest request) in /_/Octokit/Http/HttpClientAdapter.cs:line 147
   at Octokit.Internal.HttpClientAdapter.Send(IRequest request, CancellationToken cancellationToken, Func`2 preprocessResponseBody) in /_/Octokit/Http/HttpClientAdapter.cs:line 48
   at Octokit.Connection.RunRequest(IRequest request, CancellationToken cancellationToken, Func`2 preprocessResponseBody) in /_/Octokit/Http/Connection.cs:line 761
   at Octokit.Connection.Run[T](IRequest request, CancellationToken cancellationToken, Func`2 preprocessResponseBody) in /_/Octokit/Http/Connection.cs:line 752
   at Octokit.Connection.Delete(Uri uri, Object data) in /_/Octokit/Http/Connection.cs:line 602
```

### Versions

latest version : octokit.net v9.1.1 (did not try previous versions)

### Relevant log output

_No response_

### Code of Conduct

b3b00:
Resolves #2870 

----

### Before the change?
When trying to delete a file an exception is raised stating `System.ArgumentException: The value cannot be null or empty. (Parameter 'mediaType')`. And the file is not deleted

### After the change?
No exception is raised and the file is deleted.

### Pull request checklist

----

### DIFF
--- a/Octokit/Http/Connection.cs
+++ b/Octokit/Http/Connection.cs
@@ -417,7 +417,7 @@ namespace Octokit
             TimeSpan timeout,
             CancellationToken cancellationToken,
             string twoFactorAuthenticationCode = null,
-            Uri baseAddress = null, 
+            Uri baseAddress = null,
             Func<object, object> preprocessResponseBody = null)
         {
             Ensure.ArgumentNotNull(uri, nameof(uri));
@@ -596,7 +596,8 @@ namespace Octokit
                 Method = HttpMethod.Delete,
                 Body = data,
                 BaseAddress = BaseAddress,
-                Endpoint = uri
+                Endpoint = uri,
+                ContentType = "application/json"
             };
             var response = await Run<object>(request, CancellationToken.None).ConfigureAwait(false);
             return response.HttpResponse.StatusCode;
@@ -722,12 +723,12 @@ namespace Octokit
 
             return new ApiResponse<byte[]>(response, response.Body as byte[]);
         }
-        
+
         async Task<IApiResponse<Stream>> GetRawStream(IRequest request)
         {
             request.Headers.Add("Accept", AcceptHeaders.RawContentMediaType);
             var response = await RunRequest(request, CancellationToken.None).ConfigureAwait(false);
-            
+
             return new ApiResponse<Stream>(response, response.Body as Stream);
         }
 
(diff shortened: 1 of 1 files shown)


kfcampbell:
Hmm...I'm uncomfortable making this change with such a broad impact. Can it be scoped down to the one request necessary?

b3b00:
@kfcampbell , I agree that this can be dangerous. I will look at a way to narrow this fix to the delete request and make it available to other kind of request if the issue appears on other. Thanks for your comment

nickfloyd:
@b3b00 if you get the chance post what your use (code) of the SDK here looks like - specifically I am trying to find out which resource client you're using so that we can help you craft a solution.  We have some mechanisms in place to support adding headers to the request - adding where you have will definitely have some side effects for all of the other implementations that use `SendData`.

b3b00:
Hello @nickfloyd, my code : 

```c#
     public override async Task DeleteNote(string noteName)
        {
            if (gitHubClient != null)
            {
                var content = await NoteExists(noteName);
                if (content.exists)
                {
                    // this is were the exception happens
                    var deleteRequest =
                        new DeleteFileRequest($"DendrOnline : delete note {noteName}", content.content.Sha);
                    await gitHubClient.Repository.Content.DeleteFile(RepositoryId,$"notes/{noteName}.md" , deleteRequest);
                }
            }
        }
```

Issue #2876, opened 2024-02-03
[BUG]: Failed API calls update response cache


daverant:
### What happened?

The response cache in `CachingHttpClient` is called even when GitHub returns a non-2xx and non-304 status code. It's likely that implementations of `IResponseCache` will only want to cache data for successful responses. Without this check, it's easy for implementations to miss this check and flush valid responses from the cache when receiving a non-2xx response, resulting in an increase of subsequent cache misses.

This is an opinionated change - it's reasonable to expect implementations to cater for this themselves, but feels like it would help users more than hinder (I can't think of a use case for caching non-successful responses).

### Versions

octokit.net v9.1.2, net7.0

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

kfcampbell:
I'm generally okay with this approach. @nickfloyd are you? 

I'm also curious how redirect status codes (like a 301 or 307) affect caching.

PR #2877, opened 2024-02-03
BREAKING CHANGE (behavior): Modify caching to only attempt to update the response cache if a 2xx response code is received from GitHub

daverant:
Resolves https://github.com/octokit/octokit.net/issues/2876

Apologies if I've filed a bug when it should come under feature/enhancement!
I couldn't find any docs on response caching, but happy to update them if there are any.

----

### Before the change?

* If the `CachingHttpClient` received a non-200 and non-304 response from GitHub, it still attempts to update the underlying response cache, resulting in a higher number of subsequent cache misses if `IResponseCache` implementations don't guard for it themselves.

### After the change?

* Only attempts to update the response cache if a 2xx response code is received from GitHub.

### Pull request checklist

* I don't believe this is classed as a breaking change, but it is a change in behaviour that could impact users who rely on it calling the response cache for non-200 responses.

- [x] Yes
- [ ] No

### DIFF
--- a/Octokit.Tests/Caching/CachingHttpClientTests.cs
+++ b/Octokit.Tests/Caching/CachingHttpClientTests.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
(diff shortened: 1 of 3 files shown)


nickfloyd:
I am going to classify this as a breaking change due to the change in behavior in potential expectations that you point out.  This is a great change, thank you for doing it! ❤️

nickfloyd (approved):

PR #2885, opened 2024-02-14
Add 'Pending' member to WorkflowJobStatus enum

### LINKED ISSUE #2882: [BUG]: WorkflowJobStatus enum does not contain "pending" value.
### What happened?

I get an exception when trying to read the value of `WorkflowJobStep.Status` as it sends 'pending' but that's not a valid value for that Enum. It seems very similar to this issue from last year: https://github.com/octokit/octokit.net/issues/2656

### Versions

Using Octokit 9.1.2 on .NET 8 running on Windows.

### Relevant log output

_No response_

### Code of Conduct

redoz:
Resolves #2882

----

### Before the change?
* The 'pending' value on the `WorkflowJobStep.Status` would not deserialize properly.

### After the change?
* The 'pending' value on the `WorkflowJobStep.Status` should deserialize properly.

### Pull request checklist

_I didn't see any tests/docs relating to this enum, so I didn't create any for this PR, let me know if I missed something and should add them._

- [ ] Yes
- [x] No

----

### DIFF
--- a/Octokit/Models/Response/WorkflowJobStatus.cs
+++ b/Octokit/Models/Response/WorkflowJobStatus.cs
@@ -12,5 +12,7 @@ namespace Octokit
         Completed,
         [Parameter(Value = "waiting")]
         Waiting,
+        [Parameter(Value = "pending")]
+        Pending,
     }
 }


nickfloyd (approved):
Hey @redoz thanks for the contributions here! ❤️ Looks like we're missing quite a few here ([docs](https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28), [.NET generated SDK](https://github.com/octokit/dotnet-sdk/blob/main/src/GitHub/Repos/Item/Item/Actions/Workflows/Item/Runs/GetStatusQueryParameterType.cs#L15))

PR #2890, opened 2024-02-22
Change "IssueId" to `long`

### LINKED ISSUE #2889: [BUG]: Error deserializing new issues
### What happened?

I've begun noticing errors deserializing the issue ID number when receiving issue webhooks. The value appears to be an `int` in the models but it seems GitHub has finally crossed the threshold and needs a `long`. My internal issue ID number was `2147620872`, and the log was for a comment created webhook request.

### Versions

1.0.0 (we haven't upgraded in a while) but bug appears to still be present in the main branch

### Relevant log output

```shell
JSON integer 2147620872 is too large or small for an Int32. Path 'issue.id', line 1, position 612.
  at Newtonsoft.Json.JsonTextReader.ParseReadNumber(ReadType readType, Char firstChar, Int32 initialPosition)\n   at Newtonsoft.Json.JsonTextReader.ReadNumberValue(ReadType readType)\n   at Newtonsoft.Json.JsonTextReader.ReadAsInt32()
```

### Code of Conduct

Leh2:
Resolves #2889 

----

### Before the change?

* New issues couldn't be handled by the client because GitHub IDs have surpassed `int.Max`.

### After the change?

* New issues know works in the client

### Pull request checklist

Consumers using the "Id" will experience a breaking change and need to update their type to `long`.

- [x] Yes
- [ ] No

----

### DIFF
--- a/Octokit/Models/Response/Issue.cs
+++ b/Octokit/Models/Response/Issue.cs
@@ -45,7 +45,7 @@ namespace Octokit
         /// <summary>
         /// The internal Id for this issue (not the issue number)
         /// </summary>
-        public int Id { get; private set; }
+        public long Id { get; private set; }
 
         /// <summary>
         /// GraphQL Node Id
--- a/Octokit.Tests.Integration/Clients/ProjectCardsClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/ProjectCardsClientTests.cs
@@ -453,7 +453,7 @@ public class ProjectCardsClientTests
         return result;
     }
 
-    private static async Task<ProjectCard> CreateIssueCardHelper(IGitHubClient githubClient, int issueId, int columnId)
+    private static async Task<ProjectCard> CreateIssueCardHelper(IGitHubClient githubClient, long issueId, int columnId)
     {
         var newCard = new NewProjectCard(issueId, ProjectCardContentType.Issue);
         var result = await githubClient.Repository.Project.Card.Create(columnId, newCard);
--- a/Octokit.Tests.Integration/Reactive/ObservableProjectCardsClientTests.cs
+++ b/Octokit.Tests.Integration/Reactive/ObservableProjectCardsClientTests.cs
@@ -441,7 +441,7 @@ public class ObservableProjectCardsClientTests
         return result;
     }
 
-    private static async Task<ProjectCard> CreateIssueCardHelper(IObservableGitHubClient githubClient, int issueId, int columnId)
+    private static async Task<ProjectCard> CreateIssueCardHelper(IObservableGitHubClient githubClient, long issueId, int columnId)
     {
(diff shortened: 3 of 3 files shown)


Octokit/Models/Response/Issue.cs
```diff
@@ -45,7 +45,7 @@ public Issue(string url, string htmlUrl, string commentsUrl, string eventsUrl, i
         /// <summary>
         /// The internal Id for this issue (not the issue number)
         /// </summary>
-        public int Id { get; private set; }
+        public long Id { get; private set; }
```
arxange1, line 48: `int id` in the constructor to be fixed as well

Leh2: Good catch - fixed in https://github.com/octokit/octokit.net/pull/2890/commits/b1d53cc49bdaf560a71431113babaecc355308cb and https://github.com/octokit/octokit.net/pull/2890/commits/cc2dd92b072310f32bff72bb69156f068259f546

arxange1 (approved):

JimSuplizio:
Crazy question, related to the PR after it gets merged. What's the timeline between this getting merged and a release being up on nuget?

JimSuplizio (changes requested):
Unfortunately, there's a lot more that needs to be done than just changing the Id in Issue.cs. For example, the [IIssuesClient.cs](https://github.com/octokit/octokit.net/blob/main/Octokit/Clients/IIssuesClient.cs#L306C21-L306C27) uses int all over the place for the Issue number (Id). Further, does the GitHub API even support [long when you're updating](https://docs.github.com/en/rest/issues/issues?apiVersion=2022-11-28#update-an-issue) an issue? As far as I can tell, the Latest Rest API version, which is what Octokit is using, only documents Int, not long.

Leh2:
> Unfortunately, there's a lot more that needs to be done than just changing the Id in Issue.cs. For example, the [IIssuesClient.cs](https://github.com/octokit/octokit.net/blob/main/Octokit/Clients/IIssuesClient.cs#L306C21-L306C27) uses int all over the place for the Issue number (Id). Further, does the GitHub API even support [long when you're updating](https://docs.github.com/en/rest/issues/issues?apiVersion=2022-11-28#update-an-issue) an issue? As far as I can tell, the Latest Rest API version, which is what Octokit is using, only documents Int, not long.

Issue number is not the same as issue id.

danmoseley:
are there any other values that will roll over in the near future and should change to long now as well?

kfcampbell (approved):

kfcampbell:
Thank you for the quick fix! Please report back if something seems amiss.

Leh2:
> Thank you for the quick fix! Please report back if something seems amiss.

Just tested the new package, and everything works as expected again 👍 Thank you for the NuGet release 💪

PR #2891, opened 2024-02-22
Issue id can be bigger than max int value.

thaystg:
The call stack of the exception that I got:
image

The issue id that I'm getting is: `"id":2149421784`


Leh2:
Duplicate of https://github.com/octokit/octokit.net/pull/2890 :)

PR #2902, opened 2024-04-10
fix: Allow `zip` MIME type

### LINKED ISSUE #2898: [BUG]: Download Artifact - Content Stores Differing Content-Types Cause Errors
### What happened?

Afternoon!

When calling `ActionsArtifactClient.DownloadArtifact`, the download is often, or always, redirected to another server that holds the actual artifact.

In my experience, if the artifact happens to sit on _https://pipelinesghubeus10.actions.githubusercontent.com_, then the `Content-Type` returned is `application/zip` and all is well, but if the artifact happens to reside on _https://productionresultssa10.blob.core.windows.net_, then the `Content-Type` is `zip` and it all starts to go wrong.

There's a check in `HttpClientAdapter.BuildResponse` for the content type, which fails, then the response is read as a string, not a stream, and `Connection.GetRawStream` attempts to cast the string as a stream, understandably gets null, and returns null to the caller.

Easy to fix badly; perhaps harder to fix well.

### Versions

Octokit 10.0.0

### Relevant log output

_No response_

### Code of Conduct

thomhurst:
Fixes #2898

### DIFF
--- a/Octokit/Http/HttpClientAdapter.cs
+++ b/Octokit/Http/HttpClientAdapter.cs
@@ -81,6 +81,7 @@ namespace Octokit.Internal
                 AcceptHeaders.RawContentMediaType,
                 "application/zip" ,
                 "application/x-gzip" ,
+                "zip" , // Not a standard MIME type but see issue #2898
                 "application/octet-stream"};
 
             var content = responseMessage.Content;


Octokit/Http/HttpClientAdapter.cs
```diff
@@ -81,6 +81,7 @@ protected virtual async Task<IResponse> BuildResponse(HttpResponseMessage respon
                 AcceptHeaders.RawContentMediaType,
                 "application/zip" ,
                 "application/x-gzip" ,
+                "zip" , // Not a standard MIME type but see issue #2898
```
nickfloyd, line 84: While I am not super excited about not using the proper representation (asper the mdm web docs and iana) - I get the situation here.  Thank you for making sure this is commented with context. ❤️

nickfloyd (approved):
Thanks again for this change @thomhurst

PR #2905, opened 2024-04-13
fix: "zip" Content-Type resulting in null Stream for Artifacts on Blob Storage

### LINKED ISSUE #2898: [BUG]: Download Artifact - Content Stores Differing Content-Types Cause Errors

thomhurst:
Fixes #2898 

Sorry, the last one didn't actually solve it because the Content-Type wasn't being set properly by the `System.Net` classes - Probably due to an invalid format.

### DIFF
--- a/Octokit.Tests/Http/HttpClientAdapterTests.cs
+++ b/Octokit.Tests/Http/HttpClientAdapterTests.cs
@@ -177,6 +177,27 @@ namespace Octokit.Tests.Http
 
                 Assert.Equal("application/json", response.ContentType);
             }
+            
+            // See #2898 for why this is necessary
+            // Non standard MIME Content-Type coming from Blob Storage when downloading artifacts
+            [Fact]
+            public async Task SetsZipContentType()
+            {
+                var memoryStream = new MemoryStream();
+                var streamContent = new StreamContent(memoryStream);
+                streamContent.Headers.TryAddWithoutValidation("Content-Type", "zip");
+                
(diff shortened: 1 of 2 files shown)


jack0x539 (approved):

nickfloyd (approved):
Thanks for the docs describing why this is necessary, tests and validation here.  ❤️ I'm going to take a note to see if I can determine why blob storage is behaving like this - I just don't feel good about the behavior of storage in this case. 

Thank you for the quick turn around on this.

nickfloyd:
Hey @thomhurst Absolutely... the new patch is building / releasing now.  Thanks again for the efforts here!

PR #2911, opened 2024-04-23
Fix #2693 PageCount doesn't work if the query gets constructed with page number as the first query parameter 

### LINKED ISSUE #2693: [BUG]: Specifying PageCount doesn't work if the query gets constructed with page number as the first query parameter
### What happened?

I tried requesting 5 pages of file changes in a PR, for example:
```cs
var apiOptions = new ApiOptions
{
    PageCount = 5,
};
var pullRequest = await GitHubClient.PullRequest.Get(repoOwnerName, repositoryName, pullRequestNumber);
```

which actually requests all pages.

This is because of [this code in `Pagination.cs`](https://github.com/octokit/octokit.net/blob/f10f5f36a2d4b2eb416e5ea7406cb6500b1d71db/Octokit/Helpers/Pagination.cs#L59)
```cs
return uri.Query.Split('&')
    .Select(keyValue =>
    {
        var indexOf = keyValue.IndexOf('=');
        if (indexOf > 0)
        {
            var key = keyValue.Substring(0, indexOf);
            var value = keyValue.Substring(indexOf + 1);
            return new KeyValuePair<string, string>(key, value);
        }

        //just a plain old value, return it
        return new KeyValuePair<string, string>(keyValue, null);
    })
    .ToDictionary(x => x.Key, x => x.Value);
```
which incorrectly includes the initial `?`, which you can see the effect of in this debugger screenshot
image
and `TryGetValue` naturally returns false.

### Versions

Octokit.net 5.0.2

### Relevant log output

_No response_

### Code of Conduct

andrew-from-toronto:
Resolves #2693

### Before the change?

* PageCount did not work if the query gets constructed with page number as the first query parameter

### After the change?

* PageCount now works if the query gets constructed with page number as the first query parameter

### Pull request checklist
- [X] Tests for the changes have been added (for bug fixes / features)
- [X] Docs have been reviewed and added / updated if needed (for bug fixes / features)

- [ ] Yes
- [X] No

### DIFF
--- a/Octokit.Tests/Http/PaginationTests.cs
+++ b/Octokit.Tests/Http/PaginationTests.cs
@@ -1,0 +1,46 @@
+using System;
+using Octokit.Models.Request.Enterprise;
+using Xunit;
+
+namespace Octokit.Tests.Http
+{
+    public class PaginationTests
+    {
+        public class TheShouldContinueMethod
+        {
+            [Fact]
+            public void HandlesMissingUri()
+            {
+                var result = Pagination.ShouldContinue(null, null);
+                Assert.False(result);
+            }
+            
+            [Fact]
+            public void HandlesIsDone()
+            {
+                var uri = new Uri("http://example.com");
+                var options = new ApiOptionsExtended { IsDone = true };
+                var result = Pagination.ShouldContinue(uri, options);
+                Assert.False(result);
+            }
+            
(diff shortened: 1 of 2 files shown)


kfcampbell (approved):
Thank you for your attention to testing!

PR #2937, opened 2024-06-20
docs: Provide easy alternative to create App JWT token

rasmus:
This adds a simple code snippet for creating the JWT token required to authenticate GitHub Apps. All dependencies live in the `System` namespace.

- `System.IdentityModel.Tokens.Jwt`
- `System.Security.Claims`
- `System.Security.Cryptography`

----

### Before the change?

Developers would think that the only alternative was to import another dependency.

### After the change?

Developers are able to create JWT tokens for their GitHub Apps using dependencies they likely already have imported.

### Pull request checklist
- ~Tests for the changes have been added (for bug fixes / features)~

- [ ] Yes
- [x] No

----

### DIFF
--- a/docs/github-apps.md
+++ b/docs/github-apps.md
@@ -100,7 +100,38 @@ That concludes the walkthrough!
 ### A Note on JWT Tokens
 Octokit.net aims to have no external dependencies, therefore we do not currently have the ability to generate/sign JWT tokens for you, and instead expect that you will pass in the appropriately signed JWT token required to authenticate the `GitHubApp`.
 
-Luckily one of our contributors [@adriangodong](https://github.com/adriangodong) has created a library `GitHubJwt` ( [GitHub](https://github.com/adriangodong/githubjwt) | [NuGet](https://www.nuget.org/packages/githubjwt) ) which you can use as per the following example.
+In order to create the token, you can create it manually using the following snippet.
+
+``` csharp
+var rsaPrivateKey = "..."; // RSA private key from the App configuration page
+var appId = 1; // The GitHub App Id
+
+using var rsa = RSA.Create();
+rsa.ImportFromPem(rsaPrivateKey);
+var signingCredentials = new SigningCredentials(new RsaSecurityKey(rsa), SecurityAlgorithms.RsaSha256)
+{
+    CryptoProviderFactory = new CryptoProviderFactory
+    {
+        CacheSignatureProviders = false
+    }
+};
+
+var now = DateTime.UtcNow;
+var expiresAt = now + TokenLifetime;
+var jwt = new JwtSecurityToken(
+    notBefore: now,
+    expires: now + TimeSpan.FromMinutes(10),
+    signingCredentials: signingCredentials,
+    claims: new[]
+    {
+        new Claim("iat", new DateTimeOffset(now).ToUnixTimeSeconds().ToString(), ClaimValueTypes.Integer),
+        new Claim("iss", appId.ToString(), ClaimValueTypes.Integer),
+    }
+);
+var token = new JwtSecurityTokenHandler().WriteToken(jwt);
+```
+
+Alternatively, one of our contributors [@adriangodong](https://github.com/adriangodong) has created a library `GitHubJwt` ( [GitHub](https://github.com/adriangodong/githubjwt) | [NuGet](https://www.nuget.org/packages/githubjwt) ) which you can use as per the following example.
 
 ``` csharp
 // Use GitHubJwt library to create the GitHubApp Jwt Token using our private certificate PEM file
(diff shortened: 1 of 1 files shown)


docs/github-apps.md
```diff
@@ -100,7 +100,38 @@ That concludes the walkthrough!
 ### A Note on JWT Tokens
 Octokit.net aims to have no external dependencies, therefore we do not currently have the ability to generate/sign JWT tokens for you, and instead expect that you will pass in the appropriately signed JWT token required to authenticate the `GitHubApp`.
 
-Luckily one of our contributors [@adriangodong](https://github.com/adriangodong) has created a library `GitHubJwt` ( [GitHub](https://github.com/adriangodong/githubjwt) | [NuGet](https://www.nuget.org/packages/githubjwt) ) which you can use as per the following example.
+In order to create the token, you can create it manually using the following snippet.
+
+``` csharp
+var rsaPrivateKey = "..."; // RSA private key from the App configuration page
```
kfcampbell, line 106: Is this the key itself or the path to the key? I think this could be made more clear.

docs/github-apps.md
```diff
@@ -100,7 +100,38 @@ That concludes the walkthrough!
...
+var appId = 1; // The GitHub App Id
+
+using var rsa = RSA.Create();
```
kfcampbell, line 109: What's the advantage of the `using` statement here?

docs/github-apps.md
```diff
@@ -100,7 +100,38 @@ That concludes the walkthrough!
...
+    claims: new[]
+    {
+        new Claim("iat", new DateTimeOffset(now).ToUnixTimeSeconds().ToString(), ClaimValueTypes.Integer),
+        new Claim("iss", appId.ToString(), ClaimValueTypes.Integer),
+    }
+);
+var token = new JwtSecurityTokenHandler().WriteToken(jwt);
+```
+
+Alternatively, one of our contributors [@adriangodong](https://github.com/adriangodong) has created a library `GitHubJwt` ( [GitHub](https://github.com/adriangodong/githubjwt) | [NuGet](https://www.nuget.org/packages/githubjwt) ) which you can use as per the following example.
```
kfcampbell, line 134:
```suggestion
Alternatively, one of our contributors [@adriangodong](https://github.com/adriangodong) has created a library `GitHubJwt` ([GitHub](https://github.com/adriangodong/githubjwt) | [NuGet](https://www.nuget.org/packages/githubjwt)) which you can use as per the following example.
```

I've never liked the spacing on that line; this is an opportunity to fix it.

kfcampbell (commented):
Thanks for the contribution, this is great! I would rather encourage people to use system libraries. I had a couple quick questions and a suggestion before shipping.

docs/github-apps.md
rasmus: Since the `RSA` object implements the `IDisposable` interface, its usually required to invoke the `.Dispose()` when the object leaves scope and is no longer needed. There could be some resources not managed by the runtime [that needs cleaning up](https://learn.microsoft.com/en-us/dotnet/api/system.idisposable?view=net-8.0). I haven't decompiled the code to see what's actually there. The syntax for declaring `using` was made available in [.NET 8](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/using).

rasmus:
@kfcampbell do. I added a code comment regarding the required using statements as well and commented on the `using` statement

nickfloyd (approved):

colbylwilliams:
@rasmus first, thank you for [this](https://github.com/octokit/octokit.net/pull/2937).  This token is one of the most confusing parts of getting started with GitHub Apps.

I noticed an [announcement](https://github.blog/changelog/2024-05-01-github-apps-can-now-use-the-client-id-to-fetch-installation-tokens/) on the GitHub blog directing GitHub Apps to starting using the **`clientId`** instead of the **`appId`** in the `iss` when minting these tokens. 

From the announcement:

> To date, GitHub Apps have had two different IDs to manage – the application ID and the client ID. The application ID was only used to mint a JWT, subsequently used to fetch an installation token. The client ID is used with the OAuth flow to sign in users and request installations. These two values equally identify the application and the question of which one to use where caused unnecessary developer friction. You can now use the client ID in the place of the application ID when minting JWTs.

It might be worth updating the sample to use the **`clientId`**.

cc: @kfcampbell @nickfloyd

nickfloyd:
> It might be worth updating the sample to use the **`clientId`**.

Thanks for following up on this @colbylwilliams, you're correct about the shift.  We've made this move in the newly [generated SDKs](https://github.com/octokit/dotnet-sdk) as well.

PR #2939, opened 2024-06-20
[FEAT]: Create a GitHub App from a manifest

### LINKED ISSUE #2938: [FEAT]: Create a GitHub App from a manifest
### Describe the need

Add support for the [Creating a GitHub App from a manifest](https://docs.github.com/en/rest/apps/apps?apiVersion=2022-11-28#create-a-github-app-from-a-manifest).  This is the final step when using the [GitHub App Manifest flow](https://docs.github.com/en/apps/sharing-github-apps/registering-a-github-app-from-a-manifest)

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

_No response_

### Code of Conduct

colbylwilliams:
- Resolves #2938

----

### Before the change?

* Does not support [Creating a GitHub App from a manifest](https://docs.github.com/en/rest/apps/apps?apiVersion=2022-11-28#create-a-github-app-from-a-manifest). This is the final step when using the [GitHub App Manifest flow](https://docs.github.com/en/apps/sharing-github-apps/registering-a-github-app-from-a-manifest)

### After the change?

* Adds support for [Creating a GitHub App from a manifest](https://docs.github.com/en/rest/apps/apps?apiVersion=2022-11-28#create-a-github-app-from-a-manifest).
* Adds missing InstallationPermissions:
  * [`repository_custom_properties`](https://docs.github.com/rest/authentication/permissions-required-for-github-apps#repository-permissions-for-custom-properties)
  * [`organization_copilot_seat_management`](https://docs.github.com/rest/authentication/permissions-required-for-github-apps#organization-permissions-for-github-copilot-business)
  * [`organization_custom_properties`](https://docs.github.com/rest/authentication/permissions-required-for-github-apps#organization-permissions-for-custom-properties)

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/Octokit.Reactive/Clients/ObservableGitHubAppsClient.cs
+++ b/Octokit.Reactive/Clients/ObservableGitHubAppsClient.cs
@@ -175,5 +175,19 @@ namespace Octokit.Reactive
 
             return _client.GetUserInstallationForCurrent(user).ToObservable();
         }
+
+        /// <summary>
+        /// Creates a GitHub app by completing the handshake necessary when implementing the GitHub App Manifest flow.
+        /// https://docs.github.com/apps/sharing-github-apps/registering-a-github-app-from-a-manifest
+        /// </summary>
+        /// <remarks>https://docs.github.com/rest/apps/apps#create-a-github-app-from-a-manifest</remarks>
+        /// <param name="code">Temporary code in a code parameter.</param>
+        [ManualRoute("POST", "/app-manifests/{code}/conversions")]
+        public IObservable<GitHubAppFromManifest> CreateAppFromManifest(string code)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(code, nameof(code));
+
+            return _client.CreateAppFromManifest(code).ToObservable();
+        }
     }
 }
--- a/Octokit.Reactive/Clients/ObservableGitHubAppsClient.cs
+++ b/Octokit.Reactive/Clients/ObservableGitHubAppsClient.cs
@@ -182,7 +182,6 @@ namespace Octokit.Reactive
         /// </summary>
         /// <remarks>https://docs.github.com/rest/apps/apps#create-a-github-app-from-a-manifest</remarks>
         /// <param name="code">Temporary code in a code parameter.</param>
-        [ManualRoute("POST", "/app-manifests/{code}/conversions")]
         public IObservable<GitHubAppFromManifest> CreateAppFromManifest(string code)
         {
             Ensure.ArgumentNotNullOrEmptyString(code, nameof(code));
--- a/Octokit/Models/Response/GitHubApp.cs
+++ b/Octokit/Models/Response/GitHubApp.cs
@@ -13,10 +13,11 @@ namespace Octokit
     {
         public GitHubApp() { }
 
-        public GitHubApp(long id, string slug, string name, User owner, string description, string externalUrl, string htmlUrl, DateTimeOffset createdAt, DateTimeOffset updatedAt, InstallationPermissions permissions, IReadOnlyList<string> events)
+        public GitHubApp(long id, string slug, string nodeId, string name, User owner, string description, string externalUrl, string htmlUrl, DateTimeOffset createdAt, DateTimeOffset updatedAt, InstallationPermissions permissions, IReadOnlyList<string> events)
         {
             Id = id;
             Slug = slug;
+            NodeId = nodeId;
             Name = name;
             Owner = owner;
             Description = description;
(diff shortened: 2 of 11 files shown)


Octokit.Reactive/Clients/ObservableGitHubAppsClient.cs
```diff
@@ -175,5 +175,19 @@ public IObservable<Installation> GetUserInstallationForCurrent(string user)
...
             return _client.GetUserInstallationForCurrent(user).ToObservable();
         }
+
+        /// <summary>
+        /// Creates a GitHub app by completing the handshake necessary when implementing the GitHub App Manifest flow.
+        /// https://docs.github.com/apps/sharing-github-apps/registering-a-github-app-from-a-manifest
+        /// </summary>
+        /// <remarks>https://docs.github.com/rest/apps/apps#create-a-github-app-from-a-manifest</remarks>
+        /// <param name="code">Temporary code in a code parameter.</param>
+        [ManualRoute("POST", "/app-manifests/{code}/conversions")]
```
kfcampbell, line 185: Do you mind explaining why the `ManualRoute` decorator is required here?

Octokit/Models/Response/GitHubApp.cs
```diff
@@ -31,57 +32,62 @@ public GitHubApp(long id, string slug, string name, User owner, string descripti
         /// <summary>
         /// The Id of the GitHub App.
         /// </summary>
-        public long Id { get; private set; }
+        public long Id { get; protected set; }
```
kfcampbell, line 35: What's the reason for (and the ramifications of) the permissions change here?

Octokit/Models/Response/GitHubAppFromManifest.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace Octokit
+{
+    /// <summary>
+    /// Represents a GitHub application from a manifest.
+    /// https://docs.github.com/rest/apps/apps#create-a-github-app-from-a-manifest
+    /// </summary>
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
```
kfcampbell, line 11: What's this for?

kfcampbell (commented):
Thanks for contributing! I had a few (hopefully quick) questions.

Octokit/Models/Response/GitHubApp.cs
colbylwilliams: To support inheritance.

`GitHubAppFromManifest` inherits from `GitHubApp`.  The `SimpleJson`/`SimpleJsonSerializer` this project uses won't populate any of the  properties inherited from `GitHubApp` on `GitHubAppFromManifest` objects if the property is `private set;` (i.e. `id`, `slug`, `name`, etc. are all `null`).

Another example is [`License`](https://github.com/octokit/octokit.net/blob/main/Octokit/Models/Response/License.cs) subclasses [`LicenseMetadata`](https://github.com/octokit/octokit.net/blob/main/Octokit/Models/Response/LicenseMetadata.cs).  All properties on `LicensesMetadata` are `protected set`

As far as ramifications, I can't think of any that would be negative.  From the docs:

> Only code in the same class or in a derived class can access this type or member.

Honestly, it would be a big help if the standard in this library was to use `protected set` instead of `private set`.  I've often tried to create subclasses of types from this library ( like`Repository` or `GitHubApp`) to quickly add and deserialize properties being returned by the API but missing from this library's types.  Unfortunately, this is not possible due to the issue above.

Octokit/Models/Response/GitHubAppFromManifest.cs
colbylwilliams: When creating a new GitHub App using the [Manifest flow](https://docs.github.com/en/apps/sharing-github-apps/registering-a-github-app-from-a-manifest#about-github-app-manifests), GitHub redirects the user to the external API with a [`code` query parameter that the service can exchange](https://docs.github.com/en/apps/sharing-github-apps/registering-a-github-app-from-a-manifest#3-you-exchange-the-temporary-code-to-retrieve-the-app-configuration) for all the details about the new GitHub App, including the Client Id, Client Secret, auto-generated Webhook Secret, PEM, etc.

As such, it's critical to deserialize the additional values on this type and save them for the new GitHub App's backing service to use.

From GitHub's [official REST API spec](https://raw.githubusercontent.com/github/rest-api-description/main/descriptions/api.github.com/api.github.com.yaml):

```openapi
"/app-manifests/{code}/conversions":
    post:
      summary: Create a GitHub App from a manifest
      description: Use this endpoint to complete the handshake necessary when implementing
        the [GitHub App Manifest flow](https://docs.github.com/apps/building-github-apps/creating-github-apps-from-a-manifest/).
        When you create a GitHub App with the manifest flow, you receive a temporary
        `code` used to retrieve the GitHub App's `id`, `pem` (private key), and `webhook_secret`.
      tags:
      - apps
      operationId: apps/create-from-manifest
      externalDocs:
        description: API method documentation
        url: https://docs.github.com/rest/apps/apps#create-a-github-app-from-a-manifest
      parameters:
      - name: code
        in: path
        required: true
        schema:
          type: string
      responses:
        '201':
          description: Response
          content:
            application/json:
              schema:
                allOf:
                - "$ref": "#/components/schemas/integration"
                - type: object
                  properties:
                    client_id:
                      type: string
                    client_secret:
                      type: string
                    webhook_secret:
                      type: string
                      nullable: true
                    pem:
                      type: string
                  required:
                  - client_id
                  - client_secret
                  - webhook_secret
                  - pem
                  additionalProperties: true
              examples:
                default:
                  "$ref": "#/components/examples/integration-from-manifest"
        '404':
          "$ref": "#/components/responses/not_found"
        '422':
          "$ref": "#/components/responses/validation_failed_simple"
```

Octokit.Reactive/Clients/ObservableGitHubAppsClient.cs
colbylwilliams: leftover copy/pasta.  removed

Octokit/Models/Response/GitHubApp.cs
nickfloyd: This is probably one of the best arguments for using `protected` that I have heard so far.  Just for some context on why they are all private, you can read this [issue](https://github.com/octokit/octokit.net/issues/2472) and see this [PR](https://github.com/octokit/octokit.net/pull/2565) and [release](https://github.com/octokit/octokit.net/discussions/2567) if you're interested. 

The gist was that the library was pretty inconsistent when it came to how things were implemented and even a couple of years ago we knew we wanted to head to a generative approach.  We went with the more restrictive pattern of using `private` accessors as a clean slate starting point so that we'd have some consistently and so that we could generate source using that consistency.  

Given that we are close to GA'ing our [Generated .NET SDK](https://github.com/octokit/dotnet-sdk) and for the points you made, it makes sense to make  mutating these properties more permissive.

Please feel free to push forward on that front if you have time.  As a word of caution, the custom serializer that has been built makes use of property accessors and you might run into a couple issues around the serialization tests - but they should be relatively low risk.

nickfloyd (approved):

Issue #2893, opened 2024-03-05
[BUG]: Int32 overflow for Github object ids


Apolloluy:
### What happened?

Running 
```csharp
var envs = await github.Repository.Environment.GetAll(owner, "API_CICD");
```
Returns 
![image](https://github.com/octokit/octokit.net/assets/44653204/9ea8a7ab-d27b-4210-9322-78cb663a76dc)

`System.OverflowException: 'Value was either too large or too small for an Int32.'`
When attempting to parse the Github.id field in the response.

### Versions

Octokit.nupkg 10.0.0 (.net)
![image](https://github.com/octokit/octokit.net/assets/44653204/8d2ca5aa-9662-4558-9662-449445bbd4a5)

### Relevant log output

```shell
System.OverflowException
  HResult=0x80131516
  Message=Value was either too large or too small for an Int32.
  Source=System.Private.CoreLib
  StackTrace:
   at System.Convert.ThrowInt32OverflowException()

Value = 2398648970
```

```json
{
			"id": 2398648970,
			"node_id": "EN_kwDOJhYfOc6O-HqK",
			"name": "test",
			"url": "https://api.github.com/repos/ContextureHIE/API_CICD/environments/test",
			"html_url": "https://github.com/ContextureHIE/API_CICD/deployments/activity_log?environments_filter=test",
			"created_at": "2024-03-01T22:23:36Z",
			"updated_at": "2024-03-01T22:23:36Z",
			"can_admins_bypass": true,
			"protection_rules": [
				{
					"id": 16441898,
					"node_id": "GA_kwDOJhYfOc4A-uIq",
					"type": "required_reviewers",
					"prevent_self_review": false,
					"reviewers": [
						{
							"type": "User",
							"reviewer": {
								"login": "Apolloluy",
								"id": 44653204,
								"node_id": "MDQ6VXNlcjQ0NjUzMjA0",
								"avatar_url": "https://avatars.githubusercontent.com/u/44653204?v=4",
								"gravatar_id": "",
								"url": "https://api.github.com/users/Apolloluy",
								"html_url": "https://github.com/Apolloluy",
								"followers_url": "https://api.github.com/users/Apolloluy/followers",
								"following_url": "https://api.github.com/users/Apolloluy/following{/other_user}",
								"gists_url": "https://api.github.com/users/Apolloluy/gists{/gist_id}",
								"starred_url": "https://api.github.com/users/Apolloluy/starred{/owner}{/repo}",
[...]

PR #2941, opened 2024-06-24
[BREAKING CHANGES]: int to long Ids for PreReceiveHook, Deployment Environments, Repository, Org Team, Repo Invitations, Public Key, Project Cards, Organization Invitation, Migrations, GpgKey, Deployment, Authorizations, Accounts / Profiles, Codespace / Workspaces

nickfloyd:
Relates to #2893

We have uncovered a series of issues around the IDs used across our API surface (Issues, Pull Requests, deployments, etc.. are all effected).

The GitHub database schemas were changed to accommodate for the growing IDs - migrating many of them from int to bigint - this SDK was lagging behind in those changes so when API calls began to return data that was too large for the types defined in our SDKs  - their apps, integration, and automation began to break.

----

### Before the change?

* Overflow issues would happen when calling an API that returned a `long` (e.g. a release ID) but the class had this defined as an `int` - this would cause the SDK to throw an overflow exception. 

### After the change?

* We have updated the `int` types to be `long` across the SDKs - this will allow for the larger IDs to be returned without causing an overflow exception. 

### Pull request checklist

- [x] Yes
- [ ] No

----

### DIFF
--- a/Octokit.AsyncPaginationExtension/Extensions.cs
+++ b/Octokit.AsyncPaginationExtension/Extensions.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using Octokit.Models.Request.Enterprise;
 
 namespace Octokit.AsyncPaginationExtension
 {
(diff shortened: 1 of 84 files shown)


kfcampbell (approved):
Awesome job! This is super helpful for our users.

kfcampbell (approved):

PR #2962, opened 2024-08-27
Issue ID should be a long

### LINKED ISSUE #2351: GitHub IDs have exceeded Int32.MaxValue; Octokit uses int

kzu:
According to the documentation at https://docs.github.com/en/rest/issues/issues.

Fixes #2351 for issue ids.

- [X] Yes
- [ ] No

----


kfcampbell:
This was already done in #2890. You've changed issue _number_, which is the "2351" in the linked issue `https://github.com/octokit/octokit.net/issues/2351`. The issue number increments from 1 for each new repository, and can safely be described with an integer.

kzu:
image

vs

image


either one or the other is the wrong type.

PR #2965, opened 2024-08-30
Fix runner group access return type

### LINKED ISSUE #2964: [BUG]: Runner group access APIs return empty objects
### What happened?

These API responses aren't being deserialized to the proper type so all properties are unset

### Versions

13.0.1

### Relevant log output

_No response_

### Code of Conduct

MatisseHack:
Resolves #2964

----

### Before the change?

* Runner group access APIs would return empty objects because they weren't wrapped in a response object

### After the change?

* Runner group access APIs return properly deserialized responses

### Pull request checklist

I didn't see existing tests for deserialization, but I'm happy to add tests if there is an established pattern to follow.

- [x] Yes
- [ ] No

The return type of several functions changed (although these functions would have returned empty objects before)

----


kfcampbell (approved):
Thanks!

PR #2988, opened 2025-01-03
Add optional cancellation token for oauth client

zoriya:
I did not make an issue, but there is already multiples PR for other parts of the codebase so I guess this is fine?

----

### Before the change?
We could never cancel login requests (especially device flow requests, which waits for the user to do something in a browser window).

### After the change?
We can now cancel such requests.

### Pull request checklist

- [ ] Yes
- [x] No

----


nickfloyd (approved):
Hey @zoriya thanks for the contributions here! ❤️

PR #3040, opened 2025-09-29
Fix new tree deletion

### LINKED ISSUE #2836: [BUG]: Creating a new tree with a null sha tree object fails
### What happened?

I'd like to delete multiple files within one commit, so was constructing a tree with multiple objects with a null sha, as per [documentation](https://stackoverflow.com/a/67101960). I get an API error `Must supply either tree.sha or tree.content. Request will be rejected if both are present.`. When I run an identical API request via Postmark with sha set to null, it succeeds in creating the tree. When I run the request with sha not present, I see that API error.

I imagine its something in the serializer stripping out null fields, when in this case I would like them to be preserved.

I can't use the client.Repository.Content.DeleteFile api, as that creates a new commit for every file you delete.

Example failing code below.
```
var toDelete = new List<string>() { "file1", "file2" };
var repository = await client.Repository.Get("ORGID", "REPONAME");
var headOfDefault = await client.Git.Reference.Get(repository.Id, $"heads/{repository.DefaultBranch}");
var tree = new NewTree() {
    BaseTree = headOfDefault.Object.Sha
};
foreach (var file in toDelete) {
    tree.Tree.Add(
        new NewTreeItem() 
        {
            Type = TreeType.Blob,
            Mode = FileMode.File,
            Path = file
            Sha = null
        });
}
var newTree = await client Git.Tree.Create(repository.Id, tree);
```

### Versions

Octokit 9.0.0 running on net6.0

### Relevant log output

_No response_

### Code of Conduct

zoriya:
Fixes #2836

----

### Before the change?

* When creating a tree with a null content & sha, we got this error from github: `Must supply either tree.sha or tree.content. Request will be rejected if both are present`
* Since `null` are not serialized, gh received neither a `sha` nor a `content` value in the tree and that caused the error.

### After the change?

* Actually serialize the `null` value of the `sha` to allow deletion of tree items.

### Pull request checklist

I made an integration test but running it is a bit of a pain, I noticed there's no CI on my fork and i am opening this PR to get CI runs x)

- [ ] Yes
- [x] No

----

### DIFF
--- a/Octokit.Tests.Integration/Clients/TreeClientTests.cs
+++ b/Octokit.Tests.Integration/Clients/TreeClientTests.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Threading.Tasks;
 using Octokit;
 using Octokit.Tests.Integration;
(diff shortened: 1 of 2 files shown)


zoriya:
Note that i also tested this in one of our project and i can confirm it works.

zoriya:
We've been using this PR in production for almost a year, would be great to have this merged!

PR #1418, opened 2016-07-11
search repos with underscore in name

shiftkey:
Fixes #1406

I've kept the failing test around redirected repositories, but I'm not sure when the backend fix will land for the search APIs.

cc @dsplaisted, who found the fix originally

Issue #1406, opened 2016-06-24
Can't search repos with underscore in repo name


dsplaisted:
The following test fails:

```c#
    [IntegrationTest]
    public async Task SearchForFileNameInCodeWithoutTermWithUnderscore()
    {
        var request = new SearchCodeRequest()
        {
            FileName = "readme.md",
            Repos = new RepositoryCollection { "Cultural-Rogue/_51Wp.XinFengSDK.Demo" }
        };

        var repos = await _gitHubClient.Search.SearchCode(request);
    }
```

It fails with the following exception:

```
Octokit.RepositoryFormatException : The list of repositories must be formatted as 'owner/name' - these values don't match this rule: Cultural-Rogue/_51Wp.XinFengSDK.Demo
at Octokit.SearchCodeRequest.MergedQualifiers() in C:\git\octokit.net\Octokit\Models\Request\SearchCodeRequest.cs:line 204
   at Octokit.BaseSearchRequest.get_TermAndQualifiers() in C:\git\octokit.net\Octokit\Models\Request\BaseSearchRequest.cs:line 87
   at Octokit.BaseSearchRequest.get_Parameters() in C:\git\octokit.net\Octokit\Models\Request\BaseSearchRequest.cs:line 106
   at Octokit.SearchClient.SearchCode(SearchCodeRequest search) in C:\git\octokit.net\Octokit\Clients\SearchClient.cs:line 67
   at SearchClientTests.<SearchForFileNameInCodeWithoutTermWithUnderscore>d__8.MoveNext() in C:\git\octokit.net\Octokit.Tests.Integration\Clients\SearchClientTests.cs:line 99
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
[...]

PR #3083, opened 2026-07-11
Fix RepositoryFormatException for repository names that start with a hyphen

### LINKED ISSUE #3082: [BUG]: SearchIssuesRequest throws RepositoryFormatException for repository names that start with a hyphen
### What happened?

Any search request that uses the `Repos` collection (`SearchIssuesRequest`, `SearchCodeRequest`, ...) throws `RepositoryFormatException` **client-side, before any HTTP request is made**, when the repository *name* part starts with a hyphen — even though such names are valid on GitHub and GitHub's Search API accepts the corresponding `repo:` qualifier.

#### Repro

```csharp
var client = new GitHubClient(new ProductHeaderValue("repro"));

var request = new SearchIssuesRequest
{
    Repos = new RepositoryCollection { "myorg/-repo-with-leading-hyphen" }
};

// Throws Octokit.RepositoryFormatException before any HTTP call
await client.Search.SearchIssues(request);
```

**Expected:** the qualifier `repo:myorg/-repo-with-leading-hyphen` is sent to the search API.

**Actual:**

> `Octokit.RepositoryFormatException: The list of repositories must be formatted as 'owner/name' - these values don't match this rule: myorg/-repo-with-leading-hyphen`

#### Root cause

The validation regex in [`Octokit/Helpers/StringExtensions.cs#L142`](https://github.com/octokit/octokit.net/blob/main/Octokit/Helpers/StringExtensions.cs#L142):

```csharp
static readonly Regex nameWithOwner = new Regex("[a-z0-9.-]{1,}/[a-z0-9.-_]{1,}", ...);
```

In the *name* part `[a-z0-9.-_]`, the `.-_` sequence is not three literal characters — it is a character **range** from `.` (0x2E) to `_` (0x5F). The literal hyphen `-` (0x2D) sits just **below** the range start, so it is not in the class at all. A name whose first character is `-` therefore cannot match immediately after the `/`, and `IsNameWithOwnerFormat()` returns `false`. (Names with hyphens in the *middle* still pass only because the regex is unanchored, so the prefix before the first hyphen satisfies `{1,}`.)

This regressed in #1418 (the fix for #1406, repo names starting with an underscore): appending `_` after the previously **trailing** literal `-` (`[a-z0-9.-]` → `[a-z0-9.-_]`) accidentally turned `.-` + `_` into the `.-_` range — fixing leading underscores while silently breaking leading hyphens. As a side effect the accidental range also admits characters that can never appear in a repository name (`/`, `:`, `;`, `<`, `=`, `>`, `?`, `@`, `[`, `\`, `]`, `^`).

#### GitHub itself is fine with these names

* Repository names starting with `-` can be created on github.com; we hit this in production against a real (private) organization repository whose name starts with `-`, returned by GitHub's own *list organization repositories* API.
* The server-side search endpoint parses the qualifier without trouble: `GET /search/issues?q=repo:owner/-nonexistent-repo+type:pr` returns the standard `422` *"The listed users and repositories cannot be searched..."* validation response — exactly the same as for any nonexistent repo, not a query-parse error.

#### Suggested fix

Reorder the class so all three specials are literals — `.-_` → `_.-`:

```csharp
static readonly Regex nameWithOwner = new Regex("[a-z0-9.-]{1,}/[a-z0-9_.-]{1,}", ...);
```

A `-` or `.` at the class edges and `_` mid-class are all literal, so leading-hyphen (and leading-underscore, leading-dot) names validate, and the accidental junk range disappears. I'll open a PR with this change plus unit tests and link it here.

### Versions

Octokit.net 14.0.0 (still present on current `main`); observed on .NET 8 / .NET 10 — platform-independent (pure regex).

### Relevant log output

```shell
Octokit.RepositoryFormatException: The list of repositories must be formatted as 'owner/name' - these values don't match this rule: <org>/-<repo-name>
   at Octokit.SearchIssuesRequest.MergedQualifiers() in /_/Octokit/Models/Request/SearchIssuesRequest.cs:line 399
   at Octokit.BaseSearchRequest.get_Parameters() in /_/Octokit/Models/Request/BaseSearchRequest.cs:line 117
   at Octokit.SearchClient.SearchIssues(SearchIssuesRequest search) in /_/Octokit/Clients/SearchClient.cs:line 58
```

### Code of Conduct

asaf-clover:
Fixes #3082

## Problem

Search requests that use the `Repos` collection throw `RepositoryFormatException` client-side for repository names that start with a hyphen (e.g. `myorg/-my-repo`), even though such names are valid on GitHub and the server-side search API accepts the corresponding `repo:` qualifier.

## Root cause

In the `nameWithOwner` validation regex, the name-part character class `[a-z0-9.-_]` does not contain three literal punctuation characters — `.-_` is a character **range** (0x2E–0x5F). The literal `-` (0x2D) sits just below the range start, so it is not in the class, and any name whose *first* character is `-` fails to match right after the `/` (mid-name hyphens still pass only because the regex is unanchored).

This regressed in #1418 (the fix for #1406, leading-underscore names): appending `_` after the previously *trailing* literal `-` turned `.-` + `_` into the `.-_` range.

## Fix

Reorder the class so all three specials are literals — `.-_` → `_.-`:

```diff
-        static readonly Regex nameWithOwner = new Regex("[a-z0-9.-]{1,}/[a-z0-9.-_]{1,}",
+        static readonly Regex nameWithOwner = new Regex("[a-z0-9.-]{1,}/[a-z0-9_.-]{1,}",
```

Leading-hyphen names now validate; leading-underscore (#1406) and leading-dot (`.github`) names keep working. As a side effect the accidental range no longer admits characters that can never appear in a repository name (`/ : ; < = > ? @ [ \ ] ^`), so the validation also becomes slightly stricter in the correct direction.

## Tests

- `StringExtensionsTests.TheIsNameWithOwnerFormatMethod` — new theory covering leading-hyphen, leading-underscore, leading-dot, regular names, and invalid input.
- `SearchClientTests.TestingTheRepoQualifier_RepoNameStartingWithHyphen` — end-to-end: `SearchIssues` sends `repo:octokit/-repo-starting-with-hyphen` instead of throwing.
- Full `Octokit.Tests` unit suite: 4550 passed / 4 failed — the 4 failures are the `CanPopulateObjectFromSerializedData` (BinaryFormatter) tests, which fail identically on unmodified `main` under the .NET 10 runtime used locally (BinaryFormatter is disabled in .NET 9+); they are unrelated to this change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Octokit.Tests/Clients/SearchClientTests.cs
+++ b/Octokit.Tests/Clients/SearchClientTests.cs
@@ -1587,6 +1587,21 @@ namespace Octokit.Tests.Clients
             }
 
             [Fact]
+            public void TestingTheRepoQualifier_RepoNameStartingWithHyphen()
+            {
+                var connection = Substitute.For<IApiConnection>();
+                var client = new SearchClient(connection);
+                var request = new SearchIssuesRequest("something");
+                request.Repos.Add("octokit", "-repo-starting-with-hyphen");
+
+                client.SearchIssues(request);
+
+                connection.Received().Get<SearchIssuesResult>(
+                    Arg.Is<Uri>(u => u.ToString() == "search/issues"),
+                    Arg.Is<Dictionary<string, string>>(d => d["q"] == "something+repo:octokit/-repo-starting-with-hyphen"));
+            }
+
+            [Fact]
             public async Task ErrorOccursWhenSpecifyingInvalidFormatForRepos()
             {
                 var connection = Substitute.For<IApiConnection>();
--- a/Octokit.Tests/Helpers/StringExtensionsTests.cs
+++ b/Octokit.Tests/Helpers/StringExtensionsTests.cs
@@ -32,6 +32,22 @@ namespace Octokit.Tests.Helpers
             }
         }
 
+        public class TheIsNameWithOwnerFormatMethod
+        {
+            [InlineData("octokit/octokit.net", true)]
+            [InlineData("owner/-repo-starting-with-hyphen", true)]
+            [InlineData("owner/_repo_starting_with_underscore", true)]
+            [InlineData("owner/.github", true)]
+            [InlineData("my-org/my.repo_name-1", true)]
+            [InlineData("haha-business", false)]
+            [InlineData("", false)]
+            [Theory]
+            public void ProperlyDetectsNameWithOwnerStrings(string data, bool expected)
+            {
+                Assert.Equal(expected, data.IsNameWithOwnerFormat());
+            }
+        }
+
         public class TheToRubyCaseMethod
         {
             [Theory]
--- a/Octokit/Helpers/StringExtensions.cs
+++ b/Octokit/Helpers/StringExtensions.cs
@@ -139,7 +139,7 @@ namespace Octokit
         // the rule:
         // Username may only contain alphanumeric characters or single hyphens
         // and cannot begin or end with a hyphen
-        static readonly Regex nameWithOwner = new Regex("[a-z0-9.-]{1,}/[a-z0-9.-_]{1,}",
+        static readonly Regex nameWithOwner = new Regex("[a-z0-9.-]{1,}/[a-z0-9_.-]{1,}",
 #if HAS_REGEX_COMPILED_OPTIONS
             RegexOptions.Compiled |
 #endif


asaf-clover:
@G-Rath 
Hi,
Can you please help me push this PR.
Thanks!

asaf-clover:
@octokitbot

